PR #182, opened 2026-02-02
fix(sync): make family_camp_derived idempotent

adamflagg:
## Summary
- Fix family_camp_derived sync to use upsert pattern instead of delete-all-rebuild
- Each run now correctly reports Skipped (not Created) when data is unchanged
- Adds orphan deletion for records no longer in source data

## Changes
- Add ProcessedKeys tracking for orphan detection  
- Add preload functions to load existing records by composite key
- Add comparison helpers for each record type (adults, registrations, medical)
- Convert write functions to upsert logic (create/update/skip)
- Remove clearExistingRecords function

## Test plan
- [x] Added idempotency tests defining expected behavior
- [x] All Go tests pass (`go test ./sync/...`)
- [ ] Manual verification: run sync twice, second run shows Skipped=N instead of Created=N

### DIFF
--- a/pocketbase/sync/family_camp_derived.go
+++ b/pocketbase/sync/family_camp_derived.go
@@ -27,14 +27,22 @@ type FamilyCampDerivedSync struct {
 	DryRun         bool // Dry run mode (compute but don't write)
 	Stats          Stats
 	SyncSuccessful bool
+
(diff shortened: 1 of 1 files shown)


adamflagg:
Merged locally - included in main

PR #186, opened 2026-02-02
fix(sync): debug propagation, dietary idempotency, phone field size

adamflagg:
## Summary
- Fix debug flag not propagating to all sync services during unified sync
- Add idempotency to camper_dietary sync (skip unchanged records on re-run)
- Increase bus_pickup_phone and bus_alt_phone field max from 50 to 200 chars

## Test plan
- [ ] Run sync with `?debug=true` and verify debug output appears for all services
- [ ] Run camper_dietary sync twice; second run should show `Updated: 0, Skipped: N`
- [ ] Run quest_registrations sync; verify no truncation errors for long phone numbers


adamflagg:
Merged locally - included in main

PR #575, opened 2026-03-14
fix(frontend): resolve all ESLint warnings with per-case design decisions

### LINKED ISSUE #571: fix(frontend): resolve remaining ESLint warnings requiring design decisions
## Context

PR #572 resolved ~885 of 943 ESLint warnings via mechanical fixes. The remaining ~55 warnings require per-case design decisions and were intentionally skipped.

### `@typescript-eslint/no-non-null-assertion` (35 warnings)
Removing `!` non-null assertions requires deciding HOW to handle the null case for each occurrence:
- Early return?
- Throw an error?
- Optional chain with default?
- Restructure the logic?

Each case needs individual evaluation of what the correct null-handling strategy is.

### `@typescript-eslint/restrict-template-expressions` (16 warnings)
Template literals containing non-string types need per-case decisions about:
- Which `.toString()` or type coercion to use
- Whether to add a type guard instead
- Whether the template expression should be restructured

### `@typescript-eslint/no-unnecessary-type-assertion` (2 warnings)
Redundant `as X` casts that slipped through the bulk cleanup. These may be straightforward to remove but need verification that removing the cast doesn't change behavior.

### `@typescript-eslint/no-unnecessary-condition` (2 warnings)
Redundant conditionals that slipped through the bulk cleanup. Need per-case evaluation.

## Why these were skipped

These are not mechanical find-and-replace fixes. Each warning requires understanding the surrounding logic and making a design choice about the correct approach. Lumping them into a bulk cleanup PR would risk introducing bugs.

## Suggested approach

Work through these in small batches grouped by file/component, with test verification for each change.

adamflagg:
## Summary
- Resolves all 58 remaining ESLint warnings that required per-case design decisions (58 → 0)
- Audits existing eslint-disable comments — all 76 confirmed appropriate, 3 new added with justification
- Updates issue triage doc to reflect verified state of all 6 open issues

## Warning fixes by category

| Category | Count | Fix strategy |
|----------|-------|-------------|
| `compareYear!` narrowing | 17 | Replace `isComparing` with `compareYear !== null` in JSX for TypeScript narrowing |
| `restrict-template-expressions` | 16 | `String()` wrapping for `unknown`/`{}`/`never` in template literals |
| `tooltip.item!` in callbacks | 6 | Extract to `const tooltipItem` via IIFE for callback narrowing |
| Guards / early returns | 6 | sessionUtils Map/array, GradeEligibility, FeedbackModal |
| Map.get()! double lookup | 4 | Store in local variable |
| `set-state-in-effect` (dialogs) | 2 | Move selection clearing from useEffect into onChange handlers |
| `Partial<Record<>>` type correction | 2 | SessionBunkHeatmap: fix type lie, add `?? []` fallbacks |
| `getByRole<T>()` generic | 2 | SnapshotDateSelector.test: use generic instead of `as` cast |
| Justified suppressions | 3 | App.tsx, sankeyColors.ts (constant arrays), ScenarioContext (async sync) |

## Design decisions documented

See `docs/superpowers/specs/2026-03-14-eslint-design-decisions-spec.md` for rationale on each pattern.

Closes #571

## Test plan
- [x] ESLint: 0 warnings, 0 errors
- [x] TypeScript: `tsc --noEmit` passes
- [x] All 2266 tests pass (165 test files)
- [x] Pre-push hooks pass (vitest + type-check + eslint)
- [x] Only 3 new eslint-disable comments added (all with invariant justification)

## Summary by CodeRabbit

* **Documentation**
  * Updated triage/design notes and ESLint decisions.

* **Bug Fixes**
  * Prevented feedback form submission without a selected category.
  * Improved error text rendering for failed operations.

* **Improvements**
  * Comparison sections now show only when an explicit comparison period is selected.
  * More robust handling/display of missing or non-string data across charts and reports.
  * Selection behavior refined in search/filter dialogs to avoid stale selections.

### DIFF
--- a/docs/reference/issue-triage.md
+++ b/docs/reference/issue-triage.md
@@ -1,7 +1,7 @@
 # Issue Triage
 
 Open issues grouped by code area with dependencies and suggested attack order.
-Last updated: 2026-03-14 (6 open issues, all verified against current code).
+Last updated: 2026-03-14 (6 open issues, all verified against current code; #576 opened during review).
 
 ---
 
@@ -48,10 +48,9 @@ Last updated: 2026-03-14 (6 open issues, all verified against current code).
 
 | # | Title | Type |
 |---|-------|------|
-| 571 | Resolve remaining ESLint warnings requiring design decisions | tech-debt |
 | 573 | Audit 57 `eslint-disable` comments added in ESLint cleanup for proper type fixes | tech-debt |
 
-**Interplay:** Both spawned from PR #572 (~320 ESLint warnings fixed). #571 has ~58 remaining warnings needing design decisions (non-null-assertion, restrict-template-expressions, setState-in-effect). #573: actual eslint-disable count is 111 (not 57 as originally estimated), with ~54 being `@typescript-eslint` rules. Address #571 first, then #573.
+**Interplay:** Spawned from PR #572 (~320 ESLint warnings fixed). #571 resolved by PR #575 (all 58 design-decision warnings fixed). #573 remains: actual eslint-disable count is 111 (not 57 as originally estimated), with ~54 being `@typescript-eslint` rules — 76 audited and confirmed appropriate in PR #575, deeper type refactor needed to remove `as` cast patterns.
 
 ---
 
@@ -69,7 +68,8 @@ Last updated: 2026-03-14 (6 open issues, all verified against current code).
 10. ~~**Group 13**~~ — ✅ Complete (PR #563) — #559 fixed; #560 closed as stale
 11. ~~**Group 3**~~ — ✅ Complete (PR #566) — #437, #472; spawned #562, #567
 12. ~~**Group 10**~~ — ✅ Complete (PRs #564, #565, #568, #569, #570, #572) — All 8 issues closed; spawned #571, #573
-13. **Groups 4, 8, 14, 15** — Independent items, sprinkle in anytime
+13. ~~**Group 15 (1/2)**~~ — ✅ Complete (PR #575) — #571 resolved (58 warnings); #573 remains (eslint-disable audit)
+14. **Groups 4, 8, 14, 15** — Independent items, sprinkle in anytime
 
 ## Completed Groups
 
@@ -92,3 +92,4 @@ Last updated: 2026-03-14 (6 open issues, all verified against current code).
 | Group 13: Frontend auth & state gaps (#559, #560) | #563 | 2026-03-14 | #559 fixed; #560 closed as stale |
 | Group 3: Velocity frontend refactors (#437, #472) | #566 | 2026-03-14 | Spawned #562 (full-options eval), #567 (mutual exclusivity) |
 | Group 10: Tests & docs (#536, #528, #485, #442, #552, #435, #495, #421) | #564, #565, #568, #569, #570, #572 | 2026-03-14 | All 8 issues closed; ESLint cleanup (885 warnings fixed); spawned #571, #573 |
+| Group 15 (1/2): ESLint design decisions (#571) | #575 | 2026-03-14 | 58 warnings resolved with per-case design decisions; #573 remains; spawned #576 (GradeEligibility threshold bug) |


adamflagg:
## Code Review -- PR #575

**Reviewer:** Claude Opus 4.6 (1M context) | **Score cutoff:** 70 | **Score: 95/100**

### Summary

This is a clean, well-structured lint cleanup PR. All 58 ESLint warnings are resolved with thoughtful per-case design decisions, documented in a local spec. The implementation is faithful to the approved spec, all 2266 tests pass, ESLint reports 0 warnings/errors, and TypeScript compiles cleanly. Only 3 new `eslint-disable` suppressions were added, all with clear invariant justifications.

### What was done well

- **compareYear narrowing pattern** (17 instances): Replacing `isComparing` with `compareYear !== null` only at JSX sites where `compareYear` is referenced is the minimal correct fix. `isComparing` is preserved elsewhere for readability (CSS classes, simple show/hide). This avoids introducing discriminated unions or wrapper types for a straightforward narrowing need.

- **Tooltip IIFE const extraction** (CssStackedHorizontalBarChart): Extracting `tooltip.item` to a `const tooltipItem` after the guard check is the right call. Optional chaining would silently produce zeros in arithmetic; assertions would just suppress the warning without narrowing. The IIFE keeps the JSX expression inline without introducing a separate component.

- **Event-driven selection clearing** (MergeDialog, ResolveDialog): Removing the `useEffect` and moving `setSelectedEntry(null)` into the `onChange` handlers is a textbook improvement. The old pattern was a classic "effect as event handler" anti-pattern. The new approach is more predictable -- selection clears at the source of the change, not as a reaction to derived state.

- **Justified suppressions**: All 3 new `eslint-disable` comments have clear invariant documentation. The `App.tsx` and `sankeyColors.ts` cases (compile-time constant arrays) are genuinely correct -- defensive fallback values would introduce magic constants that could silently drift. The `ScenarioContext` suppression correctly identifies that `useEffectEvent` already prevents the re-render loop the lint rule guards against.

- **SessionBunkHeatmap `Partial<Record<>>`**: Correcting the type from `Record<BunkCategory, string[]>` to `Partial<Record<BunkCategory, string[]>>` fixes a genuine type lie -- not all categories are guaranteed to have entries. The `?? []` fallbacks at the usage sites are the correct consequence.

- **Spec document**: The design decisions spec is thorough, well-reasoned, and properly gitignored (local-only under `docs/superpowers/`).

### Findings

| # | Finding | Severity | Verdict |
|---|---------|----------|---------|
| 1 | `FeedbackModal` early return in `mutationFn` silently resolves with `undefined` | Suggestion | Acceptable |
| 2 | `useVelocityChartData` still has redundant `.get()` calls on lines not touched | Suggestion | Out of scope |
| 3 | Issue triage doc updates are bundled into a lint fix commit | Suggestion | Acceptable |

### Finding Details

**#1 -- FeedbackModal mutationFn early return**

The `if (!category) return` guard in the `mutationFn` will cause the mutation to resolve successfully with `undefined` if somehow triggered without a category. This is safe in practice because `canSubmit` already guards the button, but `onSuccess` would still fire (showing "Feedback submitted" toast without actually submitting). A `throw` would be more defensive, but since the guard is truly unreachable in normal flow, this is acceptable as-is.

**#2 -- useVelocityChartData adjacent Map.get() calls**

Lines 126-131 still do multiple `mMap.get(wn)` and `fMap.get(wn)` calls for the non-cancelled fields (`enrolled`, `gross_enrolled`, `weekly_new`). These could also be consolidated into the local variables (`mVal`, `fVal`) for consistency. However, these lines don't have ESLint warnings (they use `?.` not `!`), so they're outside the scope of this PR.

**#3 -- Issue triage bundled with lint fixes**

The `docs/reference/issue-triage.md` changes (updating interplay descriptions for groups 4 and 5) are included in the same commit as the lint fixes. Since this PR is squash-merged, it's a minor scope expansion. The triage updates are accurate -- the ~320 warning count correction and the eslint-disable count update (111 not 57) reflect reality.

### Verdict

**Approve.** No critical or important issues found. The implementation is consistent, well-documented, and faithful to the spec. All verification criteria pass (0 ESLint warnings, TypeScript clean, 2266/2266 tests green). The 3 new suppressions are justified and the mechanical fixes are correct throughout.

PR #656, opened 2026-03-18
fix(sync): pass orchestrator session data to resolution pipeline

adamflagg:
## Summary
- **Fixes session-based matching at production scale** — `bulk_get_sessions_for_persons()` silently fails at >150 person IDs due to PocketBase OR clause length limits, causing all exact matches to default to "different session" confidence (0.6975 vs 0.9075)
- **Passes orchestrator's pre-loaded `_person_sessions`** into the resolution pipeline via `set_person_sessions()`, bypassing the broken query entirely
- **Adds chunking (100 IDs/chunk)** to the fallback `bulk_get_sessions_for_persons()` path for non-orchestrator callers (tests, CLI tools)

## Changes
| File | Change |
|------|--------|
| `resolution_pipeline.py` | Add `_person_sessions` field + `set_person_sessions()` setter; branch `batch_resolve()` session loading between pre-loaded and fallback paths |
| `orchestrator.py` | Wire `set_person_sessions()` after `_load_person_sessions()` |
| `attendee_repository.py` | Extract `_bulk_get_sessions_chunk()`, wrap with chunking at `_BULK_CHUNK_SIZE = 100` |
| `test_resolution_pipeline.py` | 5 new tests: pre-loaded sessions used, multi-session picks first, fallback without setter, session context derivation, chunking at 250 IDs |

## Test plan
- [x] Existing 14 tests pass unchanged
- [x] 5 new tests pass (TDD: written first, verified failing, then implemented)
- [x] `lefthook run pre-push` passes all checks (ruff, mypy, pytest 3155 passed, vitest, eslint, go-test, etc.)
- [ ] On dev: force reprocess 200+ requests, verify exact match confidence distribution shows 0.9075 for same-session matches

### DIFF
--- a/bunking/sync/bunk_request_processor/resolution/resolution_pipeline.py
+++ b/bunking/sync/bunk_request_processor/resolution/resolution_pipeline.py
@@ -54,7 +54,8 @@ class ResolutionPipeline:
         """Set pre-loaded person-to-session mappings from orchestrator.
 
         When set, batch_resolve() uses these instead of calling
-        bulk_get_sessions_for_persons() which fails at >150 person IDs.
+        bulk_get_sessions_for_persons(), avoiding per-chunk DB round-trips.
+        An empty dict is treated as "not set" so the fallback path still runs.
         """
         self._person_sessions = person_sessions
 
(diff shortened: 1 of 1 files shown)


adamflagg:
### Code review

Found 1 issue:

1. **Empty dict on exception bypasses fallback path.** `_load_person_sessions()` sets `self._person_sessions = {}` on exception (line 955), then `set_person_sessions({})` is unconditionally called (line 691). In `batch_resolve()`, the guard `if self._person_sessions is not None:` (line 232) evaluates `True` for `{}`, so the pre-loaded path runs with no data and the DB fallback is never reached. If the initial session load fails, ALL resolution loses session context silently.

   Fix: either guard with `if self._person_sessions:` (truthy check), or only call `set_person_sessions` when the load succeeded.

   https://github.com/adamflagg/kindred/blob/3d77fc4222ae62edf35ab3ce8905652379f1c69c/bunking/sync/bunk_request_processor/orchestrator/orchestrator.py#L690-L691

   https://github.com/adamflagg/kindred/blob/3d77fc4222ae62edf35ab3ce8905652379f1c69c/bunking/sync/bunk_request_processor/resolution/resolution_pipeline.py#L231-L242

   https://github.com/adamflagg/kindred/blob/3d77fc4222ae62edf35ab3ce8905652379f1c69c/bunking/sync/bunk_request_processor/orchestrator/orchestrator.py#L953-L956

Issue #653, opened 2026-03-18
fix(frontend): debug pipeline page shows inconsistent status casing and source field names


adamflagg:
## Problem

The debug pipeline page (`/summer/debug/pipeline`) displays inconsistent status values and source field names due to a mismatch between backend data conventions and frontend display logic.

### 1. Status casing: "resolved" vs "RESOLVED"

The orchestrator defaults to uppercase (`RESOLVED`, `PENDING` at `orchestrator.py:1488`), but when a matching `bunk_request` exists in the DB, it overwrites with the DB's lowercase value (lines 1492-1493). The frontend's `getStatusClasses()` in `PipelineBatchList.tsx` only matches uppercase — lowercase values fall through to default gray styling.

**DB evidence:**
```
PENDING: 1     | pending: 1449
RESOLVED: 155  | resolved: 204
               | declined: 66
```

### 2. Source field: display names vs canonical codes

Backend stores CampMinder display names (`"Share Bunk With"`, `"RetParent-Socializewithbest"`, `"BunkingNotes Notes"`). Frontend `getSourceFieldClasses()` expects canonical codes (`"bunk_with"`, `"not_bunk_with"`, `"socialize_with"`). All values fall through to default gray styling. Filter dropdown is also mismatched.

**DB evidence:**
```
1686 | Share Bunk With
  96 | BunkingNotes Notes
  90 | RetParent-Socializewithbest
   3 | Do Not Share Bunk With
```

### 3. Method display (cosmetic, caused by #1)

Not actually a data bug — methods like `age_preference` and `sibling_household_lookup` are correct, but the gray styling from lowercase status makes them look out of place.

## Proposed Fix

Normalize in `trace_collector.py` when writing to `debug_pipeline_summary`:
- **Status**: `.lower()` consistently (matches `bunk_requests` convention), update frontend to match lowercase
- **Source field**: Map display names → canonical codes using `SourceField` constants

### Files involved

**Backend:**
- `bunking/sync/bunk_request_processor/shared/constants.py` (SourceField definitions)
- `bunking/sync/bunk_request_processor/orchestrator/orchestrator.py` (lines 1488-1508)
- `bunking/sync/bunk_request_processor/debug/trace_collector.py` (lines 261-276)

**Frontend:**
- `frontend/src/components/pipeline-debug/PipelineBatchList.tsx` (lines 25-63, 109-131)

Issue #616, opened 2026-03-17
refactor(frontend): eliminate derived thresholdId state in GradeEligibilityConfig


adamflagg:
## Problem

`GradeEligibilityConfig` stores `thresholdId` in `useState` and syncs it via `useEffect` from `thresholdRecords?.[0]?.id`. This is a classic derived-state anti-pattern — two sources of truth for the same value, with risk of divergence during a render cycle.

The stale-state bug fixed in PR #614 (#576) was a direct symptom of this pattern: when `thresholdRecords` changed but the effect hadn't re-run yet, `thresholdId` held a stale value.

## Fix

Remove the `thresholdId` useState and useEffect entirely. In `handleSave`, read `thresholdRecords?.[0]?.id` directly to decide create vs update. This eliminates the entire class of stale-ID bugs.

## Context

Found during code quality review of PR #614.

FocusMode-coder:
Hey! I may be able to help with this bug.

I build custom fixes for Python projects and automation systems.

If you'd like help resolving this issue quickly feel free to reach out.

Happy to take a look 👍

PR #663, opened 2026-03-19
fix(frontend,sync): combined tech-debt cleanup — #619, #654, #658

### LINKED ISSUE #619: fix(frontend): remove 4 remaining defensive eslint-disable comments
## Context

PR #618 (#573 audit) removed 45 eslint-disable comments but kept 4 defensive guards that ESLint flags as `no-unnecessary-condition`. These are runtime-correct but the TypeScript types don't reflect the runtime nullability.

### AppLayout.tsx — 3 syncStatus guards (lines ~734, 737, 750)
```typescript
// eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- syncStatus may be undefined before query resolves
syncStatus?.bunk_assignments?.end_time
```
TS thinks `syncStatus` is always defined (destructured from React Query), but it can be `undefined` before the query resolves.

**Fix:** Widen the query hook return type to include `undefined`, or add a proper loading guard before the render block. ~30 min.

### useVelocityChartData.ts — 1 week_number null guard (line ~313)
```typescript
// eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- week_number may be null at runtime despite Required<> type
.filter((marker) => marker.week_number != null)
```
`Required<>` makes `week_number` non-nullable in the type, but the API can return `null`.

**Fix:** Remove `Required<>` wrapper or make `week_number` explicitly nullable in the response type. ~15 min.

## Total LoE
~45 min for all 4. Independent of each other and of #377 (which covers the other 4 kept comments about ExpandType).

### LINKED ISSUE #654: refactor(frontend): eliminate derived configId state in GradeEligibilityConfig rows
## Problem

`GradeEligibilityConfig` stores `configId` inside the `rows` state array (via `buildRows` → `useEffect` → `setRows`). This is the same derived-state anti-pattern that was just fixed for `thresholdId` in PR #652 (#616).

The `rows` state is rebuilt from `sessions` + `configRecords` query data inside a `useEffect`. If the query data refetches between when `rows` was last set and when `handleSave` runs, the `row.configId` values could be stale — pointing to a previous year's config records or records that were deleted/recreated.

This is the exact same class of bug that caused #576 (stale `thresholdId` leading to wrong create-vs-update decision).

## Current code

```typescript
// buildRows embeds configId into state (line 89-90)
const existing = configData?.find((r) => r.config_key === String(cmId))
configId: existing?.id,

// handleSave reads configId from stale rows state (line 151-155)
if (row.configId) {
  await pb.collection('config').update(row.configId, payload)
} else {
  await pb.collection('config').create(payload)
}
```

## Proposed fix

In `handleSave`, look up each row's config record ID directly from `configRecords` query data at save time, rather than reading from `rows` state:

```typescript
for (const row of rows) {
  const existingConfig = configRecords?.find((r) => r.config_key === String(row.cm_id))
  const payload = { /* ... */ }
  if (existingConfig?.id) {
    await pb.collection('config').update(existingConfig.id, payload)
  } else {
    await pb.collection('config').create(payload)
  }
}
```

This mirrors the fix applied to `thresholdId` in #616. The `configId` field can then be removed from the `SessionRow` interface and `buildRows`.

## Context

Found during code review of PR #652. Pre-existing issue, same pattern as #576/#616.

### LINKED ISSUE #658: fix(sync): inconsistent expand null-safety in attendee_repository session filtering
## Description

In `attendee_repository.py`, the `_bulk_get_sessions_chunk()` method (and previously `bulk_get_sessions_for_persons()`) uses an inconsistent pattern for accessing `item.expand`:

```python
session = getattr(item, "expand", {}).get("session") if hasattr(item, "expand") else None
```

If `item.expand` is `None` (attribute exists but is falsy), `getattr(item, "expand", {})` returns `None`, then `.get("session")` raises `AttributeError` since `None` has no `.get()`.

The safer pattern used elsewhere in the same file (`_get_session_cm_id`):
```python
expand = getattr(item, "expand", None)
session = expand.get("session") if expand else None
```

## Origin

Flagged by CodeRabbit on PR #646 but never addressed. Carried forward into `_bulk_get_sessions_chunk()` in PR #656.

## Fix

Replace the inline ternary with the two-step safe pattern matching `_get_session_cm_id`.

adamflagg:
## Summary
- **fix(sync):** Safe expand null-check in `attendee_repository.py` — `getattr(item, "expand", {}).get("session")` raises `AttributeError` when `item.expand` is `None` (#658)
- **fix(frontend):** Remove 4 remaining defensive eslint-disable comments with proper type fixes — 3 syncStatus guards in AppLayout, 1 week_number in useVelocityChartData (#619)
- **refactor(frontend):** Eliminate derived `configId` state in GradeEligibilityConfig rows — same pattern as #616/#617 thresholdId fix (#654)
- **docs:** Update issue triage — remove fixed issues, add new #653

Closes #619
Closes #654
Closes #658

Supersedes #659, #660, #661

## Test plan
- [x] Python unit tests pass (3187 tests) — new test for expand null-safety bug
- [x] Frontend tests pass (2397 tests) — new tests for configId fresh lookup at save time
- [x] TypeScript compiles clean
- [x] ESLint clean (4 fewer disable comments)
- [x] All pre-push hooks pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved grade eligibility configuration record lookup to more reliably determine when to update versus create entries.

* **Tests**
  * Added comprehensive test coverage for configuration handling edge cases and null data scenarios.

* **Refactor**
  * Enhanced defensive programming patterns with improved null checks and explicit type guards throughout the codebase.

### DIFF
--- a/frontend/src/hooks/useVelocityChartData.test.ts
+++ b/frontend/src/hooks/useVelocityChartData.test.ts
@@ -370,7 +370,7 @@ describe('useVelocityChartData', () => {
             phase: 'open',
             date: '2024-11-01',
             label: 'Open Reg',
-            week_number: null,
+            week_number: null as unknown as number,
           },
         ],
       })
--- a/frontend/src/hooks/useVelocityChartData.ts
+++ b/frontend/src/hooks/useVelocityChartData.ts
@@ -1,5 +1,5 @@
 import { useMemo } from 'react'
-import type { PhaseMarker, VelocityResponse } from '../types/velocity'
+import type { VelocityResponse } from '../types/velocity'
 import { resolveSessionAlias } from '../utils/sessionAliases'
 import {
   sortSessionDataByCampThenQuest,
@@ -308,14 +308,15 @@ export function useVelocityChartData(
   // Phase lines with week_number for X-axis positioning
   const phaseLines = useMemo(() => {
     if (!data?.phase_markers) return []
-    return data.phase_markers
-      .filter(
-        (marker): marker is PhaseMarker & { week_number: number } => marker.week_number != null
-      )
-      .map((marker) => ({
-        ...marker,
-        weekNumber: marker.week_number,
-      }))
+    return (
+      data.phase_markers
+        // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- week_number is int (non-nullable) per API schema, but defensive guard kept as belt-and-suspenders
+        .filter((marker) => marker.week_number != null)
+        .map((marker) => ({
+          ...marker,
+          weekNumber: marker.week_number,
+        }))
+    )
   }, [data?.phase_markers])
 
   // Phase day offsets for ReferenceArea bands on daily cumulative charts
--- a/frontend/src/types/velocity.ts
+++ b/frontend/src/types/velocity.ts
@@ -55,7 +55,7 @@ export interface PhaseMarker {
   phase: string
   date: string
   label: string
-  week_number: number | null
+  week_number: number
 }
 
 export interface SessionGenderBreakdown {


adamflagg:
### Code review

Found 2 issues:

1. `PhaseMarker.week_number` widened from `number` to `number | null`, but the Python API schema defines it as non-nullable `int` (no `Optional`, no `None` default). The backend `_build_phase_markers()` always assigns a non-null `int` from `_week_number()`. Widening the frontend type beyond the actual API contract creates a false impression that null is a valid response. The original eslint-disable in PR #618 deliberately kept the type as `number` to match the API contract, using the runtime guard as belt-and-suspenders.

https://github.com/adamflagg/kindred/blob/9cde0f327940722ed2cd36f0df10de09cdcf5b49/frontend/src/types/velocity.ts#L57-L59

2. The 3 syncStatus eslint-disable comments in `AppLayout.tsx` were explicitly audited and kept in PR #618 as "justified runtime-defensive guards" (React Query timing). This PR removes them and restructures the guards. While the replacement code is functionally correct and type-safe (`bunk_assignments` is non-optional on `SyncStatusResponse`), it reverses a deliberate prior decision without documenting why the assessment changed.

https://github.com/adamflagg/kindred/blob/9cde0f327940722ed2cd36f0df10de09cdcf5b49/frontend/src/layouts/AppLayout.tsx#L733-L755

PR #670, opened 2026-03-19
fix(frontend): enforce MetricsFilterOptions mutual exclusivity (#567, #562)

### LINKED ISSUE #562: Evaluate migrating all metrics hooks to full options objects
After #437 migrates positional-param hooks to the hybrid `(year, { ...options })` pattern, evaluate whether going further to full single-object params `({ year, sessionTypes?, ... })` across all metrics hooks would add meaningful value.

**Context:**
- `useVelocity` and `useForecast` already use hybrid `(year, params)` pattern
- `useRetentionTrends` and `useDrilldownAttendees` already use full options objects
- #437 will bring `useRetentionMetrics`, `useRegistrationMetrics`, `useHistoricalTrends`, `useWaitlistMetrics`, `useCancellationMetrics` to hybrid

**Questions to answer:**
- Does the inconsistency between hybrid and full-options hooks cause confusion?
- Would full-options enable better TypeScript discriminated unions (e.g., mutually exclusive `sessionCmId` vs `duration`)?
- Is the churn worth it given the hybrid pattern is already clear?

Low priority — only pursue if the answer to the above is clearly yes.

### LINKED ISSUE #567: Enforce mutual exclusivity of sessionCmId and duration in MetricsFilterOptions
Currently `MetricsFilterOptions` allows both `sessionCmId` and `duration` to be set simultaneously, and the hooks serialize both to query params when present. The API may handle this gracefully, but the TypeScript interface doesn't enforce the mutual exclusivity.

**Options:**
1. TypeScript discriminated union (`{ sessionCmId: number } | { duration: string } | {}`)
2. Runtime validation in hooks (throw if both set)
3. Leave as-is if the API handles it correctly

Pre-existing behavior — discovered during #437/#472 refactor (PR #566). Low priority — evaluate whether it causes real issues before investing.

adamflagg:
## Summary
- **#567**: `MetricsFilterOptions` is now a discriminated union type that prevents setting both `sessionCmId` and `duration` at compile time (backend already rejects with 422, UI already prevents it, now TypeScript catches it too)
- **#562**: `useComparisonMetrics` now accepts an optional `options?: MetricsFilterOptions` parameter, matching the hybrid `(year, options?)` pattern used by all other metrics hooks
- Added `metricsFilter()` builder function that narrows loose `number | null` / `string | undefined` context params into the discriminated union — updated all 7 call sites

## Test plan
- [x] 29 tests in `useMetrics.test.ts` (type enforcement, metricsFilter helper, hook source analysis)
- [x] All 2421 frontend tests pass
- [x] Zero TypeScript errors (`tsc --noEmit`)
- [x] ESLint clean (no errors)
- [ ] CI passes

Closes #567
Closes #562

## Summary by CodeRabbit

* **New Features**
  * Comparison and historical metric queries now accept and propagate consolidated filter options (session type, session ID, or duration).

* **Improvements**
  * UI pages and session context now expose and forward a single normalized filter object for consistent filtering and query keys.
  * Query key generation expanded to include filter inputs for more accurate cache differentiation.

* **Tests**
  * Added compile-time and runtime tests ensuring filter behavior and mutual-exclusivity rules.

### DIFF
--- a/frontend/src/contexts/MetricsSessionContext.tsx
+++ b/frontend/src/contexts/MetricsSessionContext.tsx
@@ -18,6 +18,7 @@ import { type ReactNode, useCallback, useMemo, useState } from 'react'
 import { useSearchParams } from 'react-router'
 import { useCurrentYear } from '../hooks/useCurrentYear'
 import { useMetricsSessions } from '../hooks/useMetricsSessions'
+import { metricsFilter } from '../hooks/useMetrics'
 import { MetricsSessionContext, type MetricsSessionContextType } from '../hooks/useMetricsSession'
 import {
   CAMP_SESSION_TYPES,
@@ -134,6 +135,17 @@ export function MetricsSessionProvider({ children }: { children: ReactNode }) {
   // Duration param for API calls
   const durationParam = selectedDuration ?? undefined
 
+  // Pre-built filter options for metrics hooks
+  const filterOptions = useMemo(
+    () =>
+      metricsFilter({
+        sessionTypes: sessionTypesParam,
+        sessionCmId: selectedSessionCmId,
+        duration: durationParam,
+      }),
+    [sessionTypesParam, selectedSessionCmId, durationParam]
+  )
+
   // Set duration filter (clears session and view params)
   const setSelectedDuration = useCallback(
     (duration: DurationCategory | null) => {
@@ -240,6 +252,7 @@ export function MetricsSessionProvider({ children }: { children: ReactNode }) {
       selectedDuration,
       setSelectedDuration,
       durationParam,
+      filterOptions,
       durationGroups,
       expandedRetention,
       setExpandedRetention,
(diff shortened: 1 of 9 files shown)


adamflagg:
### Code review

Found 2 issues:

1. `useComparisonMetrics` sends `session_cm_id` and `duration` query params to the backend `/api/metrics/comparison` endpoint, but that endpoint only accepts `year_a`, `year_b`, and `session_types`. FastAPI silently ignores unknown query params, so a future caller passing `sessionCmId` or `duration` options would get unfiltered data with no error — silent wrong results. Either add backend support for these params, remove them from the hook's query function, or add a code comment noting the backend gap.

https://github.com/adamflagg/kindred/blob/b438acda450b41946f7236699e0b3268bc417210/frontend/src/hooks/useMetrics.ts#L143-L149

2. The `metricsFilter` test "should prioritize sessionCmId over duration when both are non-null" documents behavior for input that the discriminated union type was specifically designed to prevent. The helper's loose input type allows it, but no typed caller can ever reach this code path. The test provides false confidence by specifying behavior for an impossible scenario — consider removing it or reframing as a defensive-coding comment in the implementation rather than a test assertion.

https://github.com/adamflagg/kindred/blob/b438acda450b41946f7236699e0b3268bc417210/frontend/src/hooks/useMetrics.test.ts#L187-L197

PR #904, opened 2026-04-11
refactor(graph): simplify nested try/catch in bunk bubble rendering

witharyank:
- Removed redundant nested try/catch in drawBunkBubbles
- Simplified error handling to a single try/catch block
- Preserved existing error tracking with failedBunks

Improves readability and aligns with unit bubble pattern.

## Summary by CodeRabbit

* **Refactor**
  * Simplified bubble rendering flow with centralized error handling, improving robustness of visual rendering and ensuring failed items are logged and tracked rather than aborting individual render attempts.

### DIFF
--- a/frontend/src/components/graph/bubbleRenderer.ts
+++ b/frontend/src/components/graph/bubbleRenderer.ts
@@ -257,34 +257,34 @@ export function drawBunkBubbles(
       const bunkName = bunksData?.[parseInt(bunkId, 10)] ?? `Bunk ${bunkId}`
       const bunkColor = getBunkColor(parseInt(bunkId, 10))
 
-      try {
-  const nodeIds = nodes.map((n) => `#${n.id()}`).join(', ')
-  const nodeCollection = cy.$(nodeIds)
+   try {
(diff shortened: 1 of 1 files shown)


adamflagg:
Thanks for the contribution! Unfortunately this PR has a structural bug introduced by the refactor that I need to flag before closing.

**Missing closing braces**

The `forEach` callback and the outer `if (showBunks)` block are never closed. The diff removes `})` and `}` that wrap the entire forEach body, which means everything from the `updateStatus` call through the event listener registration (`cy.on('pan zoom resize', updatePoppers)`) ends up running once per bunk group iteration instead of once after all bunks are processed. TypeScript's `tsc --noEmit` would catch this as a syntax error.

There's also a minor indentation issue on the `try` keyword (3 spaces vs the surrounding 6).

**Superseded by #905**

The same nested try/catch simplification was included as part of a broader cleanup PR (#905) which handles the braces correctly and also fixes three other issues. I'm going to close this one in favor of that work.

Appreciate you taking a look at the code!</body>

PR #976, opened 2026-04-23
fix(frontend): repair empty scenario creation and scenario copy

adamflagg:
## Summary
- **#17A** — "Start empty scenario" threw `Cannot read properties of undefined (reading 'session')`. Root cause: `pb.collection('saved_scenarios').create()` wasn't fetching the `session` expand, but `savedScenarioToScenario()` assumed `expand.session` existed. Fix: pass `{ expand: 'session' }` on create **and** harden the transform to tolerate missing expand (defense in depth).
- **#17B** — Copying a scenario produced an error and only persisted ~2/3 of the assignments (verified via local DB: a test scenario had 97/147 assignments = 66%, with loss uniform across every bunk and bunk_plan — not a filter bug). Root cause: `copyScenarioToScenario` fired all creates in parallel via `Promise.all(createPromises)` — SQLite write contention caused ~50 creates to fail before `Promise.all` rejected, with the earlier successes already persisted. Fix: rewrote to mirror the existing `copyProductionToScenario` pattern in the same file — sequential `for...of await`, per-item try/catch, error accumulator, aggregate throw.
- **Refactor** — Extracted `savedScenarioToScenario` to its own module (`frontend/src/contexts/scenarioTransform.ts`) so the context file contains only the React component, removing an `eslint-disable react-refresh/only-export-components` workaround.

## Test plan
- [ ] Start an empty scenario — no error thrown, scenario loads
- [ ] Copy a scenario with 100+ campers — all assignments persist, no error
- [ ] Compare row counts in `bunk_assignments_draft` for source vs copy — should match exactly
- [ ] Existing scenario flows (start from CampMinder production) still work

### DIFF
--- a/frontend/src/contexts/ScenarioContext.tsx
+++ b/frontend/src/contexts/ScenarioContext.tsx
@@ -4,35 +4,11 @@ import { ScenarioContext, type Scenario, type ScenarioContextType } from '../hoo
 import { useSavedScenarios } from '../hooks/useSavedScenarios'
 import { useCreateScenario, useDeleteScenario } from '../hooks/useSavedScenariosMutation'
 import { useUpdateScenario, useClearScenario } from '../hooks/useScenarioOperations'
-import type { SavedScenario } from '../types/app-types'
 import { useYear } from '../hooks/useCurrentYear'
+import { savedScenarioToScenario } from './scenarioTransform'
 
 interface ScenarioProviderProps {
   children: ReactNode
-}
-
-// Convert SavedScenario to Scenario format
-// Exported for direct unit testing. Must tolerate records created without
-// `{ expand: 'session' }` (e.g. older flows) — missing expand must not throw.
-// eslint-disable-next-line react-refresh/only-export-components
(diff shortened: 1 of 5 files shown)


adamflagg:
### Code review

Reviewed:
- Bug A fix: `{ expand: 'session' }` correctly passed to `create()`; `savedScenarioToScenario` hardened against missing expand with documented rationale
- Bug B fix: sequential `for...of await` with per-item try/catch mirrors the existing `copyProductionToScenario` pattern; intentional parity
- Refactor: clean extraction to `scenarioTransform.ts` removes the `eslint-disable react-refresh/only-export-components` workaround correctly
- Tests cover both bugs (TDD-style: concurrency probe verifies max-in-flight = 1, error accumulation test verifies all items attempted after partial failure)
- The `updateScenario` path in `useScenarioOperations.ts` does not request `expand: 'session'` and would return `session_cm_id: 0` from `savedScenarioToScenario`, but this is pre-existing and outside the diff scope; the new defensive fallback prevents the crash that was the root cause

PR #977, opened 2026-04-23
fix(api,frontend): source social graph from active scenario and cache per-scenario

adamflagg:
## Summary
- **#30** — The social-network graph was always rendering CampMinder production bunk/session assignments, ignoring the active scenario. When viewing a scenario with campers bunked but no corresponding production assignments, the graph rendered nothing.
- **Root cause:** `useSocialGraphData` → `socialGraph` service → `GET /api/sessions/{id}/social-graph` → `OptimizedSocialGraphBuilder` all operated without a `scenario_id`. The builder hardcoded its query to the `bunk_assignments` (CampMinder) collection.
- **Fix (commit 1):** Thread `scenario_id` end-to-end. The hook pulls `currentScenario` from `ScenarioContext` (same pattern as `ValidateBunkingButton`). When `scenario_id` is present, the builder queries `bunk_assignments_draft` filtered by scenario. Production path is unchanged when no scenario is active.
- **Per-scenario caching (commit 2):** The prior commit bypassed both `GraphCacheManager` (backend) and `GraphCacheService` (frontend) when a scenario was active, because those caches were keyed only by `(session_cm_id, year)` and would otherwise mix prod/scenario results. Re-keyed both caches with a `scenario_id` slug (`"prod"` for production, scenario id otherwise). Existing prefix-based invalidation in both caches handles the new keys without modification.

## Test plan
- [ ] Open a scenario that has campers bunked (and where production has nobody bunked for the same session) — graph renders using the scenario's assignments
- [ ] Switch between two scenarios repeatedly — second and subsequent views hit the cache (no full rebuild)
- [ ] Production graph (no scenario active) still renders exactly as before
- [ ] Moving a camper invalidates both prod and scenario cache entries for that session

## Release Notes

* **New Features**
  * Social graph endpoints now accept optional scenario parameters to reflect draft bunking assignments
  * Frontend social graph components automatically incorporate active scenario context when fetching and caching data

* **Tests**
  * Added comprehensive test coverage for scenario-aware caching, data sourcing validation, and cache invalidation across both frontend and backend

### DIFF
--- a/api/routers/social_graph.py
+++ b/api/routers/social_graph.py
@@ -59,6 +59,10 @@ async def get_session_social_graph(
     include_historical: Annotated[bool, Query(description="Include historical data")] = False,
     layout: Annotated[str, Query(description="Layout algorithm: force, circle, hierarchical")] = "force",
     edge_types: Annotated[str | None, Query(description="Comma-separated edge types to include")] = None,
+    scenario_id: Annotated[
+        str | None,
+        Query(description="Scenario ID — when set, source bunk assignments from bunk_assignments_draft"),
+    ] = None,
     user: AuthUser = Depends(get_current_user),
 ) -> SocialGraphResponse:
     """Get the full social graph for a session using NetworkX analysis.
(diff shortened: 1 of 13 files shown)


adamflagg:
### Code review

Found 1 issue:

1. Bunk graph endpoint silently ignores the new `scenario_id` parameter on `get_bunk_graph`/`cache_bunk_graph`

This PR adds `scenario_id` support to `GraphCacheManager.get_bunk_graph` and `cache_bunk_graph`, but the bunk-level router endpoint and frontend service are not updated to pass it through. When a scenario is active, the session-level graph correctly sources bunk assignments from `bunk_assignments_draft`, but switching to the bunk subgraph view silently falls back to production assignments. No code comment marks this gap as intentional.

Router (line 346, no scenario_id passed):
https://github.com/adamflagg/kindred/blob/d90c191f93e67d48a8669ec63db597d9e090dead/api/routers/social_graph.py#L345-L360

Frontend service (no scenarioId parameter):
https://github.com/adamflagg/kindred/blob/d90c191f93e67d48a8669ec63db597d9e090dead/frontend/src/services/socialGraph.ts#L41-L56

PR #1016, opened 2026-04-24
fix(frontend): guard LayoutWorker against stale Cytoscape instance

adamflagg:
## Summary

- Adds instance-token guard: each layout job carries a `token` (issued by `makeLayoutToken()`) that is stored in `layoutTokenRef` and echoed back by the worker. The `onmessage` handler calls `applyLayoutPositions()` which silently drops any response whose token ≠ `layoutTokenRef.current`, preventing stale messages from reaching a replaced cy instance.
- Adds `cy.destroyed()` guard inside `applyLayoutPositions` for the narrow window between a matching token and a torn-down instance.
- Adds `cy.destroyed()` guard in the `onLayoutComplete` 500ms timer, which prevents `adjustLabelPositions` / `drawBunkBubbles` from running against a cy that was replaced while the timer was pending.
- Extracts guard logic to `layoutWorkerGuards.ts` (pure functions) so it can be unit-tested without mounting the full component.

Closes scoreboard item #54

## Root cause

`SocialNetworkGraph` destroys and recreates the Cytoscape instance every time `graphData`, `viewMode`, `bunksData`, `showEdges`, or `showLabels` changes (destroy/recreate pattern confirmed by Step 0 grep). The `LayoutWorker` is long-lived and reused. When a new render starts before the in-flight worker result arrives, the old cy is destroyed but the worker message fires `cy.batch()` on the closed-over (now-destroyed) `cy` variable, causing:

```
TypeError: Cannot read properties of null (reading 'notify')
    at id.endBatch → Array.forEach → notify
```

followed by bubble creation failing with "Cytoscape instance is not valid".

## Test plan

**Automated (Vitest — 9 new tests in `layoutWorkerLifecycle.test.ts`):**
- [ ] `makeLayoutToken` issues unique positive-integer tokens
- [ ] `isStaleLayoutMessage` returns correct boolean
- [ ] `applyLayoutPositions` applies positions when token matches and cy is alive
- [ ] `applyLayoutPositions` drops stale-token result (cy.batch not called)
- [ ] `applyLayoutPositions` drops result when cy is destroyed (token matches)
- [ ] `applyLayoutPositions` treats missing token as stale
- [ ] `applyLayoutPositions` is a no-op for non-`positions` message types

## Manual validation (dev/preview)

1. Log in and navigate to the Social Network Graph page (any session with camper data).
2. Open the browser developer console.
3. **Rapid control toggles**: Quickly toggle Unit Bubbles, Bunk Bubbles, Locked-only, and Ego-filter controls in rapid succession while the layout is computing.
4. **Rapid scenario switching**: Have 2+ scenarios available. Switch between them quickly (before the spinner disappears).
5. **Repeat 3+4 together**: Toggle a control immediately after switching scenarios.

**Expected (after fix):**
- No `TypeError: Cannot read properties of null (reading 'notify')` in console.
- No `Cytoscape instance is not valid, cannot create bubbles` errors.
- Bunk/unit bubbles render correctly after the final scenario/control settles.
- Nodes do not collapse into a tight cluster (layout applied successfully).

**Expected (before fix):** The above errors appeared reproducibly with fast toggling.

### DIFF
--- a/frontend/src/components/SocialNetworkGraph.tsx
+++ b/frontend/src/components/SocialNetworkGraph.tsx
@@ -46,8 +46,9 @@ interface SocialNetworkGraphProps {
   sessionCmId: number
 }
 
-// Import worker types
+// Import worker types and lifecycle guards
 import type { LayoutWorkerOutput } from '../workers/layoutWorker'
+import { makeLayoutToken, isStaleLayoutMessage } from '../workers/layoutWorkerGuards'
 
 // Module-level constant: request + sibling edges are always-on (not user-toggleable).
 // Hoisted outside the component so the reference is stable across renders —
@@ -68,6 +69,10 @@ export default function SocialNetworkGraph({ sessionCmId }: SocialNetworkGraphPr
   // into a new position shortly after labels appear).
   const hasMountedExpandRef = useRef(false)
   const hasMountedBubblesRef = useRef(false)
+  /** Monotonically-increasing token issued per layout job.
+   *  The worker echoes it back; stale responses (old token ≠ current) are dropped
+   *  before they can call cy.batch() on a destroyed instance. */
+  const layoutTokenRef = useRef<number>(0)
   useYear() // Ensure year context is available
 
   // Create refs object for bubble rendering - memoized to avoid recreation on every render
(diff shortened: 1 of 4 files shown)


adamflagg:
### Code review

Reviewed:
- Token lifecycle guards in `layoutWorkerGuards.ts`: token comparison and `cy.destroyed()` checks are correctly ordered and cover the primary crash path (`endBatch → null.notify`)
- `handleMessage` in `SocialNetworkGraph.tsx`: the `positions` path delegates to `applyLayoutPositions` (both guards applied); the `error` path's token check is a net improvement over the pre-PR code, and the missing `cy.destroyed()` guard on that branch is pre-existing
- Module-level `_tokenCounter` is safe: `SocialNetworkGraph` is only ever mounted once (via `FriendGroupsView`)
- Test coverage in `layoutWorkerLifecycle.test.ts` covers all four `ApplyResult` discriminants including the legacy-no-token case

Issue #1040, opened 2026-04-27
fix(graph): node circles don't reflect live request status (stale until remount)


adamflagg:
## Problem

The Social Network Graph's node-border colors (request satisfaction status — `satisfied` / `isolated` / `no_requests`) are derived server-side and cached under React Query key `['social-graph', sessionCmId, year, scenarioId]` (see `frontend/src/hooks/useSocialGraphData.ts`).

This query key is **never invalidated** by any mutation path. So after staff approve/decline a request, drag-drop a camper, or run the solver, the graph circles stay on their pre-mutation colors until either:

- the user navigates away and back (route remount triggers refetch), or
- React Query's default `staleTime` expires.

This is the graph-side analog of issue #48 / PR #1017 — that fix added broad invalidation of `['all-bunk-requests']` so the bunking-board card alert (yellow triangle) updates live. The social graph was not in scope for #48 and remains stale.

## Repro (preview/dev)

1. Open the Social Network Graph for a session that has at least one camper with a red "None" border (has requests, none satisfied).
2. Without leaving the page, go to the Requests tab in another window/tab and approve that camper's request, OR drag them into the same cabin as their requestee on the bunking board.
3. Switch back to the graph — the border is still red.
4. Refresh the page — the border updates correctly.

## Suggested fix

Mirror PR #1017's pattern: add `queryClient.invalidateQueries({ queryKey: ['social-graph'] })` to the same mutation paths that #1017 already touches:

- `useCamperMovement` (drag-drop, both null-response and normal branches)
- `useSolverOperations` (auto-apply + legacy-confirm blocks)
- `RequestReviewPanel` (single + bulk approve/decline)
- Merge/Split request modals (already invalidate `all-bunk-requests` — extend to `social-graph`)

After-confirm (not optimistic) style, matching #987/#1017.

## Notes

- The graph's satisfaction calc is recalculated server-side in `bunking/graph/social_graph_builder.py` (separate code path from the bunking-board alert). Invalidation triggers a server refetch + re-render — that's the right shape here.
- Bunk-level subgraph cache (`['bunk-social-graph']`) likely has the same problem; include it in the invalidation list.

## Manual validation (dev/preview)

- [ ] Approve a request from the Requests tab → graph node border updates without refresh
- [ ] Drag a camper to a different bunk on the board → graph border updates without refresh
- [ ] Run solver auto-apply → all newly-satisfied campers' borders flip green without refresh
[...]

PR #1017, opened 2026-04-24
fix(frontend): invalidate camper-card alerts on request + assignment mutations

adamflagg:
## Summary

The "none satisfied" yellow-triangle alert on camper cards is **client-derived** from the `['all-bunk-requests', sessionCmId, year]` React Query cache, with satisfaction computed by **bunk membership** (`personSet.has(req.requestee_id)`) — not by request status.

After dragging campers together (or running the solver) the membership changed, but the cache wasn't invalidated, so the alert stayed stale until a page reload. Fix: invalidate `['all-bunk-requests']` in the two assignment-mutating paths.

Closes scoreboard item #48.

## Investigation summary

**Alert derivation:** Client-derived. `BunkRequestProvider` (`providers/BunkRequestProvider.tsx`) fetches bunk requests under `['all-bunk-requests', sessionCmId, year]` (staleTime 1 min). `CamperCard` calls `getSatisfiedRequestInfo()` which filters by `personSet.has(req.requestee_id)` — pure bunk-membership check. Status field is not read.

**Mutation paths touched and query key invalidated on each:**

| Path | File | Why it needs invalidation |
|------|------|--------------------------|
| Drag-drop move (both null-response and normal branches) | `hooks/session/useCamperMovement.ts` | Bunk membership changes |
| Solver apply (auto-apply block) | `hooks/session/useSolverOperations.ts` | Bunk membership changes |
| Solver apply (legacy confirm block) | `hooks/session/useSolverOperations.ts` | Bunk membership changes |

**Paths that already invalidated `['all-bunk-requests']` (no change needed):**
- `MergeRequestsModal` — changes request count / merged_into
- `SplitRequestModal` — changes request count

**Paths NOT invalidated (intentional — do not affect the alert):**
- Single approve/decline (`updateRequestMutation` in `RequestReviewPanel.tsx`) — only mutates `status`, which the alert doesn't read
- Bulk approve/decline (`bulkUpdateMutation` in `RequestReviewPanel.tsx`) — same
- `useScenarioOperations.useClearScenario` / `copyScenarioToScenario` / `useDeleteScenario` — don't change request records

## Test plan

3 Vitest tests in `frontend/src/hooks/session/__tests__/alert-invalidation.test.ts` covering drag-drop (with and without `changed:false` response), solver-apply, and a regression guard for the already-correct merge/split pattern. Full suite: green.

## Manual validation (dev/preview)

1. Log in, go to bunking board for a session.
[...]

PR #1042, opened 2026-04-27
refactor(api): delete OptimizedSocialGraphBuilder _calculate_node_metrics override

adamflagg:
## Summary

`OptimizedSocialGraphBuilder._calculate_node_metrics` overrode the parent and silently dropped the **component-size** and **satisfaction-status** phases since the initial commit. The override only added `round(centrality, 2)` (cosmetic — sub-pixel size delta in the frontend; no consumer depends on the rounding) and `nx.clustering(graph.to_undirected())` (no consumer reads the `clustering` field — declared in TS but never rendered or used in logic).

Meanwhile `OptimizedSocialGraphBuilder` is the GUI path: 3 of 4 production callers in `api/routers/social_graph.py` instantiate it (the friends-page social graph endpoint among them). The parent is only invoked from the ego-network endpoint, which doesn't surface `satisfaction_status` to the frontend at all.

Net effect: `satisfaction_status` was never populated on any node the GUI rendered, and every border fell back to a single default color.

This PR **deletes the override**. The optimized class now inherits the parent's full four-phase metric calculation. The parent retains its defensive `is_directed()` branch (using `nx.weakly_connected_components` for directed graphs) so it remains DiGraph-safe for any future direct caller.

## What changed vs the original diagnosis

The first attempt at this fix (earlier commits on this branch) patched the parent's `_calculate_node_metrics` to handle directed graphs, on the theory that `nx.connected_components` was raising and the exception was being swallowed. That was wrong — no exception ever fired on the GUI path because the override never invoked the parent body. The parent fix is still useful as defense in depth, but the actual user-visible bug needed the override deleted.

## Why no override is needed

- **Rounding**: frontend uses `centrality` only as `10 + centrality * 40` for node-pixel size. Full-precision floats produce visually indistinguishable sizes (sub-pixel delta).
- **`to_undirected()` clustering**: `clustering` is declared on the TS type but no UI element reads it. The directed-vs-undirected formula difference is invisible.

The override was solving a problem nobody had while inadvertently creating the satisfaction-status bug.

## TDD

Tests in `tests/unit/bunking/graph/test_satisfaction_status.py` were retargeted at `OptimizedSocialGraphBuilder` directly so the regression anchor exercises the actual production class:

- `test_optimized_builder_sets_satisfaction_status` — every node gets a non-null `satisfaction_status` on the optimized builder.
- `test_optimized_builder_classifies_correctly` — same `satisfied` / `isolated` / `no_requests` semantics on the DiGraph as the parent produces on a Graph.
- `test_parent_builder_on_digraph_still_works` — defensive: the parent's directed-graph branch stays covered.

Red phase verified: `test_optimized_builder_sets_satisfaction_status` failed before the override was deleted with `'satisfaction_status' missing — OptimizedSocialGraphBuilder override is dropping the satisfaction loop`.

Full unit-test sweep: **3687 passed, 23 skipped (all pre-existing skip reasons)**.

## Manual validation (dev/preview)

Spin up the worktree's dev server (`cd ~/kindred-worktrees/fix-graph-satisfaction-payload && ./scripts/start_dev.sh`) and walk through:

1. Open the friends graph for any session that has bunk requests.
2. **Expected before**: every node border renders the same dark navy color.
3. **Expected after**: borders differentiate per the legend — green for `satisfied` (≥1 request landed in the cabin), red for `isolated` (has requests, none satisfied), gray for `no_requests` (no request edges at all).
4. Switch sessions; legend should match the borders.
5. Move a camper into the same cabin as their requestee → after the existing #1017 invalidation fires, the border should flip color (unrelated to this fix but confirms staleness path still works).
6. Layout, node positions, sizes, and edges should be unchanged compared to current main — only border colors change.

Closes scoreboard #36 / issue #1040.

### DIFF
--- a/bunking/graph/optimized_graph_builder.py
+++ b/bunking/graph/optimized_graph_builder.py
@@ -283,18 +283,6 @@ class OptimizedSocialGraphBuilder(SocialGraphBuilder):
 
         return sibling_edges
 
-    def _calculate_node_metrics(self) -> None:
-        """Calculate centrality and clustering coefficients with rounding."""
-        # Calculate degree centrality
-        centrality = nx.degree_centrality(self.graph)
-        for node in self.graph.nodes():
-            self.graph.nodes[node]["centrality"] = round(centrality.get(node, 0.0), 2)
-
-        # Calculate clustering coefficient
-        clustering = nx.clustering(self.graph.to_undirected())
-        for node in self.graph.nodes():
-            self.graph.nodes[node]["clustering"] = round(clustering.get(node, 0.0), 2)
-
     def update_node_position(
         self, person_cm_id: int, new_bunk_cm_id: int, session_cm_id: int, year: int
     ) -> dict[str, Any]:
--- a/bunking/graph/social_graph_builder.py
+++ b/bunking/graph/social_graph_builder.py
@@ -816,8 +816,15 @@ class SocialGraphBuilder:
         clustering = nx.clustering(self.graph)
         nx.set_node_attributes(self.graph, clustering, "clustering")
 
-        # Connected component size
-        components = list(nx.connected_components(self.graph))
+        # Connected component size — use weakly_connected_components for directed graphs
+        # (OptimizedSocialGraphBuilder uses nx.DiGraph; the parent path uses nx.Graph).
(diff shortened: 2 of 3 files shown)


adamflagg:
@coderabbitai review

PR #1066, opened 2026-04-29
refactor(sync): clear() builtin + modernization framework expansion

adamflagg:
## Summary

Two related changes shipped together. The first commit implements rank #1 of the Go modernization execution order; the second commit applies a review pass on the modernization framework itself (backlog + prompt docs) — same files, tightly coupled.

## Commit 1: `clear()` builtin refactor

Implements rank #1 of `docs/reference/modernization-backlog.md` §2c.

```go
// Before
for k := range b.ProcessedKeys {
    delete(b.ProcessedKeys, k)
}

// After
clear(b.ProcessedKeys)
```

Same semantics, one runtime intrinsic instead of N hash lookups. Adds locking tests for both `ClearProcessedKeys` and `ClearFieldDiffStats` (no prior coverage). Tests written first, verified passing on the old implementation, re-verified after the rewrite.

Also annotates the `strings.Index() != -1 → strings.Contains` row in §2a as a **FALSE POSITIVE** — the audit pattern-matched on the regex without inspecting the body, but the single hit at `sync/base_sync.go:1084` uses `idx` as a slice boundary 7 lines later. `strings.Contains` would discard a value the next 7 lines depend on.

## Commit 2: modernization framework review fixes

Applies 14 findings from a review pass on the framework docs:

**Backlog edits** (`modernization-backlog.md`):
- Execution-model preamble: each section governs its own order; no cross-section ranking
- Format-note banners on §1 (Python), §3 (Frontend) noting they predate the §a/§b/§c redo
- Note on §4 (Infrastructure): hardening checklist, not a language-version audit
- §2b: ungate `testing/synctest` (1.25 GA shipped, toolchain is 1.26); rephrase `encoding/json/v2` gate to "ecosystem signal"
- §2c: extract retired false-positive to its own subsection; renumber live items 1–16; mark #1 (`clear()`) as ✓ shipped; ungate #15 (synctest); reword #16 gate
- Drop "Suggested overall PR order" (conflicted with §2c, had a dual-`4.` numbering bug)

**Prompt consolidation** — replaces `modernization-coverage-prompt.md` + `modernization-execution-prompt.md` with a single `modernization-prompts.md`:
- Two parts (A: build/upgrade a section, B: execute row-by-row), shared rules section, per-language version-feature appendix
- Promotes the false-positive lesson out of a footnote into "Rule 1: Verify in context, not against the regex"
- Adds "Rule 2: Three independent versions" (toolchain/installed/idiom triad) and "Rule 3: Tests are the spec" (TDD clarified per row: behavioral rows = failing test first; pure rewrites = existing tests + pre-push)
- Reorganizes Part A so the version-feature appendix is the *tool* used by Steps 1–3, not an afterthought
- Adds upgrade-in-place mode with explicit preserve rules (✓ shipped markers, retired subsection, skipped reasons) replacing the destructive "replace it" instruction
- Adds `skipped (reason)` row state and section-done criterion to Part B
- Fixes `npx tsc -v` → `cd frontend && npx tsc -v`

## Test plan

- [x] `go test ./sync/ -run 'TestBaseSyncService_Clear' -v` — 2 passed
- [x] `go test ./sync/` — 1995 passed
- [x] `lefthook run pre-push` (clear() commit) — ruff-check, shellcheck, pb-js-lint, go-build, mypy, type-check all green
- [x] `lefthook run pre-push` (docs commit) — all hooks correctly skipped (no code-triggering files)

## Summary by CodeRabbit

* **Documentation**
  * Reworked the modernization backlog to use per-section execution ordering, reorganized the Go rewrite matrix, added precision for artifact types, and removed the global PR-order list.
  * Added a new modernization guidance page detailing an end-to-end process, audit rules, execution loop, PR conventions, and language-specific checklists.

* **Refactor**
  * Simplified internal map-reset implementation to a clearer reset behavior.

* **Tests**
  * Added tests verifying map-reset behavior remains safe and reusable after clearing.

### DIFF
--- a/docs/reference/modernization-backlog.md
+++ b/docs/reference/modernization-backlog.md
@@ -62,34 +62,78 @@ Baseline is strong: Pydantic v2 everywhere, `datetime.UTC` adopted, modern gener
 
 ## 2. Go (project is on 1.26)
 
-Baseline is excellent: 784 `slog` uses, 395 `fmt.Errorf("%w", ...)` wrappings, no `ioutil`. The dominant gap is an `interface{}` → `any` codemod the codebase never got.
+**Toolchain:** `pocketbase/go.mod` declares `go 1.26.0`; locally installed `go1.26.0`. 1.26 is the current stable as of April 2026 — already on latest, no version bump pending.
 
-| Impact | Feature | Where | Count | Notes |
-|--------|---------|-------|-------|-------|
-| **HIGH** | `interface{}` → `any` | `sync/api.go` (108 occurrences), `sync/base_sync.go` (42), `sync/households.go` (35+), widespread | **744** | Single mechanical pass. Alias since Go 1.18. |
-| **HIGH** | `map[string]interface{}` → `map[string]any` | same hotspots | **603** | Same PR as the `any` codemod. |
-| **HIGH** | `fmt.Sprintf` inside `slog` call sites | `campminder/client.go:1008`, `sync/financial_transactions.go:106`, `sync/orchestrator.go:784,844` | 4 | Defeats structured logging — pass raw values as key/value attrs instead. |
-| **HIGH** | String-based error matching | `sync/rate_limited_sheets_writer.go` (`strings.Contains(err.Error(), "googleapi: Error 429")`), `sync/scheduler_test.go`, `workbook_manager_test.go`, `orchestrator_test.go` | 4 | Define sentinel errors (e.g. `ErrRateLimited`) and use `errors.Is` / `errors.As`. The Google 429 detection is the production-impact one. |
-| **MEDIUM** | `sort.Slice` → `slices.SortFunc` (Go 1.21) | `sync/normalize_geographic_test.go`, `sync/workbook_manager.go`, `sync/multi_workbook_ordering.go` (2×) | 3 | |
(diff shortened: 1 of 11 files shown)


adamflagg:
### Code review

Found 1 issue:

1. **Broken `grep -E` alternation pattern in modernization-prompts.md** — Line 78 uses `\|` inside `-E` (extended regex), which is treated as a literal pipe character, not the alternation operator. The command silently matches nothing and the table entry produces no output when agents run it.

https://github.com/adamflagg/kindred/blob/bcf0e553c7faaa39b885327f88d2dc72d3892b0f/docs/reference/modernization-prompts.md#L76-L80

Fix: change `'^(go\|toolchain) '` to `'^(go|toolchain) '`.

Verified by running both forms against a sample `go.mod` — the escaped form returns nothing; the unescaped form correctly matches both the `go` and `toolchain` lines.

---

Also noted (below threshold, not blocking):

- `modernization-backlog.md:134` marks rows as `✓ shipped #1066` while the PR is still open. This is self-referential and will be accurate at merge time, but CodeRabbit flagged it as a minor doc-accuracy concern.
- `TestBaseSyncService_ClearFieldDiffStats` (base_sync_test.go) omits the repeat-clear idempotency check that `TestBaseSyncService_ClearProcessedKeys` includes. Minor asymmetry; `clear()` on an empty map is a documented Go no-op so this isn't a behavioral gap.

PR #1087, opened 2026-04-30
fix(frontend): gate CSV pipeline indicator on actual upload context

### LINKED ISSUE #1085: bug(frontend): CsvPipelineIndicator stuck on 'Resolving camper names and calculating requests'; also wraps and uses wrong font size
## Symptoms (prod, observed today)

1. The header indicator is stuck on **"Resolving camper names and calculating requests"** with a spinner. There has been no recent activity in the prod container logs to suggest matching is actually running.
2. The visual side-effects of the long indicator text:
   - The "Assignments synced …" and "Requests synced …" badges in the secondary nav wrap onto two lines.
   - The indicator text appears in a larger font than the rest of the secondary nav row.

## Root cause hypothesis — no terminal state for `matching`

`derivePhase()` in `frontend/src/services/csvPipelineStatus.ts:33-64` decides the UI phase from two inputs:

- `bunk_requests` sync job status (from `/api/custom/sync/status`)
- The latest `debug_pipeline_runs` row (created by the matching/trace collector)

```ts
if (sync.status === 'completed') {
  if (debugIsFresh && debug) return doneFromDebug(debug)
  return { phase: 'matching', startedAt: sync.startedAt }
}
```

When the sync finished cleanly (`status=completed`, `end_time` set) but **no debug_pipeline_runs row exists with `created >= sync.startedAt`**, the function returns `phase: 'matching'` indefinitely. The hook polls every 2s (`useCsvPipelineStatus.ts`) waiting for a fresh debug row that may never arrive.

`debug_pipeline_runs` is written by `bunking/sync/bunk_request_processor/debug/trace_collector.py:207-239` in `flush()`. Notably:

```py
if not self._traces:
    return ""
```

If the matching step short-circuits (no traces collected — e.g., nothing to process, processor disabled, or the run threw before flushing), no row is ever created, and the indicator remains stuck in `matching` forever. There is no client-side timeout and no server-side "matching failed" signal.

## Suggested fix direction

- **Add a terminal timeout** to `derivePhase()` for the `matching` branch — if `now - sync.startedAt > N` (e.g. 5–10 min) and still no fresh debug row, fall through to either `done` (with a "no traces written" note) or `error`.
- **Or** have the request processor write a debug_pipeline_runs row even on the empty/no-op path so the UI can derive a real terminal state.
- Consider distinguishing "matching is genuinely running" from "matching is expected to follow" — the current code can't tell, because `sync.status=completed` covers both.

## UI bugs in `CsvPipelineIndicator.tsx` and `AppLayout.tsx`

`frontend/src/components/CsvPipelineIndicator.tsx:59` — the indicator button uses `text-sm`, while the surrounding secondary-nav badges in `frontend/src/layouts/AppLayout.tsx:711` are `text-xs`. The size mismatch is what makes the indicator look "wrong font size" relative to its neighbors.

`frontend/src/layouts/AppLayout.tsx:697-738` — the secondary nav row has no `flex-wrap` / `min-w-0` / truncation handling. When the indicator's right-side copy is long (e.g. "Resolving camper names and calculating requests"), the synced-badges container gets squeezed and wraps. Options:
- Tighten the indicator text (e.g. "Matching CSV requests…").
- Add `whitespace-nowrap` + truncation to the synced badges, or `min-w-0` + `truncate` on the right side.
- Drop indicator text size to `text-xs` to match the row.

## Code paths

- `frontend/src/components/CsvPipelineIndicator.tsx`
- `frontend/src/services/csvPipelineStatus.ts:33-64` (`derivePhase`)
- `frontend/src/hooks/useCsvPipelineStatus.ts`
- `frontend/src/layouts/AppLayout.tsx:697-745`
- `bunking/sync/bunk_request_processor/debug/trace_collector.py:207-239` (`flush`)

adamflagg:
## Summary

Closes #1085. The CSV pipeline indicator was stuck on **"Resolving camper names and calculating requests"** after every nightly `bunk_requests` cron run, with the long copy also pushing the "Assignments synced / Requests synced" badges onto two lines and rendering at the wrong font size.

### Root cause

`derivePhase()` treated any completed `bunk_requests` sync without a fresh `debug_pipeline_runs` row as "matching in progress." The nightly cron runs that sync on its own — no CSV import, no matching/processor step that writes a debug row — so the indicator polled forever for a row that never arrives. The same shape applied to imports where matching crashed or short-circuited (empty traces never get flushed).

### Fix

- **Gate every non-idle phase on a CSV upload context.** `BunkRequestsUpload` writes a localStorage marker (`csvUploadStartedAt`) in `onMutate` before the upload fetch starts, clears it on `onError`. The hook reads the marker and passes it to `derivePhase`, which only attributes a sync to a CSV upload when the marker is within 10 min of `sync.startedAt`. Cron `bunk_requests` runs are now invisible to the indicator.
- **Stuck-matching safety net.** If a CSV-upload-triggered sync finishes but no debug row arrives within 10 min, surface as `error` rather than spin indefinitely.
- **UI cleanup.** Indicator copy is now `text-xs` (was `text-sm`) to match the secondary nav row. Shorter strings ("Importing CSV…" / "Matching CSV requests…") stop squeezing the synced badges. The synced spans get `whitespace-nowrap` so they no longer break onto two lines.

### Test plan

- [x] Unit tests added for the new gating: cron returns idle for running/completed/failed, CSV-context flow returns importing → matching → done → error correctly, stuck-matching safety net transitions to error.
- [x] Hook tests cover: idle without context, importing with context, idle when context is missing on a running sync, done with fresh debug row + context.
- [x] Existing grace-window boundary tests retained (now passing CSV context).
- [x] Pre-push verification: `vitest` (3308 / 3308 passed), `eslint`, `tsc`, prettier — all green.
- [ ] Manual: upload a CSV in dev → indicator shows Importing → Matching → Done. Trigger a fake cron-shape sync (no localStorage marker) → indicator stays idle.

## Summary by CodeRabbit

* **New Features**
  * CSV uploads now display separate status from automatic background syncs

* **Bug Fixes**
  * Fixed status text from wrapping across multiple lines
  * Added timeout detection for stalled import operations to prevent indefinite loading

* **UI/Style**
  * Streamlined pipeline status display with more compact icons and clearer labels

### DIFF
--- a/frontend/src/components/BunkRequestsUpload.tsx
+++ b/frontend/src/components/BunkRequestsUpload.tsx
@@ -2,7 +2,7 @@ import { type ChangeEvent, useState, useRef } from 'react'
 import { Upload, Loader2, FileText, AlertCircle, CheckCircle } from 'lucide-react'
 import { useApiWithAuth } from '../hooks/useApiWithAuth'
 import { syncService, type UploadError } from '../services/sync'
-import { CSV_UPLOAD_STORAGE_KEY } from '../services/csvPipelineStatus'
+import { clearCsvUploadMarker, markCsvUploadStarted } from '../services/csvPipelineStatus'
 import { useMutation, useQueryClient } from '@tanstack/react-query'
 import toast from 'react-hot-toast'
 import { useCurrentYear } from '../hooks/useCurrentYear'
@@ -26,7 +26,7 @@ export default function BunkRequestsUpload({ compact = false }: BunkRequestsUplo
       // Mark this browser session as having initiated a CSV upload so the
       // CsvPipelineIndicator can attribute the next bunk_requests sync to
       // the upload (rather than the nightly cron, which also runs that sync).
-      localStorage.setItem(CSV_UPLOAD_STORAGE_KEY, new Date().toISOString())
+      markCsvUploadStarted()
     },
     onSuccess: () => {
       toast.success(
@@ -45,7 +45,7 @@ export default function BunkRequestsUpload({ compact = false }: BunkRequestsUplo
     onError: (error: UploadError) => {
       // Upload failed before the sync could start — clear the marker so a
       // subsequent cron run isn't falsely attributed to this upload.
-      localStorage.removeItem(CSV_UPLOAD_STORAGE_KEY)
+      clearCsvUploadMarker()
       if (error.missing_columns) {
         toast.error(
           <div>
--- a/frontend/src/components/BunkRequestsUpload.tsx
+++ b/frontend/src/components/BunkRequestsUpload.tsx
@@ -2,6 +2,7 @@ import { type ChangeEvent, useState, useRef } from 'react'
 import { Upload, Loader2, FileText, AlertCircle, CheckCircle } from 'lucide-react'
 import { useApiWithAuth } from '../hooks/useApiWithAuth'
 import { syncService, type UploadError } from '../services/sync'
+import { CSV_UPLOAD_STORAGE_KEY } from '../services/csvPipelineStatus'
 import { useMutation, useQueryClient } from '@tanstack/react-query'
 import toast from 'react-hot-toast'
 import { useCurrentYear } from '../hooks/useCurrentYear'
@@ -21,6 +22,12 @@ export default function BunkRequestsUpload({ compact = false }: BunkRequestsUplo
 
   const uploadMutation = useMutation({
     mutationFn: (file: File) => syncService.uploadBunkRequestsCSV(file, fetchWithAuth, currentYear),
+    onMutate: () => {
+      // Mark this browser session as having initiated a CSV upload so the
+      // CsvPipelineIndicator can attribute the next bunk_requests sync to
+      // the upload (rather than the nightly cron, which also runs that sync).
+      localStorage.setItem(CSV_UPLOAD_STORAGE_KEY, new Date().toISOString())
+    },
     onSuccess: () => {
       toast.success(
         "Importing CSV — this may take a few minutes. The icon next to the Upload Requests button will update when it's done.",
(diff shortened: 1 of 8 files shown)


adamflagg:
### Code review

Found 3 issues:

1. `isSyncFromCsvUpload` is symmetric — `Math.abs(syncAt - csvAt) <= 10 min` accepts syncs that started up to 10 minutes BEFORE the upload, not just after. A cron run at 02:00 plus a user upload at 02:05 will match: the cron's `done` state gets attributed to the user's upload — exactly the bug this PR claims to fix, just inverted in time. No test covers `syncAt < csvAt` within the proximity window. Suggest `syncAt >= csvAt && syncAt - csvAt <= CSV_UPLOAD_PROXIMITY_MS`.

https://github.com/adamflagg/kindred/blob/1ffc61f67c13840f3ba090376b0d94ac7c13a931/frontend/src/services/csvPipelineStatus.ts#L43-L50

2. `csvUploadStartedAt` is a client-clock timestamp (`new Date().toISOString()` in `BunkRequestsUpload.onMutate`) but `sync.startedAt` is a server-clock timestamp from `/api/custom/sync/status`. Comparing them with a 10-minute proximity window means any client whose clock differs from the server by more than 10 minutes (camp tablets/laptops without strict NTP) renders every CSV upload invisible — `derivePhase` returns `idle` permanently for legitimate user uploads.

https://github.com/adamflagg/kindred/blob/1ffc61f67c13840f3ba090376b0d94ac7c13a931/frontend/src/services/csvPipelineStatus.ts#L43-L50

3. `POST_UPLOAD_POLL_WINDOW_MS = 5 min` (hook) is half of `CSV_UPLOAD_PROXIMITY_MS = 10 min` (service). If the backend takes longer than 5 minutes to start the sync, `pollIntervalForPhase` stops polling while `derivePhase` would still attribute it correctly — the indicator misses the transition. The two windows should share a single constant. (Also flagged by CodeRabbit as Major.)

https://github.com/adamflagg/kindred/blob/1ffc61f67c13840f3ba090376b0d94ac7c13a931/frontend/src/hooks/useCsvPipelineStatus.ts#L16-L18

PR #1123, opened 2026-05-04
chore(frontend): batch trivial deletions and sentinel cleanup

### LINKED ISSUE #981: refactor(frontend): remove unreachable legacy branch in GraphCacheService.getBunkGraph
## Context

Surfaced during shrink-it review of PR #977 (social graph scenario sourcing). User opted to skip in that PR; tracking here for follow-up.

## Problem

\`GraphCacheService.getBunkGraph\` (\`frontend/src/services/GraphCacheService.ts:97-100\`) retains a fallback \`else\` branch that constructs a legacy key format:

\`\`\`ts
bunk-\${bunkCmId}-\${sessionCmId}
\`\`\`

…when the \`year\` argument is omitted. After PR #977, the only caller (\`BunkSocialGraphModal\`) always passes \`year\`, so the fallback branch is unreachable.

## Scope

- Remove the \`else\` branch and make \`year\` a required parameter on \`getBunkGraph\` / \`cacheBunkGraph\` / related signatures
- Remove any helper code that existed only to support the legacy key format
- Update tests (if any target the legacy branch)
- Verify no external consumers rely on the old shape

## Why not in #977

PR #977 was scoped to threading \`scenario_id\` end-to-end. Removing the legacy branch is a pure cleanup with no behavior consequence; widening the PR for it invited unnecessary review burden.

### LINKED ISSUE #1002: refactor(frontend): deprecate combined 'loading' flag on ScenarioContext
`ScenarioContext` exposes three flags today: `isLoading` (initial fetch), `isMutating` (in-flight create/update/delete), and `loading` (`isLoading || isMutating`). PR #992 switched the scenario-list placeholder rendering from `loading` to `isLoading` to fix the list-vanish bug, but `loading` is retained 'for backward compatibility' with consumers like `SessionView` that still use it as a generic busy signal.

- Audit all consumers of `loading` — is there a legitimate use case that needs the combined flag, or can each caller be migrated to `isLoading` / `isMutating` explicitly?
- If every consumer can be split, drop `loading` from the context type entirely.
- Low-risk refactor, out of scope for #992's bug fix.

Label: `tech-debt`, scope: `frontend`.

### LINKED ISSUE #1003: fix(frontend): "Export Moved" button label misleading when changeFilter is 'all'
On the Scenario Comparison page, the CSV export button is labeled "Export Moved" regardless of the active `changeFilter` value. When `changeFilter === 'all'`, the button is visible and the user may reasonably expect 'all rows (moved + added + removed + unchanged)' — but the export always emits only `filteredChanges.moved`.

- Two possible fixes: (a) update the button label to reflect the filter ("Export Moved" when filter is 'moved', "Export All Changes" when filter is 'all', etc.); (b) change the export payload to honor the filter (riskier — requires CSV schema decisions for rows without `from`/`to` bunks).
- Option (a) is the minimum-surprise fix.

Label: `bug`, scope: `frontend`.

### LINKED ISSUE #1004: refactor(frontend): extract shared buildRequesteeUpdate builder for AllCamperRequestsModal + RequestReviewPanel
## Problem

`handleTargetChange` in `AllCamperRequestsModal.tsx` (~L242–253) and the equivalent update code in `RequestReviewPanel.tsx` (~L1200–1213 and L1408–1420) both construct the same `pbUpdates` shape:

- Same `requestee_id` field
- Same `as unknown as number` cast
- Same conditional `status = 'resolved'` / `confidence_score = 1.0` block

This is duplicated business logic that must be kept in sync manually across two files.

## Proposed Fix

Consolidate into a shared `buildRequesteeUpdate(updates: Partial<BunkRequest>): Partial<BunkRequestsRecord>` helper, likely in `frontend/src/utils/bunkRequestHelpers.ts`.

## Notes

- Related but separate from issue #997 (which tracks the demote-on-cleared-requestee bug). Both issues can be fixed in the same refactor, but they are distinct concerns.
- Scope: `frontend`
- No behavior change expected — pure extraction refactor.

### LINKED ISSUE #1011: refactor(frontend): replace useSyncStatusAPI empty-object sentinel with discriminated return shape
## Context

Surfaced during scan-it on PR #1009 (fix for feedback item #55 — fresh-login \`end_time\` crash).

## Problem

\`useSyncStatusAPI\` catches 401 responses and returns an empty-object sentinel:

\`\`\`ts
return {} as unknown as SyncStatusResponse;
\`\`\`

This lies to the type system. Every field in \`SyncStatusResponse\` is non-optional (\`bunk_assignments: SyncStatus\`, \`bunk_requests: SyncStatus\`), so TypeScript believes every consumer has a fully populated object — but the sentinel has none of those fields. Every call site must remember to guard with \`?.\` or it crashes at runtime during the render window before \`pb.afterSend\` queues the async \`/login\` redirect.

## Why it matters

This is the root cause of the fresh-login crash class. PR #1009 patched the two known crashes (\`AppLayout.tsx:690\` and \`ScenarioManagementModal.tsx:143\`) by adding \`?.\`, but any future consumer of \`useSyncStatusAPI\` will hit the same footgun — the type signature provides no warning.

Example of the latent hazard in the codebase today: \`SyncTab.tsx:454\` reads \`syncStatus._daily_sync_running ?? syncStatus._historical_sync_running\` without optional chaining. It happens to no-op safely (\`undefined ?? undefined\` is \`undefined\`, falsy), but that's luck, not design.

## Proposed fix

Change the return shape to force guards at every call site. Two options:

**Option A — discriminated union**:

\`\`\`ts
type SyncStatusResult =
  | { authenticated: true; data: SyncStatusResponse }
  | { authenticated: false };
\`\`\`

**Option B — nullable return**:

\`\`\`ts
function useSyncStatusAPI(): { data: SyncStatusResponse | null; ... }
\`\`\`

Either approach forces consumers to narrow before field access. The compiler catches every missing guard at build time instead of production.

## Scope

Migration touches every consumer of \`useSyncStatusAPI\`. Likely a medium-sized refactor PR. Worth doing once the open post-feedback PRs are drained so rebase churn is minimal.

🤖 Surfaced via scan-it on PR #1009

adamflagg:
## Summary

Batch cleanup of 5 frontend issues. Three were already resolved on main before this PR landed; this PR implements the remaining two plus updates tests.

- **#981** — Remove unreachable legacy key branch in `GraphCacheService.getBunkGraph` — **already resolved on main** (year is now a required parameter, no else branch)
- **#1002** — Remove combined `loading` flag from `ScenarioContext`; migrate sole consumer (`SessionView`) to derive `isLoading || isMutating` locally
- **#1003** — Fix "Export Moved" button label when `changeFilter === 'all'` — extracts `getExportButtonLabel()` helper so the label reflects the active filter
- **#1004** — Extract shared `buildRequesteeUpdate` helper — **already resolved on main** (`computeTargetUpdate` in `requestEditableHelpers.ts` is used by both `AllCamperRequestsModal` and `RequestReviewPanel` via `RequestEditableHeader`)
- **#1011** — Replace empty-object sentinel in `useSyncStatusAPI` with `null` — **already resolved on main** (hook already returns `null` on 401, tests already verify this)

## Test plan

- [x] `cd frontend && npx vitest run` — 3562 tests, 0 failures
- [x] `cd frontend && npx tsc --noEmit` — no errors
- [x] `lefthook run pre-push` — all checks pass (ruff, shellcheck, pb-js-lint, go-build, mypy, tsc)

Closes #981
Closes #1002
Closes #1003
Closes #1004
Closes #1011

### DIFF
--- a/frontend/src/components/ScenarioManagementModal.test.tsx
+++ b/frontend/src/components/ScenarioManagementModal.test.tsx
@@ -3,8 +3,8 @@
  * staff testing (April 2026). See ScenarioContext.test.tsx for full context.
  *
  * The modal must keep rendering the scenario cards while a delete mutation
- * is pending — only the initial query-fetch state should swap the list for
- * a "Loading scenarios..." placeholder.
+ * is pending — only the initial query-fetch (isLoading) should swap the list
+ * for a "Loading scenarios..." placeholder, not isMutating.
  */
 import { describe, it, expect, vi } from 'vitest'
 import { render, screen } from '@testing-library/react'
@@ -47,7 +47,6 @@ function makeContext(overrides: Partial<ScenarioContextType>): ScenarioContextTy
     currentScenario: null,
     isProductionMode: true,
     scenarios,
-    loading: false,
     isLoading: false,
     isMutating: false,
     error: null,
@@ -77,18 +76,17 @@ function renderModal(ctx: ScenarioContextType) {
 describe('ScenarioManagementModal loading state', () => {
   it('shows "Loading scenarios..." during initial query fetch', () => {
     // Matches what ScenarioContext produces when the query is first loading:
-    // loading=true AND isLoading=true (scenarios list not yet fetched).
-    renderModal(makeContext({ loading: true, isLoading: true, scenarios: [] }))
+    // isLoading=true and scenarios list not yet fetched.
+    renderModal(makeContext({ isLoading: true, scenarios: [] }))
     expect(screen.getByText('Loading scenarios...')).toBeInTheDocument()
   })
 
   it('keeps scenario cards visible while a delete mutation is pending', () => {
-    // Simulates real context state mid-delete: the combined loading flag
-    // is true (because isMutating is true), but the query has already
-    // resolved (isLoading=false). Prior to the fix, the modal consumed the
-    // combined `loading` and swapped the list for the placeholder — this
-    // test asserts the fix by requiring the cards to remain visible.
-    renderModal(makeContext({ loading: true, isLoading: false, isMutating: true }))
+    // Simulates real context state mid-delete: the query has already resolved
+    // (isLoading=false) but a mutation is pending (isMutating=true). The modal
+    // uses isLoading (not isMutating) to show the placeholder, so cards remain
+    // visible while the delete processes.
+    renderModal(makeContext({ isLoading: false, isMutating: true }))
 
     expect(screen.getByText('Dorm Cabin Plan A')).toBeInTheDocument()
     expect(screen.queryByText('Loading scenarios...')).not.toBeInTheDocument()
--- a/frontend/src/components/SessionView.tsx
+++ b/frontend/src/components/SessionView.tsx
@@ -52,8 +52,10 @@ export default function SessionView() {
     scenarios,
     loadScenarios,
     selectScenario,
-    loading: scenarioLoading,
+    isLoading: scenarioIsLoading,
+    isMutating: scenarioIsMutating,
   } = useScenario()
+  const scenarioLoading = scenarioIsLoading || scenarioIsMutating
   const { setSessionPbId: setLockGroupSessionPbId } = useLockGroupContext()
   const { hasPermission } = usePermissions()
   const canManage = hasPermission(Permission.BUNKING_MANAGE)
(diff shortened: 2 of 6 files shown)


adamflagg:
### Code review

Issue #1047, opened 2026-04-27
feat(pb): enforce one friend group per camper per scenario


adamflagg:
## Problem

Current `locked_group_members` schema (`pb_migrations/1500000025_locked_group_members.js`) has a `UNIQUE(group, attendee)` index — preventing the same attendee being added to a single group twice, but **not** preventing the same attendee from being a member of two different groups within the same scenario.

Frontend code (`useGroupMap` in `ScenarioComparisonPage.tsx`, and similar consumers on the bunking board) implicitly assumes one-group-per-camper-per-scenario by storing `Map<cm_id, LockGroupSummary>` — last-write-wins if the assumption is violated. Same goes for the camper-pill color dot: only one color can render.

## Spec

Add a backend constraint that enforces: **for any `(scenario, attendee, year)` tuple, at most one `locked_group_members` row may exist.**

Implementation options (pick one in PR review):

1. **Migration adds a generated column on `locked_group_members`** carrying the parent `scenario` id, with a UNIQUE index on `(scenario_generated, attendee, year)`. Trade-off: SQLite generated-column UNIQUE indexes are supported but require careful migration syntax.
2. **App-level guard in PocketBase Go hook** on `locked_group_members.OnRecordBeforeCreate` / `OnRecordBeforeUpdate`: lookup other rows for `(attendee, year)` with `group.scenario == this.group.scenario`; reject with 409 if any. Trade-off: hook adds a query per write; concurrent writes still racy without a unique index.
3. **Both** — DB constraint as source of truth, app-level guard for friendly error messages.

Recommendation: option 3 (constraint + hook). The hook returns a structured error the UI can render as 'Camper is already in friend group "<name>".'

## Acceptance criteria

- [ ] Attempting to add a camper to a second friend group in the same scenario fails with a clear error message naming the existing group.
- [ ] Existing data (if any duplicate state exists) is migrated by either: keeping the most recent membership and deleting the rest, OR aborting the migration with a remediation script.
- [ ] Frontend `useGroupMap` and any other consumers can safely assume one-group-per-camper without defensive last-write-wins logic.
- [ ] Tests cover: add to first group succeeds, add to second group in same scenario fails, add to a group in a different scenario succeeds (confirming the constraint is scenario-scoped, not global).

## Notes

This was deliberately deferred during PR #1014 in favor of the existing implicit assumption. Filing as a follow-up so the schema and UX are made correct rather than relying on a convention the schema doesn't enforce.

[...]

PR #1208, opened 2026-05-07
feat(frontend): scenario copy carries locked groups + warn on cross-group membership

### LINKED ISSUE #1046: feat(frontend): scenario copy should include locked friend groups
## Problem

`copyScenarioToScenario` and `copyProductionToScenario` (`frontend/src/hooks/useSavedScenariosMutation.ts:151`) only duplicate `bunk_assignments_draft` rows when creating a new scenario from an existing one. Locked friend groups (`locked_groups` + `locked_group_members`) are silently dropped — the new scenario starts with zero friend groups even though the source had them.

This is surprising: staff expect a scenario copy to preserve the full state of the source, including friend-group assignments they may have spent time curating.

## Spec

Extend `copyScenarioToScenario` (and skip in `copyProductionToScenario` since production has no scenario to source groups from) to:

1. Read all `locked_groups` rows where `scenario = fromScenarioId && year = year`.
2. For each source group, create a new `locked_groups` row pointing to `toScenarioId`, preserving `name`, `color`, `session`, `year`. Build an `oldGroupId → newGroupId` map.
3. Read all `locked_group_members` rows where `group IN (sourceGroupIds)`.
4. For each member row, create a new `locked_group_members` row using `oldGroupId → newGroupId` from the map, preserving `attendee`, `year`.

Use the existing sequential-writes-with-per-item-try/catch pattern (mirrors how `copyScenarioToScenario` handles assignments). Aggregate errors and throw at end.

## Acceptance criteria

- [ ] Copying scenario A (with N friend groups, M members total) produces scenario B with N friend groups and M members, preserving names, colors, and member sets.
- [ ] Copying from production (`fromProduction: true`) does NOT attempt to copy groups (production has none).
- [ ] If any single group or member create fails, partial state is reported via aggregate throw; user sees a clear error.
- [ ] Unit tests cover: source has 0 groups (no-op), source has 1 group with 0 members, source has multiple groups with overlapping members, member create error path.

## Out of scope

- Schema-level provenance (`parent_lock_group_id`) — not needed; UI features that match groups across scenarios use member-overlap heuristics, which work naturally for freshly copied groups.

adamflagg:
## Summary

Implements #1046 (scenario copy carries locked friend groups) and provides a UI-level warning for the cross-group membership scenario described in #1047.

### #1046 — Scenario copy includes locked friend groups (unchanged)

- `copyScenarioToScenario` now calls `copyLockedGroupsToScenario` after copying assignments.
- `copyLockedGroupsToScenario`: fetches source `locked_groups`, creates new groups in the destination scenario, builds `oldId→newId` map, fetches source members, creates members pointing to new group IDs. Sequential `for..of`/`await` with per-item `try/catch`; aggregates errors and throws after the loop.
- `copyProductionToScenario` intentionally does NOT call this (production has no locked groups to copy).

### #1047 — UI confirmation dialog (pivot from DB enforcement)

The original approach (migration 1500000100 with BEFORE INSERT/UPDATE SQLite triggers + a Go hook returning 409) was **rejected** because the migration silently deleted staff-created memberships during upgrade. Pre-existing duplicates may represent intentional staff actions — the solver merges overlapping groups transitively, so the consequence is a larger super-group at solve time, not broken data.

**New approach**: staff are informed and asked to confirm, not blocked.

- New hook `useGroupConflictConfirm`: queries `locked_group_members` filtered by `(attendee, scenario)` before any create, opens an async confirmation dialog when a camper is already in a **different** group in the same scenario, and resolves with `null` (no conflict), `'confirmed'`, or `'cancelled'`.
- New component `GroupConflictDialog`: wraps `<Modal>`, shows both the existing and new group names, explains that the solver merges overlapping groups into a super-group at solve time (not an error), and offers Cancel / Continue.
- `LockGroupContext.addCamperToGroup` (drag-into-group): replaces the silent toast-error with the conflict check; Cancel aborts, Continue proceeds with the create.
- `LockGroupActionBar.createGroupMutation` (create-new-group from action bar): runs the conflict check for each pending camper before the group is created; Cancel suppresses creation silently, Continue proceeds.
- 7 new vitest tests covering: no conflict, same-group (no dialog), dialog opens with correct names, confirmed outcome, cancelled outcome, scenario scoping, and filter content assertion.

**Note**: `#1047` is referenced but not closed here. The issue framing was "enforce one group per camper per scenario" — this PR delivers a softer version (UI warning + confirmation) rather than hard enforcement. The user will decide whether to close or repurpose #1047.

Closes #1046
Refs #1047 — pivoted to UI warning, see issue comment

## Summary by CodeRabbit

* **New Features**
  * Locked friend groups are now copied when duplicating scenarios to the destination.
  * Added confirmation dialog when attempting to add a camper to a group where the camper already has a group assignment, alerting users that overlapping groups will merge during solving.

* **Improvements**
  * Enhanced conflict detection for group assignments within draft scenarios.


adamflagg:
### Code review

Found 1 issue:

1. Same-group membership check removed — when a camper is dragged onto the group they are already in, the old code hit the `camperToGroup.has()` guard and showed a clear toast: "Camper is already in a friend group". The new code only shows the conflict dialog when `existingGroup.id !== groupId`, so the same-group case falls through, the `create()` call is made, and the DB unique-index on `(group, attendee)` returns a 409. That 409 is caught by the generic `catch` block and surfaces as the unhelpful "Failed to add camper to group" toast. Data integrity is preserved, but the UX error message degrades for this path.

https://github.com/adamflagg/kindred/blob/431520e7c644369a8109dfbe1ae3e6901e50e5f1/frontend/src/contexts/LockGroupContext.tsx#L301-L305

Fix: add an explicit same-group early-return before the cross-group conflict check, restoring the original guard for that case:
```ts
if (existingGroup && existingGroup.id === groupId) return // already in this group, no-op
```

PR #2109, opened 2026-08-07
fix(frontend): a11y interaction gaps in modal family

adamflagg:
## What

Sweeps the `modal-family` chunk of the jsx-a11y backlog (21 warnings across 10 files, all fixed with real interaction fixes except two justified `no-autofocus` suppressions).

## Before / after

```text
./node_modules/.bin/eslint <chunk files> 2>/dev/null | grep -c jsx-a11y
```

| | jsx-a11y warnings |
|---|---|
| Before | 21 |
| After | 0 |

Full tree still has **0 errors** after the change (`./node_modules/.bin/eslint src` → 489 pre-existing `@typescript-eslint` warnings, untouched, 0 errors).

## Files

- `src/components/BunkSocialGraphModal.tsx`
- `src/components/BunkSwapModal.tsx`
- `src/components/FeedbackModal.tsx`
- `src/components/NewScenarioModal.tsx`
- `src/components/PostValidationResultsModal.tsx`
- `src/components/PostValidationResultsModal.render.test.tsx`
- `src/components/PreValidationResultsModal.test.tsx`
- `src/components/ScenarioEditModal.tsx`
- `src/components/ScenarioManagementModal.tsx`
- `src/components/SolverProgressModal.tsx`

## Fixed (19) vs suppressed (2)

**Fixed — backdrop split** (`click-events-have-key-events` + `no-static-element-interactions` / `no-noninteractive-element-interactions`):
`BunkSocialGraphModal.tsx` and `BunkSwapModal.tsx` each combined "backdrop click-to-close" and "dialog content" into one element with a raw `onClick`. Split each into a decorative `aria-hidden="true"` backdrop div (click-to-close is a mouse convenience; the real close path is the header's close button) and a content wrapper with no handler — mirrors the split already used by the shared `ui/Modal.tsx` primitive (not touched here — owned by a sibling chunk).

`SolverProgressModal.tsx` already had this split; it only needed `aria-hidden="true"` added to its existing backdrop div.

**Fixed — clickable text became a real `<button>`** (`click-events-have-key-events` + `no-static-element-interactions`):
- `PostValidationResultsModal.tsx`: a clickable camper-name `<span>` in the "Families to contact" list → `<button type="button">`.
- `PostValidationResultsModal.render.test.tsx` / `PreValidationResultsModal.test.tsx`: the `CamperDetailsPanel` test mock (`<div onClick={onClose}>`) → `<button>`. Fixture JSX, not real UI — the mock stands in for the real close affordance in tests that assert on it via `data-testid`. Updated a stale test comment that described this as "not a standalone button."

**Fixed — clickable "card" became an overlay `<button>`** (same two rules), `ScenarioManagementModal.tsx`:
Two selectable cards (the "CampMinder production mode" tile, and each scenario row's name/description column) had `onClick` directly on a `<div>` with nested block content (heading, paragraph, badge). Per house style (a `<button>`'s content model is phrasing-only — see #2094), rather than converting all the nested `<div>`/`<h3>`/`<p>` to `<span>`, each card got a `position:relative` wrapper with a full-cover `absolute inset-0` `<button aria-label="...">` as the actual hit target — same visual result, real keyboard operability, no content-model violation. The scenario-row overlay is scoped to its own flex column so it doesn't cover the sibling Edit/Clear/Delete buttons.

**Fixed — mislabeled `<label>` → group label** (`label-has-associated-control`):
`FeedbackModal.tsx` ("Category") and `NewScenarioModal.tsx` ("Copy Assignments From") both had a `<label>` heading a button/radio group rather than a single control. Neither has one input to associate via `htmlFor`, so both became a `<span id="...">` + `role="group"`/`role="radiogroup"` + `aria-labelledby` on the group container — the semantically correct pattern for a group heading.

**Suppressed (2) — `no-autofocus`, justified:**
- `NewScenarioModal.tsx` (scenario name field) and `ScenarioEditModal.tsx` (scenario name field): both are the single/first field of a short create/edit form opened as a modal. Autofocus here is the correct, deliberate UX — the user should be able to start typing immediately — not an accident. Each carries an `eslint-disable-next-line jsx-a11y/no-autofocus` with a one-line reason on the preceding line.

Zero warnings silenced without a fix. All 21 either got a structural fix or a justified, commented suppression.

## Verification

- Chunk eslint: 21 → 0 (`grep -c jsx-a11y`)
- Tree-wide `eslint src`: 0 errors, 489 pre-existing warnings (unchanged rule set, untouched by this PR)
- `npm run type-check` (`tsc --noEmit` ×2): clean
- `./node_modules/.bin/prettier --check` on changed files: clean
- Full `./node_modules/.bin/vitest run`: 5921/5928 passed, 6 files / 7 tests failed. All 6 failing files are outside this chunk and untouched by it (`AllCampersView.test.tsx`, `CamperDetail.test.tsx`, `SplitRequestModal.test.tsx`, `eslintA11yLayering.guard.test.ts`, `WeekendRosterPage.scenario.test.tsx`, `DrillDownModal.test.tsx`), every failure is `Error: Test timed out in 5000ms`, and re-running the same 6 files in isolation (outside full-suite resource contention) passes 5 of them cleanly. The 6th (`eslintA11yLayering.guard.test.ts`) makes a real ESLint Node API call to resolve `eslint.config.js` and times out even alone on this machine under concurrent sibling-agent load (7 parallel jsx-a11y chunk worktrees running builds/tests simultaneously); `eslint.config.js` is untouched by this PR. All 10 of this chunk's own component test files pass cleanly (184/184 tests) run together in isolation.
- Pre-push hook (ruff, go build, pb-js-lint, shellcheck, markdownlint, api-types-freshness, mypy, pytest 6013 passed, tsc) — all green.

### DIFF
--- a/frontend/src/components/BunkSocialGraphModal.tsx
+++ b/frontend/src/components/BunkSocialGraphModal.tsx
@@ -468,16 +468,24 @@ export default function BunkSocialGraphModal({
   // Use Activity to preserve state when hidden while unmounting effects
   // The backdrop transitions in/out based on isOpen
   return (
-    <div
-      className={clsx(
-        'fixed inset-0 z-50 flex items-center justify-start p-0 transition-all duration-300 sm:p-4',
-        isOpen ? 'pointer-events-auto bg-black/50' : 'pointer-events-none bg-transparent opacity-0'
-      )}
-      onClick={(e) => e.target === e.currentTarget && onClose()}
-    >
+    <div className="fixed inset-0 z-50 flex items-center justify-start p-0 sm:p-4">
+      {/* Decorative backdrop: click-to-close is a mouse convenience, not the
+          only path to close (see the header's real close button), so this
+          stays a plain non-interactive div rather than growing keyboard
+          handling that would duplicate the close button. */}
+      <div
+        aria-hidden="true"
+        className={clsx(
+          'absolute inset-0 transition-all duration-300',
+          isOpen
+            ? 'pointer-events-auto bg-black/50'
+            : 'pointer-events-none bg-transparent opacity-0'
+        )}
+        onClick={onClose}
+      />
       <Activity mode={isOpen ? 'visible' : 'hidden'}>
         <div
-          className={`bg-card shadow-lodge-xl overflow-hidden rounded-none transition-all duration-300 sm:ml-4 sm:rounded-2xl ${
+          className={`bg-card shadow-lodge-xl relative overflow-hidden rounded-none transition-all duration-300 sm:ml-4 sm:rounded-2xl ${
             selectedCamperId
               ? 'h-full w-full sm:max-h-[95vh] sm:w-[calc(95vw-20rem)] md:w-[calc(95vw-26rem)]'
               : 'h-full w-full sm:max-h-[95vh] sm:w-[95vw]'
--- a/frontend/src/components/BunkSwapModal.tsx
+++ b/frontend/src/components/BunkSwapModal.tsx
@@ -35,16 +35,16 @@ export default function BunkSwapModal({
   const selected = candidates.find((b) => b.id === selectedId) ?? null
 
   return (
-    <div
-      role="dialog"
-      aria-modal="true"
-      aria-labelledby="bunk-swap-title"
-      className="fixed inset-0 z-[70] flex items-center justify-center bg-black/40"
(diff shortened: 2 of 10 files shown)


adamflagg:
@coderabbitai review

PR #2121, opened 2026-08-08
fix(frontend): clear stale party selection on weekend surfaces

adamflagg:
## Summary
- `HouseholdRosterTable`, `LodgingBoard`, and `LodgingMap` all kept a selected party's panel open across a weekend switch, because `selected` was never reconciled against a changed `parties` prop — the sticky-header `<select>` re-renders these components with a different roster without unmounting them.
- `FamilyDetailsPanel` renders `MedicalNarrative`, so this meant a `lodging.phi` holder switching weekends kept seeing the **previous weekend's** family and medical narrative attached to a roster that no longer contained them.
- Fixed by deriving what the panel renders (`panelParty`) from whether the selected party's `partyKey` is still present in `parties`, computed at render time rather than via a `useEffect`+`setState` that would add an extra render pass. `selected` state itself is untouched by a departure — only what gets rendered changes.
- Because `partyKey` compares content rather than array identity, a refetch that returns the same parties (new array reference, same content) correctly leaves an open panel open.

## Test plan
- [x] Added a test per surface pinning the departure case (selected party's key no longer in `parties` → panel closes)
- [x] Added a test per surface pinning the refetch-same-parties trap (new array identity, same content → panel stays open)
- [x] `npx vitest run src/components/weekend` — 28 files / 582 tests passed
- [x] Full frontend suite — passed except two failures unrelated to this diff and confirmed pre-existing/environmental (`eslintA11yLayering.guard.test.ts` timing out under machine load, passes with a longer timeout; `WeekendRosterPage.codeSplitting.test.tsx` flaked once, passed on retry). Neither imports the files this PR touches.
- [x] `npx tsc --noEmit` — clean
- [x] `./node_modules/.bin/eslint` on changed files — 0 errors, pre-existing warning classes only
- [x] Full pre-push hook suite (ruff, go build, pb-js-lint, shellcheck, markdownlint, api-types-freshness, mypy, full pytest, tsc) — all green

Closes #2062.

### DIFF
--- a/frontend/src/components/weekend/HouseholdRosterTable.tsx
+++ b/frontend/src/components/weekend/HouseholdRosterTable.tsx
@@ -14,6 +14,7 @@ import { useDismissOnDeadSpace } from '../../hooks/useDismissOnDeadSpace'
 import type { LodgingUnitRow, RosterPartyRow } from '../../types/lodging'
 import { FamilyDetailsPanel } from './FamilyDetailsPanel'
 import { HouseholdRosterRow } from './HouseholdRosterRow'
+import { partyKey } from './partyKey'
 import { attentionSections, indexUnitsByCode } from './rosterAttention'
 
 export interface HouseholdRosterTableProps {
@@ -61,7 +62,20 @@ export function HouseholdRosterTable({ parties, units = [], year }: HouseholdRos
     setRequestClose(false)
   }, [])
 
-  useDismissOnDeadSpace(selected !== null, () => {
+  // DERIVED, not effect-cleared (kindred#2062). A weekend switch re-renders
+  // this table with a different `parties` prop but never unmounts it, so a
+  // `selected` that outlived its own row's departure kept the panel — and
+  // its medical narrative — open over the new weekend's roster. Computing
+  // this at render time, rather than in a useEffect that calls setState,
+  // avoids the extra render pass React's docs warn an Effect would add here
+  // (`selected` itself is untouched; only what gets rendered changes). A
+  // refetch that returns the SAME parties (new array identity, same
+  // content) still resolves to present, because `partyKey` compares
+  // content, not identity — see partyKey.ts's own docstring.
+  const panelParty =
+    selected !== null && parties.some((p) => partyKey(p) === partyKey(selected)) ? selected : null
+
+  useDismissOnDeadSpace(panelParty !== null, () => {
     setRequestClose(true)
   })
 
@@ -129,10 +143,10 @@ export function HouseholdRosterTable({ parties, units = [], year }: HouseholdRos
         </table>
       </div>
 
-      {selected !== null && (
+      {panelParty !== null && (
         <FamilyDetailsPanel
-          party={selected}
-          unit={unitsByCode.get(selected.unit_code ?? '')}
+          party={panelParty}
+          unit={unitsByCode.get(panelParty.unit_code ?? '')}
           year={year}
           requestClose={requestClose}
           onClose={closePanel}
--- a/frontend/src/components/weekend/LodgingBoard.test.tsx
+++ b/frontend/src/components/weekend/LodgingBoard.test.tsx
@@ -651,3 +651,53 @@ describe('LodgingBoard — the details panel updates in place', () => {
     expect(second).toHaveTextContent('Chen Household')
   })
 })
+
+describe('LodgingBoard — clears a stale selection (kindred#2062)', () => {
+  // A weekend switch re-renders the board with a different `parties` prop
+  // without unmounting it, so the previously-open family's panel — including
+  // its medical narrative — stayed open over the new weekend's roster.
+  it('closes the panel when the selected party is no longer in parties', async () => {
+    const { rerender } = render(
+      <LodgingBoard
+        parties={[party({ unit_code: 'cedar-1', unit_name: 'Cedar 1' })]}
+        units={[unit()]}
+        year={2026}
+      />,
+      { wrapper }
+    )
+    await userEvent.click(screen.getByRole('button', { name: /Johnson/ }))
+    expect(screen.getByTestId('family-details-panel')).toBeInTheDocument()
+
+    rerender(
+      <LodgingBoard
+        parties={[
+          party({
+            household_cm_id: 102,
+            display_name: 'Chen',
+            sort_name: 'Chen',
+            unit_code: 'cedar-1',
+            unit_name: 'Cedar 1',
+          }),
+        ]}
+        units={[unit()]}
+        year={2026}
+      />
+    )
+    expect(screen.queryByTestId('family-details-panel')).not.toBeInTheDocument()
+  })
+
+  // The trap: a refetch that returns the SAME parties (new array identity,
+  // same content) must not close a panel out from under whoever has it open.
+  it('keeps the panel open when parties refetches with the same content', async () => {
+    const makeParties = () => [party({ unit_code: 'cedar-1', unit_name: 'Cedar 1' })]
+    const { rerender } = render(
+      <LodgingBoard parties={makeParties()} units={[unit()]} year={2026} />,
+      { wrapper }
+    )
+    await userEvent.click(screen.getByRole('button', { name: /Johnson/ }))
+    expect(screen.getByTestId('family-details-panel')).toBeInTheDocument()
+
+    rerender(<LodgingBoard parties={makeParties()} units={[unit()]} year={2026} />)
+    expect(screen.getByTestId('family-details-panel')).toBeInTheDocument()
+  })
+})
--- a/frontend/src/components/weekend/LodgingBoard.tsx
+++ b/frontend/src/components/weekend/LodgingBoard.tsx
@@ -306,8 +306,21 @@ export function LodgingBoard({
     setRequestClose(false)
   }, [])
 
+  // DERIVED, not effect-cleared (kindred#2062). A weekend switch re-renders
+  // the board with a different `parties` prop but never unmounts it, so a
+  // `selected` that outlived its own card's departure kept the panel — and
+  // its medical narrative — open over the new weekend's roster. Computing
+  // this at render time, rather than in a useEffect that calls setState,
+  // avoids the extra render pass React's docs warn an Effect would add here
+  // (`selected` itself is untouched; only what gets rendered changes). A
+  // refetch that returns the SAME parties (new array identity, same
+  // content) still resolves to present, because `partyKey` compares
+  // content, not identity — see partyKey.ts's own docstring.
+  const panelParty =
+    selected !== null && parties.some((p) => partyKey(p) === partyKey(selected)) ? selected : null
+
   // Same dead-space dismissal the summer board uses, through the same hook.
-  useDismissOnDeadSpace(selected !== null, () => {
+  useDismissOnDeadSpace(panelParty !== null, () => {
     setRequestClose(true)
   })
 
@@ -512,14 +525,14 @@ export function LodgingBoard({
         <FloatingUnplacedBadge
           parties={board.unplaced}
           onOpenParty={openParty}
-          isPanelOpen={selected !== null}
+          isPanelOpen={panelParty !== null}
           canPlace={canPlace}
         />
 
-        {selected !== null && (
+        {panelParty !== null && (
           <FamilyDetailsPanel
-            party={selected}
-            unit={unitsByCode.get(selected.unit_code ?? '')}
+            party={panelParty}
+            unit={unitsByCode.get(panelParty.unit_code ?? '')}
             year={year}
             requestClose={requestClose}
             onClose={closePanel}
--- a/frontend/src/components/weekend/LodgingMap.tsx
+++ b/frontend/src/components/weekend/LodgingMap.tsx
@@ -280,6 +280,19 @@ export function LodgingMap({ parties, units, year }: LodgingMapProps) {
     setRequestClose(false)
   }, [])
 
+  // DERIVED, not effect-cleared (kindred#2062). A weekend switch re-renders
+  // the map with a different `parties` prop but never unmounts it, so a
+  // `selected` that outlived its own mark's departure kept the panel — and
+  // its medical narrative — open over the new weekend's roster. Computing
+  // this at render time, rather than in a useEffect that calls setState,
+  // avoids the extra render pass React's docs warn an Effect would add here
+  // (`selected` itself is untouched; only what gets rendered changes). A
+  // refetch that returns the SAME parties (new array identity, same
+  // content) still resolves to present, because `partyKey` compares
+  // content, not identity — see partyKey.ts's own docstring.
+  const panelParty =
+    selected !== null && parties.some((p) => partyKey(p) === partyKey(selected)) ? selected : null
+
   // Same dead-space dismissal the summer board and the weekend board use, and
   // deliberately the same one line they use. The canvas is a bare div that a
   // pan gesture ends with a click on, and that click matches none of
@@ -290,7 +303,7 @@ export function LodgingMap({ parties, units, year }: LodgingMapProps) {
   // rather than by teaching this callback to second-guess the clicks it does
   // get. A callback that remembers things about earlier gestures is a callback
   // that can be wrong about the current one.
-  useDismissOnDeadSpace(selected !== null, () => {
+  useDismissOnDeadSpace(panelParty !== null, () => {
     setRequestClose(true)
   })
 
@@ -1014,13 +1027,13 @@ export function LodgingMap({ parties, units, year }: LodgingMapProps) {
       <FloatingUnplacedBadge
         parties={model.unplaced}
         onOpenParty={openParty}
-        isPanelOpen={selected !== null}
+        isPanelOpen={panelParty !== null}
       />
 
-      {selected !== null && (
+      {panelParty !== null && (
         <FamilyDetailsPanel
-          party={selected}
-          unit={unitsByCode.get(selected.unit_code ?? '')}
+          party={panelParty}
+          unit={unitsByCode.get(panelParty.unit_code ?? '')}
           year={year}
           requestClose={requestClose}
           onClose={closePanel}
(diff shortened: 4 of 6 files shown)


frontend/src/components/weekend/LodgingBoard.tsx
```diff
@@ -306,8 +306,21 @@ export function LodgingBoard({
...
+  // the board with a different `parties` prop but never unmounts it, so a
+  // `selected` that outlived its own card's departure kept the panel — and
+  // its medical narrative — open over the new weekend's roster. Computing
+  // this at render time, rather than in a useEffect that calls setState,
+  // avoids the extra render pass React's docs warn an Effect would add here
+  // (`selected` itself is untouched; only what gets rendered changes). A
+  // refetch that returns the SAME parties (new array identity, same
+  // content) still resolves to present, because `partyKey` compares
+  // content, not identity — see partyKey.ts's own docstring.
+  const panelParty =
```
adamflagg, line 319: **Critical: `selected` is never cleared on departure, so the PHI panel can silently reopen with no user click.**

The comment says "`selected` itself is untouched by a departure — only what gets rendered changes," but nothing else ever clears it either. `openParty` only *replaces* it (needs a different click); `closePanel` only runs through `FamilyDetailsPanel`'s `onClose`, which can't fire once the panel is unmounted.

Repro (all on `queryClient.ts` defaults — 30 min staleTime, 60 min gcTime, no override in `useWeekendRoster`, so `MetricsQueryGuard` never re-shows its spinner on a re-visited weekend):
1. Open weekend A, click a family → `selected = Johnson`, panel opens, `MedicalNarrative` fetches PHI.
2. Switch to weekend B (already cached this session): `rosterQuery.isLoading` stays `false`, so this component is never unmounted — the exact #2062 precondition. `parties` no longer contains Johnson → `panelParty = null` → panel hides. `selected` is *still* Johnson.
3. Switch back to weekend A (also cached, no unmount). `parties.some(...)` matches again → `panelParty = Johnson` → the panel reopens on its own and `useHouseholdMedical` (`staleTime: 0, gcTime: 0`) re-issues a real network fetch for a household nobody asked to see.

Same mechanism at `LodgingMap.tsx:293` and `HouseholdRosterTable.tsx:75`. Clearing `selected` (e.g. `setSelected(null)`) when `panelParty` derives to `null` would close this.

frontend/src/components/weekend/LodgingBoard.tsx
```diff
@@ -306,8 +306,21 @@ export function LodgingBoard({
...
+    selected !== null && parties.some((p) => partyKey(p) === partyKey(selected)) ? selected : null
+
   // Same dead-space dismissal the summer board uses, through the same hook.
-  useDismissOnDeadSpace(selected !== null, () => {
+  useDismissOnDeadSpace(panelParty !== null, () => {
```
adamflagg, line 323: **Critical: `requestClose` can latch `true` across the new unmount path, so a resurrected panel (see the finding at line 319) mounts already mid-exit-animation.**

`FamilyDetailsPanel.tsx:92`'s own comment states the assumption this breaks: "`requestClose` needs no equivalent — the parent's `openParty` sets it false." That held only while the panel could never reappear without `openParty` running. After this PR it can (via `panelParty` flipping null → non-null with no click).

Sequence: click dead space while a panel is open → `useDismissOnDeadSpace` sets `requestClose = true`, the 300ms slide-out starts. Before `onAnimationEnd` fires, the party departs (weekend switch) → `panelParty` goes null → the element is torn out mid-animation → `onClose`/`closePanel` never runs → both `selected` and `requestClose` are left stuck. When the party reappears (see line 319's finding), `FamilyDetailsPanel` mounts with `requestClose = true`, so `exiting` is `true` on the very first render — the household's name, share request, and `MedicalNarrative` all render fully, then immediately animate out, with no interaction and no `openParty` call in between.

Same at `LodgingMap.tsx:306` and `HouseholdRosterTable.tsx:78`.

frontend/src/components/weekend/LodgingBoard.tsx
```diff
@@ -306,8 +306,21 @@ export function LodgingBoard({
...
+  // `selected` that outlived its own card's departure kept the panel — and
+  // its medical narrative — open over the new weekend's roster. Computing
+  // this at render time, rather than in a useEffect that calls setState,
+  // avoids the extra render pass React's docs warn an Effect would add here
+  // (`selected` itself is untouched; only what gets rendered changes). A
+  // refetch that returns the SAME parties (new array identity, same
+  // content) still resolves to present, because `partyKey` compares
+  // content, not identity — see partyKey.ts's own docstring.
+  const panelParty =
+    selected !== null && parties.some((p) => partyKey(p) === partyKey(selected)) ? selected : null
```
adamflagg, line 320: **High: the ternary returns the stale `selected` object rather than the live matching row, so an open panel doesn't reflect changes to its own party.**

`... ? selected : null` hands back the captured `selected`, never the element `.some()` just matched. Confirmed via `applyPlacement` (`dragPlacement.ts:251`), which is deliberately "purely functional" and returns a *new* party object (`{ ...party, unit_code: intent.unitCode, ... }`) on a drag placement, matched by the same `partyKey`.

Repro: open a family's panel, then drag that same family's card from Cedar 1 to Pine 2. `useLodgingPlacement`'s optimistic `onMutate` rewrites the roster with the new `unit_code`/`unit_name`; `partyKey` is unchanged so `panelParty` still resolves — but to the pre-drag object. The open panel keeps showing "Cedar 1" (and `unitsByCode.get(panelParty.unit_code)` keeps resolving Cedar 1's area) while the board redraws the card in Pine 2, until the user closes and reopens the panel. Same for a scenario switch on a cached scenario where the party sits in a different unit, or any refetch that changes share text/flags for the selected party.

`parties.find((p) => partyKey(p) === partyKey(selected)) ?? null` would fix this for free — it's also what would make the membership check and the freshness fix the same operation, since `selectedKey`-based state (rather than storing the whole object) removes the need to reuse `selected` at all.

Same at `LodgingMap.tsx:294` and `HouseholdRosterTable.tsx:76`.

frontend/src/components/weekend/LodgingMap.tsx
```diff
@@ -280,6 +280,19 @@ export function LodgingMap({ parties, units, year }: LodgingMapProps) {
...
+  // `selected` that outlived its own mark's departure kept the panel — and
+  // its medical narrative — open over the new weekend's roster. Computing
+  // this at render time, rather than in a useEffect that calls setState,
+  // avoids the extra render pass React's docs warn an Effect would add here
+  // (`selected` itself is untouched; only what gets rendered changes). A
+  // refetch that returns the SAME parties (new array identity, same
+  // content) still resolves to present, because `partyKey` compares
+  // content, not identity — see partyKey.ts's own docstring.
+  const panelParty =
+    selected !== null && parties.some((p) => partyKey(p) === partyKey(selected)) ? selected : null
```
adamflagg, line 294: **Medium (efficiency): unmemoized O(n) scan lands on the map's pan hot path, and `partyKey(selected)` is recomputed once per array element instead of hoisted.**

This file already memoizes `buildMapModel` with an explicit comment explaining why: "panning updates `view` on every pointermove, and an unmemoised call would re-run buildBoard... on every frame of a drag" (line ~180). `panelParty` is computed bare at render, so with a panel open, every `onPointerMove` during a pan (which calls `setView` → re-render) walks all parties again — up to ~123 on a full adult weekend, per this codebase's own comments elsewhere.

It's actually 2N `partyKey` calls, not N+1: `partyKey(selected)` is invoked inside the `.some()` predicate itself, so it's recomputed per element scanned rather than hoisted once above the loop. `frontend/CLAUDE.md` states: "Anything that builds a model to read a number belongs in `useMemo`." A `useMemo` on `[parties, selected]`, with the needle key hoisted (`const key = partyKey(selected)`), fixes both at once. Same unmemoized/per-iteration pattern at `LodgingBoard.tsx:320` and `HouseholdRosterTable.tsx:76`.

frontend/src/components/weekend/HouseholdRosterTable.tsx
```diff
@@ -61,7 +62,20 @@ export function HouseholdRosterTable({ parties, units = [], year }: HouseholdRos
...
+  // this table with a different `parties` prop but never unmounts it, so a
+  // `selected` that outlived its own row's departure kept the panel — and
+  // its medical narrative — open over the new weekend's roster. Computing
+  // this at render time, rather than in a useEffect that calls setState,
+  // avoids the extra render pass React's docs warn an Effect would add here
+  // (`selected` itself is untouched; only what gets rendered changes). A
+  // refetch that returns the SAME parties (new array identity, same
+  // content) still resolves to present, because `partyKey` compares
+  // content, not identity — see partyKey.ts's own docstring.
+  const panelParty =
```
adamflagg, line 75: **Medium (reuse): the exact same derivation and an 11-line comment are triplicated verbatim across `HouseholdRosterTable.tsx`, `LodgingBoard.tsx`, and `LodgingMap.tsx`, with no shared hook.**

Only the nouns change (row/card/mark). `partyKey.ts`'s own docstring is the precedent this PR should have followed but didn't: "shared by every weekend surface that lists parties... One definition, because four drifted into two variants and one of them was wrong." A natural extraction target sits right next to it — e.g. `usePanelParty(parties, selected)` in `frontend/src/components/weekend/`, returning `panelParty` with the rationale documented once as its JSDoc. Each of the three call sites would shrink to one line, and a future fix to the staleness guard (e.g. the `.find`-vs-`.some` issue flagged elsewhere in this diff) would only need applying once instead of three times with a real chance of missing a copy.

frontend/src/components/weekend/LodgingBoard.test.tsx
```diff
@@ -651,3 +651,53 @@ describe('LodgingBoard — the details panel updates in place', () => {
     expect(second).toHaveTextContent('Chen Household')
   })
 })
+
+describe('LodgingBoard — clears a stale selection (kindred#2062)', () => {
```
adamflagg, line 655: **Low-medium: the new tests don't pin the invariants this PR actually depends on.**

Both new cases here (and the matching ones in `HouseholdRosterTable.test.tsx`/`LodgingMap.test.tsx`) assert only `queryByTestId('family-details-panel')` presence/absence across a disjoint-roster rerender and a same-content-refetch rerender. Neither test would fail against the shipped `.some()`-plus-stale-object implementation *or* a corrected `.find()`-based one, and neither covers:
- A third rerender back to the *original* parties (would catch the "panel resurrects on return" bug flagged elsewhere in this diff).
- A rerender where the selected party's own content changes, e.g. a different `unit_code` (would catch the stale-object bug at line 320 of `LodgingBoard.tsx`).
- The `isPanelOpen`/`useDismissOnDeadSpace` prop changes — reverting either back to `selected !== null` still passes the full suite (verified by mutation-testing both reversions locally).
- The actual PHI fetch: all three test files mock `useHouseholdMedical` to a constant that ignores its arguments, so `MedicalNarrative`'s fetch — the harm named in this PR's own description — is never exercised by any assertion here.

frontend/src/components/weekend/LodgingMap.tsx
```diff
@@ -280,6 +280,19 @@ export function LodgingMap({ parties, units, year }: LodgingMapProps) {
...
+  // the map with a different `parties` prop but never unmounts it, so a
+  // `selected` that outlived its own mark's departure kept the panel — and
+  // its medical narrative — open over the new weekend's roster. Computing
+  // this at render time, rather than in a useEffect that calls setState,
+  // avoids the extra render pass React's docs warn an Effect would add here
+  // (`selected` itself is untouched; only what gets rendered changes). A
+  // refetch that returns the SAME parties (new array identity, same
+  // content) still resolves to present, because `partyKey` compares
+  // content, not identity — see partyKey.ts's own docstring.
+  const panelParty =
```
adamflagg, line 293: **Medium: `pinnedKey`/`dwellKey` (declared at line 224, outside this diff's hunks so anchoring the comment here) carry the exact same never-cleared-latch problem this PR just fixed for `selected`, left unaddressed in the same file.**

`openCluster` (line 399) derives correctly via `clusters.find(...)`, so the peek popover hides when its cluster stops existing — but `pinnedKey`/`dwellKey` themselves are never reset when that happens, the same shape `selected` had before this PR's fix. With "Empty rooms" unchecked (`showEmpty` false, filtering `parties.length > 0`), a roster refetch that empties a room dissolves its cluster and the pinned peek disappears; a later refetch that re-fills the same room re-mints the identical `clusterKey` (sorted unit ids) and the popover reopens over that cabin with no user action. A `ResizeObserver`-driven re-cluster (line ~252) can reach the same state without any roster change at all.

Worth noting because the natural "deeper fix" for the `panelParty` staleness bugs elsewhere in this PR is to switch to key-based state and derive via `.find` — which is exactly this file's own `pinnedKey`/`dwellKey` → `openCluster` pattern. That pattern needs the same clearing fix, not just imitating as-is.

frontend/src/components/weekend/LodgingBoard.tsx
adamflagg, line 319: **Altitude: storing the full party object in `selected` (declared at line 127, outside this diff's hunks so anchoring here) rather than its key is one layer too shallow — it's the root cause of the stale-object finding on line 320 below.**

If this were `const [selectedKey, setSelectedKey] = useState<string | null>(null)` with `panelParty = selectedKey === null ? null : (parties.find((p) => partyKey(p) === selectedKey) ?? null)`, the presence check and the "get fresh data" step would be the same operation — the object-staleness bug disappears for free, not just the departure bug this PR targets. This file's own `LodgingMap` sibling already uses this shape for its unit popover (`pinnedKey`/`dwellKey` → `openCluster` via `.find`), though see the separate finding on that pattern in `LodgingMap.tsx` — it has the same never-cleared latch problem and would need fixing too if used as the model.

frontend/src/components/weekend/HouseholdRosterTable.tsx
```diff
@@ -129,10 +143,10 @@ export function HouseholdRosterTable({ parties, units = [], year }: HouseholdRos
         </table>
       </div>
 
-      {selected !== null && (
+      {panelParty !== null && (
```
adamflagg, line 146: **Low: this file now carries two disagreeing identity functions for the same rows (row `key=` prop is at line 134, outside this diff's hunks so anchoring here).**

The newly-imported `partyKey` (`||`, falls through household → person → display_name, used at line 75) disagrees with the pre-existing hand-rolled row key at line 134 (`` `${party.grain}-${String(party.household_cm_id ?? 0)}-${String(party.person_cm_id ?? 0)}` ``, using `??` and both ids). For two unresolved households (both ids 0, different display names), that row key produces the identical React key (`household-0-0`) for both — a duplicate key that makes React reconcile the two rows as one and reuse state/DOM across them — while `partyKey` correctly treats them as distinct for the new panel-membership check. Now that `partyKey` is imported into this file, there's no remaining reason for a second, disagreeing identity scheme to sit beside it.

frontend/src/components/weekend/LodgingBoard.test.tsx
adamflagg, line 655: **Low (altitude): `FamilyDetailsPanel` (a file this PR doesn't touch, so noting it here) has no structural enforcement of "only render a party present in the current roster" — the guarantee lives entirely in three separate callers' discipline.**

`party: RosterPartyRow` is a required, non-null prop with no membership check of its own, and this component is the one place PHI actually renders (`MedicalNarrative`). A future fourth weekend surface that copies the *old* two-state pattern (`useState` + `{selected !== null && <FamilyDetailsPanel party={selected} .../>}`, i.e. main's code before this PR) reintroduces #2062 with no compile error, no lint error, and none of this PR's new tests would catch it, since they're scoped to the three existing call sites. A deeper fix would have `FamilyDetailsPanel` take `partyKey: string | null` + `parties: RosterPartyRow[]` and do the lookup itself, making it impossible for any caller to hand it a stale or absent-from-roster party.

adamflagg (commented):
Extra-high-effort recall review (10 parallel finder angles + direct verification + a gap sweep). The fix is real and does close the reported bug for the common case, but the derivation it adds (`panelParty`) is narrower than the PR's own framing suggests, and two of its side effects are privacy-relevant. 12 findings posted inline; 2 more below couldn't be anchored inline because `partyKey.ts` isn't part of this diff.

**Most severe, independently confirmed by code-reading (not just plausible):**
- `selected` is never cleared when a party departs (by design, per the PR's own comment) — so if the party reappears (e.g. switching back to a previously-visited, still-cached weekend — `useWeekendRoster` inherits the app's 30 min staleTime with no override), the PHI panel **silently reopens with no user click** and re-fetches the medical narrative. See the comment on `LodgingBoard.tsx:319`.
- `requestClose` isn't reset on that same unmount path, so a resurrected panel can mount already mid-exit-animation, flashing another family's panel. `FamilyDetailsPanel.tsx:92`'s own comment documents the assumption this breaks. See `LodgingBoard.tsx:323`.
- The `.some()` check returns the **stale `selected` object**, not the live matching row — confirmed via `dragPlacement.ts`'s `applyPlacement`, which returns a new party object on every drag placement. An open panel doesn't reflect a placement change to its own party. See `LodgingBoard.tsx:320`.
- `partyKey` (used as the presence test) carries no session/weekend identifier — confirmed against `RosterParty`'s generated type, which has no session field — so a household enrolled in **two weekends** produces the same key in both, and the panel survives a weekend switch instead of closing. This is the literal #2062 symptom surviving the fix for that case. (Not inline-commentable: `partyKey.ts` isn't touched by this diff.)
- Related: `partyKey`'s documented `display_name` fallback (for unresolved households, `household_cm_id = 0`) can match two *different* households across weekends. (Also not inline-commentable, same reason.)

**Cleanup:** the same 2-line derivation plus an 11-line comment is triplicated verbatim across all three files with no shared hook, despite `partyKey.ts`'s own docstring citing exactly this "one definition" precedent in this directory. The derivation is also unmemoized (`frontend/CLAUDE.md`: "Anything that builds a model to read a number belongs in `useMemo`"), worst on `LodgingMap` where it re-runs every pointermove during a pan.

**Test gap:** the new tests only cover disjoint-roster-closes and identical-content-refetch-stays-open. Mutation-testing confirms the `isPanelOpen`/`useDismissOnDeadSpace` prop changes aren't pinned by any test, and all three test files mock `useHouseholdMedical` to a constant, so the PHI fetch — the stated motivation for this PR — is never exercised by any assertion.

None of this should block landing the fix for the reported case, but the "panel silently reopens with PHI" and "same-household-two-weekends bypasses the fix entirely" findings seem worth addressing before calling #2062 closed.

PR #2124, opened 2026-08-08
feat(frontend): widen camper journey to include family camp and reach 2017-2021

adamflagg:
> ## ✅ Unblocked — the `years_at_camp` premise was falsified
>
> Measured against production: `years_at_camp` counts **summer** attendance and is **correct**. Our `camp_sessions` covers only 2017–2026, so where it disagrees it is **higher 42.6%** of the time (CampMinder sees pre-2017 history we never synced) and lower only 3.3%. Of 6,482 people showing `0` with real attendance, **90.6% have no summer session at all** — 3,877 family-only, 1,273 adult-only. Showing `0` is right for them.
>
> Owner decision: **keep CampMinder's value and widen the journey to include family anyway.** The header measures summer; the timeline shows attendance. Different things, both correct.
>
> The genuine residual — **609 people who have summer attendance and still show `0`** — is tracked in #2123 and is not a blocker on this PR.

## Summary
- Reverses the "no family" curation on `CAMPER_JOURNEY_TYPES`, documented at `sessionTypePredicates.ts:38-43`. That exclusion mirrored the All Campers page and was meant to keep the journey timeline uncluttered for multi-session staff kids, but measured against production it hid ~45% of enrolled attendance and left the journey completely blank for over a thousand people in some years (2018/2023/2024) — a real data-loss cost with no offsetting benefit, since `fetchCamperJourney` already had no year floor. Family rows now render with a muted de-emphasis tag in `CampJourneyTimeline` (and in `CamperDetailsPanel`, the board popout — a second consumer of the same widened data) instead of being hidden, so the original noise concern is addressed visually rather than by dropping the data. `CAMPER_DETAIL_TYPES` (a separate consumer, the camper detail page's current-year fetch) is unchanged and keeps its no-family curation.
- Widens `CurrentYearContext`'s year-navigation window from a fixed 5-year lookback to a floor of 2017 (where summer data starts), so the list/board surfaces can reach 2017-2021 alongside the journey timeline, which already read prior years directly.
- Code review (floor 60) on the first pass of this PR found a real regression the widening reopened: a lone family-camp bunk assignment could attach to an unrelated same-year summer row via `fetchCamperJourney`'s year-fallback — the exact leak `fb1a88d2` closed for current-year views. Confirmed against production data (2,218 `family`-type `bunk_assignments` rows exist) and closed with a boundary guard + regression test. See the review comment thread and the follow-up PR comment for the full list of smaller fixes from that pass.

## Notes
- The pinning test at `sessionTypePredicates.test.ts` (`CAMPER_JOURNEY_TYPES ... not.toContain('family')`) was deliberately restated, not deleted — both the test and the `sessionTypePredicates.ts` doc comment now explain the reversal and its reasoning.
- **#2123 is a merge blocker, not a follow-up to get to later** — see the banner above. It presents both fix branches without choosing.
- Filed #2149 as a genuine, lower-priority follow-up: a **pre-existing** bug (not caused by this PR) where `CAMPER_DETAIL_TYPES` still excludes family, so 500+/year campers whose only current-year enrollment is family camp get "Unable to load camper details" today. #2113's text explicitly deferred deciding `CAMPER_DETAIL_TYPES` alongside `CAMPER_JOURNEY_TYPES`; #2149 is what that deferral pointed at.
- #2114 (sync fix for 2021 Family Camp sessions misclassified as `other`) is a separate, already-filed PR track — this PR is not blocked on it, but 2021 family camp stays invisible here until it lands.

Closes #2113.

### DIFF
--- a/frontend/src/components/camper/CampJourneyTimeline.tsx
+++ b/frontend/src/components/camper/CampJourneyTimeline.tsx
@@ -5,6 +5,7 @@
 import { TreePine, Home } from 'lucide-react'
 import { getSessionDisplayNameFromString } from '../../utils/sessionDisplay'
 import { getStatusIndicator } from '../../utils/enrollmentFilter'
+import { isFamilySessionType } from '../../utils/sessionTypePredicates'
 import type { HistoricalRecord } from '../../hooks/camper/types'
 
 interface CampJourneyTimelineProps {
@@ -87,7 +88,7 @@ export function CampJourneyTimeline({
                         muted tag instead of being hidden, so a reader can
                         visually skip a run of family rows without losing the
                         underlying data. */}
-                    {record.sessionType === 'family' && (
+                    {isFamilySessionType(record.sessionType) && (
                       <span className="bg-muted text-muted-foreground flex-shrink-0 rounded px-1.5 py-0.5 text-[10px] font-medium">
                         Family
                       </span>
--- a/frontend/src/components/camper/CampJourneyTimeline.tsx
+++ b/frontend/src/components/camper/CampJourneyTimeline.tsx
@@ -6,7 +6,6 @@ import { TreePine, Home } from 'lucide-react'
 import { getSessionDisplayNameFromString } from '../../utils/sessionDisplay'
 import { getStatusIndicator } from '../../utils/enrollmentFilter'
 import type { HistoricalRecord } from '../../hooks/camper/types'
-import { getCampTagline } from '../../config/branding'
 
 interface CampJourneyTimelineProps {
   history: HistoricalRecord[]
@@ -27,8 +26,11 @@ export function CampJourneyTimeline({
           <TreePine className="h-5 w-5" />
           Camp Journey
         </h2>
+        {/* Program-agnostic count (#2113): the branding tagline reads as
+            summer-only ("summers at camp") and the journey now includes
+            family camp and teen years, so it's no longer paired here. */}
         <p className="text-forest-200 mt-1 text-sm">
-          {yearsAtCamp} {getCampTagline()}
+          {yearsAtCamp} {yearsAtCamp === 1 ? 'year' : 'years'} at camp
         </p>
       </div>
 
@@ -79,6 +81,18 @@ export function CampJourneyTimeline({
                       {getSessionDisplayNameFromString(record.sessionName, record.sessionType)}
                     </span>
 
+                    {/* Family-camp de-emphasis tag (#2113): family rows were
+                        excluded from the journey because they made it noisy
+                        for multi-session staff kids. Now included, they get a
+                        muted tag instead of being hidden, so a reader can
+                        visually skip a run of family rows without losing the
+                        underlying data. */}
+                    {record.sessionType === 'family' && (
+                      <span className="bg-muted text-muted-foreground flex-shrink-0 rounded px-1.5 py-0.5 text-[10px] font-medium">
+                        Family
+                      </span>
+                    )}
+
                     {/* Status indicator for non-enrolled */}
                     {statusIndicator && (
                       <span
@@ -123,7 +137,9 @@ export function CampJourneyTimeline({
         ) : (
           <div className="py-4 text-center">
             <TreePine className="text-muted-foreground/50 mx-auto mb-2 h-8 w-8" />
-            <p className="text-muted-foreground text-sm">First summer at camp!</p>
+            {/* Program-agnostic (#2113): a family-camp-only or teen-only
+                first year should not read as "summer" */}
+            <p className="text-muted-foreground text-sm">First year at camp!</p>
           </div>
         )}
       </div>
--- a/frontend/src/contexts/CurrentYearContext.test.tsx
+++ b/frontend/src/contexts/CurrentYearContext.test.tsx
@@ -3,6 +3,7 @@
  * and the isYearReady flag that prevents premature query firing.
  */
 import { describe, it, expect } from 'vitest'
+import { calculateAvailableYears as realCalculateAvailableYears } from './CurrentYearContext'
 
 // Test the year calculation logic without needing full React context
 describe('Year calculation logic', () => {
@@ -43,6 +44,28 @@ describe('Year calculation logic', () => {
 
     it('should return empty array when base year is 0 (not ready)', () => {
       expect(calculateAvailableYears(0)).toEqual([])
+    })
+  })
+
+  // #2113: the real calculateAvailableYears (exported from CurrentYearContext)
+  // must reach back to 2017, where summer data starts, instead of a fixed
+  // 5-year window that stopped list/board year navigation at 2022.
+  describe('calculateAvailableYears (real implementation, #2113 widening)', () => {
+    it('reaches back to 2017 from the current base year', () => {
+      const years = realCalculateAvailableYears(2026)
+      expect(years[0]).toBe(2026)
+      expect(years[years.length - 1]).toBe(2017)
+      expect(years).toHaveLength(10)
+      expect(years).not.toContain(2016)
+    })
+
+    it('stays descending and contiguous', () => {
+      const years = realCalculateAvailableYears(2020)
+      expect(years).toEqual([2020, 2019, 2018, 2017])
+    })
+
+    it('returns empty array when base year is 0 (not ready)', () => {
+      expect(realCalculateAvailableYears(0)).toEqual([])
     })
   })
 
--- a/frontend/src/contexts/CurrentYearContext.test.tsx
+++ b/frontend/src/contexts/CurrentYearContext.test.tsx
@@ -28,26 +28,11 @@ describe('Year calculation logic', () => {
     })
   })
 
-  describe('calculateAvailableYears', () => {
-    function calculateAvailableYears(baseYear: number, count: number = 5): number[] {
-      if (baseYear === 0) return []
-      return Array.from({ length: count }, (_, i) => baseYear - i)
-    }
-
-    it('should generate 5 years descending from base year', () => {
-      expect(calculateAvailableYears(2026)).toEqual([2026, 2025, 2024, 2023, 2022])
-    })
-
-    it('should work with any base year', () => {
-      expect(calculateAvailableYears(2024, 3)).toEqual([2024, 2023, 2022])
-    })
-
-    it('should return empty array when base year is 0 (not ready)', () => {
-      expect(calculateAvailableYears(0)).toEqual([])
-    })
-  })
-
-  // #2113: the real calculateAvailableYears (exported from CurrentYearContext)
+  // #2113: replaces a stale local re-implementation of calculateAvailableYears
+  // that asserted the old, now-replaced fixed-5-year-window contract (a
+  // private closure shadowing the real export — it stayed green regardless
+  // of the real implementation, per code review on #2113). This block tests
+  // the real calculateAvailableYears (exported from CurrentYearContext)
   // must reach back to 2017, where summer data starts, instead of a fixed
   // 5-year window that stopped list/board year navigation at 2022.
   describe('calculateAvailableYears (real implementation, #2113 widening)', () => {
@@ -66,6 +51,15 @@ describe('Year calculation logic', () => {
 
     it('returns empty array when base year is 0 (not ready)', () => {
       expect(realCalculateAvailableYears(0)).toEqual([])
+    })
+
+    // Code-review finding on #2113: a bogus baseYear below the data floor
+    // (e.g. a misconfigured backend _configured_year) must not fabricate a
+    // single-year result — that would assert an out-of-range year as
+    // "available with data" when the function's own contract says data
+    // starts at EARLIEST_AVAILABLE_YEAR.
+    it('returns empty array for a nonzero base year below the data floor', () => {
+      expect(realCalculateAvailableYears(2010)).toEqual([])
     })
   })
 
--- a/frontend/src/contexts/CurrentYearContext.tsx
+++ b/frontend/src/contexts/CurrentYearContext.tsx
@@ -5,10 +5,21 @@ import { useSyncStatusAPI } from '../hooks/useSyncStatusAPI'
 
 const STORAGE_KEY = 'bunking-current-year'
 
-// Calculate available years based on a base year
-function calculateAvailableYears(baseYear: number): number[] {
+/**
+ * Earliest year with historical attendee data — summer data starts in 2017
+ * (#2113). Was a fixed 5-year window, which blocked list/board year
+ * navigation to 2017-2021 even though the underlying data (and the camper
+ * journey timeline, which reads prior years directly) goes back that far.
+ */
+export const EARLIEST_AVAILABLE_YEAR = 2017
+
+// Calculate available years: descending from baseYear back through the
+// earliest year with data, rather than a fixed-size window (#2113).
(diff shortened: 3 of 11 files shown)


frontend/src/hooks/camper/fetchCamperJourney.ts
```diff
@@ -1,9 +1,9 @@
 /**
  * Shared prior-year journey source. Lists every prior year a camper was
- * ENROLLED (curated types: summer + teen, no family), labeling each row with its
- * bunk/day-group when an assignment exists. Sourcing from attendees (not
- * bunk_assignments) is what surfaces 2022 (a CampMinder export gap), teens, and
- * family camp uniformly.
+ * ENROLLED (curated types: summer + teen + family, see #2113), labeling each
```
adamflagg, line 3: **[Correctness — CONFIRMED] Family-camp bunk assignments now leak onto unrelated same-year rows via the year-fallback**

`CAMPER_JOURNEY_TYPES` (widened by this PR to include `family`) drives `typeFilter` here, which now gates **both** the `attendees` query (line 62) and this `bunk_assignments` query. The old comment above still claims this restriction prevents "a lone family-camp bunk in a year [being] attached to a summer row via the year-fallback below" — that protection came entirely from `family` being *absent* from the type set. It no longer holds.

The fallback at line 125 (`if (!match && yearAssignments.length === 1) match = yearAssignments[0]`) keys on "exactly one assignment *in the year*", not "one assignment for this record's session". Concrete trigger:

```
2019 attendees: main "Session 2" (cm_id 100, no summer bunk synced)
                family "Winter Family Weekend" (cm_id 900)
2019 bunk_assignments: [ {session cm_id 900, bunk "Cabin FC-2"} ]   ← only one
```
For the `main` row: exact match on cm_id 100 fails → `yearAssignments.length === 1` → attaches "Cabin FC-2" (the family lodging unit) to the summer row. The timeline renders the camper's summer session with the family-camp cabin as if it were their bunk. It leaks the other direction too (a lone summer bunk borrowing onto a family row), and can also *suppress* a real bunk label once the year now has 2 assignments instead of 1.

This is exactly the leak `fb1a88d2` closed for `useCamperEnrollment.ts` (which still guards with `buildSummerSessionTypeFilter()` for its assignments query) — reopened here for the journey. The only regression test for this (`fetchCamperJourney.test.ts`, "no family-camp bunk leak") was rewritten to assert the *new* leaky filter string rather than kept as a behavioral guard; no test exercises a same-year main+family pair.

Fix: keep `family` in the attendees filter but exclude it from the assignments filter (or any non-matching-session-type from the fallback), or scope the length-1 fallback to `match.expand.session?.session_type === session?.session_type`.

frontend/src/hooks/camper/fetchCamperJourney.ts
```diff
@@ -1,9 +1,9 @@
...
+ * row with its bunk/day-group when an assignment exists. Sourcing from
```
adamflagg, line 4: **[Correctness — PLAUSIBLE, pre-existing gap newly exercised] AG-collapse sentinel guard is missing here, unlike its current-year twin**

`enrolledByYear` is seeded from `att.expand.session?.cm_id` for *every* attendee this year (line 75-79) with no `> 0` guard, then an AG row is dropped via `!enrolledByYear.get(att.year)?.has(session.parent_id)` (line 86) — the comment claims "an unmatched parent_id (e.g. 0) is never present in enrolledByYear → kept," but that's only true if no *other* session that year happens to have `cm_id === 0`.

Compare the current-year twin, `frontend/src/hooks/camper/agCollapse.ts`, which explicitly guards this: `e.parentId > 0 && enrolledCmIds.has(e.parentId)`, with a doc comment stating *"Without the guard a cm_id-less session (sessionCmId 0) would seed 0 into the enrolled set and silently collapse an unrelated parentless AG row."* This function has no equivalent guard.

This bug is pre-existing (not introduced by this diff), but this PR widens the `attendees` query to include `family`, meaningfully increasing the number of distinct session rows pooled into `enrolledByYear` per year — raising the odds that some session with a `cm_id === 0` sentinel (a real, documented default per the sync convention) lands in the set and silently drops an unrelated AG-only row for that year, blanking it from the journey. Worth adding the same `> 0` guard here for consistency with `agCollapse.ts`, especially now that more rows flow through this path.

frontend/src/components/camper/CampJourneyTimeline.tsx
```diff
@@ -27,8 +26,11 @@ export function CampJourneyTimeline({
           <TreePine className="h-5 w-5" />
           Camp Journey
         </h2>
+        {/* Program-agnostic count (#2113): the branding tagline reads as
+            summer-only ("summers at camp") and the journey now includes
+            family camp and teen years, so it's no longer paired here. */}
         <p className="text-forest-200 mt-1 text-sm">
-          {yearsAtCamp} {getCampTagline()}
+          {yearsAtCamp} {yearsAtCamp === 1 ? 'year' : 'years'} at camp
```
adamflagg, line 34: **[Correctness — CONFIRMED, cross-file] The header count is relabeled to sound program-agnostic while the underlying number stays summer-only**

`yearsAtCamp` here is `camper.years_at_camp`, CampMinder's `YearsAtCamp` field (synced in `pocketbase/sync/persons.go`), which excludes family/adult/b'mitzvah attendance — it counts summer + SCIT/TLI only. Before this PR the pairing was `{yearsAtCamp} {getCampTagline()}` (e.g. "3 summers at camp") — summer-scoped wording next to a summer-scoped number, internally consistent. This diff changes the wording to `{yearsAtCamp} {yearsAtCamp === 1 ? 'year' : 'years'} at camp` at the exact moment the body below starts rendering visible, tagged `Family` rows.

Failure scenario: a camper with 2 summers and 5 distinct family-camp years now renders **"2 years at camp"** directly above a 7-row timeline. The old wording was merely narrow; the new wording is a specific, false claim about what's being counted.

This is related to, but distinct from, the already-filed #2123 (which tracks `years_at_camp` numerically disagreeing with fetched history in general) — the wording problem here is new in this diff and would persist even after #2123's numeric fix, since it's about what noun pairs with the number, not the number itself. Worth at least keeping "summers" (or deriving the count from `history` instead of `camper.years_at_camp`).

frontend/src/components/camper/CampJourneyTimeline.tsx
```diff
@@ -79,6 +81,18 @@ export function CampJourneyTimeline({
                       {getSessionDisplayNameFromString(record.sessionName, record.sessionType)}
                     </span>
 
+                    {/* Family-camp de-emphasis tag (#2113): family rows were
+                        excluded from the journey because they made it noisy
+                        for multi-session staff kids. Now included, they get a
+                        muted tag instead of being hidden, so a reader can
+                        visually skip a run of family rows without losing the
+                        underlying data. */}
+                    {record.sessionType === 'family' && (
```
adamflagg: **[Correctness — CONFIRMED, cross-file] The PR's own noise mitigation (de-emphasis tag) is applied to only one of two live consumers of the widened journey**

`CampJourneyTimeline.tsx` gets this "Family" tag, but `CamperDetailsPanel.tsx` — the board pop-in/popout panel (mounted from `RightPanelContainer.tsx`, `CamperPopout/index.tsx`, `BunkingBoardByArea.tsx`, i.e. actively used, unlike the orphaned `CamperTooltip.tsx`) — renders the same widened `fetchCamperJourney()` output (`historicalData.map(...)` around line 1043 of that file) with **no** visual differentiation between family and summer rows.

The PR body's own justification for reversing the "no family" exclusion is: *"Family rows now render with a muted de-emphasis tag in `CampJourneyTimeline` instead of being hidden, so the original noise concern is addressed visually rather than by dropping the data."* That's true for exactly one of the two live render surfaces. A staffer opening the board popout for a multi-session staff kid gets exactly the undifferentiated, noisy list the original exclusion existed to prevent — just via a different component.

Given `record.sessionType === 'family'` is a raw inline check (see the tag below), the fix (add the same tag to `CamperDetailsPanel.tsx`'s row renderer, or factor a shared row/tag component) is small but was missed.

adamflagg: **[Convention — CONFIRMED] Inline `session_type === 'family'` check violates this module's own stated rule, on an untyped field**

`sessionTypePredicates.ts` (the file this same PR edits to widen `CAMPER_JOURNEY_TYPES`) states at its top: *"All session_type semantics belong here. Never write inline `session_type === 'ag'` checks in components or other utils — import a named predicate instead."* This line is exactly that pattern with `'family'` in place of `'ag'`, and there is no `isFamilyCampSessionType`/equivalent predicate defined anywhere to import instead.

Compounding: `HistoricalRecord.sessionType` (`frontend/src/hooks/camper/types.ts`) is typed as plain `string`, not a narrow literal union. A typo (`'Family'`) or a future schema rename compiles clean with no error, and the de-emphasis tag silently stops rendering — reintroducing the "noisy for multi-session staff kids" problem this exact tag exists to prevent, with no visible signal that anything broke.

frontend/src/contexts/CurrentYearContext.test.tsx
```diff
@@ -46,6 +47,28 @@ describe('Year calculation logic', () => {
     })
   })
```
adamflagg: **[Test hygiene — CONFIRMED] Stale shadow re-implementation of `calculateAvailableYears` still asserts the old, replaced 5-year-window contract**

This file (`describe('calculateAvailableYears', ...)`, starting a few lines above this hunk) defines a **local** closure — `function calculateAvailableYears(baseYear, count = 5)` — and asserts `calculateAvailableYears(2026)` equals `[2026, 2025, 2024, 2023, 2022]`. It is never bound to the real exported `calculateAvailableYears` from `CurrentYearContext.tsx`, so it can't fail regardless of what the real (now 2017-floored) implementation does — it green-lights the old 5-year contract forever, right next to the new `describe('calculateAvailableYears (real implementation, #2113 widening)', ...)` block added by this PR that correctly imports and tests the real export.

A future reader grepping this file for the year-window contract hits the stale, still-passing "should generate 5 years descending from base year" test and may conclude the window is still fixed at 5, or "fix" the real implementation to satisfy it. Since the real function is now exported and tested directly, the old shadow block (and its private re-implementation) should be deleted rather than left alongside.

frontend/src/contexts/CurrentYearContext.tsx
```diff
@@ -5,10 +5,21 @@ import { useSyncStatusAPI } from '../hooks/useSyncStatusAPI'
...
+ */
+export const EARLIEST_AVAILABLE_YEAR = 2017
+
+// Calculate available years: descending from baseYear back through the
+// earliest year with data, rather than a fixed-size window (#2113).
+// eslint-disable-next-line react-refresh/only-export-components
+export function calculateAvailableYears(baseYear: number): number[] {
   if (baseYear === 0) return []
-  return Array.from({ length: 5 }, (_, i) => baseYear - i)
+  const length = Math.max(1, baseYear - EARLIEST_AVAILABLE_YEAR + 1)
```
adamflagg: **[Correctness — PLAUSIBLE] `Math.max(1, …)` fabricates a bogus year instead of the more honest empty result**

`const length = Math.max(1, baseYear - EARLIEST_AVAILABLE_YEAR + 1)` — if `baseYear` is ever a nonzero value below 2017 (e.g. a misconfigured/anomalous backend `_configured_year`), this clamps `length` to 1 and returns `[baseYear]`, i.e. asserts that out-of-range year as an "available year with data" — contradicting the function's own documented floor (`EARLIEST_AVAILABLE_YEAR`, "earliest year with historical attendee data").

Without the clamp, `Array.from({length: negative})` already coerces to length 0 and naturally returns `[]` (confirmed: `Array.from({length: -1}, (_,i)=>i)` → `[]`), which is arguably the more correct behavior for an out-of-range `baseYear` — closer to "not ready" than to "exactly one valid year." A misconfigured backend would then look like a working one-season deployment instead of surfacing as empty/not-ready. Backend clamps (`pocketbase/sync/year.go`) make this unreachable in practice today, but the dead defensive code produces the wrong answer if it ever does become reachable.

frontend/src/hooks/camper/fetchCamperJourney.test.ts
```diff
@@ -73,27 +73,35 @@ describe('fetchCamperJourney', () => {
...
+    expect(filter).not.toContain('"bmitzvah"') // still excluded — not a journey type
     expect(filter).toContain('session.session_type = "scit"')
   })
 
-  it('restricts the bunk_assignments query to journey session types (no family-camp bunk leak)', async () => {
-    // A summer enrollment with no summer bunk + a lone family-camp bunk that year
-    // must NOT have the family bunk attached via the length===1 fallback. The
-    // query itself excludes non-journey (family) session types — mirrors the
-    // fb1a88d2 fix applied to current-year views in useCamperEnrollment.
+  it('restricts the bunk_assignments query to journey session types (family included, others still excluded)', async () => {
```
adamflagg, line 97: **[Test coverage gap — CONFIRMED] The widening ships with zero behavioral coverage of the mixed same-year case**

Both `#2113` tests in this block assert only on the outgoing filter *string* (`toContain('session.session_type = "family"')`), never on the actual join/label output. No test in this file constructs a year with both a `main` and a `family` attendee plus exactly one assignment — the precise scenario that trips the fallback bug at `fetchCamperJourney.ts:125` (see that line's finding). The comment here even states the intended behavior as settled ("a lone family-camp bunk legitimately participates in the fallback like any other journey type") without a test pinning it either way.

Net effect: the one regression test that used to guard this class of leak (the old "no family-camp bunk leak" test) was replaced with a filter-string assertion, not a behavioral one — so neither the current bug nor a future fix to it will show up as a test failure here. A test with `[main-2023-nobunk, family-2023-cm_id900]` attendees and one `assignment(2023, 900, 'Cabin')`, asserting the `main` row's `bunkName` is `undefined`, is the missing case.

frontend/src/utils/sessionTypePredicates.ts
```diff
@@ -36,11 +36,32 @@ export const QUEST_SESSION_TYPES = ['quest'] as const
...
+ * was pure data loss, not curation. The noise concern is still real; it's
+ * handled with a visual de-emphasis on family rows in CampJourneyTimeline
+ * rather than by hiding the data.
+ *
+ * CAMPER_DETAIL_TYPES below is a separate consumer (the camper detail page's
+ * current-year fetch) and keeps the original no-family curation — #2113 did
+ * not decide that one.
  */
-export const CAMPER_JOURNEY_TYPES = ['main', 'embedded', 'ag', 'quest', 'scit', 'tli'] as const
+export const CAMPER_JOURNEY_TYPES = [
```
adamflagg: **[Reuse/Altitude — CONFIRMED, minor] `CAMPER_JOURNEY_TYPES`/`CAMPER_DETAIL_TYPES` are hand-duplicated array literals**

Before this PR these two arrays were identical (`['main','embedded','ag','quest','scit','tli']`), documented with a shared "summer + teen, no family" comment. This diff makes them diverge by hand-typing a near-duplicate 7-element array (`[...,'family']`) rather than deriving one from the other. Nothing enforces the relationship structurally — only prose comments — so a future addition to one (e.g. a new summer session type) can silently forget the other.

Consider `export const CAMPER_JOURNEY_TYPES = [...CAMPER_DETAIL_TYPES, 'family'] as const` so "journey = detail + family" is expressed in code, not just in a comment.

PR #2128, opened 2026-08-08
fix(pb): close guardUnitYear's one-directional dependent gap

adamflagg:
## Summary

`guardUnitYear` checks a `lodging_units` row's own `year` against the parent it points AT (area, parent_unit), but nothing previously re-validated the rows that point BACK at a parent whose `year` gets edited in place. Two gaps, both named in kindred#2039:

- `lodging_areas` carried no year binding at all — a superuser or `bunking.manage` PATCH editing an existing area's own `year` hit no check, silently stranding every `lodging_units` row pointing at it.
- Editing a `lodging_units` row's own `year` was checked against its own area/parent_unit, but not against the rows depending on it: `lodging_availability`, `lodging_assignments`, `lodging_assignments_draft`, `lodging_slot_merges`.

This adds `guardYearImmutable`, bound via `OnRecordUpdate` on `lodging_areas` and `lodging_units`, which refuses to change a row's `year` in place once any dependent row references it. Cascading the new year onto every dependent row would be a materially bigger change than any other guard in this file makes, and nothing in the product needs to re-season a row in place — `copyAreas`/`copyUnits` (roll-forward) already create a new record for the target year rather than mutating the source, so a categorical refusal costs the roll-forward nothing.

The comparison is `e.Record.Original().GetInt("year")` against the new value, **not** whether `year` is present in the write — `LodgingUnitForm.handleSubmit` (`frontend/src/components/admin/lodging/LodgingUnitForm.tsx`) sends `year: openedYear` on every save, routine edits included, so a presence test would block ordinary saves on any unit with a dependent. A test pins this: a routine edit that resends the unchanged year, on a unit with a real dependent row, must succeed.

`lodging_units.parent_unit` is deliberately left uncovered as a dependent — scoped to the four collections the issue names; a child unit pointing at a re-seasoned parent via `parent_unit` is the same shape of gap and isn't addressed here.

## Test plan
- [x] `go build ./...`
- [x] `go test ./...` (full suite, exit 0)
- [x] `scripts/dev/verify-lodging-schema.sh` — PocketBase boots against a throwaway DB, all migrations apply
- [x] `scripts/dev/verify-lodging-seed.sh` — seed path exercises the hooks end-to-end
- [x] New tests: routine same-year resend is allowed; area/unit year change is refused when a dependent exists (availability single-relation, assignments multi-relation, area→unit)
- [x] Pre-push hooks green (ruff, go-build, mypy, pb-js-lint, shellcheck, markdownlint, api-types-freshness, type-check, pytest 6020 passed)

Closes #2067.

## Folded in: #2146 — no test enforced the two ref maps staying in sync

`yearImmutableRefs` is a hand-maintained reverse of `yearScopedRefs` and nothing checked the correspondence. Adding a year-scoped table to `yearScopedRefs` alone is silent — the new table's own writes are still checked, so its tests pass, while `guardYearImmutable` stops covering a parent's year edit against it.

`TestYearRefMapsAgreeOnEveryRelation` walks both maps in both directions. The one deliberate exception, `lodging_units.parent_unit` (scoped out by #2067, already documented in both maps), is allowlisted with its reason — and the allowlist is itself checked, so the test also fails if that pair ever gains a mirror or stops being declared without the allowlist following. No production behaviour change; the map comments are repointed from the issue to the test that now enforces them.

Mutation-checked all four assertion paths (drop a mirror; add a bogus dependent; mirror the allowlisted pair; drop the allowlisted pair from `yearScopedRefs`) — each fails with its own message, `go test` exit 1.

Closes #2146.

### DIFF
--- a/pocketbase/lodging/hooks.go
+++ b/pocketbase/lodging/hooks.go
@@ -97,11 +97,12 @@ type yearScopedRef struct {
 //     point AT it: lodging_availability, lodging_assignments,
 //     lodging_assignments_draft, lodging_slot_merges.
 //
-// What that does NOT close: lodging_units is ALSO a valid TARGET via
-// parent_unit (line 111 below), the identical shape to the lodging_areas
-// bullet above, but yearImmutableRefs carries no reverse entry for it -- see
-// that map's own doc comment for why, and kindred#2146 for the maintenance
-// hazard this un-mirrored entry is one instance of.
+// What that does NOT close: lodging_units is ALSO a valid TARGET via the
+// parent_unit entry below, the identical shape to the lodging_areas bullet
+// above, but yearImmutableRefs carries no reverse entry for it -- see that
+// map's own doc comment for why. It is the single allowlisted exception in
+// TestYearRefMapsAgreeOnEveryRelation, so it stays a stated scoping choice
+// rather than becoming a silent one (kindred#2146).
 //
 // Cascading the new year onto every dependent row on a parent's write would
 // be a materially bigger change than this guard makes anywhere else, so
@@ -130,10 +131,15 @@ var yearScopedRefs = map[string][]yearScopedRef{
 // real id (see countAssignments's own comment). The reverse mistake is not
 // silent the same way -- checked directly, a joined "unit.id ?= {:id}" against
 // a single relation still matches the dependent row correctly, so writing it
-// that way is merely non-idiomatic, not wrong. kindred#2146 tracks the
-// hand-typed-per-entry risk this comment describes, including mechanizing the
-// derivation instead of leaving it to be gotten right by hand at each entry
-// below.
+// that way is merely non-idiomatic, not wrong.
+//
+// Hand-typing per entry buys that correctness at the cost of the two maps
+// drifting apart unnoticed, so it is paid for by a test rather than by an
+// abstraction: TestYearRefMapsAgreeOnEveryRelation walks both maps and fails
+// the moment an entry in one has no counterpart in the other (kindred#2146).
+// The filter string itself is still on the author to get right -- the test
+// checks that an edge is declared from both ends, not that its filter uses
+// the arity-correct operator.
 type dependentRef struct {
 	collection string
 	filter     string
@@ -149,8 +155,10 @@ type dependentRef struct {
 // collections its own body names below. A child unit pointing at a
 // re-seasoned parent via parent_unit is the same shape of gap and is not
 // covered by this guard. It is also the one entry in yearScopedRefs with no
-// hand-typed reverse here -- kindred#2146 tracks that as a maintenance
-// hazard in this map's own shape, not just as a scoping choice.
+// hand-typed reverse here, and therefore the sole allowlisted exception in
+// TestYearRefMapsAgreeOnEveryRelation (kindred#2146) -- which also fails if
+// the exception is ever mirrored or dropped without the allowlist following,
+// so this paragraph cannot quietly go out of date.
 var yearImmutableRefs = map[string][]dependentRef{
 	collectionAreas: {
 		{collection: collectionUnits, filter: "area = {:id}"},
--- a/pocketbase/lodging/hooks.go
+++ b/pocketbase/lodging/hooks.go
@@ -100,7 +100,7 @@ type yearScopedRef struct {
 // What that does NOT close: lodging_units is ALSO a valid TARGET via
 // parent_unit (line 111 below), the identical shape to the lodging_areas
 // bullet above, but yearImmutableRefs carries no reverse entry for it -- see
-// that map's own doc comment for why, and kindred#2148 for the maintenance
+// that map's own doc comment for why, and kindred#2146 for the maintenance
 // hazard this un-mirrored entry is one instance of.
 //
 // Cascading the new year onto every dependent row on a parent's write would
@@ -130,9 +130,10 @@ var yearScopedRefs = map[string][]yearScopedRef{
 // real id (see countAssignments's own comment). The reverse mistake is not
 // silent the same way -- checked directly, a joined "unit.id ?= {:id}" against
 // a single relation still matches the dependent row correctly, so writing it
-// that way is merely non-idiomatic, not wrong. kindred#2148 tracks
-// mechanizing this derivation instead of leaving it to be gotten right by
-// hand at each entry below.
+// that way is merely non-idiomatic, not wrong. kindred#2146 tracks the
+// hand-typed-per-entry risk this comment describes, including mechanizing the
+// derivation instead of leaving it to be gotten right by hand at each entry
+// below.
 type dependentRef struct {
 	collection string
 	filter     string
@@ -148,7 +149,7 @@ type dependentRef struct {
 // collections its own body names below. A child unit pointing at a
 // re-seasoned parent via parent_unit is the same shape of gap and is not
 // covered by this guard. It is also the one entry in yearScopedRefs with no
-// hand-typed reverse here -- kindred#2148 tracks that as a maintenance
+// hand-typed reverse here -- kindred#2146 tracks that as a maintenance
 // hazard in this map's own shape, not just as a scoping choice.
 var yearImmutableRefs = map[string][]dependentRef{
 	collectionAreas: {
--- a/pocketbase/lodging/hooks.go
+++ b/pocketbase/lodging/hooks.go
@@ -325,9 +325,9 @@ func guardYearImmutable(e *core.RecordEvent, refs []dependentRef) error {
 		if len(records) > 0 {
 			return apis.NewBadRequestError(
 				fmt.Sprintf(
-					"Cannot change %q's year from %d to %d: %s row(s) reference it. "+
-						"Use the roll-forward to move it to a new season instead of "+
-						"re-seasoning an existing row in place.",
+					"Cannot change %q's year from %d to %d: at least one %s row "+
+						"references it. Use the roll-forward to move it to a new "+
+						"season instead of re-seasoning an existing row in place.",
 					e.Record.GetString("code"), oldYear, newYear, ref.collection,
 				),
 				nil,
--- a/pocketbase/lodging/hooks.go
+++ b/pocketbase/lodging/hooks.go
@@ -84,24 +84,23 @@ type yearScopedRef struct {
 // SCOPE: this guard is one-directional. It checks the row being written
 // against the OUTGOING relations listed in its own entry above; it never
 // re-checks rows elsewhere that point back AT the thing just written. Two
-// gaps follow, both residual as of kindred#2039, not regressions:
+// gaps followed from that, both residual as of kindred#2039, not
+// regressions, and both closed below by guardYearImmutable (kindred#2067)
+// rather than by extending this map:
 //
 //   - lodging_areas is a valid TARGET above (lodging_units.area resolves into
-//     it) but is not itself a key in this map, so it carries no binding at
-//     all. A superuser or bunking.manage PATCH that edits an existing area
-//     row's own `year` in place hits no check, and every lodging_units row
-//     already pointing at that area -- written for whatever year it was
-//     created -- is now silently cross-season.
-//   - Editing a lodging_units row's own `year` is checked against that row's
-//     OWN area/parent_unit, but nothing re-validates the rows that point AT
-//     it: lodging_availability, lodging_assignments,
-//     lodging_assignments_draft, lodging_slot_merges. Those keep whatever
-//     year they were written with, now possibly disagreeing with the unit's
-//     new one.
+//     it) but is not itself a key in this map, so editing an existing area
+//     row's own `year` in place hit no check here, leaving every
+//     lodging_units row already pointing at it silently cross-season.
+//   - Editing a lodging_units row's own `year` was checked against that
+//     row's OWN area/parent_unit, but nothing re-validated the rows that
+//     point AT it: lodging_availability, lodging_assignments,
+//     lodging_assignments_draft, lodging_slot_merges.
 //
-// Closing either gap means cascading to every DEPENDENT row on a parent's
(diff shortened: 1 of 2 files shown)


pocketbase/lodging/hooks.go
```diff
@@ -113,6 +112,42 @@ var yearScopedRefs = map[string][]yearScopedRef{
...
+// covered by this guard.
+var yearImmutableRefs = map[string][]dependentRef{
+	collectionAreas: {
+		{collection: collectionUnits, filter: "area = {:id}"},
+	},
+	collectionUnits: {
+		{collection: collectionAvailability, filter: "unit = {:id}"},
+		{collection: collectionAssignments, filter: "units.id ?= {:id}"},
+		{collection: collectionAssignmentsDraft, filter: "units.id ?= {:id}"},
+		{collection: collectionSlotMerges, filter: "unit = {:id}"},
```
adamflagg, line 170: **Test-coverage gap, confirmed by mutation test.** Of the four `yearImmutableRefs[collectionUnits]` entries, only `lodging_availability` (single-relation, `unit = {:id}`) and `lodging_assignments` (multi-relation, `units.id ?= {:id}`) get a dedicated `TestGuardYearImmutableRejects...` test. `lodging_assignments_draft` (line 146) and `lodging_slot_merges` (line 147) have none.

Verified empirically: mutating the `lodging_assignments_draft` filter from `"units.id ?= {:id}"` to the broken bare form `"units = {:id}"` — exactly the silent-zero-match mistake this file's own `dependentRef` comment warns about — leaves `go test ./lodging/...` fully green. A regression on either of these two filters (e.g. a future edit that drops the `.id` join, or gets `=` vs `?=` backwards) would ship silently: a unit's year could change while a scenario draft or slot-merge override still references it.

Worth a test per entry, the same way `TestMultiRelationAnyMatchFilter` pins `countAssignments`'s filter shape.

pocketbase/lodging/hooks.go
```diff
@@ -113,6 +112,42 @@ var yearScopedRefs = map[string][]yearScopedRef{
...
+// lodging_units.parent_unit is deliberately absent as a dependent of
+// collectionUnits here -- kindred#2067 scoped this fix to the four
+// collections its own body names below. A child unit pointing at a
+// re-seasoned parent via parent_unit is the same shape of gap and is not
+// covered by this guard.
+var yearImmutableRefs = map[string][]dependentRef{
+	collectionAreas: {
+		{collection: collectionUnits, filter: "area = {:id}"},
+	},
+	collectionUnits: {
```
adamflagg, line 166: **Residual gap, disclosed but worth a tracking issue.** `yearImmutableRefs[collectionUnits]` has no entry for `lodging_units` itself via `parent_unit` (documented as deliberate at lines 134-138, and in the PR body). But the SCOPE comment above `yearScopedRefs` now reads (line 88): *"Two gaps followed from that... **both closed below by guardYearImmutable**"* — that's only true for the two gaps as originally scoped (which never mentioned `parent_unit`), not for every outgoing relation `yearScopedRefs` tracks for `lodging_units` (which includes `parent_unit`, line 111).

Concretely: unit A is `parent_unit` of unit B, and A has no rows in the four watched dependent tables. PATCHing A's `year` succeeds silently — B is left pointing at a cross-season parent, the identical bug shape this PR closes for the other four relations. Worse, B is then locked out of unrelated future edits, since the pre-existing `guardUnitYear` will refuse B's next save once A.year and B.year disagree.

Given `yearImmutableRefs` is table-driven, closing this costs one more `dependentRef{collection: collectionUnits, filter: "parent_unit = {:id}"}` entry, not the cascade the guard's design note argues against — worth a fast follow-up issue rather than leaving it implicit in two comments.

pocketbase/lodging/hooks.go
```diff
@@ -113,6 +112,42 @@ var yearScopedRefs = map[string][]yearScopedRef{
...
+// guardYearImmutable binds to, every OTHER collection that may hold a foreign
+// key pointing back at it. yearScopedRefs asks what a row points AT;
+// yearImmutableRefs asks what points AT a row.
+//
+// lodging_units.parent_unit is deliberately absent as a dependent of
+// collectionUnits here -- kindred#2067 scoped this fix to the four
+// collections its own body names below. A child unit pointing at a
+// re-seasoned parent via parent_unit is the same shape of gap and is not
+// covered by this guard.
+var yearImmutableRefs = map[string][]dependentRef{
```
adamflagg, line 162: **Altitude: two hand-maintained, structurally-mirrored maps with nothing enforcing they stay in sync.** `yearImmutableRefs` is described as "the REVERSE of `yearScopedRefs`" (line 129), and today every `{field, target}` in `yearScopedRefs` has a hand-typed reverse `{collection, filter}` here — except `lodging_units.parent_unit`, which is exactly the entry a mechanical transpose would have produced and is the gap flagged in the sibling comment on this thread.

No test asserts the two maps are transposes of each other. A future engineer adding a new year-scoped table to `yearScopedRefs` (a new placement table with a `unit`/`units` relation into `lodging_units`) has no signal that `yearImmutableRefs` needs a matching reverse entry, and the reverse's filter string is hand-typed rather than derived from the schema, so getting `=` vs `.id ?=` backwards is compile-clean and matches zero rows.

pocketbase/lodging/hooks.go
```diff
@@ -209,6 +282,61 @@ func guardUnitYear(app core.App, rec *core.Record) error {
...
+			map[string]any{"id": e.Record.Id},
+		)
+		if err != nil {
+			return fmt.Errorf("checking %s for rows depending on %s: %w",
+				ref.collection, e.Record.Id, err)
+		}
+		if len(records) > 0 {
+			return apis.NewBadRequestError(
+				fmt.Sprintf(
+					"Cannot change %q's year from %d to %d: %s row(s) reference it. "+
```
adamflagg: **Message reads oddly: "row(s)" never resolves to a count.** `%s row(s) reference it` interpolates `ref.collection` (e.g. `lodging_units`), not a number — because the existence check is capped at `limit 1` (line 317), so there's no count to report. The result: *"Cannot change "test-unit-a"'s year from 2027 to 2028: lodging_units row(s) reference it."*

Compare `guardUnitDelete` and `guardAliasDelete` in this same file, which both query unlimited and interpolate a real `%d` count (`"%d lodging placement(s) reference it"`). Either drop the pluralization here (`"at least one lodging_units row references it"`) or make the message's shape match its siblings' — as written it implies a count is coming and then doesn't supply one.

pocketbase/lodging/hooks.go
```diff
@@ -158,6 +193,44 @@ func wireHooks(app core.App) {
...
+
+	// guardYearImmutable's dependents run in the OPPOSITE direction from the
+	// loop above (see yearImmutableRefs's doc comment), so it gets its own
+	// loop rather than being folded into it. Same existence-check guard on
+	// the collection being bound to, since lodging_areas is not otherwise
+	// bound in this function. UPDATE only: a create cannot have a stale
+	// dependent, because nothing can reference an id before the record that
+	// owns it exists (same reasoning as guardUnitParentCycle's absent create
+	// binding).
+	for name := range yearImmutableRefs {
```
adamflagg, line 239: **Reuse: three near-identical "does this collection exist? warn and skip" stanzas now live in `wireHooks`** — the pre-existing loop for `yearScopedRefs` (line 173), this loop's own parent-collection check (line 206), and its nested dependent-collection check (line 222) all repeat the same `if _, err := app.FindCollectionByNameOrId(x); err != nil { app.Logger().Warn(...); continue }` shape, five-plus lines each. A small `collectionExists(app, name) bool` (or a variant returning the warn-and-skip decision) would collapse each site to one line while keeping each call site's distinct log message, and remove the risk of the three copies drifting apart on a future edit.

pocketbase/lodging/hooks.go
```diff
@@ -113,6 +112,42 @@ var yearScopedRefs = map[string][]yearScopedRef{
...
+// key pointing back at it. yearScopedRefs asks what a row points AT;
+// yearImmutableRefs asks what points AT a row.
+//
+// lodging_units.parent_unit is deliberately absent as a dependent of
+// collectionUnits here -- kindred#2067 scoped this fix to the four
+// collections its own body names below. A child unit pointing at a
+// re-seasoned parent via parent_unit is the same shape of gap and is not
+// covered by this guard.
+var yearImmutableRefs = map[string][]dependentRef{
+	collectionAreas: {
```
adamflagg, line 163: **Simplification: `dependentRef.filter` is hand-typed per entry even though it's fully determined by (field name, single-vs-multi).** Two entries below are the literal duplicate string `"unit = {:id}"` (availability, slot_merges) and two are the literal duplicate `"units.id ?= {:id}"` (assignments, assignments_draft). The doc comment above (lines 119-123) itself frames getting `=` vs `.id ?=` backwards as "silent" and dangerous — a `dependentRef{field string, multi bool}` with the join syntax centralized in one place (derived once, tested once) removes the one spot this needs to be right per entry, instead of five, and would have made the untested `assignments_draft`/`slot_merges` entries (see the coverage-gap comment on this thread) inherit correctness from the tested ones automatically instead of needing their own pinning test.

pocketbase/lodging/hooks.go
```diff
@@ -113,6 +112,42 @@ var yearScopedRefs = map[string][]yearScopedRef{
...
 }
 
+// dependentRef names one collection that may hold a foreign key pointing back
+// at the row guardYearImmutable is checking, and the filter to find it. The
+// filter is baked in per entry rather than derived from a single/multi flag,
+// for the same reason countAssignments's filter is hardcoded rather than
+// assembled (see its comment): `?=` plus the `.id` join is only correct for a
+// MULTI-valued relation (`units`, MaxSelect 20); a single relation (`unit`,
+// `area`, MaxSelect 1) stores the id as a plain string and needs a plain `=`.
+// Getting that backwards is silent -- both a bare "units = {:id}" and a
```
adamflagg: **Doc-comment claim doesn't hold up under test.** This comment states: *"both a bare `\"units = {:id}\"` and a joined `\"unit.id ?= {:id}\"` compile and match ZERO rows against a real id."* The first half (bare `=` on a multi-relation) is correct and is exactly what `countAssignments`'s comment already establishes. The second half — a joined `.id ?=` form used against a single-valued (MaxSelect 1) relation — was checked empirically: mutating `lodging_availability`'s filter from `"unit = {:id}"` to `"unit.id ?= {:id}"` still correctly matches the dependent row and produces the same refusal. So the joined form isn't actually broken on a single relation; only the reverse mistake (bare `=` on a multi-relation) is. Worth correcting so a future reader doesn't take the "measured, not assumed" framing at face value for a claim that wasn't actually measured the way `TestMultiRelationAnyMatchFilter` measures the multi-relation half.

pocketbase/lodging/hooks.go
```diff
@@ -209,6 +282,61 @@ func guardUnitYear(app core.App, rec *core.Record) error {
...
+// in this environment -- see the bind-time loop's comment -- rather than
+// looked up here from yearImmutableRefs directly.
+func guardYearImmutable(e *core.RecordEvent, refs []dependentRef) error {
+	oldYear := e.Record.Original().GetInt("year")
+	newYear := e.Record.GetInt("year")
+	if oldYear == newYear {
+		return e.Next()
+	}
+
+	for _, ref := range refs {
```
adamflagg, line 346: **Minor/plausible efficiency note.** When a `year` change is actually attempted on a `lodging_units` row, this loop issues up to 4 sequential `FindRecordsByFilter` round-trips (one per dependent collection) before either refusing or succeeding — including the "legal" case (a unit with zero dependents), which pays all 4 queries serially before the write is allowed through. Low impact given this is an admin-only, low-frequency path over an in-process SQLite DB, but worth noting if this guard's dependent list grows, or if it turns out to be reachable from a higher-throughput path later.

pocketbase/lodging/hooks_test.go
```diff
@@ -1504,3 +1504,147 @@ func TestGuardUnitYearSkipsAUnitsSelfReference(t *testing.T) {
...
+	if err := app.Save(rec); err != nil {
+		t.Fatalf("refused a routine edit that resent the unchanged year: %v", err)
+	}
+}
+
+// TestGuardYearImmutableRejectsAUnitYearChangeWithAnAvailabilityDependent is
+// the single-relation case (`lodging_availability.unit`, MaxSelect 1) --
+// kindred#2067's second gap: guardUnitYear checks a unit's own OUTGOING
+// relations (area, parent_unit) but never the rows pointing back AT it.
+func TestGuardYearImmutableRejectsAUnitYearChangeWithAnAvailabilityDependent(t *testing.T) {
```
adamflagg, line 1604: **Minor reuse note.** `TestGuardYearImmutableAllowsARoutineEditThatResendsTheSameYear` and this test both inline byte-for-byte identical `lodging_availability` row construction (`core.NewRecord(mustCollection(t, app, "lodging_availability")); .Set("year", 2027); .Set("unit", unit); .Set("session", seedSession(t, app))`). Not a bug, but a small `seedAvailability(t, app, unit, year)` helper (mirroring `seedUnit`/`seedArea`) would remove the duplication for these two and any future test needing an availability dependent.

adamflagg (commented):
Automated review (xhigh effort, 10 finder angles + 1-vote verify + gap sweep). No confirmed correctness bugs or crashes — the filter cardinality in `yearImmutableRefs` was checked entry-by-entry against the actual PocketBase migration schema and matches, the `oldYear == newYear` change-detection is correct (not a presence test), the bind-time existence-check loops degrade gracefully on missing collections, and no legitimate existing write path (Go sync, migrations, frontend admin forms) is broken by the new guard. 9 findings survived verification, all cleanup/test-coverage/altitude in nature — most actionable is a confirmed (mutation-tested) test-coverage gap on 2 of 4 dependent filters, and a disclosed-but-real residual gap on `lodging_units.parent_unit` as a dependent. See inline comments.

PR #2133, opened 2026-08-08
fix(api): log the exception swallowed by lodging create-race recovery

adamflagg:
## Summary

`place_party`, `set_availability`, and `set_slot_merge` in `LodgingWriteService` each recover from a failed create by treating **any** non-refusal `ClientResponseError` (not only the unique-constraint one an actual race produces) as "we lost a race": re-read, and update whatever row is found. The write is idempotent either way, so a spurious recovery still lands the row where the caller asked — but a create that failed for an unrelated reason (a `500`, a malformed relation id, a transient PocketBase fault) is now reported to the caller as a plain success, and nothing anywhere records that the original error happened.

The guard's width is deliberate and stays as-is: narrowing to a guessed status would break the recovery for the unique-constraint status PocketBase actually returns, which the code's own comment says is not settled between 400 and 409 — and a pinned test (`test_a_lost_placement_race_is_still_recovered` etc.) locks that in.

## Change

Log the swallowed exception at `WARN` in the branch that actually recovers (the raced row was found and the update is about to proceed) at all three call sites:
- `place_party` (`api/services/lodging_write_service.py:174`)
- `set_availability` (`api/services/lodging_write_service.py:470`)
- `set_slot_merge` (`api/services/lodging_write_service.py:538`)

The `raced is None` branch is untouched — it already re-raises through `pb_error_to_http` and reaches the caller as a real error response, so it was never the silent case.

`_seed_failure` (the fourth similar-looking method) is intentionally **not** touched: it already refuses first and tests `held > copied`, which is the correct version of this guard, not an instance of the bug — see its own docstring and `kindred#1936`.

## Test plan

- [x] Added 4 new tests (`TestARecoveredRaceLogsTheErrorItSwallowed`) confirming each of the three sites logs a WARN when a non-refusal (`500`) create failure is spuriously recovered, and confirming the `raced is None` branch does *not* log (it already surfaces as a real error).
- [x] Verified tests fail before the fix (`git stash` the implementation, tests red) and pass after.
- [x] `uv run pytest tests/ -q` — 6051 passed, 23 skipped
- [x] `uv run ruff check .`, `uv run mypy . --explicit-package-bases` — clean
- [x] `lefthook run pre-push --force` — all hooks pass (ruff, mypy, go build, shellcheck, pb-js-lint, markdownlint, api-types-freshness, pytest, tsc)

Closes #2043.

### DIFF
--- a/api/services/lodging_write_service.py
+++ b/api/services/lodging_write_service.py
@@ -107,6 +107,33 @@ class LodgingWriteService:
             raise SessionNotFoundError(f"No weekend session {session_cm_id} in {year}")
         return str(getattr(session, "id", ""))
 
+    def _log_recovered_race(self, what: str, exc: ClientResponseError, **context: Any) -> None:
+        """WARN that a create failure was treated as a lost race (kindred#2043).
+
+        The guard this backs is deliberately wide -- ANY non-refusal status
+        reaches it, not only the unique-constraint one an actual race
+        produces -- so this is the only trace left of a create that may have
+        failed for an unrelated reason (a 500, a malformed relation id, a
+        transient fault) once the recovery reports success.
+
+        Context is inlined into the message rather than passed as `extra`:
+        `bunking.logging_config.ISO8601Formatter.format` only ever renders
+        `record.getMessage()`, so an `extra={}` payload is silently dropped
+        and never reaches log output -- the same trap `api/routers/scenarios.py`
+        documents and works around at its own `logger.error` call. `str(exc)`
+        is deliberately NOT included: PocketBase's error bodies are not
+        guaranteed single-line, and a multi-line value here would break the
+        `key=value` shape the rest of this codebase's logs use.
+
+        Call this only once the recovery's own update has SUCCEEDED. Logging
+        it any earlier -- e.g. right after the raced row is found -- would
+        claim a recovery that the update call could still go on to fail,
+        which reaches the caller as an error despite the log saying
+        "Recovered".
+        """
+        fields = " ".join(f"{key}={value}" for key, value in context.items())
+        logger.warning(f"Recovered a {what} create as a lost race: status={exc.status} {fields}")
+
     async def place_party(self, request: PlacementWriteRequest) -> LodgingWriteResponse:
         """Upsert one party's placement inside a scenario.
 
@@ -185,23 +212,16 @@ class LodgingWriteService:
                     )
                     if raced is None:
                         raise pb_error_to_http(exc) from exc
-                    # The guard above is deliberately wide (see REFUSAL_STATUSES'
-                    # comment) -- ANY non-refusal status reaches here, not only
-                    # the unique-constraint one an actual race produces. The
-                    # write is idempotent either way, but a create that failed
-                    # for an unrelated reason now looks like a plain success. Log
-                    # it so that failure is still traceable (kindred#2043).
-                    logger.warning(
-                        "Recovered a draft-assignment create as a lost race",
-                        extra={
-                            "year": request.year,
-                            "session_cm_id": request.session_cm_id,
-                            "scenario": request.scenario,
-                            "status": exc.status,
-                            "error": str(exc),
-                        },
-                    )
                     record = await self.repository.update_draft_assignment(str(raced.id), data)
+                    self._log_recovered_race(
+                        "draft-assignment",
+                        exc,
+                        year=request.year,
+                        session_cm_id=request.session_cm_id,
+                        scenario=request.scenario,
+                        household_cm_id=request.household_cm_id,
+                        person_cm_id=request.person_cm_id,
+                    )
                 except ClientResponseError as retry_exc:
                     raise pb_error_to_http(retry_exc) from retry_exc
         return LodgingWriteResponse(record_id=str(getattr(record, "id", "")))
@@ -493,19 +513,14 @@ class LodgingWriteService:
                     )
                     if raced is None:
                         raise pb_error_to_http(exc) from exc
-                    # Same width, same reason to log as place_party's own
-                    # recovery above -- see the comment there (kindred#2043).
-                    logger.warning(
-                        "Recovered an availability create as a lost race",
-                        extra={
-                            "year": request.year,
-                            "session_cm_id": request.session_cm_id,
-                            "unit_id": request.unit_id,
-                            "status": exc.status,
-                            "error": str(exc),
-                        },
-                    )
                     record = await self.repository.update_availability(str(raced.id), data)
+                    self._log_recovered_race(
+                        "availability",
+                        exc,
+                        year=request.year,
+                        session_cm_id=request.session_cm_id,
+                        unit_id=request.unit_id,
+                    )
                 except ClientResponseError as retry_exc:
                     raise pb_error_to_http(retry_exc) from retry_exc
         return LodgingWriteResponse(record_id=str(getattr(record, "id", "")))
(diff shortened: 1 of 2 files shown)


api/services/lodging_write_service.py
```diff
@@ -185,6 +185,22 @@ async def place_party(self, request: PlacementWriteRequest) -> LodgingWriteRespo
                     )
                     if raced is None:
                         raise pb_error_to_http(exc) from exc
+                    # The guard above is deliberately wide (see REFUSAL_STATUSES'
+                    # comment) -- ANY non-refusal status reaches here, not only
+                    # the unique-constraint one an actual race produces. The
+                    # write is idempotent either way, but a create that failed
+                    # for an unrelated reason now looks like a plain success. Log
+                    # it so that failure is still traceable (kindred#2043).
+                    logger.warning(
```
adamflagg: **`extra={...}` is silently dropped by this codebase's log formatter — the PR doesn't actually achieve kindred#2043's stated goal.**

`ISO8601Formatter.format()` (`bunking/logging_config.py`) builds the emitted line only from `record.getMessage()`; it never reads `extra` fields off the `LogRecord`. Verified empirically:

```
$ uv run python3 -c "from bunking.logging_config import configure_logging, get_logger; configure_logging(source='api'); get_logger('x').warning('msg', extra={'year':2026,'status':500,'error':'boom'})"
WARNING msg
```

Only the bare message reaches stdout — `year`, `session_cm_id`, `scenario`, `status`, `error` all vanish (same at lines 498 and 578). So exactly in the case this PR is meant to fix — a create failing for a reason that is *not* contention — the log line is just `"Recovered a draft-assignment create as a lost race"`, with none of the diagnostic detail needed to actually trace it.

This is a known, already-documented limitation with an established in-repo workaround: `api/routers/scenarios.py` (~line 685) has *"The custom formatter in bunking/logging_config.py only emits record.getMessage(), so any extra={} payload is dropped. Inline diagnostic context into the message so it actually reaches log output."* This PR repeats the broken `extra=` pattern instead of following that precedent (or fixing the formatter once).

Root `CLAUDE.md` documents the intended format as `... LEVEL message key=value...`, which none of these three calls actually produce.

tests/unit/api/services/test_lodging_write_service.py
```diff
@@ -649,6 +649,96 @@ async def test_a_lost_availability_race_is_still_recovered(
...
+        self, write_service: LodgingWriteService, repo: MagicMock
+    ) -> None:
+        """Status 500 -- not the unique-constraint 400 the race actually
+        produces -- to prove the guard's own width is what gets logged."""
+        repo.create_draft_assignment = AsyncMock(
+            side_effect=ClientResponseError("boom", status=500, data={}, url="", is_abort=False, original_error=None)
+        )
+        repo.find_draft_assignment = AsyncMock(side_effect=[None, SimpleNamespace(id="draft_winner")])
+
+        with patch("api.services.lodging_write_service.logger") as mock_logger:
```
adamflagg: These tests replace `logger` outright with `patch("api.services.lodging_write_service.logger")` and assert only `mock_logger.warning.called`, so they never exercise the real `ISO8601Formatter`. They pass whether or not the `extra={...}` payload actually reaches the emitted log line — which it doesn't (see the comment on `lodging_write_service.py:194`). That's how the formatter-drop bug ships green: the test checks that logging was *attempted*, not that the diagnostic context the PR exists to add is actually observable in output.

api/services/lodging_write_service.py
```diff
@@ -185,6 +185,22 @@ async def place_party(self, request: PlacementWriteRequest) -> LodgingWriteRespo
...
+                        "Recovered a draft-assignment create as a lost race",
+                        extra={
+                            "year": request.year,
+                            "session_cm_id": request.session_cm_id,
+                            "scenario": request.scenario,
+                            "status": exc.status,
+                            "error": str(exc),
+                        },
+                    )
                     record = await self.repository.update_draft_assignment(str(raced.id), data)
```
adamflagg, line 215: The `logger.warning("Recovered ... as a lost race", ...)` call (line 194) fires *before* `update_draft_assignment` here is awaited, still inside the `try` that also guards this update. The docstring above `place_party` already says the recovery itself races — "the winner's row can vanish again before the update lands" — so if this `update_draft_assignment` call raises `ClientResponseError`, it's caught by `except ClientResponseError as retry_exc` a few lines down and turned into a real error response to the caller. The log has already asserted "Recovered" past tense by that point, even though the caller actually got a failure. Same ordering at lines 498/507 (`set_availability`) and 578/588 (`set_slot_merge`). Minor, but worth moving the log after a successful `update_*`, or rewording to "attempting recovery" rather than "Recovered".

api/services/lodging_write_service.py
```diff
@@ -185,6 +185,22 @@ async def place_party(self, request: PlacementWriteRequest) -> LodgingWriteRespo
...
+                    # The guard above is deliberately wide (see REFUSAL_STATUSES'
+                    # comment) -- ANY non-refusal status reaches here, not only
+                    # the unique-constraint one an actual race produces. The
+                    # write is idempotent either way, but a create that failed
+                    # for an unrelated reason now looks like a plain success. Log
+                    # it so that failure is still traceable (kindred#2043).
+                    logger.warning(
+                        "Recovered a draft-assignment create as a lost race",
+                        extra={
+                            "year": request.year,
```
adamflagg: This block logs `year`/`session_cm_id`/`scenario`/`status`/`error` but not `household_cm_id`/`person_cm_id` — the only fields that identify *which party's* create raced. Both are already in scope a few lines up (used to build `data`) and are guaranteed non-ambiguous by `PartyGrainRequest._exactly_one_grain`. Once the formatter-drop issue (see comment on this same block) is fixed, an operator still won't be able to tell which of two concurrent placements produced a given warning without cross-referencing timestamps.

api/services/lodging_write_service.py
```diff
@@ -545,6 +573,19 @@ async def set_slot_merge(self, request: SlotMergeRequest) -> LodgingWriteRespons
                     )
                     if raced is None:
                         raise pb_error_to_http(exc) from exc
+                    # Same width, same reason to log as place_party's own
+                    # recovery above -- see the comment there (kindred#2043).
+                    logger.warning(
```
adamflagg: This ~14-line `logger.warning(..., extra={...})` block is copy-pasted a third time (lines 194 and 498 are the other two), varying only the message string and 1-2 keys. The comments even say so explicitly ("Same width, same reason to log as place_party's own recovery above"). Given all three sites share the same shape (message, year, session_cm_id, status, error, plus a per-site identifier), a small private helper (e.g. `self._log_recovered_race(kind, extra)`) on `LodgingWriteService` would avoid tripling the maintenance surface — and would be one place to fix the `extra=` drop issue instead of three.

api/services/lodging_write_service.py
```diff
@@ -185,6 +185,22 @@ async def place_party(self, request: PlacementWriteRequest) -> LodgingWriteRespo
...
+                    # for an unrelated reason now looks like a plain success. Log
+                    # it so that failure is still traceable (kindred#2043).
+                    logger.warning(
+                        "Recovered a draft-assignment create as a lost race",
+                        extra={
+                            "year": request.year,
+                            "session_cm_id": request.session_cm_id,
+                            "scenario": request.scenario,
+                            "status": exc.status,
+                            "error": str(exc),
```
adamflagg: Minor: `"error": str(exc)` duplicates `"status": exc.status` — `ClientResponseError.__str__` (`pocketbase/errors.py`) already renders a 6-line block including `Status: {status}`, `URL`, `Data`, etc. Not consequential right now since `extra=` is dropped entirely (see the top comment on this block), but worth knowing for whichever fix lands: if these fields get inlined into the message string per `scenarios.py`'s precedent, a raw multi-line `str(exc)` will break the single-line `key=value` log format documented in root `CLAUDE.md`.

adamflagg (commented):
Automated review (xhigh effort, recall mode). Headline finding: the `extra={...}` payload on all three new `logger.warning(...)` calls is silently dropped by this codebase's `ISO8601Formatter` (`bunking/logging_config.py` only reads `record.getMessage()`), so none of the diagnostic context (year/session_cm_id/scenario/status/error) this PR adds actually reaches the log output — verified empirically. This is a known limitation with an established in-repo workaround already used at `api/routers/scenarios.py`, which this PR doesn't follow. 6 inline findings below, ranked by severity.

PR #2150, opened 2026-08-08
feat(frontend): weekend map loses its control bar, gains summer's Visual Guide

adamflagg:
## Stacked PR

This PR is **stacked on #2121** (`feature/fix-2062-stale-party-selection`, issue #2062) and should merge **after** it — this branch was created off #2121's tip because both touch `LodgingMap.tsx`. The diff below is scoped to this issue only; #2121's changes are included because the branch is built on top of it.

## Summary

Staff feedback on the weekend lodging map (issue #1997), settled across three rounds of mockup review:

**Deleted entirely:**
- The `Fade map` slider and its `fade` state (the scrim itself stays, pinned at `DEFAULT_FADE`)
- The whole `Highlight` fieldset (`No highlight` / `Near bathhouse` / `Staff cabins` radios) — with #1993 excluding unreleased staff housing from the map, "Staff cabins" matched zero marks on every 2026 weekend
- The `Area tint` checkbox, `areaTintBoxes()`, `TintBox`, and the tint render block — mark **hue** is a different channel and stays
- The entire control bar, including `Fit all` and the zoom +/− buttons, and the `{view.k}×` readout

**Replaced / moved:**
- Reset view is now **double-click on the canvas** — the mechanism already existed (`onDoubleClick` → `resetView`), it just had no advertised affordance once `Fit all` went. The corner hint now says so.
- `Empty rooms` **demotes into the legend strip** rather than disappearing. It hides 25 of 76 marks on the busiest 2026 weekend, and per the file's own accessibility note it is the map's **last keyboard-reachable control** — still a real `<input type="checkbox">` behind a real `<label>`, not reinvented as a div.
- The `staff-default`, `near bathhouse` and `area colour` legend rows move off the map's own legend into a new shared **Visual Guide**.

**New — shared Visual Guide for the weekend:**
Summer already has a "Visual Guide" modal (`BunkingLegend.tsx`) behind a `HelpCircle` button in `SessionHeader`. This PR extracts its modal shell + `LegendEntry` primitive into a `sections`-parameterised `VisualGuideModal`, so `BunkingLegend` (summer) and the new `WeekendLegendButton` render entirely different content through the one component — no forked file. `WeekendLegendButton` is mounted once in `WeekendRosterPage`'s header, so it covers Roster, Board and Map from a single button rather than living only on the map tab.

The weekend's "Room Indicators" section documents the three rows moved off the map (staff cabin, near bathhouse, area colour) plus the board's own shared-consent amber ring, since it carries the same signal.

**Chunk-graph note:** `WeekendLegendButton` mounts unconditionally and eagerly in the weekend header, so `BunkingLegend.tsx` deliberately **duplicates** two colour literals rather than importing them from `LodgingMap.tsx` / `MapUnitPopover.tsx` — importing either would drag `LodgingMap`'s lazy chunk into the eager bundle, which `WeekendRosterPage.chunkGraph.test.ts` (a real `vite build` + chunk-graph inspection, kindred#2057) exists to catch. Ran it explicitly to confirm: still green.

**Also fixed:** `countMapUnits`' docstring in `mapModel.ts` quoted a pre-#1993 measurement ("103 positioned rooms render as 75 marks"). Corrected to the current live figure (76 rooms / 76 marks on 2026's busiest weekend, per the issue's own re-measurement).

**Post-review fixes (kindred#2157, kindred#2158):** code review flagged that the demoted `Empty rooms` checkbox landed inside `<dl data-testid="map-legend">` as `div > label > input + text`, breaking the dt/dd term-definition pattern the rest of the legend follows — a description list has no defined semantics for a live form control as a child. It now lives in a sibling `role="group"`, with the `<dl>` switched to `display: contents` so its dt/dd children still lay out in the same flex-wrap row with no visual change. Also swept `BunkingLegend.tsx`'s dead default export (confirmed via `git grep` — only its own test file imported it) while the file was already open, folding its test coverage into cases that click `BunkingLegendButton` instead.

## Tests

Followed the issue's own "Tests that move or die" table:
- Radios, highlight-replacement and area-tint tests **deleted** with the controls they exercised.
- The fade-slider test **deleted**; replaced with a test that the scrim still renders at a fixed opacity with no control for it.
- The two `Empty rooms` tests **stay**, plus a new structural test pinning that the checkbox lives in a sibling `role="group"` next to the legend, not inside it (moved from the earlier `within(legend)` assertion per kindred#2157).
- The double-click-reset tests (`:513-537` in the issue's numbering) were **untouched** — same mechanism, no changes needed.
- Legend-content tests updated: `staff-default`/`near bathhouse`/`area colour` assertions removed from `LodgingMap.test.tsx` and added (with new tests) to `BunkingLegend.test.tsx` for `WeekendLegendButton`.
- New test in `WeekendRosterPage.test.tsx` pins the guide button living in the header once, reachable from any tab.
- Fixed one **collateral** break in `WeekendRosterPage.test.tsx` (`tab component state survives switching (#2004)`) — its zoom-persistence test drove zoom via the now-deleted "Zoom in" button; rewired it to drive zoom via wheel (same mechanism `LodgingMap.test.tsx`'s own zoom tests use) and assert on the backdrop's transform instead of the deleted `{k}×` readout.
- `BunkingLegend.test.tsx`'s former direct-render cases (`describe('BunkingLegend', ...)`) rerouted through `BunkingLegendButton`'s click, so the dead default export could be deleted without losing coverage (kindred#2158).

## Verification

- `lefthook run pre-push` (full): ruff, go build, mypy, pb-js-lint, shellcheck, markdownlint, api-types-freshness, **pytest (6020 passed)**, tsc type-check — all green.
- `npx vitest run` (full suite, including the real-build `WeekendRosterPage.chunkGraph.test.ts`), rerun after the post-review fixes: **418 files / 5955 tests passed**.
- `npx tsc --noEmit`: clean.
- `./node_modules/.bin/eslint` on touched files: 0 errors (2 pre-existing warnings, unrelated).
- Pushed and verified `git rev-list --count origin/...HEAD` = 0.

⚠️ This PR targets a feature branch (#2121's), not `main`, so `ci.yml` does not run on it — these local runs are its only verification.

Closes #1997.
Closes #2157.
Closes #2158.

### DIFF
--- a/frontend/src/components/BunkingLegend.tsx
+++ b/frontend/src/components/BunkingLegend.tsx
@@ -11,9 +11,20 @@ import {
   History,
 } from 'lucide-react'
 
-interface LegendEntryProps {
+import { BATHHOUSE_BLUE, CONSENT_AMBER } from './weekend/mapColors'
+
+/** What one legend row and one legend section share: an icon and a title.
+ *  A row's body is `children` (rendered inline by `LegendEntry` below); a
+ *  section's is `content` (one or more rows, rendered by `VisualGuideModal`).
+ *  Different field names because the two nest — a section's `content` IS a
+ *  list of entries — so reusing `children` for both would blur which level
+ *  a given JSX blob belongs to. */
+interface LegendIconTitle {
   icon: React.ReactNode
   title: string
+}
+
+interface LegendEntryProps extends LegendIconTitle {
   children: React.ReactNode
   iconClass?: string
 }
@@ -30,9 +41,7 @@ function LegendEntry({ icon, title, children, iconClass }: LegendEntryProps) {
   )
 }
 
-interface LegendSection {
-  icon: React.ReactNode
-  title: string
+interface LegendSection extends LegendIconTitle {
   content: React.ReactNode
 }
 
@@ -349,19 +358,6 @@ export function BunkingLegendButton() {
   return <LegendButton sections={CAMPER_SECTIONS} />
 }
 
-/**
- * DUPLICATED, not imported, from `LodgingMap.tsx` / `MapUnitPopover.tsx`.
- *
- * This button mounts unconditionally in the weekend header (WeekendRosterPage,
- * outside the lazy tab panels), so an import from either would drag
- * `LodgingMap`'s whole lazy chunk into the eager bundle — exactly what
- * `WeekendRosterPage.chunkGraph.test.ts` exists to catch (kindred#2057).
- * A colour literal is cheap to keep in sync by hand; a chunk-graph regression
- * is not.
- */
-const WEEKEND_BATHHOUSE_BLUE = '#2563eb'
-const WEEKEND_CONSENT_AMBER = '#fbbf24'
-
 const WEEKEND_SECTIONS: LegendSection[] = [
   {
     icon: <Home className="h-4 w-4" />,
@@ -383,7 +379,7 @@ const WEEKEND_SECTIONS: LegendSection[] = [
         <LegendEntry
           icon={
             <span
-              style={{ backgroundColor: WEEKEND_BATHHOUSE_BLUE }}
+              style={{ backgroundColor: BATHHOUSE_BLUE }}
               className="h-3 w-3 rounded-full ring-1 ring-white"
             />
           }
@@ -413,7 +409,7 @@ const WEEKEND_SECTIONS: LegendSection[] = [
         <LegendEntry
           icon={
             <div
-              style={{ boxShadow: `0 0 0 2px #fff, 0 0 0 4.5px ${WEEKEND_CONSENT_AMBER}` }}
+              style={{ boxShadow: `0 0 0 2px #fff, 0 0 0 4.5px ${CONSENT_AMBER}` }}
               className="h-5 w-5 rounded-full bg-white"
             />
           }
--- a/frontend/src/components/BunkingLegend.tsx
+++ b/frontend/src/components/BunkingLegend.tsx
@@ -330,15 +330,6 @@ const CAMPER_SECTIONS: LegendSection[] = [
   },
 ]
 
-interface BunkingLegendProps {
-  isOpen: boolean
-  onClose: () => void
-}
-
-export default function BunkingLegend({ isOpen, onClose }: BunkingLegendProps) {
-  return <VisualGuideModal isOpen={isOpen} onClose={onClose} sections={CAMPER_SECTIONS} />
-}
-
 /** The open/close state and trigger both `BunkingLegendButton` and
  *  `WeekendLegendButton` share; only the `sections` they open differ. */
 function LegendButton({ sections }: { sections: LegendSection[] }) {
--- a/frontend/src/components/BunkingLegend.tsx
+++ b/frontend/src/components/BunkingLegend.tsx
@@ -11,11 +11,6 @@ import {
   History,
 } from 'lucide-react'
 
-interface BunkingLegendProps {
-  isOpen: boolean
-  onClose: () => void
-}
-
 interface LegendEntryProps {
   icon: React.ReactNode
   title: string
@@ -35,7 +30,30 @@ function LegendEntry({ icon, title, children, iconClass }: LegendEntryProps) {
   )
 }
 
-export default function BunkingLegend({ isOpen, onClose }: BunkingLegendProps) {
+interface LegendSection {
+  icon: React.ReactNode
+  title: string
+  content: React.ReactNode
+}
+
+interface VisualGuideModalProps {
+  isOpen: boolean
+  onClose: () => void
+  sections: LegendSection[]
+}
+
+/**
+ * The shared "Visual Guide" shell — header, section list and "Got it"
+ * footer — behind whichever `HelpCircle` button mounts it.
+ *
+ * Parameterised on `sections` rather than forked (kindred#1997): summer and
+ * the weekend document entirely different signals — friend groups and AG
+ * grade ratios have no weekend equivalent — but the modal chrome, the
+ * section shape and the `LegendEntry` primitive are the one shell both use.
+ * Copying this file to give the weekend its own guide would be the thing
+ * CLAUDE.md §4 names by name.
+ */
+function VisualGuideModal({ isOpen, onClose, sections }: VisualGuideModalProps) {
   if (!isOpen) return null
 
   return (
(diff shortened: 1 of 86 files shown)


frontend/src/components/BunkingLegend.tsx
```diff
@@ -35,7 +30,30 @@ function LegendEntry({ icon, title, children, iconClass }: LegendEntryProps) {
...
+ * footer — behind whichever `HelpCircle` button mounts it.
+ *
+ * Parameterised on `sections` rather than forked (kindred#1997): summer and
+ * the weekend document entirely different signals — friend groups and AG
+ * grade ratios have no weekend equivalent — but the modal chrome, the
+ * section shape and the `LegendEntry` primitive are the one shell both use.
+ * Copying this file to give the weekend its own guide would be the thing
+ * CLAUDE.md §4 names by name.
+ */
+function VisualGuideModal({ isOpen, onClose, sections }: VisualGuideModalProps) {
```
adamflagg, line 65: **Cleanup (not a correctness bug from this PR): `VisualGuideModal` reimplements the app's shared `Modal.tsx` shell, minus Escape-to-close and viewport portal.**

`frontend/src/components/ui/Modal.tsx` already provides `isOpen`/`onClose`/`header`/`footer` slots plus:
- an Escape-key listener (`Modal.tsx` `useEffect` on `keydown`)
- `createPortal(..., document.body)` so it escapes any ancestor stacking context (its own comment names `z-[60] CamperDetailsPanel` as the reason)

`VisualGuideModal` here has neither — no Escape handler, and it renders inline via plain `fixed inset-0 z-50` rather than a portal. This gap predates this PR (confirmed against the parent commit `5e1d48a3`), so it's not a regression. But this PR is exactly the kind of change that entrenches it: it factors the modal into a shared shell and gives it a *second* production caller (`WeekendLegendButton`), so the missing Escape/portal behavior — inconsistent with every other modal in the app — now ships on two surfaces instead of one, at the moment it would have been cheapest to fold in `Modal`'s existing `header`/`footer` slots instead.

Not blocking, but worth a follow-up rather than letting a third caller entrench it further.

frontend/src/components/BunkingLegend.tsx
```diff
@@ -54,257 +72,363 @@ export default function BunkingLegend({ isOpen, onClose }: BunkingLegendProps) {
...
+ *
+ * This button mounts unconditionally in the weekend header (WeekendRosterPage,
+ * outside the lazy tab panels), so an import from either would drag
+ * `LodgingMap`'s whole lazy chunk into the eager bundle — exactly what
+ * `WeekendRosterPage.chunkGraph.test.ts` exists to catch (kindred#2057).
+ * A colour literal is cheap to keep in sync by hand; a chunk-graph regression
+ * is not.
+ */
+const WEEKEND_BATHHOUSE_BLUE = '#2563eb'
+const WEEKEND_CONSENT_AMBER = '#fbbf24'
```
adamflagg: **Altitude: the color duplication this PR's comment justifies is real and currently in sync, but a shared leaf module would remove the drift risk without reintroducing the chunk-graph problem.**

`WEEKEND_BATHHOUSE_BLUE`/`WEEKEND_CONSENT_AMBER` duplicate `BATHHOUSE_BLUE` (`LodgingMap.tsx`) and `CONSENT_AMBER` (`MapUnitPopover.tsx`) byte-for-byte today — verified. The PR's own comment defends this by citing `WeekendRosterPage.chunkGraph.test.ts`: importing either constant from those files would drag `LodgingMap`'s lazy chunk into the eager header bundle.

But the codebase already has the right pattern for exactly this situation — `frontend/src/utils/dispositionColors.ts`, `yearColors.ts`, and `retentionColors.ts` are tiny, zero-import leaf modules holding shared color constants, importable from anywhere (including an eager surface) without pulling anything else along. A `BATHHOUSE_BLUE`/`CONSENT_AMBER` pair moved to such a leaf module would satisfy the chunk-graph test exactly as well as the duplication does, while removing the two-sources-of-truth risk the comment concedes ("cheap to keep in sync by hand").

It's not just the color literal that's duplicated, either — the CSS recipe around it is copy-pasted too: the "Near Bathhouse" swatch (lines 384-389 below) repeats `rounded-full ring-1 ring-white`, and the "Sharing Not Consented" swatch (lines 413-417) repeats the exact `0 0 0 2px #fff, 0 0 0 4.5px ${...}` box-shadow geometry from `LodgingMap.tsx`'s `halo` computation — a shared design decision (ring width, offset), not just a color.

frontend/src/components/weekend/LodgingMap.tsx
```diff
@@ -903,23 +677,46 @@ export function LodgingMap({ parties, units, year }: LodgingMapProps) {
...
           <dl
             data-testid="map-legend"
             className="text-muted-foreground flex flex-wrap items-center gap-x-4 gap-y-1.5 text-[11px]"
           >
+            {/* DEMOTED from the control bar, not deleted (kindred#1997): it
+                hides 25 of 76 marks on the busiest 2026 weekend, and it is the
+                map's LAST keyboard-reachable control — see the note at
+                `:16-26` above. Still a real `<input>` behind a real `<label>`,
+                per that note, not re-invented as a div. */}
+            <div className="flex items-center gap-1.5">
```
adamflagg: **Cleanup: the demoted `Empty rooms` checkbox breaks the `<dl>`'s content model and carries a redundant wrapper.**

Every other child of `<dl data-testid="map-legend">` follows `div > dt.sr-only + dd` (a term/definition pair). This entry is just `div > label > input + text` — no `dt`/`dd`. A description list has no defined semantics for a live form control, so this is the one entry that breaks the pattern the rest of the strip establishes; a screen reader / accessibility-tree walker sees an anomalous child.

Separately, the wrapping `<div className="flex items-center gap-1.5">` (line 704) wraps a single `<label>` child that already declares the identical `inline-flex items-center gap-1.5` (line 705) — the outer div does nothing here, unlike sibling entries where a wrapper groups an icon span plus `dt`/`dd`.

Given the file's own comment above this block cites keeping the checkbox as "the map's LAST keyboard-reachable control," a small adjacent `<div role="group" aria-label="Map controls">` sibling to the `<dl>` (still visually one row via the existing flex-wrap) would keep the control reachable without stretching the `<dl>`'s semantics, and would drop the redundant div in the same move.

frontend/src/components/BunkingLegend.tsx
```diff
@@ -54,257 +72,363 @@ export default function BunkingLegend({ isOpen, onClose }: BunkingLegendProps) {
...
+  },
+]
+
+interface BunkingLegendProps {
+  isOpen: boolean
+  onClose: () => void
 }
 
-export function BunkingLegendButton() {
+export default function BunkingLegend({ isOpen, onClose }: BunkingLegendProps) {
```
adamflagg: **Cleanup, pre-existing but worth a note since this PR touches this exact file heavily: the default-exported `BunkingLegend({isOpen, onClose})` is dead in production.**

Confirmed via `git grep` at HEAD and at the parent commit (`5e1d48a3`, before this PR): the only importer of the default export, on either side of this PR, is `BunkingLegend.test.tsx` itself. Every production caller goes through `BunkingLegendButton`/`WeekendLegendButton` → the internal `LegendButton`, never through this default export directly.

Not introduced by this diff, so not blocking — but since this PR already restructures the file top-to-bottom (introducing `VisualGuideModal`/`LegendButton`), it was a natural point to either delete the unused default export (folding its one remaining test into the `LegendButton`/`BunkingLegendButton` coverage) or to actually wire it in somewhere. Leaving it means the file now carries three superficially-similar named things (`BunkingLegend`, `BunkingLegendButton`, `LegendButton`) where one is vestigial, which reads as load-bearing on a skim.

frontend/src/components/BunkingLegend.tsx
```diff
@@ -35,7 +30,30 @@ function LegendEntry({ icon, title, children, iconClass }: LegendEntryProps) {
   )
 }
 
-export default function BunkingLegend({ isOpen, onClose }: BunkingLegendProps) {
+interface LegendSection {
```
adamflagg: **Minor simplification: `LegendEntryProps` and `LegendSection` are two parallel "icon + title + content" shapes.**

```ts
interface LegendEntryProps { icon; title; children; iconClass? }   // line 14
interface LegendSection     { icon; title; content }               // line 33
```

Only the content-field name differs (`children` vs `content`) and `LegendEntryProps` carries the extra `iconClass?`. Not worth a large refactor, but a shared base (`{icon, title}`) extended by each would make the relationship between "one legend row" and "one legend section of rows" explicit in the types rather than left to the reader to notice they rhyme.

adamflagg (commented):
xhigh-effort recall review (5 correctness angles + 3 cleanup angles + altitude + conventions, each independently verified, plus a fresh gap-sweep pass — 11 finder passes total).

**No correctness bugs found.** Every removed-behavior check (dead state, dangling refs to `fade`/`areaTint`/`highlight`/`zoomBy`/`HIGHLIGHTS`/`legendHues`, cross-file call-site breakage, the deliberately-duplicated `WEEKEND_BATHHOUSE_BLUE`/`WEEKEND_CONSENT_AMBER` vs. their source constants, the chunk-graph guard, rules-of-hooks, hook deps, the rewritten wheel-zoom persistence test) came back clean and verified against the actual post-PR file content (`git show 8e8a154c7...`), not just the diff hunks. This is an unusually well-defended PR — the tests and in-code comments anticipate most of what a review would otherwise flag.

The 5 inline comments below are cleanup/altitude findings only (no correctness issues survived verification): a pre-existing Escape-key/portal gap in the shared `VisualGuideModal` now extended to a second surface, a color+CSS-recipe duplication that a leaf constants module (matching this repo's own `dispositionColors.ts`/`yearColors.ts` precedent) would remove without reintroducing the chunk-graph problem, a `<dl>` content-model/redundant-wrapper nit on the demoted `Empty rooms` checkbox, a dead default-exported `BunkingLegend`, and a minor type-shape duplication. None are blocking.

PR #2151, opened 2026-08-08
fix(sync): auto-cleanup a cancelled household's lodging placement

adamflagg:
> ## ✅ Data-loss risk resolved
>
> A post-hoc review found `deleteLodgingOrphans(year)` ran unconditionally, while `handleLodgingAssignmentsSync`'s `?year=` and the orchestrator's historical-sync re-registration both call `Sync()` with an arbitrary year. Since this service reads only local PocketBase tables (no CampMinder client), a stale or incomplete local `attendees` snapshot for a past season would read as "nobody enrolled" and permanently delete real historical placements — with no `lodging_assignment_history` row, so the deletion was also unrecoverable.
>
> **Fixed structurally, not by moving the call site:**
> - `deleteLodgingOrphans` now only runs when the resolved year equals `CAMPMINDER_SEASON_ID` (`activeSeasonYear()`) — the season this deployment is actively syncing. Every other year (any `?year=` backfill, any historical re-registration) keeps the pre-#2028 create/update-only behavior that was safe before this pass existed. `deleteLodgingOrphans` also re-checks `SyncSuccessful` itself, matching the guard every other derived sync's own orphan pass uses (`BaseSyncService.DeleteOrphans` and friends) — the check travels with the delete even if the call site moves.
> - The delete now writes a `lodging_assignment_history` row (`old_unit` → `""`, `source_field = "orphan_sweep"`) before removing the placement, so a hard delete is auditable like every other way this table changes.
> - `deleteLodgingOrphans` now calls `findLodgingEnrollmentOrphans` — the same predicate `stranded_assignment_cleanup.go`'s production audit already runs against this table — instead of a hand-rolled copy, so the audit log and the actual deletion can no longer silently disagree.
>
> New test: `TestLodgingAssignmentsSyncSkipsOrphanSweepForHistoricalYear` drives `Sync()` through the exact shape the orchestrator's historical path and the API's `?year=` both use (`s.Year` set to a season other than `CAMPMINDER_SEASON_ID`) and asserts zero deletions. Mutation-checked: reasoning and transcript in the PR comment below.

## Summary

Mirrors summer's `stranded_assignment_cleanup` pattern for lodging (#2028): a household or person who cancels **after** being placed kept its `lodging_assignments` and `lodging_assignments_draft` rows forever — CampMinder never clears the source cabin custom-field value on cancellation, so the ingest only ever walked values forward and never revisited a row once the party's enrolled-session candidates went empty.

**Stacked on #2120** (`feature/fix-2061-registryless-season`) — this branch was created off it because both touch `lodging_assignments_sync.go`. Base is set to that branch, not `main`.

### Mechanism (mirrors summer exactly, per the owner's settled ruling on #2028)

- **`lodging_assignments_draft`** — `stranded_assignment_cleanup.go` gains a lodging enrollment-orphan pass. It reuses `BuildHouseholdSessionIndex` / `BuildPersonSessionIndex` (not a second attendees query) to find a draft row whose household/person is no longer actively enrolled in the session it names, and nulls `units` on it — across every scenario for the year. `staff_touched` and `source` are left exactly as they are; only the placement clears, the same as summer's `bunk` + `bunk_plan` nulling at `:251-253`. `staff_touched` is **not** a skip guard here — a cancelled household is gone whether or not staff moved it.
- **`lodging_assignments` (the mirror)** — `LodgingAssignmentsSync` gains `deleteLodgingOrphans`, called after the write passes (and after the `DryRun` check, so a dry run deletes nothing), gated to the active season (see above). It's driven by absence from the *same* enrolled-session indexes `syncHouseholdGrain`/`syncPersonGrain` already build (captured as `s.householdSessionIndex` / `s.personSessionIndex`), exactly as `bunk_assignments.deleteOrphans()` is driven by absence from its own CampMinder pull. `stranded_assignment_cleanup.go`'s lodging pass only **audits and warns** on this table (new `Stats.LodgingProdAuditWarnings`) — it never deletes it, the same split of responsibility bunk already has (`msgOrphanProd` vs. `bunk_assignments.go`'s own `deleteOrphans()`).
- **Per-session zero-enrolled guard, carried over verbatim** — a session with zero reliably-enrolled parties of a grain is left completely untouched by both passes. An attendee-sync hiccup is indistinguishable from "everyone cancelled" from inside this code, so it must never read as license to sweep.
- **No new `lodging_ingest_issues` kind, no queue surface.** Both passes are silent (log + `Stats` counters only), matching summer.

### Scope ceiling

**The cascade stops at `lodging_assignments` + `lodging_assignments_draft`, sync year only.** Neither touched file references `family_camp_registrations` anywhere (verified by grep). That table's `cabin_assignment` column is the *only* pre-2026 placement history that exists anywhere in the database — the same enrolment predicate applied there would delete 43 rows of 2025 history, which is unrecoverable. This PR nulls a draft field and deletes mirror rows going forward; it does not, and must not, touch historical placement data.

## Test plan

- [x] STRICT TDD: wrote failing Go tests first (`go vet` confirmed `undefined: lodgingOrphanCandidate` etc. before any implementation), then implemented.
- [x] `go build ./...` — clean.
- [x] `go vet ./...` — clean.
- [x] `golangci-lint run ./...` — 0 issues.
- [x] `go test ./sync/...` — all green, including the historical-year regression test and the orphan-delete history-write test.
- [x] Mutation-check: removing the per-session `reliable` guard in either `deleteLodgingOrphans` or `findLodgingEnrollmentOrphans` turns the corresponding zero-enrolled tests red (`TestLodgingAssignmentsSyncSkipsMirrorDeletionWhenSessionHasZeroEnrolled`, `TestStrandedAssignmentCleanup_LodgingEnrollmentGateSkipsPerSession`, `TestFindLodgingEnrollmentOrphans`); confirmed, then reverted. Removing the `activeSeasonYear` gate turns `TestLodgingAssignmentsSyncSkipsOrphanSweepForHistoricalYear` red; removing the history write turns `TestLodgingAssignmentsSyncWritesHistoryOnOrphanDelete` red. Both confirmed, then reverted — transcript in the PR comment below.
- [x] `lefthook run pre-push` via full `git push` — pytest, mypy, tsc, go-build, ruff, shellcheck, markdownlint, api-types-freshness all green.
- [x] Push verified: `git rev-list --count origin/...HEAD` = `0`.

## Folded in: #2160 — per-run Stats reset

`StrandedAssignmentCleanupSync.Sync()` never reset `s.Stats`, unlike every sibling derived sync (`LodgingAssignmentsSync.Sync()` does `s.Stats = Stats{}`), and the orchestrator holds one instance for the process lifetime — so `Updated`/`Errors` climbed across every run since boot, and the assigned-not-incremented audit counters carried a stale value whenever a later run short-circuited on the zero-`bunk_plans` gate before the audit ran. Folded here because this PR added more writes into those same counters (`Stats.Updated` in the lodging draft sweep, the new `Stats.LodgingProdAuditWarnings`).

- STRICT TDD: two tests written first, both confirmed red before the one-line fix — `TestStrandedAssignmentCleanup_StatsResetBetweenRuns` (one instance, two `Sync()` runs, second must report `Updated=0`; failed with `got 1`) and `TestStrandedAssignmentCleanup_ProdAuditWarningsClearWhenRunGatesOut` (a gated-out run must report `ProdAuditWarnings=0`, not echo the prior run; failed with `got 1`). Green after.
- **#2160's second half is deliberately NOT folded.** Surfacing `lodging_prod_audit_warnings` needs `frontend/src/hooks/useSyncStatusAPI.ts` and `frontend/src/components/admin/SyncTab.tsx`, neither of which this branch touches — it would put new frontend surface area on an otherwise all-Go PR. Re-filed as its own S issue so #2160 can close with this fix.

Closes #2028.
Closes #2160.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

### DIFF
--- a/pocketbase/sync/lodging_assignments_sync.go
+++ b/pocketbase/sync/lodging_assignments_sync.go
@@ -43,6 +43,13 @@ type LodgingAssignmentsSync struct {
 	personsByHouseholdCMID  map[int][]*core.Record
 	enrolledByPersonSession map[string]int // "<personPBID>|<sessionPBID>" -> count
 	adultsByHouseholdPBID   map[string]int
+
+	// householdSessionIndex / personSessionIndex are the SAME indexes
+	// syncHouseholdGrain / syncPersonGrain build to attribute values, captured
+	// here so deleteLodgingOrphans can reuse them rather than re-deriving the
+	// enrolled set (#2028).
+	householdSessionIndex map[int][]SessionWindow
+	personSessionIndex    map[int][]SessionWindow
 }
 
 // NewLodgingAssignmentsSync builds the service. Year 0 means "resolve from the
@@ -145,6 +152,15 @@ func (s *LodgingAssignmentsSync) Sync(ctx context.Context) error {
 		return nil
 	}
 
+	// #2028: remove mirror rows for a party no longer actively enrolled in the
+	// session they name -- e.g. a household that cancelled after being placed.
+	// CampMinder never clears the source custom-field value on cancellation, so
+	// the ingest above never revisits an existing row once its Candidates go
+	// empty; this is the only pass that does.
+	if delErr := s.deleteLodgingOrphans(year); delErr != nil {
+		return delErr
+	}
+
 	priorCounts, err := s.valueCountsByCMID(year-1, fieldTargets)
 	if err != nil {
 		return err
@@ -190,6 +206,7 @@ func (s *LodgingAssignmentsSync) Sync(ctx context.Context) error {
 		"year", year,
 		"created", s.Stats.Created,
 		"updated", s.Stats.Updated,
+		"deleted", s.Stats.Deleted,
 		"skipped", s.Stats.Skipped,
 		"errors", s.Stats.Errors,
 		"queued_issues", issuesCreated+issuesUpdated,
@@ -206,6 +223,7 @@ func (s *LodgingAssignmentsSync) syncHouseholdGrain(
 	if err != nil {
 		return fmt.Errorf("building household session index: %w", err)
 	}
+	s.householdSessionIndex = sessionIndex
 	householdCMIDs, err := s.cmIDsByPBID("households", year)
 	if err != nil {
 		return err
@@ -272,6 +290,7 @@ func (s *LodgingAssignmentsSync) syncPersonGrain(
 	if err != nil {
 		return fmt.Errorf("building person session index: %w", err)
 	}
+	s.personSessionIndex = sessionIndex
 	personCMIDs, err := s.cmIDsByPBID("persons", year)
 	if err != nil {
 		return err
@@ -802,6 +821,69 @@ func (s *LodgingAssignmentsSync) partySize(in *ingestContext, sessionID string) 
 		adultCount += s.adultsByHouseholdPBID[hh]
 	}
 	return enrolled + adultCount
+}
+
+// deleteLodgingOrphans removes lodging_assignments rows for a party no longer
+// actively enrolled in the session the row names -- #2028's mirror-table half.
+//
+// Driven by absence from BuildHouseholdSessionIndex / BuildPersonSessionIndex
+// -- specifically s.householdSessionIndex / s.personSessionIndex, the SAME
+// indexes syncHouseholdGrain / syncPersonGrain just built above, not
+// re-derived -- exactly as bunk_assignments.deleteOrphans() is driven by
+// absence from its own CampMinder pull. A household that cancels after being
+// placed keeps its cabin value in household_custom_values (CampMinder does not
+// clear it), so ingestValue's Candidates go empty and the row is never
+// revisited by the write path; this pass is the only thing that ever will
+// revisit it.
+//
+// staff_touched is NOT a guard here, unlike upsertAssignment's write-path
+// skip: that skip protects a staff move from being overwritten by a
+// CONFLICTING campminder_sync value, but a cancelled household is not
+// attending regardless of who last touched its placement -- the same ruling
+// #2028 makes for the draft-null pass in stranded_assignment_cleanup.go.
+//
+// Per-session guard: a session with zero reliably-enrolled parties of a grain
+// is left untouched -- an attendee-sync hiccup must not read as "everyone
+// cancelled" and empty the whole session's placements.
+func (s *LodgingAssignmentsSync) deleteLodgingOrphans(year int) error {
+	rows, err := findAllRecords(s.App, "lodging_assignments", fmt.Sprintf("year = %d", year))
+	if err != nil {
+		return fmt.Errorf("querying lodging_assignments for orphan sweep: %w", err)
+	}
+
+	householdReliable := reliableEnrolledSessions(s.householdSessionIndex)
+	personReliable := reliableEnrolledSessions(s.personSessionIndex)
+
+	for _, rec := range rows {
+		sessionID := rec.GetString("session")
+		hhCMID := rec.GetInt("household_cm_id")
+		personCMID := rec.GetInt("person_cm_id")
+
+		var stillEnrolled, reliable bool
+		switch {
+		case hhCMID > 0:
+			reliable = householdReliable[sessionID]
+			stillEnrolled = sessionIndexHasWindow(s.householdSessionIndex[hhCMID], sessionID)
+		case personCMID > 0:
+			reliable = personReliable[sessionID]
+			stillEnrolled = sessionIndexHasWindow(s.personSessionIndex[personCMID], sessionID)
+		default:
+			continue // grain-less row -- not this pass's concern
+		}
+
+		if !reliable || stillEnrolled {
+			continue
+		}
+
+		if delErr := s.App.Delete(rec); delErr != nil {
+			s.Stats.Errors++
+			slog.Error("lodging_assignments_sync: deleting cancelled-party mirror row",
+				"id", rec.Id, "household_cm_id", hhCMID, "person_cm_id", personCMID, "error", delErr)
+			continue
+		}
+		s.Stats.Deleted++
+	}
+	return nil
 }
 
 // cmIDsByPBID maps a collection's PB record id -> its CampMinder id for one
--- a/pocketbase/sync/lodging_assignments_sync.go
+++ b/pocketbase/sync/lodging_assignments_sync.go
@@ -92,6 +92,31 @@ func (s *LodgingAssignmentsSync) Sync(ctx context.Context) error {
 	if s.resolver, err = NewAliasResolver(s.App); err != nil {
 		return fmt.Errorf("building alias resolver: %w", err)
 	}
+
+	// A season with no lodging_units rows at all cannot resolve anything:
+	// Resolve is all-or-nothing on (code, year), so every alias's stored
+	// member id -- still pointing at whatever year it was last authored
+	// against -- misses. Left unguarded that unresolves every cabin: each
+	// distinct raw string queues an unresolved_alias work-queue item that
+	// never self-clears (IssueRecorder.Flush only sets is_resolved on
+	// create), and writeHistory on the unresolved path is unconditional, so
+	// every run appends another lodging_assignment_history row per household.
+	// Skip and return nil -- one unseeded season must not fail the whole sync
+	// run. Two distinguishable messages, because they call for different
+	// action: nothing has ever been loaded for this year, vs. a prior season
+	// exists but this one has not been carried forward yet. See #2061.
+	if !s.resolver.HasUnitsForYear(year) {
+		if s.resolver.HasAnyUnits() {
+			slog.Warn("lodging_assignments_sync: skipping -- registry has not been rolled forward to this season yet",
+				"year", year)
+		} else {
+			slog.Warn("lodging_assignments_sync: skipping -- no lodging registry has ever been loaded for this season",
+				"year", year)
+		}
+		s.SyncSuccessful = true
+		return nil
+	}
+
 	s.issues = NewIssueRecorder(s.App, year)
 
 	fieldTargets, err := LodgingFieldDefIDs(s.App)
--- a/pocketbase/sync/lodging_assignments_sync.go
+++ b/pocketbase/sync/lodging_assignments_sync.go
@@ -18,6 +18,12 @@ const serviceNameLodgingAssignments = "lodging_assignments"
 // distinguish them from staff placements.
 const sourceCampMinderSync = "campminder_sync"
 
+// sourceFieldOrphanSweep labels a lodging_assignment_history row written by
+// deleteLodgingOrphans, so a hard delete driven by cancelled enrolment reads
+// distinctly in the audit trail from a row driven by one of the CampMinder
+// field names (fieldNameFamilyCampCabin / fieldNameReportableFamilyCampCabin).
+const sourceFieldOrphanSweep = "orphan_sweep"
+
 // LodgingAssignmentsSync derives lodging_assignments from the two CampMinder
 // cabin custom fields.
 //
@@ -157,8 +163,31 @@ func (s *LodgingAssignmentsSync) Sync(ctx context.Context) error {
 	// CampMinder never clears the source custom-field value on cancellation, so
 	// the ingest above never revisits an existing row once its Candidates go
 	// empty; this is the only pass that does.
-	if delErr := s.deleteLodgingOrphans(year); delErr != nil {
-		return delErr
+	//
+	// GATED to the season this deployment is actively configured for
+	// (CAMPMINDER_SEASON_ID), never merely to whatever `year` this particular
+	// call carries. handleLodgingAssignmentsSync's ?year= query param and the
+	// orchestrator's historical re-registration (opts.Year) both drive this
+	// Sync() with an arbitrary year, and s.householdSessionIndex /
+	// s.personSessionIndex above are built from the LOCAL attendees table for
+	// THAT year, not from CampMinder. Before this guard, a stale or partial
+	// local snapshot for a season that already happened read as "nobody is
+	// enrolled", and the sweep hard-deleted real historical placements with no
+	// way back. Every year outside the active season keeps the pre-#2028
+	// create/update-only behavior that was safe before this pass existed.
+	// deleteLodgingOrphans repeats the SyncSuccessful half of this guard
+	// itself, matching the idiom every other derived sync in this package
+	// uses before its own orphan pass (e.g. FamilyCampDerivedSync.
+	// deleteOrphanedAdults, NormalizeGeographicSync, StaffSkillsSync) -- so the
+	// check travels with the delete even if this call site is ever moved.
+	if active := s.activeSeasonYear(); year != active {
+		slog.Info("lodging_assignments_sync: skipping orphan sweep -- year is not the actively configured season",
+			"year", year, "configured_season", active)
+	} else {
+		s.SyncSuccessful = true
+		if delErr := s.deleteLodgingOrphans(year, now); delErr != nil {
+			return delErr
+		}
 	}
 
 	priorCounts, err := s.valueCountsByCMID(year-1, fieldTargets)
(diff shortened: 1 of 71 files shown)


pocketbase/sync/lodging_assignments_sync.go
```diff
@@ -804,6 +823,69 @@ func (s *LodgingAssignmentsSync) partySize(in *ingestContext, sessionID string)
...
+// staff_touched is NOT a guard here, unlike upsertAssignment's write-path
+// skip: that skip protects a staff move from being overwritten by a
+// CONFLICTING campminder_sync value, but a cancelled household is not
+// attending regardless of who last touched its placement -- the same ruling
+// #2028 makes for the draft-null pass in stranded_assignment_cleanup.go.
+//
+// Per-session guard: a session with zero reliably-enrolled parties of a grain
+// is left untouched -- an attendee-sync hiccup must not read as "everyone
+// cancelled" and empty the whole session's placements.
+func (s *LodgingAssignmentsSync) deleteLodgingOrphans(year int) error {
```
adamflagg: **Destructive delete driven by a possibly-incomplete local snapshot, no audit trail.**

`deleteLodgingOrphans` permanently deletes `lodging_assignments` rows based on `s.householdSessionIndex`/`s.personSessionIndex`, which are built from whatever `attendees` rows currently exist locally for `year` — not a fresh CampMinder pull. The per-session `reliable` guard only protects against a session going to **zero** enrolled parties; it does nothing for a **partial** gap (some but not all of a session's attendee rows missing/stale).

Two concrete ways this bites:
- `handleLodgingAssignmentsSync` (api.go:1969) accepts an unrestricted `?year=2017..2050` and runs `Sync()` — including this delete pass — against whatever local attendees snapshot exists for that year, without re-syncing attendees first.
- Any current-year run where the `attendees` source sync partially fails (some rows land, some don't) but still reports success: the session stays "reliable" (non-empty), so genuinely-still-enrolled households whose attendee row is simply missing get `s.App.Delete(rec)`'d at line 878.

Unlike every other mutation path in this file (`writeHistory` on the placement/label-change paths), the delete here writes **no** `lodging_assignment_history` row — only an ephemeral log line. A false-positive sweep is silent and unrecoverable.

Worth at least: writing a history row on delete, and/or gating the delete pass on some completeness signal for the attendees snapshot (not just "session isn't totally empty").

pocketbase/sync/stranded_assignment_cleanup.go
```diff
@@ -175,7 +235,10 @@ func (s *StrandedAssignmentCleanupSync) Sync(_ context.Context) error {
 			return fmt.Errorf("stranded_assignment_cleanup: year resolution failed: %w", err)
 		}
 	}
-	return reconcileStrandedAssignments(s.App, year, &s.Stats)
+	if err := reconcileStrandedAssignments(s.App, year, &s.Stats); err != nil {
```
adamflagg, line 266: **`StrandedAssignmentCleanupSync.Sync()` never resets `s.Stats` between runs — this PR adds more accumulation into the same broken counter.**

`orchestrator.go:1876` registers a single `StrandedAssignmentCleanupSync` instance for the process lifetime, and every daily/manual run calls that same instance's `Sync()`. Every other sync service in this package resets `s.Stats = Stats{}` at the top of `Sync()` (e.g. `lodging_assignments_sync.go:81`, this PR's own sibling code) — this one never has.

Concretely: day 1 sweeps 2 bunk drafts + 3 lodging drafts → `Stats.Updated = 5`. Day 2 finds zero orphans of either kind → `GetStats()` still reports `Updated = 5` (plus whatever `Errors` accumulated since process start), because nothing ever clears it. This is surfaced to admins via `useSyncStatusAPI.ts` (`stranded_assignment_cleanup: SyncStatus`) — the Sync tab shows a monotonically growing, wrong "updated" count that no longer reflects the current run, and now mixes bunk-draft-nulls with lodging-draft-nulls in the same never-cleared total.

This bug predates this PR (it's true for the bunk-only pass already), but this diff touches `Sync()` directly and adds a second source of writes into the same unreset field, making it worse. Worth a `s.Stats = Stats{}` at the top of `Sync()` while this function is already being touched.

pocketbase/sync/stranded_assignment_cleanup.go
```diff
@@ -314,3 +377,113 @@ func reconcileStrandedAssignments(app core.App, year int, stats *Stats) error {
...
+//  1. lodging_assignments_draft: null `units` on rows whose party is no longer
+//     enrolled, across every scenario for the year. staff_touched and source
+//     are left exactly as they are -- only the placement clears, mirroring
+//     the bunk + bunk_plan nulling above. staff_touched is NOT a skip guard
+//     here: a cancelled household is gone whether or not staff moved it.
+//  2. lodging_assignments (the mirror): audit only. Logs and counts; never
+//     deletes. Deletion belongs to LodgingAssignmentsSync.deleteLodgingOrphans,
+//     driven by absence from the CampMinder cabin value -- the same split of
+//     responsibility bunk_assignments already has with the pass above.
+func reconcileLodgingOrphans(app core.App, year int, stats *Stats) error {
```
adamflagg, line 425: **`reconcileLodgingOrphans` hard-fails on an index-build error, unlike this file's own fail-soft convention for the same class of failure.**

If `BuildHouseholdSessionIndex` or `BuildPersonSessionIndex` errors (transient DB hiccup), this function returns the error immediately (lines 401/406), which `Sync()` (line 238-241) propagates straight out. `orchestrator.go`'s `runSyncAndWait` then marks the **entire** `stranded_assignment_cleanup` job `Failed` for the day — even though `reconcileStrandedAssignments` (the bunk pass, called just before) already ran, wrote its draft-nulling changes, and WAL-checkpointed successfully.

Compare `loadEnrolledAttendeeSets` (used by the bunk pass a few lines above), which handles the identical failure mode by logging + incrementing `stats.Errors` and returning **empty maps**, letting the per-session guard naturally no-op the sweep while `Sync()` still returns `nil`. `reconcileLodgingOrphans` should probably do the same — degrade gracefully (skip the lodging sweep for this run, but don't fail the whole job and don't discard the bunk pass's already-committed success).

pocketbase/sync/lodging_assignments_sync.go
```diff
@@ -804,6 +823,69 @@ func (s *LodgingAssignmentsSync) partySize(in *ingestContext, sessionID string)
...
+
+	householdReliable := reliableEnrolledSessions(s.householdSessionIndex)
+	personReliable := reliableEnrolledSessions(s.personSessionIndex)
+
+	for _, rec := range rows {
+		sessionID := rec.GetString("session")
+		hhCMID := rec.GetInt("household_cm_id")
+		personCMID := rec.GetInt("person_cm_id")
+
+		var stillEnrolled, reliable bool
```
adamflagg: **Two hand-synced copies of "is this row orphaned."**

`deleteLodgingOrphans` re-derives the household/person orphan predicate inline (hhCMID/personCMID switch + reliable-session guard + `sessionIndexHasWindow`) instead of calling `findLodgingEnrollmentOrphans` (stranded_assignment_cleanup.go:156), which computes the identical thing and is already used by `reconcileLodgingOrphans`'s prod audit to log "N orphans detected" for these same rows.

A future change to the reliability guard or the grain-dispatch logic applied to only one copy would make the audit-log warning count and the actual deletion count silently disagree — worth converting this loop to build `lodgingOrphanCandidate`s and call the shared function, the same way the prod audit already does.

pocketbase/sync/orchestrator.go
```diff
@@ -235,6 +235,9 @@ type Stats struct {
 	AlreadyProcessed int `json:"already_processed,omitempty"`
 	// ProdAuditWarnings counts bunk_assignments rows found stranded but not cleared (observe-only).
 	ProdAuditWarnings int `json:"prod_audit_warnings,omitempty"`
+	// LodgingProdAuditWarnings counts lodging_assignments rows found enrollment-orphaned
+	// but not cleared (observe-only; deletion is LodgingAssignmentsSync's job, #2028).
+	LodgingProdAuditWarnings int `json:"lodging_prod_audit_warnings,omitempty"`
```
adamflagg, line 240: `Stats.LodgingProdAuditWarnings` has no frontend consumer. Its sibling `ProdAuditWarnings` is rendered by `SyncTab.tsx:295-298` as a "N prod⚠" badge, but `frontend/src/hooks/useSyncStatusAPI.ts` only declares `prod_audit_warnings?: number` — `lodging_prod_audit_warnings` isn't in the type and isn't read anywhere in `frontend/src/`. The new audit signal this PR adds is currently invisible on the admin Sync tab even though the parallel bunk signal is visible on the same card.

pocketbase/sync/lodging_assignments_sync.go
```diff
@@ -804,6 +823,69 @@ func (s *LodgingAssignmentsSync) partySize(in *ingestContext, sessionID string)
 	return enrolled + adultCount
 }
 
+// deleteLodgingOrphans removes lodging_assignments rows for a party no longer
```
adamflagg, line 872: **(Altitude) The root cause generalizes past lodging; this PR fixes only the lodging symptom.**

The premise here — "CampMinder never clears a custom-field value on cancellation, so a forward-only ingest can't detect cancellation on its own" — isn't lodging-specific. `camper_dietary.go`'s `deleteOrphans` (line 559), and the same shape in `camper_transportation.go`, `staff_skills.go`, `household_demographics.go`, `quest_registrations.go`, and `family_camp_derived.go`, all key orphan detection on "does this party's value still exist in the current custom-value scan" — not "is this party still actively enrolled." If CampMinder leaves those source fields non-empty after cancellation the same way it does the cabin field (plausible, same platform behavior), a camper who cancels after being recorded with a dietary need, transportation request, quest registration, staff skill, or demographic value keeps that derived row forever too — the exact bug #2028 describes, just unfixed in six other places.

Not asking this PR to fix all six, but worth filing a follow-up issue now rather than letting each sibling get rediscovered and hand-fixed independently (as this PR itself had to hand-write a second "reliable session" guard next to the bunk-side one that already existed — see the other comment on `reliableEnrolledSessions`).

pocketbase/sync/lodging_session_attribution.go
```diff
@@ -268,6 +268,35 @@ func findAllRecords(app core.App, collection, filter string, params ...dbx.Param
...
+func sessionIndexHasWindow(windows []SessionWindow, sessionID string) bool {
+	for _, w := range windows {
+		if w.ID == sessionID {
+			return true
+		}
+	}
+	return false
+}
+
+// reliableEnrolledSessions returns the set of session PB ids with at least one
```
adamflagg, line 283: `reliableEnrolledSessions`'s doc comment says it's "the per-session guard #2028 shares with `stranded_assignment_cleanup.go`'s `findEnrollmentOrphans`," but they never call into each other — they're two independent implementations of "does this session have ≥1 enrolled party":

- `loadEnrolledAttendeeSets` (stranded_assignment_cleanup.go:122) derives `enrolledSessions` from a direct `attendees` scan with a hardcoded `status_id = 2` literal.
- `reliableEnrolledSessions` derives the same fact indirectly, via `BuildHouseholdSessionIndex`/`BuildPersonSessionIndex`'s named `statusIDActiveEnrolled` constant (lodging_session_attribution.go:191).

If "actively enrolled" ever needs a second qualifying status, the hardcoded literal is the one likely to get missed, and a reader trusting this comment would wrongly assume a fix to one propagates to the other.

pocketbase/sync/stranded_assignment_cleanup.go
```diff
@@ -125,6 +133,55 @@ func loadEnrolledAttendeeSets(app core.App, year int, stats *Stats) (enrolledSes
...
+// findLodgingEnrollmentOrphans returns the candidates whose party (household or
+// person, by CampMinder id) is absent from its named session in
+// householdIndex / personIndex -- the household/person-grain analog of
+// findEnrollmentOrphans above, built from BuildHouseholdSessionIndex /
+// BuildPersonSessionIndex rather than a second attendees query, per #2028's
+// insistence on reuse over re-derivation. A candidate naming neither id is
+// skipped (grain-less); a session with zero reliably-enrolled parties of that
+// grain is skipped entirely, mirroring findEnrollmentOrphans' own per-session
+// guard.
+func findLodgingEnrollmentOrphans(
```
adamflagg, line 178: `findLodgingEnrollmentOrphans`'s two switch cases (`c.HouseholdCMID > 0` / `c.PersonCMID > 0`) repeat the identical three-step pattern — skip if `!reliable[c.SessionID]`, append if `!sessionIndexHasWindow(index[cmID], c.SessionID)` — differing only in which index/reliable map is consulted. A small helper (`lodgingCandidateOrphaned(cmID, sessionID, reliable, index) bool`) would collapse both arms to one line each, and — combined with the earlier comment on `deleteLodgingOrphans` — could be the single place both copies of this logic actually live.

pocketbase/sync/lodging_assignments_sync.go
```diff
@@ -43,6 +43,13 @@ type LodgingAssignmentsSync struct {
 	personsByHouseholdCMID  map[int][]*core.Record
 	enrolledByPersonSession map[string]int // "<personPBID>|<sessionPBID>" -> count
 	adultsByHouseholdPBID   map[string]int
+
+	// householdSessionIndex / personSessionIndex are the SAME indexes
+	// syncHouseholdGrain / syncPersonGrain build to attribute values, captured
+	// here so deleteLodgingOrphans can reuse them rather than re-deriving the
+	// enrolled set (#2028).
+	householdSessionIndex map[int][]SessionWindow
```
adamflagg, line 57: `householdSessionIndex`/`personSessionIndex` are permanent struct fields whose only valid lifetime is a single `Sync()` call — they exist purely to carry data from `syncHouseholdGrain`/`syncPersonGrain` to `deleteLodgingOrphans`. That's a temporal-coupling contract enforced only by the doc comment above, not the type system: nothing stops `deleteLodgingOrphans` from being called (directly, from a test, or after a future reordering of `Sync()`) before these fields are populated, in which case every row would silently read as "not enrolled anywhere" (a nil-map read returns the zero value) rather than failing loudly. Passing both indexes as explicit parameters to `deleteLodgingOrphans(year, householdIndex, personIndex)` would make this a compile-time-checked dependency instead of a runtime one.

pocketbase/sync/stranded_assignment_cleanup.go
```diff
@@ -314,3 +377,113 @@ func reconcileStrandedAssignments(app core.App, year int, stats *Stats) error {
...
+
+	// --- Draft sweep ---
+	drafts, err := app.FindRecordsByFilter("lodging_assignments_draft", yearFilter, "", 0, 0)
+	if err != nil {
+		stats.Errors++
+		return fmt.Errorf("stranded_assignment_cleanup: query lodging_assignments_draft: %w", err)
+	}
+	draftByID := make(map[string]*core.Record, len(drafts))
+	draftCandidates := make([]lodgingOrphanCandidate, 0, len(drafts))
+	for _, d := range drafts {
```
adamflagg: The draft-candidate loop here and the prod-candidate loop at line 453-462 duplicate the same "skip if already unplaced, else build a `lodgingOrphanCandidate`" logic verbatim. A shared `lodgingCandidateOf(rec) (lodgingOrphanCandidate, bool)` helper would collapse both ~9-line blocks to a couple of lines each — as written, a field added to `lodgingOrphanCandidate` has to be added in both places, and an editor touching only the draft loop (the one with the write-back) can leave the prod audit's projection silently out of sync with it.

pocketbase/sync/stranded_assignment_cleanup.go
adamflagg, line 425: (Efficiency) `reconcileLodgingOrphans` rebuilds `BuildHouseholdSessionIndex`/`BuildPersonSessionIndex` from scratch (full `attendees`+`persons` scans), duplicating scans `LodgingAssignmentsSync.deleteLodgingOrphans` already performed moments earlier in the same sync run for the same year. `syncJobMeta` runs `lodging_assignments` before `stranded_assignment_cleanup` in the transform phase, and `s.householdSessionIndex`/`s.personSessionIndex` were captured specifically "so `deleteLodgingOrphans` can reuse them rather than re-deriving the enrolled set" (lodging_assignments_sync.go:47-50) — but that intent doesn't survive the job boundary; the next job re-derives the identical indexes from zero. Not a correctness issue (different service instance, same DB), just doubled I/O that a year-scoped cache passed through the orchestrator would avoid.

adamflagg (commented):
Automated recall-mode review (xhigh effort, 10 finder angles + verification sweep). 11 findings posted inline — 3 correctness/reliability (an unaudited destructive delete driven by a possibly-incomplete local attendee snapshot, a Stats-never-reset counter this PR compounds, and a hard-fail that skips the lodging sweep and mis-reports the whole job as Failed), plus reuse/duplication, an observability gap, and an altitude note that the fix doesn't generalize past lodging. No language-pitfall or wrapper-correctness issues found (Go idioms, filter syntax, WAL checkpoints, and CampMinder-ID keying all check out), and no CLAUDE.md violations found.

Issue #2301, opened 2026-08-13
refactor(sync): evaluate whether the four standalone sync services should embed BaseSyncService


adamflagg:
## ⚠️ CORRECTED 2026-08-16 — phase D triage. The survey is DONE. Answer: do not embed.

Verified against `main` @ `a946565a`.

**The evaluation this issue asks for has been carried out, and the answer is "do not embed `BaseSyncService`".**
This issue's own text accepts that as a valid outcome — *"If the survey shows the base class does not fit,
closing this with that finding written down is a good result."*

**⛔ Do not perform the refactor.** The owner's standing instruction is *"I want to wake up to fewer issues, not
more"*, and a speculative refactor with hedged benefits is exactly what that excludes.

### Two real defects surfaced by the survey — fold these into one small PR, then close

1. **Gate `deleteOrphans` on `SyncSuccessful`** in `pocketbase/sync/staff_applications.go` and
   `pocketbase/sync/staff_vehicle_info.go`. An orphan sweep that runs after a partial fetch can delete rows the
   feed simply did not return.
2. **Correct `pocketbase/sync/orphan_guard.go`'s "ten services" comment to nine** — `camper_history` was removed
   in #2366 (`fix(sync): remove the camper_history table, its sync, and dead consumers`).

Classification: **MECHANICAL, S.** Owner approved closing on this basis 2026-08-16, after the PR.

⚠️ **File collisions:** `pocketbase/sync/orphan_guard.go` is shared with **#2394** and **#2307**;
`pocketbase/sync/base_sync.go` with **#2394**. #2394 is the larger change — sequence behind it or take the
files first and let it wait, but not both at once.

---

## ⚠️ CORRECTED 2026-08-14 — phase C triage

Verified against `main` @ `3eb44ba4` and the production snapshot. Verdict: **STALE_BODY** / NO_CODE.

> The four services are correctly identified and still standalone, but the body's central evidence has expired: the interim fix it describes as future work shipped in #2296/#2299, so SyncSuccessful is now set right after extraction and read by each service's own sweep. The sentences "it is therefore always false during their own sweep" and "nothing ever reads it — write-only dead state" are both false against main. The survey (step 1) is still the right deliverable and its blockers are gone; a genuine finding for it is that financial_aid_applications.go carries a sixth `fieldEquals` copy the body does not count.

### Corrections to the body below

[...]

PR #2366, opened 2026-08-14
fix(sync): remove the camper_history table, its sync, and dead consumers

### LINKED ISSUE #2369: chore(sync): remove the camper_history table and its sync — every consumer has migrated off it
## Decision

**Remove the `camper_history` table, its sync service, and the resulting dead code.** Owner-directed 2026-08-14, conditional on an audit confirming no live consumer. That audit ran and confirmed it; this issue is the record, and the artifact PR #2366 should cite.

## Why it exists and why it goes

`camper_history` is a 31-column, 36,687-row denormalised table spanning 2017-2026. It was built to flatten camper data into one queryable shape. Every consumer has since migrated off it, file by file, without anyone retiring the table.

**Its columns break four ways:**

| group | columns | where the truth lives |
|---|---|---|
| person copies | `first_name`, `last_name`, `age`, `gender`, `grade`, `city`, `state`, `school`, `synagogue`, `household_id` | `persons` |
| session copies | `session_name`, `session_type`, `division_name`, `status`, `enrollment_date` | `attendees` + `camp_sessions` |
| bunk copies | `bunk_cm_id`, `bunk_name` | `bunk_assignments` |
| normalised | `city_normalized`, `school_normalized`, `congregation_normalized` | **already on `persons`** as `normalized_city` / `normalized_school` / `normalized_congregation` |
| genuinely unique | `is_returning_summer`, `is_returning_family` | nowhere else |

The normalised trio agrees with `persons` **2675/2675, 2580/2580 and 1390/1392** on 2026 rows — they are duplicates under transposed names. `years_at_camp` agrees **1868/1868** wherever `persons` has it set; the apparent disagreements are exactly the rows where `persons` holds `0` and the sync substituted a computed fallback.

That leaves **two booleans** as the table's entire irreducible content — and both are per-(person, year) facts stored at per-(person, session, year) grain, so even they are duplicated **1.25-1.38x** (2026: 2,872 rows for 2,296 people).

## Audit result — no live consumer

Four live reads of the collection, **all inside the deletion's own blast radius**:

1. `camper_history.go:257` — the writer's own collection handle.
2. `camper_history.go:1096-1097` — `preloadExistingRecords`, self-referential idempotency compare + orphan sweep.
3. `normalize_geographic.go:1001` — see the blocker below.
4. `multi_workbook_export.go:378` via `table_exporter.go:777` — the Google Sheets "Camper History" tab. **Owner has confirmed nobody reads that sheet.**

**No product, API, UI, reporting or solver surface reads it.** Verified negatives:

- `grep -rn "collection(['\"\`]camper_history" frontend/ api/ bunking/ tests/ pocketbase/ scripts/` → **no matches**.
- The frontend hook that *sounds* like a reader is not one: `useCamperHistory` builds from `attendees` plus `fetchCamperJourney`, which queries `camp_sessions`, `attendees` and `bunk_assignments`. `queryKeys.camperHistory` and the `'camper-history'` react-query prefix are **cache keys, not the collection** — they must survive.
- Python has four hits, all docstrings. `historical_service.py` documents having moved off it; `metrics_repository.py:232` says its approach works "without requiring the camper_history sync".

**No schema relation.** Against the live DB: zero collections carry `col_camper_history` in `fields`; zero reference it in any list/view/create/update/delete rule; zero view collections reference it. Its own two relations (`persons`, `camp_sessions`) are outbound and die with it. The only residue is 2 rows in `sync_runs` with `service='camper_history'` — a plain TEXT column, harmless audit history, no FK.

## ⚠️ The blocker that makes this not a schema-only drop

`normalize_geographic.go` Step 6 (`Sync()` :213-215 → `updateCamperHistoryNormalized` :980-1063) **reads** the collection and returns the wrapped error. **Dropping the collection without removing Step 6 fails `normalize_geographic` on every daily and unified sync run.** They must ship together.

Two orchestrator comments are stale and should go with it: `:1777` "depends on camper_history" and `:2386` "computes from camper_history". There is no input dependency — only the write-back.

## Reversibility

The down migration recreates the **schema only**; the 36,687 rows are not recoverable from it. That is acceptable because the data is **recomputable**: `camper_history.go` computes purely from PocketBase (`attendees`, `persons`, `bunk_assignments`, `camp_sessions`, `divisions`, `households`, `custom_field_defs`) with no CampMinder client. `attendees` spans the identical years within ~1% (2017: 3,162 vs 3,159 … 2026: 2,891 vs 2,872). A regretted deletion is a re-run from a restored binary, not a data-loss event.

## Relationship to #2301

**#2301 does NOT authorize this**, and its survey argues the other way: it names `camper_history` as one of three "Shape B" services where embedding `BaseSyncService` would be "the closest to a real drop-in of any of the ten". That recommendation is **moot for this service** once the table goes — #2301 needs correcting so a future agent is not sent to refactor a service that no longer exists. The embedding candidate set drops from three to two (`normalize_geographic`, `staff_skills`).

## Scope

`camper_history.go` + its test · `normalize_geographic.go` Step 6 and its fixture · `orchestrator.go` registration/ordering + two stale comments · `api.go` endpoint · `table_exporter.go` entry · frontend sync-admin plumbing (`SyncTab.tsx`, `syncTypes.ts`, `useRunIndividualSync.ts`, `useSyncStatusAPI.ts`, `useSyncCompletionToasts.ts`) · a DROP migration · regenerated `pocketbase-types.ts`.

Implemented by PR #2366.

adamflagg:
## Summary

Removes the `camper_history` PocketBase collection, its Go sync writer, and every consumer.

An audit (see #2369) found this table has **zero live consumers** outside its own writer's
blast radius: no product surface, API endpoint, UI component, reporting query, or solver path
reads `camper_history`, and no other collection carries a schema relation into it. The only two
reads were the writer's own idempotency/orphan-sweep comparison, and a `*_normalized` column
rewrite inside `normalize_geographic.go` — both removed here.

## What this removes

- `pocketbase/sync/camper_history.go` (1,191 lines) and its test (2,709 lines) — the writer
- Its registration and ordering in `orchestrator.go` (`syncJobMeta`, the daily job order,
  `GetDefaultUnifiedSyncJobs`, and both registration paths — `InitializeSyncServices` and
  `RunSyncWithOptions`)
- Its endpoint (`POST /api/custom/sync/camper-history`) and handler in `api.go`
- Its entry in `table_exporter.go` — the Google Sheets export (owner confirmed nobody reads
  that tab either)
- The frontend sync-admin plumbing: `SyncTab.tsx`, `syncTypes.ts`, `useRunIndividualSync.ts`,
  `useSyncStatusAPI.ts`, `useSyncCompletionToasts.ts`, and the now-dead
  `useCamperHistorySync.ts` hook (deleted outright)
- A migration (`1500000157_delete_camper_history.js`) dropping the collection
- Regenerated `frontend/src/types/pocketbase-types.ts` against a scratch DB built from this
  branch's migrations. The check that enforces this is the **CI** `pocketbase-types` freshness
  job (`.github/workflows/ci.yml`, added by #2278), which boots PocketBase against
  `pb_migrations/` and diffs the regenerated file — not the `api-types-freshness` pre-push
  hook, which covers `src/types/api-generated/` from the Python schemas and is unrelated
- Several stale docs/comments that named the removed collection or its deleted handler as a
  live example (the `pocketbase-migrations` skill's own reference doc, `sync-layer.md`,
  `tables.md`, `go-sync-patterns.md`, `sync-id-conventions.md`, `metrics-module.md`,
  `api/schemas/metrics.py`, `bunking/metrics/__init__.py`)

**Left alone, on purpose:** `orphan_guard.go` and `base_sync.go` each carry one doc-comment
mention of `camper_history` in a list of sibling services. Both files are locked by a
concurrently-in-flight PR (#2325), so those two comments go mildly stale here rather than being
touched — noted for whoever lands next in that file.

## The hard part the initial audit under-stated

`normalize_geographic.go`'s Step 6 (`updateCamperHistoryNormalized`) does
`FindRecordsByFilter("camper_history", …)`, and `Sync()` wraps and returns that error. Dropping
the collection without also removing this step would have failed `normalize_geographic` on
**every daily and unified sync run** — not a hypothetical, a guaranteed production break the
very next time that service ran.

Fixed here by removing Step 6 outright. The **persons-normalization half survives unchanged**
(Step 6, renumbered) — `normalize_geographic.go` still writes `persons.normalized_city` /
`normalized_school` / `normalized_congregation`, which is what `registration_service.py`'s
`by_school`/`by_city`/`by_synagogue` breakdowns actually read.

TDD proof this is real, not assumed: `TestNormalizeGeographicSyncPropagatesSweepRefusal`'s test
fixture built a `camper_history` collection specifically because `Sync()` needed it to exist.
Removing that collection from the fixture *before* touching production code turned the test RED
with exactly the failure this bug would cause in production:

```
Sync error = "updating camper history: querying camper_history page 1: sql: no rows in result set"
```

Fixing `normalize_geographic.go` turned it GREEN again — now without needing the collection to
exist at all, proving persons-normalization keeps working with `camper_history` gone.

A second, purpose-built test (`TestCamperHistoryServiceFullyRemoved`) pins the orchestrator side:
`camper_history` is absent from `getDailySyncJobs`, `GetDefaultUnifiedSyncJobs` (both
`includeCustomValues` settings), `syncJobMeta`, and a fresh orchestrator's service registry.

## Confirmed NOT a false positive: the "CamperHistory" UI feature

`frontend/src/{contexts/CamperHistoryContext.ts, providers/CamperHistoryProvider.tsx,
hooks/useCamperHistory.ts, hooks/camper/useCamperHistory.ts}` and their consumers
(`CamperCard`, `CamperDetail`, `CamperTooltip`, `SessionView`, `CamperDetailsPanel`, etc.) share
the name but are a completely unrelated, live feature — a "camper's history at this camp" panel
sourced directly from `bunk_assignments` / `attendees` / `camp_sessions`, never from the
`camper_history` PocketBase collection. Read every one of these files before excluding them from
this PR; none of them changed.

## Data loss — read before merging

**36,687 rows are dropped and are NOT recoverable from this migration's down arm alone.**
SQLite has no undo for `DROP TABLE`; the down migration recreates the empty *schema* only (all
30 fields, all 10 indexes, matching the collection's final live shape). That loss is acceptable
because the table was itself a computed cache — every field on it is re-derivable from
`attendees` + `persons` + `bunk_assignments` + `household_custom_values`, the same source tables
the writer read to populate it in the first place. Nothing on it was a system of record.

## Verification

- `go build ./...` and `go test ./...` — green (full `pocketbase` module, all packages)
- `npx tsc --noEmit` and `npx vitest run` — green (6,825 frontend tests)
- `uv run pytest tests/` — green (6,337 passed / 23 skipped, all server-hitting)
- `bash scripts/pre-push-verify.sh` — all checks green: ruff format/check, mypy, pytest,
  go build, **golangci-lint**, go test, prettier, **eslint**, tsc, vitest
- **PocketBase binary boot, not just `go test`**: built the real binary and ran
  `scripts/dev/lib/pb-harness.sh`'s `pb_harness_boot` against a fresh scratch DB with every
  migration (including `1500000157`) applied — confirmed `camper_history` absent from
  `_collections`, 72 total collections landed (not just PocketBase's own defaults), and
  `normalize_geographic`'s surviving read paths (`persons`, `normalized_mappings`, `attendees`,
  `bunk_assignments`) intact.
- **Down migration verified live**, not just read: ran `migrate down 1` (`printf 'y\n' | ...`,
  interactive per this repo's convention) against that same booted DB. Recreated
  `camper_history` with all 31 columns (30 fields + `id`) and all 10 indexes, confirmed via
  `PRAGMA table_info` and a `sqlite_master` index-name query.
- Pushed branch verified by the remote ref moving (`git rev-list --count
  origin/<branch>..HEAD` → 0), not by tool exit code.

See #2369.

## Summary by CodeRabbit

* **Changes**
  * Removed Camper History from synchronization, status tracking, individual sync actions, exports, and generated data types.
  * Removed the Camper History collection from the data model; historical records will not be recreated during rollback.
  * Updated metrics to use attendee and person data, with historical no-data messaging clarified.
  * Updated documentation and tests to reflect the revised sync, export, and metrics behavior.

Closes #2369.


adamflagg:
## Post-scan review of the three commits pushed after the first scan

Reviewed `18eb5af8` (reference repointing) and `d13e25b2` (the up-arm guard fix). `d0e2f1f9` was
context only. Everything below was re-derived from the tree rather than read off the commit
messages — two of the claims in them did not survive that.

### 1. The guard fix — `extract_created_collections`

**The premise is real and the fix works.** `1500000157_delete_camper_history.js` has no
`new Collection` in its up arm at all (`app.delete`), and recreates the collection in its down arm
for reversibility. The whole-file scan read that down arm as a create.

Independently re-derived, not taken from the commit message:

| Check | Result |
|---|---|
| Down-arm forms actually present | `}, (app) => {` ×111, `}, () => {` ×4, `}, (_app) => {` ×1, `}, function (` ×1 — **117 of 117** |
| Migrations with no down arm | **0** (the commit said 4; those 4 are `}, () => {`, which `\w*` already matched) |
| Boundary vs. the true top-level `migrate(up, down)` comma, by brace/paren-depth parse | **0 mismatches across 117** |
| Files where the regex matched more than once | **0** |
| Before/after extraction across all 117 | **exactly one file changes** — `1500000157` goes from `['camper_history']` to `[]`. Confirmed. |
| Guard's own suite | 13/13 pass, `exit=0` |

**The dangerous failure mode is constructible, and I built it.** The search was unanchored and
took the *first* `}, (x) => {` anywhere in the file. A nested single-argument callback inside an
up arm matches it:

```js
seedDefaults({ retention_days: 30 }, (row) => {   // <-- boundary matched HERE, offset 176
  row.set("x", 1);
});

const collection = new Collection({ name: "brand_new_thing", ... });  // truncated out of the scan
```

Extraction returned `[]` instead of `['brand_new_thing']` — a **silent false CLEAN** on a boot that
would fail, which is the one direction this detector cannot afford (its own TEST 11 comment says
so). No migration in the tree has that shape today, so this was latent, not live.

Fixed in `06a7fca0` by anchoring to a line start with at most two spaces — where the top-level
boundary sits in every migration here (column 0 in 80 files, column 2 in the 37 prettier wraps).
Extraction is **byte-identical to the unanchored version for all 117 migrations**, and no boundary
match at all now falls back to scanning the whole file, i.e. toward the loud false positive rather
than the silent miss.

**No regression test shipped with the fix, and TEST 1 cannot substitute for one** — `1500000157` is
applied on any dev DB that has booted since, so the real-DB check no longer reaches the shape
(verified: `_migrations` has the row, `_collections` has no `camper_history`). Added two, both
mutation-checked:

- **TEST 14** — a down-arm recreate is not a create (fixture, the real `1500000157`)
- **TEST 15** — a nested single-argument callback in the up arm must not truncate the scan

<details><summary>Mutation transcript</summary>

```
A. revert the up-arm split entirely (pre-d13e25b2 logic)
   === TEST 14: a DOWN-arm recreate is not a create ===
   FAIL: expected exit 0 — a down-arm recreate is not a create, got 1
   suite_exit=1

B. drop the ^[ \t]{0,2} anchor + re.M (as-committed logic)
   === TEST 14 ===  PASS: down-arm recreate not mistaken for a create
   === TEST 15: a nested callback in the UP arm must not truncate the scan ===
   FAIL: expected exit 1 — a nested up-arm callback truncated the scan, got 0
   suite_exit=1

RESTORED
   All tests passed.   suite_exit=0        (15/15)
```

Standalone false-positive reproduction, camper_history live in `_collections`, only `1500000157`
on disk:

```
WITH the fix     GUARD_EXIT=0   (no collision reported)
WITHOUT the fix  GUARD_EXIT=1   COLLECTION ALREADY EXISTS — collection: camper_history
```
</details>

### 2. The reference repointing (`18eb5af8`)

- All four changed lines are comments — two ` * ` lines inside the migration's `/** */` header,
  two Go `//` lines. **No executable line changed.**
- **#2369 exists, is open, and is the removal record** — it names PR #2366 as the artifact that
  should cite it. **#2301 is unrelated**: the BaseSyncService-embedding survey.
- **One `#2301` reference survived the sweep.** `bunking/metrics/__init__.py:5` — a docstring this
  PR itself wrote in `52c00f0b` — still cited `#2301` for the removal. Fixed in `d3904ee0`; that
  was the last one in the tree.
- No stray closing keyword in prose. The only one is the deliberate `Closes #2369.` trailer at the
  end of the PR body; nothing else in the body or in any branch commit message carries one.

### Verified green

`bash scripts/pre-push-verify.sh` → `ALL CHECKS PASSED`, exit 0 (ruff, mypy, shellcheck,
api-types-freshness, 6149 passed / 14 skipped). Push confirmed by the remote ref moving
(`d13e25b2..d3904ee0`), then `git rev-list --count origin/feature/remove-camper-history..HEAD` → `0`.
Re-polled after the push: **CI Summary `pass`**, zero failing checks on `d3904ee0`.

### Pre-existing gap, not fixed here

`scripts/dev/test-verify-migration-history.sh` **runs nowhere automatically** — not in CI, not in
pre-push. Only `test-verify-no-hardcoded-lodging.sh` has a `Guard self-test` step. So the 15 tests
above, including the two just added, only run when someone runs them by hand. Left alone
deliberately: wiring it means either a new job plus four touches to the `CI Summary` `needs:` gate,
or a step inside `tests-migrations`, which is `detect-changes`-gated on paths a `scripts/dev/`-only
change would not match. That is a CI-topology call, and `ci.yml`'s own comment about kindred#2181
("entirely unwired before, so nothing ran it") is the argument for making it deliberately rather
than as a side effect of this PR.

### Not merging

The PR touches `frontend/src/components/admin/SyncTab.tsx` and
`frontend/src/pages/metrics/trends/TrendsOverview.tsx` — a sync control disappears and metrics
messaging changes, so it holds for the owner's eyes. Worktree
`~/kindred-worktrees/remove-camper-history`, branch `feature/remove-camper-history`, Vite `3112`,
Caddy `8312`.

PR #2373, opened 2026-08-15
ci(ci): wire the migration-history guard's orphaned self-test into CI

### LINKED ISSUE #2370: ci: the migration-history guard's self-test has no trigger — 15 tests run only by hand
## What

`scripts/dev/test-verify-migration-history.sh` — the self-test for the migration-history guard — **runs nowhere automatically.** All 15 tests only execute when someone runs them by hand.

Verified against `main`:

- **Not in CI.** `.github/workflows/ci.yml` wires exactly one guard self-test, `test-verify-no-hardcoded-lodging.sh`, under a step named "Guard self-test". There is no equivalent step for the migration-history guard.
- **Not in pre-push.** `.lefthook.yml` runs `shellcheck` over changed shell scripts, which checks syntax, not behaviour.

## Why it matters now

The guard it tests, `scripts/dev/verify-migration-history.sh`, gates local `start_dev.sh` boots and is the thing standing between a renumbered migration and a corrupted local database. It just shipped a real fix (PR #2366) for a false positive that blocked every dev server on a branch — and the review of that fix found a **second**, worse defect in the first attempt: an unanchored boundary regex that would have made the guard report **silently CLEAN on a real collision**, the one direction a collision predictor must never fail in.

Both defects are now pinned by TEST 14 and TEST 15. **Neither test will run again unless a human remembers to run it.**

That is the whole argument: the suite's value is regression protection, and a regression suite with no trigger provides none.

## Why it was not just wired up in that PR

This is a CI-topology decision, not a mechanical edit, and doing it as a side effect of a `camper_history`-removal PR is the kind of unconsidered wiring `ci.yml`'s own kindred#2181 comment warns about. Two shapes, both with a real cost:

1. **A new job**, plus four touches to the `CI Summary` `needs:` gate. Explicit and readable; adds a job to every run.
2. **A step inside `tests-migrations`.** Cheaper, but that job is `detect-changes`-gated on paths that a `scripts/dev/`-only change would not match — so the suite would still not run on the very commits most likely to break it, unless the path filter is widened too.

Option 2 without widening the filter is the trap: it looks wired and is not.

## Suggested

Option 1, plus the path filter for `scripts/dev/**` so a change to either the guard or its suite triggers it. Worth checking at the same time whether any other guard self-test in `scripts/dev/` has the same gap — `test-verify-consolidation.sh`, `test-verify-migration-schema-diff.sh`, `test-pb-harness.sh` and `test-verify-no-hardcoded-lodging.sh` all exist, and only the last is known-wired.

## Sequencing

- **Files:** `.github/workflows/ci.yml`; possibly `.lefthook.yml`.
- **Predecessors:** none — PR #2366 (which added TESTs 14 and 15) has merged.
- **Collides with:** nothing currently open.

adamflagg:
## The defect

`scripts/dev/test-verify-migration-history.sh` (15 tests) is the self-test for
`scripts/dev/verify-migration-history.sh` — the guard that gates local
`start_dev.sh` boots and stands between a renumbered migration and a
corrupted local DB. Before this PR, that self-test ran nowhere in CI.

Verified against the tree: `.github/workflows/ci.yml` wired exactly one guard
self-test (`test-verify-no-hardcoded-lodging.sh`, in the `lodging-guard` job's
"Guard self-test" step), and `.lefthook.yml` runs only `shellcheck` over
changed shell scripts — syntax, not behavior. TEST 14 and TEST 15 in the
migration-history suite pin a real defect fixed in PR #2366: an unanchored
boundary regex that made the guard report silently CLEAN on a real migration
collision — the one direction a collision predictor must never fail in. With
no trigger, a regression there could ship and nothing would catch it.

## What changed

Adopted Option 1 from the issue:

1. **New job `guard-self-tests`** ("Dev Guard Self-Tests"), wired into the
   `CI Summary` gate the same way `lodging-guard` is: added to `summary`'s
   `needs:` list, a `GUARD_SELF_TESTS` env var reading its result, and a
   success-or-skipped check in the summary step.
2. **New `guardSelfTests` path filter** on `scripts/dev/**` plus
   `.github/workflows/ci.yml`, and the job is gated on it
   (`needs.detect-changes.outputs.guardSelfTests == 'true'`) — not on
   `migrations` or `go`. That's deliberate: a change to only a script under
   `scripts/dev/` (the guard, its suite, or the shared `pb-harness.sh` lib)
   touches neither of those filters, which is exactly the documented trap
   the issue calls out for putting a step inside `tests-migrations` instead.
   `.github/workflows/ci.yml` is in the filter for the same reason the `go`
   filter above it already includes that file: the job's own shape lives
   here, so a PR that only edits the job (like this one) has to be able to
   trigger it too, or its own CI run proves nothing.
3. **Audited the other three guard self-tests** found under
   `scripts/dev/test-*.sh` and wired all three into the same job, having run
   each standalone first:
   - `test-verify-consolidation.sh` — 5/5 tests pass locally (~24s; builds a
     scratch `pocketbase` binary).
   - `test-migration-schema-diff.sh` — 7/7 tests pass locally (~0.2s). Note:
     the issue names this file `test-verify-migration-schema-diff.sh`; the
     actual filename on `main` is `test-migration-schema-diff.sh` (no
     `verify-` infix). Wired under its real name.
   - `test-pb-harness.sh` — 12/12 tests pass locally (~1.2s; also builds a
     scratch `pocketbase` binary).

   All four are self-contained (scratch dirs / scratch SQLite DBs / scratch
   Go builds via `pb_harness_build_binary`) — none needs a live PocketBase
   server or a seeded dev DB, so none carries flake risk from that.

## What I deliberately did not do

- Did not touch `tests-migrations`'s own gate (`migrations || go`) — that's
  the documented trap (option 2), not the fix.
- Did not widen the pre-existing `scripts` filter (`scripts/**/*.sh`) to
  cover this. That glob already matches `scripts/dev/**/*.sh` via `**`, but
  the only job reading that output is `go-lint`'s shellcheck step
  (syntax-only). Reusing it would have made `guard-self-tests`'s trigger an
  incidental side effect of an unrelated filter rather than a stated
  dependency, so I added a dedicated `guardSelfTests` filter instead.
- Did not touch `.lefthook.yml` — pre-push already stays fast by design
  (syntax-only shellcheck); these are behavioral suites, CI's job.

## Verification

Local, this branch, exit codes read directly (not the summary line):

```
$ bash scripts/dev/test-verify-migration-history.sh; echo exit=$?
... All tests passed.
exit=0   (15/15)

$ bash scripts/dev/test-verify-consolidation.sh; echo exit=$?
... All tests passed.
exit=0   (5/5)

$ bash scripts/dev/test-migration-schema-diff.sh; echo exit=$?
... All tests passed.
exit=0   (7/7)

$ bash scripts/dev/test-pb-harness.sh; echo exit=$?
... All tests passed.
exit=0   (12/12)
```

`python3 -c "import yaml,sys; yaml.safe_load(open('.github/workflows/ci.yml'))"`
parses clean.

`bash scripts/pre-push-verify.sh --all` passed.

**The real proof is this PR's own CI run.** This PR touches only
`.github/workflows/ci.yml`, and the `guardSelfTests` filter explicitly
includes that file (see above), so `guard-self-tests` must appear and pass
in this PR's own checks. If it doesn't appear, the filter is wrong and I'll
fix it and push again before asking for review.

Closes #2370.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -199,9 +199,19 @@ jobs:
           # must still be able to trigger it. Without this line, that PR
           # would ship a job whose own CI run never proves it runs at all,
           # which is the exact failure kindred#2370 is about.
+          # `pocketbase/pb_migrations/**` is included because
+          # test-verify-consolidation.sh (wired into this same job) runs
+          # end-to-end against the REAL migration tree, not a fixture --
+          # a migrations-only PR can add a schema-free tail migration
+          # (a pure data/RBAC rewrite like 1500000154) that breaks the
+          # suite's drift assertion. Without this line, that PR would not
+          # trigger guard-self-tests and would merge green; the red would
+          # then surface on some later, unrelated PR that happens to touch
+          # scripts/dev/ or ci.yml, misattributed to the wrong change.
           guardSelfTests:
             - 'scripts/dev/**'
             - '.github/workflows/ci.yml'
+            - 'pocketbase/pb_migrations/**'
           docker:
             - 'docker/**'
             - 'docker-compose*.yml'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,6 +90,7 @@ jobs:
       migrations: ${{ steps.filter.outputs.migrations }}
       frontend: ${{ steps.filter.outputs.frontend }}
       scripts: ${{ steps.filter.outputs.scripts }}
+      guardSelfTests: ${{ steps.filter.outputs.guardSelfTests }}
       docker: ${{ steps.filter.outputs.docker }}
       actions: ${{ steps.filter.outputs.actions }}
     steps:
@@ -179,6 +180,28 @@ jobs:
           scripts:
             - 'scripts/**/*.sh'
             - '.shellcheckrc'
+          # kindred#2370: scripts/dev/test-*.sh are the guard scripts' own
+          # self-tests (e.g. test-verify-migration-history.sh pins the
+          # unanchored-boundary regression from PR #2366) and, before this
+          # filter existed, ran nowhere in CI -- the `scripts` filter above
+          # already matches this directory via `scripts/**/*.sh`, but the
+          # only job reading that output is go-lint's shellcheck step
+          # (syntax, not behavior). A dedicated filter makes the
+          # guard-self-tests job's gate say what it actually depends on,
+          # instead of relying on unrelated incidental coverage from a
+          # filter meant for something else. `scripts/dev/**` (not just
+          # `*.sh`) also covers scripts/dev/lib/pb-harness.sh, the shared
+          # lib every guard here delegates its tool-absence contract to.
+          # `.github/workflows/ci.yml` is included for the same reason the
+          # `go` filter above includes it: the guard-self-tests job's own
+          # shape (which scripts it runs, how) lives in this file, so a PR
+          # that only edits the job -- like the one that first added it --
+          # must still be able to trigger it. Without this line, that PR
+          # would ship a job whose own CI run never proves it runs at all,
+          # which is the exact failure kindred#2370 is about.
+          guardSelfTests:
+            - 'scripts/dev/**'
+            - '.github/workflows/ci.yml'
           docker:
             - 'docker/**'
             - 'docker-compose*.yml'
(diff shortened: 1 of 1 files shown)


adamflagg:
## Review follow-up (triage)

Reviewer filed one finding; fixed in-branch.

**Finding (medium, XS, `.github/workflows/ci.yml:202`, in code this PR wrote):**
The `guardSelfTests` filter (`scripts/dev/**` + `.github/workflows/ci.yml`) didn't cover `pocketbase/pb_migrations/**`, but `test-verify-consolidation.sh` — wired into this same job — runs end-to-end against the *real* migration tree. A migrations-only PR that lands a schema-free tail migration (a pure data/RBAC rewrite, same shape as the already-merged `1500000154_drop_lodging_phi_permission.js`) breaks TEST 2's drift assertion without ever triggering `guard-self-tests`, so it merges green and the red surfaces later on an unrelated PR that happens to touch `scripts/dev/` or `ci.yml`.

Reproduced independently in a throwaway clone (not just re-trusting the reviewer's repro): copied the migration tree, added a no-op tail migration mirroring `1500000154`'s shape, ran `test-verify-consolidation.sh`:
```
=== TEST 2: drift in a real migration should exit 1 ===
FAIL: drift expected exit 1, got 0
```
Confirms the gap.

**Fix:** added `pocketbase/pb_migrations/**` to the `guardSelfTests` filter (same glob the pre-existing `migrations` filter already uses), with a comment explaining why. Re-ran all four self-tests against the real (unmodified) tree in this worktree — 15/15, 5/5, 7/7, 12/12, all exit 0 — and re-validated `bash scripts/pre-push-verify.sh --all`.

**Not done:** did not harden `test-verify-consolidation.sh` TEST 2 itself to pick a migration that provably yields structural drift instead of assuming the tail does (the reviewer's "optional" suggestion) — that's a test-quality improvement to a pre-existing script this PR didn't write, orthogonal to the CI-wiring gap being closed here, and the script already documents the assumption as a caveat in its own comments.

**Discarded:** 10 other findings from this review pass discarded as noise/non-issues (reviewer's own count); none required a response here.

CodeRabbit not requested — single-file CI-config diff, doesn't meet the cross-surface bar this wave is reserving quota for.

Issue #2289, opened 2026-08-13
perf(sync): inject the active season into the lodging orphan sweep so its eight tests can run in parallel


adamflagg:
## ⚠️ CORRECTED 2026-08-14 — phase C triage

Verified against `main` @ `3eb44ba4` and the production snapshot. Verdict: **VALID** / SCOPED.

> Every structural claim verifies: the env read is at lodging_assignments_sync.go:873, it gates a deletion at :183-191, the eight named tests are the serialGroups entry at main_test_parallelism_test.go:77-86, and the exemption guard at :304 makes a half-landing impossible. The fail-closed contract is documented at :868-871 and the injected field must preserve it (unset/invalid → 0, which can never equal a validated 2017-2050 year). The timing figures are the filer's own measurements and I did not re-measure them.

### Corrections to the body below

1. Stale figure: "all 18 serial `sync` tests together: **24.6s**". Replacement: "all serial `sync` tests together at filing time (18 of them): **24.6s**. PR #2294 has since added five more (`TestProtectThenSweepOrphans_*`, main_test_parallelism_test.go:122-128), so the group is now 23 and the total is larger; the 22.0s figure for these eight is unaffected."
2. Stale figure: "`ParseSeasonYear()` has 24 call sites". Replacement: "`ParseSeasonYear()` has 32 call sites (26 outside tests). Only `lodging_assignments_sync.go:873` needs to change."

<details><summary>Re-measured figures (denominator and scope stated)</summary>

- Serial sync tests in serialGroups on main @3eb44ba4: 23, not 18 — counted by parsing pocketbase/main_test_parallelism_test.go (groups of 6 + 8 + 2 + 1 + 1 + 5). The 18 was correct when this was filed at 06:30Z; PR #2294 (merged 17:48Z) added five more (the TestProtectThenSweepOrphans_* group at :122-128). The eight-test measurement itself is unaffected.
- Serial lodging tests: 16 (two groups of 8), for a repo total of 39 serial exemptions.
- ParseSeasonYear call sites: 32 (26 non-test), of which exactly 1 — lodging_assignments_sync.go:873 — needs to change. The body says 24; the count has grown, the conclusion has not.

</details>

---

Split out of #2281 / #2288, which parallelised the `sync` and `lodging` suites but deliberately left this alone.

## Problem

[...]

PR #2374, opened 2026-08-15
fix(scripts): scrub real lodging names from test comments, widen guard to catch them

### LINKED ISSUE #2367: chore(frontend): real lodging unit names in public test fixtures the name guard cannot see
## What

Real lodging unit names sit in `frontend/src/` test fixtures and comments on a **public** repo. The Lodging Name Guard cannot catch them, and says so about itself.

## Why the guard misses them — both by design

`scripts/dev/verify-no-hardcoded-lodging.sh` states its own scope honestly:

> this is a tripwire, not a proof. It greps for a REPRESENTATIVE SAMPLE of distinctive unit strings (NEEDLES below) — not the full ~90-unit registry, which would be both unmaintainable and prone to false positives on ordinary words ("Ridge A", "Kitty"). A leak of a unit name that is not in NEEDLES will pass.

Two consequences:

1. **Comments are dropped before matching.** A needle term in a `//` or `/** */` block passes.
2. **Ordinary-sounding names are deliberately excluded from `NEEDLES`** to avoid false positives — so any area or unit whose name reads like normal English is invisible to it.

The guard runs green today. That is correct behaviour, not a bug in the guard.

## Inventory (2026-08-14, `frontend/src/`)

**Needle terms that leaked through the comment/test blind spot:**

| file | what |
|---|---|
| `types/beds.test.ts:43` | four unit names in a comment |
| `components/weekend/HouseholdJourneyCard.test.tsx:148,160` | a full unit + room string, as a fixture value AND an assertion |

**Non-needle names, which need a policy call rather than a fix:**

| file | what |
|---|---|
| `pages/WeekendRosterPage.test.tsx` | an area name, ×4 fixture values |
| `pages/WeekendRosterPage.codeSplitting.test.tsx:138` | same area name |
| `types/lodging.test.ts:183` | same area name |
| `components/weekend/boardLayout.ts:442` | same area name, in a **non-test source comment** |

One instance (`boardLayout.test.ts`, a needle term in a comment) was fixed in passing by the PR that found it — see kindred#2339.

## The decision this needs

**Do ordinary-word area names count?** They are excluded from `NEEDLES` on purpose because guarding them produces false positives on normal English. If they count, the fix is a fixture-vocabulary change across four files plus a source comment; if they do not, only the two needle-term rows above need touching and this is small.

Worth noting what the rule is actually for: the guard's header frames it as *"the lodging registry is DATA, not code … it now lives in exactly one place — the private `config/lodging_registry.json`."* That is single-source-of-truth hygiene with a privacy flavour, not a PII rule in the sense `CLAUDE.md` §4 uses for camper/family/staff names. So this is a judgement about how much a facility name in a public test fixture matters, not an incident.

## Suggested shape

1. Replace the two needle-term instances with structural or fictional names. Low risk, no decision needed.
2. Get the ruling on area names, then sweep or close that half.
3. Consider whether the guard should scan comments in test files even though it cannot scan for every name — catching needle terms everywhere is strictly better than catching them only in non-comment production code.

## Sequencing

- **Files:** the six above, plus `scripts/dev/verify-no-hardcoded-lodging.sh` if step 3 is taken.
- **Predecessors:** none. Independent of every open sync item.
- **Collides with:** kindred#2252 and kindred#2237 both touch `frontend/src/components/weekend/`; sequence against whichever is in flight.

adamflagg:
## What was wrong

`scripts/dev/verify-no-hardcoded-lodging.sh` deliberately greps for a representative sample of
real lodging-unit strings (NEEDLES) and is honest that it drops comment-only hits and exempts test
files entirely — that is correct, spec'd behaviour, not a bug. But two real gaps sat inside that
honest scope:

1. **Two public test fixtures carried real unit names the guard's own scope statement admits it
   cannot see.**
   - `frontend/src/types/beds.test.ts` — a comment listing six real family/guest-housing unit
     names (`Tenaya, Tioga, Half Dome, El Cap, Doctor's House, Manzanita 2`), pure documentation
     that drove no assertion.
   - `frontend/src/components/weekend/HouseholdJourneyCard.test.tsx` — a real unit+wing+room
     string (`Clouds Rest Lodge - West Wing Room 12B`) present as **both** a fixture value (line
     148) and the assertion that reads it back (line 160).
2. **The guard's test-file exemption drops comment hits and fixture hits alike, with no
   distinction.** Catching needle terms in comments is strictly better than catching them only in
   production code — that half of the guard was never built.

## What changed

- `beds.test.ts`: replaced the six real unit names with a generic description
  ("a small set of named units, not camper cabins") — the comment never drove the test, which only
  exercises `full_twin_bunk` arithmetic, so nothing about test coverage changed.
- `HouseholdJourneyCard.test.tsx`: replaced `Clouds Rest Lodge - West Wing Room 12B` with the
  fictional `Fernwood Lodge - West Wing Room 12B` in **both** the fixture (line 148) and the
  assertion (line 160) — same building+wing+room shape, so it still exercises the truncate-vs-wrap
  regression (kindred#2253) it was written for.
- `scripts/dev/lib/drop_comment_hits.py`: added a `--only-comments` mode that inverts the filter —
  keep only comment/docstring hits, drop code hits — for the guard's new comment-in-test-file scan.
- `scripts/dev/verify-no-hardcoded-lodging.sh`: `frontend/src/**` test-file hits are now split —
  fixture code stays exempt (unchanged, deliberate), but comment-line hits are now scanned and fail
  the guard.
- `scripts/dev/test-verify-no-hardcoded-lodging.sh`: added TEST 10-12 pinning the new behaviour —
  a comment leak in a `frontend/src/**` test file fails (TEST 10), a fixture-code leak in the same
  tree still doesn't (TEST 11), and a comment leak in a non-frontend test file still doesn't
  (TEST 12, proving the narrowing below is deliberate, not an oversight).

## Blast radius measured before touching any fixture

Per the issue's instruction, the widened guard was run across the full default scan roots
(`pocketbase/ api/ bunking/ frontend/src/ scripts/`) **before** any fixture was edited, to see what
it would additionally catch if applied repo-wide to every test file's comments, not just
`frontend/src/**`:

- **93** NEEDLES hits total inside files matching the test-file pattern.
- Of those, **14** sit on a comment line (the rest are legitimate fixture literals — the exact
  case the exemption exists for, per the guard's own header comment about kindred#1909).
- The 14 comment hits split **3 hits / 2 files under `frontend/src/**`** (both of which are the
  two fixtures this PR already fixes — `beds.test.ts` and `MapUnitPopover.test.tsx`, the latter a
  one-line M-sized extra hit found by the measurement, fixed in this branch) and **11 hits / 4
  files under `pocketbase/`** (`lodging_alias_resolver_test.go`, `lodging_issues_test.go`,
  `lodging_assignments_sync_test.go`, `registry_alias_target_test.go`).

The `pocketbase/` side is left alone: those files sit in territory this branch has no standing
context on and did not check against sibling in-flight PRs (`#2289` touches
`lodging_assignments_sync.go` in the same package this round), so the comment-scanning widening is
narrowed to `frontend/src/**` test files only — exactly the scope the issue suggested as a fallback
if the blast radius warranted narrowing. The Go-side leaks are real and worth a future look, but
guessing at files another issue may be mid-edit on is worse than leaving them for a scoped pass.

## Deliberately NOT done

- **The four area-name sites are untouched**, per the owner's 2026-08-14 ruling recorded in the
  issue: `pages/WeekendRosterPage.test.tsx`, `pages/WeekendRosterPage.codeSplitting.test.tsx`,
  `types/lodging.test.ts`, `components/weekend/boardLayout.ts:442`. Ordinary-word area names do not
  count as needle-term leaks.
- **The 11 comment hits in `pocketbase/` Go test files** are not fixed and the guard is not widened
  to catch them — narrowed deliberately, see above.
- This is not framed as a PII incident (per the guard's own header): the lodging registry is data
  that now lives in exactly one place, `config/lodging_registry.json` — single-source-of-truth
  hygiene with a privacy flavour, not the camper/family/staff-name rule in CLAUDE.md §4.

## Verification

- `bash scripts/dev/verify-no-hardcoded-lodging.sh` — exit 0 (`verify-no-hardcoded-lodging: OK`)
  on the full tree with the fixtures fixed.
- `bash scripts/dev/test-verify-no-hardcoded-lodging.sh` — all 12 cases pass (9 pre-existing + 3
  new), exit 0.
- `cd frontend && vitest run src/types/beds.test.ts src/components/weekend/HouseholdJourneyCard.test.tsx src/components/weekend/MapUnitPopover.test.tsx`
  — 3 files, **90 tests passed**.
- `uv run pytest tests/unit/scripts/test_drop_comment_hits.py` — 2 passed (unchanged `noncode_lines`
  logic, confirms the new `--only-comments` mode didn't disturb it).

**Mutation checks, all read by exit code:**
- Re-planted a needle term (`Tioga`) inside `beds.test.ts`'s comment: guard exited **1** and named
  the file:line. Reverted: guard exited **0**.
- Changed only the assertion side of the `HouseholdJourneyCard.test.tsx` fixture (line 160) to a
  mismatched string, leaving the fixture value (line 148) alone: the real test failed with an
  explicit content mismatch, proving the assertion drives the real fixture rather than passing
  independently of it. Reverted: test passed again.
- TEST 10 was written and run against the **unmodified** guard first and observed to fail
  (`FAIL: expected exit 1 ... got 0`) before `drop_comment_hits.py`/`verify-no-hardcoded-lodging.sh`
  were changed — confirming the new test actually pins the new behaviour rather than passing
  trivially.

Closes #2367.

### DIFF
--- a/frontend/src/components/weekend/HouseholdJourneyCard.test.tsx
+++ b/frontend/src/components/weekend/HouseholdJourneyCard.test.tsx
@@ -145,7 +145,7 @@ describe('the housing states', () => {
       _row({
         year: 2025,
         housing: 'placed',
-        cabin_name: 'Clouds Rest Lodge - West Wing Room 12B',
+        cabin_name: 'Fernwood Lodge - West Wing Room 12B',
       }),
     ])
 
@@ -157,7 +157,7 @@ describe('the housing states', () => {
     // wrap instead of overflowing its row.
     expect(housing).toHaveClass('min-w-0')
     expect(housing).not.toHaveClass('truncate')
-    expect(housing).toHaveTextContent('Clouds Rest Lodge - West Wing Room 12B')
+    expect(housing).toHaveTextContent('Fernwood Lodge - West Wing Room 12B')
   })
 })
 
--- a/frontend/src/components/weekend/MapUnitPopover.test.tsx
+++ b/frontend/src/components/weekend/MapUnitPopover.test.tsx
@@ -420,7 +420,7 @@ describe('MapUnitPopover — a cluster of rooms', () => {
   })
 
   it('drops the building name the cluster shares, so cells differ visibly', () => {
-    // A browser found this: every cell read "Clouds Rest Ba…" / "Clouds Rest
+    // A browser found this: every cell read "Cedar Lodge Ba…" / "Cedar Lodge
     // La…" and truncated away the distinguishing word.
     const house = [
       mapUnit(row({ unit_id: 'u1', code: 'cedar-back', name: 'Cedar Lodge Back' })),
--- a/frontend/src/types/beds.test.ts
+++ b/frontend/src/types/beds.test.ts
@@ -39,10 +39,10 @@ describe('suggestedSleeps', () => {
   })
 
   // Twin on top, full underneath. Staff-confirmed as a real and distinct piece
-  // of furniture, found only in the family/guest housing (Tenaya, Tioga, Half
-  // Dome, El Cap, Doctor's House, Manzanita 2) and never in a camper cabin.
-  // Splitting it into a full plus a twin would total the same three people but
-  // describe two separate beds that could sit in different rooms.
+  // of furniture, found only in family/guest housing (a small set of named
+  // units, not camper cabins) and never in a camper cabin. Splitting it into
+  // a full plus a twin would total the same three people but describe two
+  // separate beds that could sit in different rooms.
   it('counts a full-over-twin bunk as three people', () => {
     expect(suggestedSleeps([{ type: 'full_twin_bunk', count: 1 }])).toBe(3)
   })
--- a/scripts/dev/lib/drop_comment_hits.py
+++ b/scripts/dev/lib/drop_comment_hits.py
@@ -1,6 +1,14 @@
 #!/usr/bin/env python3
 """Drop grep hits that sit in a comment or a docstring.
 
+With `--only-comments`, the filter inverts: it keeps ONLY the hits that sit
+in a comment or docstring and drops everything else. verify-no-hardcoded-
+lodging.sh uses this second mode for frontend/src/** test files, where a
+fixture literal (code) is a deliberate, exempted case but a needle named in
+an explanatory comment is not (kindred#2367) -- so a test file's hits split
+into "genuine leak" (comment) and "known-legitimate fixture" (code) instead
+of the normal code/comment split going the other way.
+
 Spec 3.8 forbids the lodging registry from living in source. A registry lives
 in *code* -- string literals, lists, maps. Prose that names a unit to explain a
 rule is documentation, and failing on it made the guard red on `main` for a
@@ -176,6 +184,7 @@ def noncode_lines(path: Path) -> set[int]:
 
 
 def main() -> int:
+    only_comments = "--only-comments" in sys.argv[1:]
     cache: dict[str, set[int]] = {}
     for raw in sys.stdin:
         hit = raw.rstrip("\n")
@@ -183,13 +192,16 @@ def main() -> int:
             continue
         parts = hit.split(":", 2)
         if len(parts) < 3 or not parts[1].isdigit():
-            # Not a path:lineno:text hit -- pass it through rather than eat it.
+            # Not a path:lineno:text hit -- pass it through rather than eat
+            # it, in either mode: unparseable input means we learned nothing,
+            # so keep it reportable rather than silently dropping it.
             print(hit)
             continue
         path, lineno = parts[0], int(parts[1])
         if path not in cache:
             cache[path] = noncode_lines(Path(path))
-        if lineno not in cache[path]:
+        is_comment = lineno in cache[path]
+        if is_comment == only_comments:
             print(hit)
     return 0
 
--- a/scripts/dev/test-verify-no-hardcoded-lodging.sh
+++ b/scripts/dev/test-verify-no-hardcoded-lodging.sh
@@ -375,4 +375,39 @@ else
 fi
 
 echo
+echo "=== TEST 13: a needle in a COMMENT under frontend/src/test/ (shared test helpers, not *.test.ts) must exit 1 too ==="
+# FRONTEND_TEST_PATTERN is '^frontend/src/.*(_test\.|\.test\.|/tests?/)'. A
+# file directly under frontend/src/test/ -- the repo's real shared test-helper
+# directory (mockData.ts, mocks/, testUtils.tsx, test-helpers.ts) -- has no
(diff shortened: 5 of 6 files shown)


adamflagg:
## Triage follow-up on the reviewer's findings

The reviewer's one finding (medium/XS, `scripts/dev/verify-no-hardcoded-lodging.sh:107`) verified true and is fixed in this branch, commit bf952f07.

**Finding:** `FRONTEND_TEST_PATTERN`'s trailing `/tests?/` alternative only matches once a character already precedes the slash. A file directly under `frontend/src/test/` — the repo's real shared test-helper directory (`mockData.ts`, `mocks/`, `testUtils.tsx`, `test-helpers.ts`) — has no `_test.`/`.test.` in its filename, so it never matched the pattern. It fell through to `OTHER_RAW`'s blanket test-file exemption instead of the new comment-only scan, so a needle in a comment there would still ship green — exactly the gap this PR exists to close for `frontend/src/**`.

**Reproduced before fixing:** planted `// Regression comment naming Kitty to explain a fixture.` in both a probe under `frontend/src/test/` and a probe named `*.test.ts`. The `.test.ts` probe was caught (exit 1); the `frontend/src/test/` probe was not — guard printed `OK` and exited 0.

**Fix:** anchored the pattern to a full path segment — `^frontend/src/(.*/)?tests?/|^frontend/src/.*(_test\.|\.test\.)` — instead of the bare `/tests?/`.

**TDD/mutation-check transcript:**
- Added TEST 13 to `test-verify-no-hardcoded-lodging.sh`, ran it against the pre-fix pattern first: `FAIL: expected exit 1 for a needle in a frontend/src/test/ comment, got 0` — red as expected, confirming it pins new behavior.
- Applied the fix. Full suite (13 cases) green.
- Mutation check: reverted `FRONTEND_TEST_PATTERN` to the pre-fix bare-`/tests?/` form in isolation — suite failed again at TEST 13 with the identical message. Restored the fix — green again.
- `bash scripts/dev/verify-no-hardcoded-lodging.sh` on the real tree: OK (no new leaks surfaced by the widened anchor).
- `shellcheck` clean on both modified `.sh` files.
- `bash scripts/pre-push-verify.sh`: ALL CHECKS PASSED.

**Nothing filed** — the finding was M-or-smaller and fixed in-branch per standing policy.

**No disagreements** — the finding reproduced exactly as described on an isolated tree.

CI Summary: pass (all required checks green, polled to terminal state). No GUI-impacting files in this diff (`gh pr diff --name-only` shows only `.test.ts`/`.test.tsx` and shell-script changes). CodeRabbit not requested — single-lane, test/tooling-only diff, per the per-account quota policy.

Push verified: `git fetch && git rev-list --count origin/feature/lodging-names-test-fixtures-2367..HEAD` → `0`.

PR #2376, opened 2026-08-15
fix(sync): inject the active season into the lodging orphan sweep

### LINKED ISSUE #2289: perf(sync): inject the active season into the lodging orphan sweep so its eight tests can run in parallel

adamflagg:
## The defect

`LodgingAssignmentsSync.activeSeasonYear()` (`pocketbase/sync/lodging_assignments_sync.go`) called `ParseSeasonYear()` unconditionally, so the only way to drive the `#2028` orphan sweep was through `CAMPMINDER_SEASON_ID`. That forced its eight tests to use `t.Setenv`, and Go panics on `t.Setenv` combined with `t.Parallel()`. Those eight tests were the "orphan sweep gates on CAMPMINDER_SEASON_ID" entry in `main_test_parallelism_test.go`'s `serialGroups`, which the issue reported as ~22.0s of the serial prefix under `-race`. I could not get a clean isolated re-measurement of that specific 8-test slice on unmodified `main` myself (see "Deliberately not done" below), so I'm not re-quoting that figure as something I personally verified — only citing it as the issue's stated rationale for the change.

## The fix

Added `ActiveSeasonYear` as an injectable field on `LodgingAssignmentsSync`, defaulting to `ParseSeasonYear()` when left at its zero value, so tests set the active season directly instead of reaching for the environment:

```go
func (s *LodgingAssignmentsSync) activeSeasonYear() int {
	if s.ActiveSeasonYear != 0 {
		return s.ActiveSeasonYear
	}
	active, err := ParseSeasonYear()
	if err != nil {
		return 0
	}
	return active
}
```

Fail-closed is unchanged and now pinned directly: a new `TestActiveSeasonYearFailsClosedWhenUnset` asserts that with neither the field nor the env set, `activeSeasonYear()` returns 0 — which can never equal a validated 2017-2050 year, so the deletion gate still blocks the sweep rather than accidentally allowing it. A companion `TestActiveSeasonYearFieldOverridesEnv` pins the injection point itself.

Migrated the eight orphan-sweep tests off `t.Setenv` onto `s.ActiveSeasonYear`, added `t.Parallel()` to each, and removed their `serialGroups` exemption. The historical-year test (`TestLodgingAssignmentsSyncSkipsOrphanSweepForHistoricalYear`) keeps its exact shape — `ActiveSeasonYear = 2025` while `s.Year = 2024` — since that mismatch is the whole point of the test.

## Deliberately not done

- **Deletion semantics are unchanged.** The eight tests still assert exactly what they asserted before (orphaned mirror rows, dry-run, staff-touched rows, historical-year gating) — only *how* the active season arrives changed, not *what* the sweep does. No assertion was weakened.
- **No other `ParseSeasonYear()` call site touched.** It has ~32 call sites (26 non-test); only `lodging_assignments_sync.go`'s needed the injectable field for this issue.
- **No before/after wall-clock comparison quoted.** I measured `go test -race ./sync/` on this branch at 136.4s wall (full package), but the equivalent measurement on unmodified `main` ran inside the shared main checkout while another agent's own test run was actively contending for CPU there, so I did not get a clean baseline and am not quoting one.

## Verification (all measured on this branch)

- `go test -race ./sync/ -run 'Orphan|Sweep|Season|ActiveSeasonYear' -v`: PASS
- `go test -race ./sync/` (full package): PASS, 136.4s wall
- `go test ./...` (full suite): PASS, 108.2s wall
- `go vet ./...`: clean
- `bash scripts/pre-push-verify.sh`: **ALL CHECKS PASSED**
- `TestSyncAndLodgingTestsRunInParallel` / `TestSerialTestExemptionsAreAllStillNeeded`: both PASS with the exemption removed
- **Mutation check**: changed `activeSeasonYear()`'s parse-failure branch to `return 2025` instead of `return 0`. `TestActiveSeasonYearFailsClosedWhenUnset` went RED:
  ```
  lodging_assignments_sync_test.go:933: activeSeasonYear() = 2025, want 0 (fail closed)
  --- FAIL: TestActiveSeasonYearFailsClosedWhenUnset (0.00s)
  ```
  Reverted; test passes again.

Closes #2289.

### DIFF
--- a/pocketbase/main_test_parallelism_test.go
+++ b/pocketbase/main_test_parallelism_test.go
@@ -61,31 +61,6 @@ var serialGroups = []struct {
 		},
 	},
 	{
-		// LodgingAssignmentsSync.activeSeasonYear() calls ParseSeasonYear()
-		// unconditionally to gate the #2028 orphan sweep, so these eight can
-		// only reach the sweep through the environment.
-		//
-		// They are also the most expensive tests left serial -- ~22s of the
-		// ~25s serial prefix, because each builds a fresh ~15-collection
-		// schema and that is pure CPU under the race detector. Injecting the
-		// active season instead of reading the env would recover most of it.
-		// Deliberately not done here: that is an edit to a deletion gate, and
-		// it should not ride inside a thousand-line mechanical diff where no
-		// reviewer will find it.
-		pkg:    "sync",
-		reason: "t.Setenv: orphan sweep gates on CAMPMINDER_SEASON_ID",
-		tests: []string{
-			"TestLodgingAssignmentsSyncDeletesOrphanedHouseholdMirrorRow",
-			"TestLodgingAssignmentsSyncDeletesOrphanedPersonGrainMirrorRow",
-			"TestLodgingAssignmentsSyncDeletesStaffTouchedOrphanedMirrorRow",
-			"TestLodgingAssignmentsSyncDryRunDoesNotDeleteOrphans",
-			"TestLodgingAssignmentsSyncKeepsMirrorRowForStillEnrolledHousehold",
-			"TestLodgingAssignmentsSyncSkipsMirrorDeletionWhenSessionHasZeroEnrolled",
-			"TestLodgingAssignmentsSyncSkipsOrphanSweepForHistoricalYear",
-			"TestLodgingAssignmentsSyncWritesHistoryOnOrphanDelete",
-		},
-	},
-	{
 		pkg:    "sync",
 		reason: "t.Setenv: asserts the CAMPMINDER_SEASON_ID fallback specifically",
 		tests: []string{
--- a/pocketbase/main_test_parallelism_test.go
+++ b/pocketbase/main_test_parallelism_test.go
@@ -58,6 +58,22 @@ var serialGroups = []struct {
 			"TestParseSeasonYear_BelowRange",
 			"TestParseSeasonYear_AboveRange",
 			"TestParseSeasonYear_Boundaries",
+			"TestActiveSeasonYearFailsClosedWhenUnset",
+		},
+	},
+	{
+		// #2289 review: the eight orphan-sweep tests migrated onto
+		// s.ActiveSeasonYear so they could run in parallel, but that left the
+		// #2028 year gate itself (year != active) and production's actual
+		// CAMPMINDER_SEASON_ID fallback path with zero direct coverage --
+		// s.ActiveSeasonYear always overrides both in every other test in this
+		// file. These two drive Sync() with ActiveSeasonYear left at 0, so
+		// they need the real environment variable.
+		pkg:    "sync",
+		reason: "t.Setenv: CAMPMINDER_SEASON_ID drives the #2028 gate itself, not the ActiveSeasonYear override",
+		tests: []string{
+			"TestLodgingAssignmentsSyncSkipsOrphanSweepWhenSeasonUnresolvable",
+			"TestLodgingAssignmentsSyncDeletesOrphansViaEnvFallback",
 		},
 	},
 	{
--- a/pocketbase/sync/lodging_assignments_sync.go
+++ b/pocketbase/sync/lodging_assignments_sync.go
@@ -31,12 +31,19 @@ const sourceFieldOrphanSweep = "orphan_sweep"
 // household_custom_values / person_custom_values / attendees / persons /
 // camp_sessions and the lodging registry, all from PocketBase.
 type LodgingAssignmentsSync struct {
-	App            core.App
-	Year           int  // 0 = current year from env
-	DryRun         bool // compute but do not write
-	Debug          bool
-	Stats          Stats
-	SyncSuccessful bool
+	App    core.App
+	Year   int  // 0 = current year from env
+	DryRun bool // compute but do not write
+	Debug  bool
+	// ActiveSeasonYear injects the value activeSeasonYear() returns, bypassing
+	// CAMPMINDER_SEASON_ID entirely. 0 (the zero value) means "resolve from
+	// the environment via ParseSeasonYear() at Sync time", matching Year's
+	// existing convention. Set directly by tests exercising the #2028 orphan
+	// sweep so they need not reach for t.Setenv, which cannot be combined with
+	// t.Parallel() (#2289).
+	ActiveSeasonYear int
+	Stats            Stats
+	SyncSuccessful   bool
 
 	resolver *AliasResolver
 	issues   *IssueRecorder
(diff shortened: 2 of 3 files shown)


adamflagg:
## Review round applied in-branch

All three findings from review verified true and fixed directly (all M-or-smaller — no issue filed):

**Finding 1 (medium/S)** — `TestActiveSeasonYearFailsClosedWhenUnset` pinned `activeSeasonYear()`'s return value only, not the `#2028` deletion gate (`if active := s.activeSeasonYear(); year != active`). Reproduced the reviewer's mutation myself: changing the gate to `active != 0 && year != active` (treat "unset" as "no gate") kept the whole `./sync/` package green.

**Finding 2 (low/XS)** — migrating all eight sweep tests onto `s.ActiveSeasonYear` left the `CAMPMINDER_SEASON_ID` → `ParseSeasonYear()` fallback — the path *every real production sync actually takes* — with zero coverage. Reproduced: replacing `activeSeasonYear()` with `return s.ActiveSeasonYear` (fallback deleted) also left `./sync/` green.

**Finding 3 (low/XS)** — `TestActiveSeasonYearFailsClosedWhenUnset` read the ambient environment without clearing it, so it goes spuriously red on any shell with `CAMPMINDER_SEASON_ID` exported (repo's own `.env` sets it, several `scripts/dev/verify-*.sh` export it). Reproduced: `CAMPMINDER_SEASON_ID=2026 go test ./sync/ -run ActiveSeasonYear` failed on unmodified code.

### Fix
- `TestActiveSeasonYearFailsClosedWhenUnset`: dropped `t.Parallel()`, added `t.Setenv("CAMPMINDER_SEASON_ID", "")`, moved into the existing `serialGroups` entry for `ParseSeasonYear` parsing (fixes #3).
- Added `TestLodgingAssignmentsSyncSkipsOrphanSweepWhenSeasonUnresolvable` — drives `Sync()` with `ActiveSeasonYear` left at 0 and the env cleared; asserts the cancelled household's row **survives** and `Stats.Deleted == 0` (fixes #1 — pins the gate itself).
- Added `TestLodgingAssignmentsSyncDeletesOrphansViaEnvFallback` — same fixture, `CAMPMINDER_SEASON_ID=2025` and `ActiveSeasonYear` left at 0; asserts the row **is deleted** (fixes #2 — pins the production fallback path).
- Both new Sync()-level tests are serial (`t.Setenv` cannot combine with `t.Parallel()`) in a new `serialGroups` entry — 2 tests, ~0.9s added to the serial prefix, versus the 8 tests × parallel this PR already freed.

### Mutation-check transcript (re-run against the fix)
```
$ sed -i 's/year != active {/active != 0 \&\& year != active {/' ...
$ go test ./sync/ -run TestLodgingAssignmentsSyncSkipsOrphanSweepWhenSeasonUnresolvable -v
    ...real placement deleted: sql: no rows in result set
    Stats.Deleted = 1, want 0 -- an unresolvable season must fail closed, not fail open
--- FAIL   ✓ now caught

$ # activeSeasonYear() -> return s.ActiveSeasonYear (fallback removed)
$ go test ./sync/ -run TestLodgingAssignmentsSyncDeletesOrphansViaEnvFallback -v
    cancelled household's stale mirror row still exists -- env fallback did not reach the sweep
    Stats.Deleted = 0, want 1
--- FAIL   ✓ now caught

$ CAMPMINDER_SEASON_ID=2026 go test ./sync/ -run TestActiveSeasonYearFailsClosedWhenUnset -v
--- PASS  ✓ no longer flaky on ambient env
```

Both parallelism guards (`TestSyncAndLodgingTestsRunInParallel`, `TestSerialTestExemptionsAreAllStillNeeded`) still pass. Full `go test ./...` (all packages), `go vet ./...`, and `golangci-lint run ./sync/... ./` all clean on the fix commit.

### Not filed / not changed
- The reviewer's "~32 vs 36 call sites" note on `ParseSeasonYear()` was explicitly flagged immaterial (hedged with `~`, load-bearing on nothing) — left as-is.
- No production behavior changed — `activeSeasonYear()` and the gate at line 197 are untouched; this round is coverage-only.

Push verified via `git fetch -q && git rev-list --count origin/...HEAD` = 0. CI re-running on the new commit.

PR #2377, opened 2026-08-15
fix(frontend): gate four overlays' Escape on the modalStack token, classify the rest

### LINKED ISSUE #2237: bug(frontend): sixteen overlays still handle Escape without the modalStack token gate (#2205 follow-through)

> The issue's own reproduce command still returns exactly sixteen files, and the list matches the body file-for-file against `main` @ 3eb44ba4 — so the grep is reproducible. But the grep over-reports: one of the sixteen, `weekend/FamilyDetailsPanel.tsx`, already stands down for any open overlay via `hasOpenModal()` and is not an instance of the defect at all. The genuine count is FIFTEEN. The body's two 'worth doing first' picks are also mis-ranked: `PlaceFamilyPicker` guards by focus containment plus a React synthetic handler, while the two sibling pickers install capture-phase document listeners, which are the shape that actually steals Escape from an overlay above them.

1. WRONG: "Sixteen components carry their own `Escape` handling and **none** consults `isTopOverlay`" → REPLACEMENT: "Sixteen components are returned by the grep below and none consults `isTopOverlay` — but **the real count is FIFTEEN**, because `weekend/FamilyDetailsPanel.tsx` is already guarded by the OTHER half of the same module: it calls `hasOpenModal()` at `:137` and stands down for any overlay opened on top of it (comment at `:131-136`, written for kindred#2073). Since #2234, `hasOpenModal()` reads the same widened token stack `isTopOverlay` reads (`modalStack.ts:100-102`), so that guard is correct for every overlay kind, not just `ui/Modal`. A container that stands down for everything above it does not need a token of its own. **Classify it as 'correct as-is' and do not convert it.**"
2. WRONG: "Two look clearly worth doing first, because both are weekend surfaces that demonstrably co-occur with dialogs:\n\n- `weekend/FamilyDetailsPanel.tsx`\n- `weekend/PlaceFamilyPicker.tsx` (and the two sibling pickers)" → REPLACEMENT: "The priority order should follow the HANDLER SHAPE, not the surface. Ranked by how badly each can steal a keypress:\n\n1. **`weekend/AddHouseholdPicker.tsx:118-127` and `weekend/AddToGroupPicker.tsx:57-66`** — both install `document.addEventListener('keydown', handler, true)` in the **CAPTURE** phase with `stopPropagation()`, explicitly 'to stop it before an outer modal listener reacts'. A capture-phase listener fires before every bubble-phase handler in the tree, including a token-gated overlay stacked above it, so these are the two that most reliably reproduce the defect. Do these first.\n2. **`weekend/PlaceFamilyPicker.tsx:159-168`** — a React **synthetic** `onKeyDown` on the combobox `<input>`, not a document listener. It only fires while that input holds focus, and the control closes on blur, so it is largely self-limiting. Lower priority than the body originally implied.\n3. `weekend/FamilyDetailsPanel.tsx` — **already guarded, remove from the list** (see above)."
3. WRONG: "## Measured against `main` @ `ba646482`" → REPLACEMENT: "## Re-measured against `main` @ `3eb44ba4` (2026-08-13). The grep still returns the same sixteen files, unchanged file-for-file — but see the correction above: one of them is already guarded, so **fifteen** need classifying, not sixteen."
4. WRONG (in Acceptance): "- [ ] Each of the sixteen classified: needs a token / correct as-is (with the reason recorded)." → REPLACEMENT: "- [ ] Each of the **fifteen** classified: needs a token / correct as-is (with the reason recorded). `FamilyDetailsPanel` is pre-classified 'correct as-is — guards via `hasOpenModal()` at `:137`'."

- Overlays with genuinely unconditional Escape: **15**, not 16. Denominator = the 16 files the issue's own reproduce command returns. Command: `for f in $(grep -rln "Escape" src/components/ --include=*.tsx --include=*.ts | grep -v '\.test\.' | xargs grep -Ln "isTopOverlay" | xargs grep -l "'Escape'" | sort); do echo "$(grep -c hasOpenModal $f)  $f"; done` → `2  src/components/weekend/FamilyDetailsPanel.tsx`, `0` for the other fifteen.
- Escape-handler SHAPE across the 15 unguarded (denominator 15, hand-classified by reading each handler): capture-phase `document` listener with `stopPropagation` — the shape that pre-empts an overlay above it — includes `AddHouseholdPicker.tsx:118-127` and `AddToGroupPicker.tsx:57-66`. React synthetic handler on a focused input — largely self-limiting — includes `PlaceFamilyPicker.tsx:159-168`.
- Total files under `src/components/` mentioning 'Escape' at all (non-test): 24. Of those, 3 are the #2205 converts (`ui/Modal.tsx`, `ConfirmActionPopover.tsx`, `ui/Tooltip.tsx`) plus `modalStack.ts` and the guarded panel. Command: grep -rln "Escape" src/components/ --include=*.tsx --include=*.ts | grep -v '\.test\.' | wc -l → 24

</details>

---

#2205 shipped the mechanism: `modalStack` is now a LIFO **token stack**, and an overlay that acquires a token only handles Escape when `isTopOverlay()` says it is on top. `ui/Modal`, `ConfirmActionPopover` and `ui/Tooltip` were converted.

**Nothing else was.** Every other overlay in the tree still installs its own unconditional Escape listener, so any two that can be open at once reproduce the original defect — one keypress, two overlays closed.

## Measured against `main` @ `ba646482`

Sixteen components carry their own `Escape` handling and **none** consults `isTopOverlay`:

| | |
|---|---|
| `components/CamperCard.tsx` | `components/LockGroupPanel.tsx` |
| `components/CamperDetailsPanel.tsx` | `components/metrics/DrillDownModal.tsx` |
| `components/CsvPipelineIndicator.tsx` | `components/metrics/WaitlistGradePopover.tsx` |
| `components/graph/filter/GraphFilterPopover.tsx` | `components/OptimizeBunksButton.tsx` |
| `components/RequestReviewPanel.tsx` | `components/SocialNetworkGraph.tsx` |
| `components/ui/FloatingQueueBadge.tsx` | `components/weekend/AddHouseholdPicker.tsx` |
| `components/weekend/AddToGroupPicker.tsx` | `components/weekend/FamilyDetailsPanel.tsx` |
| `components/weekend/LodgingMap.tsx` | `components/weekend/PlaceFamilyPicker.tsx` |

Reproduce it:

```
grep -rln "Escape" frontend/src/components/ --include=*.tsx --include=*.ts \
  | grep -v '\.test\.' \
  | xargs grep -Ln "isTopOverlay" \
  | xargs grep -l "'Escape'"
```

## ⚠️ This is a per-component call, not a sweep

**Do not mechanically wrap all sixteen.** An overlay only needs the token if it can be open *at the same time* as another one. A page-level drawer that can never co-occur with a dialog is correct as it is, and giving it a token adds a lifecycle it has to get right — and a token acquired but not released on unmount **silently disables Escape for every overlay after it**, which is worse than the bug being fixed. #2205 pins exactly that leak for the three converted overlays; any new adopter needs the same pin.

So each of the sixteen needs one question answered: *can this be open while another overlay is open?*

Two look clearly worth doing first, because both are weekend surfaces that demonstrably co-occur with dialogs:

- `weekend/FamilyDetailsPanel.tsx`
- `weekend/PlaceFamilyPicker.tsx` (and the two sibling pickers)

## Why this was not folded into #2205

It is pre-existing, spans sixteen files, and the decision above is a design/priority call rather than a mechanical fix — the three cases #2205 actually reproduced are all converted. Surfaced by the review pass on PR #2234.

## Acceptance

- [ ] Each of the sixteen classified: needs a token / correct as-is (with the reason recorded).
- [ ] Every adopter gets a release-on-unmount test asserting `hasOpenModal()` is false after unmount — a test that only checks "a freshly-opened overlay still receives Escape" passes even with a real leak, because a new token is always last in the stack. #2205 hit exactly that false-green.

adamflagg:
#2205 shipped the LIFO token stack in `components/ui/modalStack.ts`; #2359 added the shared
`hooks/useOverlayEscape.ts` and converted three weekend overlays. This finishes the classification
pass over what was left.

The reproduce command returned **13** files on `main` @ 8fb02894 and returns **9** on this branch.
`weekend/FamilyDetailsPanel` was pre-classified correct as-is, leaving **12** to call. Four needed a
token; eight did not, and the reason for each is now recorded at its own Escape handler.

## The defect, per adopter

**`CamperDetailsPanel` was the bad one.** It installed a capture-phase `document` listener with an
**unconditional** `stopPropagation()`. A capture-phase stop at `document` halts the event before the
bubble phase begins, and both `ui/Modal` and `ConfirmActionPopover` listen in the bubble phase and
gate on `isTopOverlay` *without* stopping anything. So the panel beat every overlay stacked above it
— including `AllCamperRequestsModal`, a `ui/Modal` the panel itself opens. One Escape closed the
panel out from under the dialog and left the dialog open. That is the wrong overlay closing, not
merely an extra one.

**`graph/filter/GraphFilterPopover`** and **`CamperCard`**'s context menu were the plain #2205
double-close: each is rendered inside a container that also handles Escape ungated (the expanded
`SocialNetworkGraph` on `window`, the expanded `ui/FloatingQueueBadge` on `document`), so one press
ran both handlers — the filter closed and the graph collapsed with it; the context menu closed and
the queue it was opened from collapsed with it.

**`RequestReviewPanel`**'s bulk-confirm dialog sits in a component that renders five other
independently-stated overlays with nothing making them mutually exclusive. Its ungated bubble
listener fired alongside whichever was genuinely on top.

## Classification of all twelve

| Component | Call | Why |
|---|---|---|
| `CamperDetailsPanel` | **needs a token** | capture-phase + unconditional stop; hosts a `ui/Modal` and is itself hosted inside `BunkSocialGraphModal` |
| `graph/filter/GraphFilterPopover` | **needs a token** | rendered inside the expanded `SocialNetworkGraph`, which closes on Escape too |
| `CamperCard` | **needs a token** | context menu opens inside the expanded `FloatingQueueBadge`, which closes on Escape too |
| `RequestReviewPanel` | **needs a token** | bulk-confirm dialog coexists with five sibling overlays, none of which stops propagation |
| `SocialNetworkGraph` | correct as-is | a container at the bottom of its own stack; both overlays it hosts are converted here and stop the key in the capture phase, before its `window` bubble listener |
| `ui/FloatingQueueBadge` | correct as-is | same — the two things that stack above it (the `CamperCard` menu, `CamperDetailsPanel` via `isPanelOpen`) now both hold tokens |
| `weekend/LodgingMap` | correct as-is — **would regress** | a peek token makes `hasOpenModal()` true, and `FamilyDetailsPanel` yields to *anything* in the stack rather than to the topmost. A peek can sit behind that panel, so a token here would stop the panel closing on Escape. Fixing it properly means converting `FamilyDetailsPanel`, which #2237 pre-classifies as out of scope |
| `weekend/PlaceFamilyPicker` | correct as-is | React synthetic handler on the combobox input — fires only while that input has focus, and the control closes on blur. Same weekend `hasOpenModal` hazard as above |
| `metrics/WaitlistGradePopover` | correct as-is | **not rendered by any production surface** — the only non-test reference in `src/` is to `transformWaitlistGradeData`, a same-prefixed function. It cannot co-occur with anything |
| `metrics/DrillDownModal` | correct as-is | hosts no overlay; its one mount site puts no other Escape-handling surface on screen |
| `CsvPipelineIndicator` | correct as-is | app-shell trigger popover dismissed by `useClickOutside` on any outside press |
| `OptimizeBunksButton` | correct as-is | header dropdown dismissed by its outside-mousedown listener; hosts no overlay |

The load-bearing observation behind six of the eight: **the overlay that can be ON TOP is the one
that needs the token.** `useOverlayEscape` stops propagation in the document capture phase while
topmost, which lands before any `document`- or `window`-bubble listener beneath it. Converting the
child therefore fixes the pair, and giving the container a token would only add a lifecycle to get
wrong — a token acquired and not released silently disables Escape for every overlay after it.

## Mutation-check evidence

Deleting `releaseOverlayToken(token)` from the unmount path of `useOverlayEscape` turned **9 tests
red across all four adopter files**. Critically, the **150 other tests in those files stayed green**,
including every "does this overlay still close on Escape?" assertion — reproducing the exact
false-green #2205 shipped, since a newly acquired token is always last in the stack whether or not
the previous one leaked. Only the `hasOpenModal()`-is-empty assertions catch a leak, and every
adopter has one.

Also added per adopter: a test that the overlay does **not** close while a token is held above it,
and, where it applies, that the key is swallowed so the surface beneath does not close too.

## Deliberately not done

- Did not convert the eight classified correct as-is; `LodgingMap` and `PlaceFamilyPicker`
  specifically because converting them without `FamilyDetailsPanel` would make Escape *worse*.
- Did not touch `weekend/FamilyDetailsPanel` (pre-classified).
- Did not delete the unmounted `metrics/WaitlistGradePopover` — recorded, not swept, as it is a
  separate question from Escape handling.
- No `aria-*`, `role`, or `sr-only` text added anywhere.

## Verification

- Full frontend suite: **450 files, 6869 tests, all passing**.
- `bash scripts/pre-push-verify.sh`: prettier, eslint, tsc, vitest all PASS.

**GUI impact:** this changes keyboard behaviour on live overlays and wants a human look before merge.

Closes #2237.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Escape-key handling across menus, panels, dialogs, and graph filters.
  * Ensured only the topmost open overlay closes when Escape is pressed.
  * Prevented Escape events from unintentionally closing underlying overlays.
  * Improved cleanup when overlays close or are removed.

* **Tests**
  * Added regression coverage for overlay stacking, keyboard dismissal, event propagation, and cleanup behavior.

### DIFF
--- a/frontend/src/components/CamperCard.test.tsx
+++ b/frontend/src/components/CamperCard.test.tsx
@@ -1,0 +1,136 @@
+/**
+ * kindred#2237 — CamperCard's right-click context menu and the kindred#2205
+ * overlay token stack.
+ *
+ * The menu is rendered by cards INSIDE the expanded `ui/FloatingQueueBadge`
+ * queue (see `FloatingUnassignedBadge`, which renders CamperCard rows straight
+ * into the badge's popover). Both surfaces listened on `document` in the
+ * bubble phase and neither stopped propagation, so a single Escape closed the
+ * context menu AND collapsed the queue it was opened from — the kindred#2205
+ * double-close, still live on the board.
+ *
+ * The menu is the overlay that sits ON TOP, so it is the one that needs the
+ * token: once it swallows the key while topmost, the badge beneath it never
+ * sees the press and needs no token of its own.
+ *
+ * Fictional data throughout.
+ */
+import { describe, it, expect, vi, beforeEach } from 'vitest'
+import { render, screen, fireEvent, waitFor } from '@testing-library/react'
+
+import CamperCard from './CamperCard'
+import { acquireOverlayToken, hasOpenModal, releaseOverlayToken } from './ui/modalStack'
+import { mockCamper } from '../test/mockData'
+import { emptyCamperSatisfaction } from '../types/satisfaction'
+
+vi.mock('../hooks/useCurrentYear', () => ({ useYear: () => 2025 }))
+
+vi.mock('@dnd-kit/sortable', () => ({
+  useSortable: () => ({
+    attributes: {},
+    listeners: {},
+    setNodeRef: vi.fn(),
+    transform: null,
+    transition: undefined,
+    isDragging: false,
+  }),
+}))
+
+vi.mock('../hooks', () => ({
+  useBunkRequestContext: () => ({
+    getSatisfiedRequestInfo: (cmId: number) => emptyCamperSatisfaction(cmId),
+  }),
+  useCamperHistoryContext: () => ({ getLastYearHistory: () => null }),
+}))
+
+vi.mock('../contexts/LockGroupContext', () => ({
+  useLockGroupContext: () => ({
+    addPendingCamper: vi.fn(),
+    removePendingCamper: vi.fn(),
+    getPendingAnimationDelay: () => 0,
+    groups: [],
+    addCamperToGroup: vi.fn(),
+    getCamperLockGroup: () => null,
+    getGroupMembers: () => [],
+    setSelectedGroupId: vi.fn(),
+    setIsLockPanelOpen: vi.fn(),
+  }),
+}))
+
+describe('CamperCard — context-menu overlay token (kindred#2237)', () => {
+  beforeEach(() => {
+    vi.clearAllMocks()
+  })
+
+  /**
+   * `handleContextMenu` defers the open by one `requestAnimationFrame` (it
+   * broadcasts `closeAllContextMenus` first and lets that settle), so every
+   * test has to wait for the menu rather than assert synchronously.
+   */
+  async function openMenu() {
+    const view = render(<CamperCard camper={mockCamper({ person_cm_id: 9001 })} isDraftMode />)
+    const card = document.querySelector('[data-camper-card], button') as HTMLElement
+    fireEvent.contextMenu(card)
+    await waitFor(() => {
+      expect(screen.getByText('View Details')).toBeInTheDocument()
+    })
+    return view
+  }
+
+  it('closes the menu on Escape when it is the only open overlay', async () => {
+    await openMenu()
+
+    fireEvent.keyDown(document, { key: 'Escape' })
+
+    await waitFor(() => {
+      expect(screen.queryByText('View Details')).not.toBeInTheDocument()
+    })
+  })
+
+  it('does NOT close once an overlay has opened on top of it', async () => {
+    await openMenu()
+
+    const topToken = acquireOverlayToken()
+    try {
+      fireEvent.keyDown(document, { key: 'Escape' })
+      expect(screen.getByText('View Details')).toBeInTheDocument()
+    } finally {
+      releaseOverlayToken(topToken)
+    }
+  })
+
+  // The defect this issue is actually about: the expanded FloatingQueueBadge
+  // beneath the menu listens on `document` too. While the menu is topmost it
+  // must swallow the key so the queue does not collapse alongside it.
+  it('swallows the key while topmost, so the queue beneath does not also close', async () => {
+    const beneath = vi.fn()
+    document.addEventListener('keydown', beneath)
+    try {
+      await openMenu()
+
+      fireEvent.keyDown(document, { key: 'Escape' })
+
+      const escapes = beneath.mock.calls.filter(([e]) => (e as KeyboardEvent).key === 'Escape')
+      expect(escapes).toHaveLength(0)
+    } finally {
+      document.removeEventListener('keydown', beneath)
+    }
+  })
+
+  // A leaked token is invisible to a "does it still close?" test — a newly
+  // acquired token is always last in the stack, so the freshly-opened menu is
+  // topmost either way. Only the emptiness of the stack catches it.
+  it('releases its overlay token on unmount, so the stack does not leak', async () => {
+    const { unmount } = await openMenu()
+    expect(hasOpenModal()).toBe(true)
+
+    unmount()
+
+    expect(hasOpenModal()).toBe(false)
+  })
+
+  it('registers no token while the context menu is closed', () => {
+    render(<CamperCard camper={mockCamper({ person_cm_id: 9002 })} isDraftMode />)
+    expect(hasOpenModal()).toBe(false)
+  })
+})
--- a/frontend/src/components/CamperCard.tsx
+++ b/frontend/src/components/CamperCard.tsx
@@ -16,6 +16,7 @@ import { useYear } from '../hooks/useCurrentYear'
 import type { Camper } from '../types/app-types'
 import { useBunkRequestContext, useCamperHistoryContext } from '../hooks'
 import { useLockGroupContext } from '../contexts/LockGroupContext'
+import { useOverlayEscape } from '../hooks/useOverlayEscape'
 import { emptyCamperSatisfaction } from '../types/satisfaction'
 
 interface CamperCardProps {
@@ -95,16 +96,17 @@ function CamperCard({
 
   // Escape dismisses the context menu — the keyboard equivalent for the
   // full-viewport backdrop's click-to-close (see the backdrop's own
-  // eslint-disable below). Matches the pattern already used for the social
-  // graph's expand overlay and several other dismissible surfaces.
-  useEffect(() => {
-    if (!showContextMenu) return
-    const handleEscape = (e: KeyboardEvent) => {
-      if (e.key === 'Escape') setShowContextMenu(false)
-    }
-    document.addEventListener('keydown', handleEscape)
-    return () => document.removeEventListener('keydown', handleEscape)
-  }, [showContextMenu])
+  // eslint-disable below).
+  //
+  // NEEDS AN OVERLAY TOKEN (kindred#2237): these cards are rendered INSIDE the
+  // expanded `ui/FloatingQueueBadge` queue (`FloatingUnassignedBadge`), which
+  // closes itself on Escape from its own ungated `document` listener. Two
+  // bubble-phase handlers, neither stopping propagation, meant one press
+  // dismissed the menu AND collapsed the queue it was opened from. The menu is
+  // the overlay on TOP, so it is the one that takes the token; once it
+  // swallows the key while topmost the badge beneath never sees the press,
+  // which is why the badge needs no token of its own.
+  useOverlayEscape(showContextMenu, () => setShowContextMenu(false))
 
   const {
     attributes,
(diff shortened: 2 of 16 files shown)


adamflagg:
@coderabbitai review

adamflagg:
## Triage close-out (#2237)

**Reviewer verdict: clean, 0 findings** (13 discarded). Every checkable body claim verified against the tree, including the mutation-check transcript (deleting `releaseOverlayToken(token)` → 9 red / 150 green, exactly the release-on-unmount/no-token-while-closed assertions across all four adopters; restored → 159/159 green). One immaterial wording note (RequestReviewPanel's comment says "below" when the hook call is above) — below the reporting floor, not fixed.

**CodeRabbit — requested explicitly** (the earlier green "CodeRabbit" check was a no-op: `Review skipped — auto reviews are disabled`). A real review landed: **4 actionable comments**.

Disposition:

1. **Fictional CampMinder IDs in new tests** (`9001`/`9002`/`12345`/`610`/`611`/`620`/`621` vs. the `1000001`/`1000002` example in `tests/CLAUDE.md`) — **disagreed, not fixed.** `tests/CLAUDE.md` lists `1000001, 1000002` as an example of "generic, not real CampMinder IDs," not an exhaustive whitelist. The existing suite already uses small synthetic integers pervasively (`AllCampersView.test.tsx`: `person_cm_id: 1..5`; `BunkingBoardByArea.panelOverlay.test.tsx`: `person_cm_id: 100`; etc.). These new values are the same kind of obviously-synthetic placeholder, not a privacy risk, and rewriting only this PR's occurrences would be inconsistent with the rest of the codebase rather than fixing anything real.

2. **"Tests dispatch Escape at `document`, so `stopPropagation()` doesn't prove the capture-vs-bubble ordering" (Major, functional correctness)** — **disagreed, not fixed.** Verified empirically in both runtimes that matter: jsdom (the actual test runtime) and real Chromium via Playwright. In both, registering a capture listener on `document` after a bubble listener, then dispatching a `keydown` with `document` as the target, the capture listener fires first and its `stopPropagation()` does suppress the bubble listener on the same node:
   ```
   jsdom:     ["capture(hook)"]   (bubble listener never runs)
   Chromium:  ["capture(hook)"]   (same)
   ```
   CodeRabbit's own sandbox script never actually tested "document as target" — both of its probes dispatched from a descendant `div`, a different scenario. Its prose conclusion came from a general web search, not a run against this codebase's runtime. The tests as written correctly pin real behavior in both jsdom and Chromium.

3. **`DrillDownModal.tsx` comment claims "it renders no overlay of its own"** — **fixed.** It does render a fixed full-viewport backdrop/modal; the actual invariant is the single mount site with no competing Escape-handling surface. Comment corrected.

4. **`OptimizeBunksButton.tsx` comment claims "hosting no overlay of its own"** — **fixed.** The portal dropdown (`role="listbox"`, `fixed z-[9999]`) is an overlay; the actual invariant is the outside-mousedown dismiss with no competing Escape handler. Comment corrected.

**Also fixed, same pattern, not separately flagged:** `CsvPipelineIndicator.tsx`'s comment had the identical defect ("hosts no overlay… has no host relationship with one") for its own `role="region"` absolute-positioned popover. Found by inspection while fixing #3/#4, in a file this PR already touched — fixed for consistency rather than left half-corrected.

All three comment fixes are documentation-only — no behavior touched. Verified: the three adopters' test files (43 tests) still green, `tsc --noEmit` clean, `scripts/pre-push-verify.sh` (prettier/eslint/tsc/vitest) all pass, pushed as `1bfdf84f`, remote ref confirmed moved (`git rev-list --count origin/...HEAD` = 0).

Nothing filed as a separate issue — both fixed findings were M-or-smaller and in files this PR already had open; both disagreed findings were refuted with direct evidence, not deferred.

Holding for GUI review per standing policy — this changes live Escape-key behavior on six surfaces (see PR body's "click through" list).

PR #2378, opened 2026-08-15
fix(sync,frontend): split discarded custom-field values from record-level Skipped

### LINKED ISSUE #2356: fix(sync,frontend): the sync toast's "N skipped" counts discarded values, not records, on two syncs
## What

The admin Sync tab's completion toast renders `N skipped` for two syncs where `N` counts **discarded field values**, not records. A reader sees *"Staff Applications: 274 created, 557 skipped"* and reasonably concludes 557 applications were skipped. What actually happened is that 557 individual custom-field **answers** were discarded across the 274 rows that were created.

## Where

Both syncs increment the same counter the same way:

| file | line | statement |
|---|---|---|
| `pocketbase/sync/camper_transportation.go` | 468 | `s.Stats.Skipped += total` |
| `pocketbase/sync/staff_applications.go` | 443 | `s.Stats.Skipped += total` |

In both, `total` is the sum over `unmappedCounts` — one increment per discarded *value*. The toast then renders it through a formatter that has no idea of the unit:

```ts
// frontend/src/hooks/useSyncCompletionToasts.ts:57
if (stats.skipped > 0) parts.push(`${stats.skipped} skipped`)
```

Both syncs are in that file's label map (`camper_transportation: 'Transportation'` at `:39`, `staff_applications: 'Staff Applications'` at `:41`), so both surface it.

## Why it reads wrong

`created`, `updated` and `errors` in the same toast are all **record** counts. `skipped` sitting beside them inherits that unit by association. The mismatch is visible in the numbers themselves: 2026 staff applications produce **557 skipped against 274 created**, which is impossible to read as records without concluding something went badly wrong.

Historical years are worse — 2017-2023 each discard roughly 2,400-3,400 values, so a backfill run will show a "skipped" count several times its own record count.

## Not a defect in either sync

Both are behaving as designed and as their issues prescribed (kindred#2318 for transportation, kindred#2271 for staff applications, which says *"Stats.Skipped += total for bookkeeping"*). The two are deliberately in step with each other. **This issue is about the display unit, and any fix must change both together** — fixing one alone re-introduces the divergence the template exists to prevent.

## Options

1. **Separate counter.** Add something like `Stats.SkippedValues`, leave `Stats.Skipped` for records, and give the toast a distinct segment (`557 values skipped`). Most correct; touches the stats struct, both syncs, the API status shape, and the toast.
2. **Label it at the toast.** Keep the counter and special-case these two sync ids in `formatStatsText` to render `N values skipped`. Cheapest; leaves the underlying unit ambiguous for any future reader of `Stats.Skipped`.
3. **Stop counting it as `Skipped`.** Drop the increment and rely on the aggregated `slog.Warn` both syncs already emit. Loses the at-a-glance signal that discards happened at all.

Option 1 is the honest one; option 2 is defensible if the stats struct is considered frozen. Either way the change is paired across both syncs.

## Sequencing

- **Files:** `pocketbase/sync/camper_transportation.go`, `pocketbase/sync/staff_applications.go`, `frontend/src/hooks/useSyncCompletionToasts.ts`, and — for option 1 — whatever defines `SubStats` plus the sync status API shape.
- **Predecessors:** kindred#2318 (merged) and kindred#2271 (PR #2355). Both must be in before this is meaningful.
- **Collides with:** anything touching those two sync services. Has frontend impact, so it needs a look on screen before merging.

adamflagg:
## The defect

The admin Sync tab's completion toast rendered `N skipped` for `camper_transportation` and
`staff_applications` as if it were a RECORD count, matching the unit of the `created`/`updated`/
`errors` segments beside it. It was not: both services' `loadPersonCustomValues` incremented the
shared `Stats.Skipped` counter once per discarded custom-field VALUE (an unmapped `BUS-*`/`App-*`
answer), while the record that value belonged to was still created. A toast reading
"274 created, 557 skipped" read as 557 dropped applications; what actually happened is 557
individual field answers discarded across some subset of the 274 rows that *were* created.

Verified in this branch (`TestLoadPersonCustomValuesCountsAndLogsUnmappedFields` /
`...UnmappedAppFields`): one record with a routed field plus two discarded field values produced
`Stats.Skipped = 2` before this change (a "2 skipped" toast segment for a run that skipped zero
records and created one).

## The fix — Option 1, a separate counter (owner-decided, not re-litigated here)

- Added `Stats.SkippedValues int` (`pocketbase/sync/orchestrator.go`) alongside the existing
  `Stats.Skipped`, which keeps its RECORD-count meaning everywhere else in the package.
- `camper_transportation.go` and `staff_applications.go` now increment `Stats.SkippedValues`
  instead of `Stats.Skipped` at their one discard site each (`loadPersonCustomValues`), and both
  completion log lines gained a `skipped_values` key alongside the existing `skipped` key.
  **Both files changed together in this PR** — they are deliberately kept in step
  (kindred#2318, kindred#2271), and fixing one alone would re-introduce that divergence.
- `Stats` is the same struct used for both a service's top-level stats and its `SubStats` entries
  (`SubStats map[string]Stats`), and `GetStats()` on both services returns `Stats` by value, so no
  separate aggregation code was needed — `SkippedValues` flows through automatically everywhere
  `Stats` already flows.
- Frontend: `SubStats` and the sync-status `summary` type (`useSyncStatusAPI.ts`) gained an
  optional `skipped_values` field. `formatStatsText` (`useSyncCompletionToasts.ts`) is now
  **exported** and renders `skipped_values` as its own `N values skipped` segment, never merged
  into the `N skipped` segment. The toast's "standard stats" branch — the actual code path
  `camper_transportation`/`staff_applications` toasts render through — previously hand-duplicated
  `formatStatsText`'s field list rather than calling it; it now calls the same function (with an
  empty label, since the outer toast message already prefixes the service's display name), so the
  fix reaches the real toast instead of only the `persons`/`households` branch that already called
  it.

## What I deliberately did NOT do

- **No `sync_runs` DB column for `SkippedValues`.** `sync_runs.go` persists several stats fields to
  dedicated PocketBase columns; adding one for this counter would need a new migration, which
  the issue didn't ask for. This mirrors the existing `DuplicateStaffStatus` counter, whose comment
  in `orchestrator.go` states the same reasoning for the same non-persistence.
  `SkippedValues` is still surfaced live on every completed run via the sync-status JSON summary
  and the completion log line — the toast this issue is about — just not archived to `sync_runs`.
- **No change to `SyncTab.tsx`'s persistent "N skip" status badge.** That's a separate, non-toast
  display (an inline summary in the admin sync table) that reads `summary.skipped` the same
  merged way the toast used to. It carries the same unit confusion but was out of this issue's
  stated scope (title and body both say "the sync toast"), and changing its wording/layout is a
  product-judgment call the issue doesn't make. Filed as a natural follow-up if wanted.
- **No `frontend/src/types/api-generated/` regeneration.** Checked first: the sync-status endpoint
  and its `SubStats`/`summary` shape are hand-typed in `useSyncStatusAPI.ts`, not part of the
  FastAPI-derived `types.gen.ts` (confirmed by grep — no `sync` status types appear there). No
  codegen freshness gate applies to this change.

## Tests

- Go: extended the existing end-to-end pins for kindred#2272/#2271
  (`TestLoadPersonCustomValuesCountsAndLogsUnmappedFields`,
  `TestLoadPersonCustomValuesCountsAndLogsUnmappedAppFields`), which already drive the real
  `loadPersonCustomValues` entry point with a routed field (record created) plus two discarded
  field values. They now assert `Stats.SkippedValues == 2` and `Stats.Skipped == 0` instead of the
  old (wrong) `Stats.Skipped == 2`. The "no discards" pins for both services now also assert
  `Stats.SkippedValues == 0`.
- Frontend: `useSyncCompletionToasts.test.ts` gained three tests driving the real exported
  `formatStatsText` (not a local reimplementation) that pin (1) a values-skipped count rendering
  as its own `"N values skipped"` segment, (2) both a record-level `skipped` and `skipped_values`
  count rendering as two distinct segments in one string (never collapsing to `559 skipped`), and
  (3) the segment being omitted entirely when `skipped_values` is zero/absent.

## Mutation-check evidence

- Go: pointed both services' discard increment back at `s.Stats.Skipped` (the pre-fix code) and
  re-ran the four `TestLoadPersonCustomValues*` pins above — both "counts and logs" tests failed:
  `Stats.SkippedValues = 0, want 2` and `Stats.Skipped = 2, want 0`, for both services. Reverted;
  full `go test ./sync/...` is green (`ok github.com/camp/kindred/pocketbase/sync 27.217s`), full
  `go test ./...` is green.
- Frontend: changed the new `skipped_values` branch in `formatStatsText` to push `` `${n} skipped`
  `` (the merge bug) instead of `` `${n} values skipped` ``. The two new tests asserting distinct
  segments failed with `Received: "...274 created, 557 skipped"` vs `Expected: "...557 values
  skipped"`. Reverted; `vitest run` on the file is green (13/13), and the full frontend suite is
  green (449 files / 6854 tests).

## GUI impact

**Yes.** The admin Sync tab's completion toast (fires after a `camper_transportation` or
`staff_applications` sync run finishes) now shows a values-skipped count as its own segment,
worded `N values skipped`, separate from the existing `N skipped`/`N created`/`N updated`/`N
errors` segments. Before this change, a run with discarded field values but zero skipped records
showed only `N created, M skipped` (M being the discarded-value count, misread as records); after,
the same run shows `N created, M values skipped` and the `skipped` segment is omitted (it's
genuinely zero). A run with both real skipped records and discarded values now shows both segments
distinctly, e.g. `N created, 2 skipped, 557 values skipped`, instead of a single collapsed number.

Closes #2356.

## Summary by CodeRabbit

* **New Features**
  * Sync status cards now show the number of skipped custom-field values as “val skip.”
  * Sync completion notifications report skipped values separately from skipped records.
  * Expanded reporting applies to transportation, staff application, person, household, and other syncs.

* **Bug Fixes**
  * Improved sync statistics so discarded unmapped values no longer inflate skipped-record counts.

### DIFF
--- a/frontend/src/components/admin/SyncTab.test.tsx
+++ b/frontend/src/components/admin/SyncTab.test.tsx
@@ -33,6 +33,10 @@ let staffLookupsStatus: unknown = idleStatus
 // covered by any existing entry above. Reset in afterEach.
 let staffStatus: unknown = idleStatus
 
+// kindred#2356: `camper_transportation` is a year sync (renderSyncCard) and needs its own
+// injectable status to pin the skipped_values badge. Reset in afterEach.
+let camperTransportationStatus: unknown = idleStatus
+
 vi.mock('../../hooks/useSyncCompletionToasts', () => ({
   useSyncCompletionToasts: () => ({
     family_camp_derived: idleStatus,
@@ -41,7 +45,6 @@ vi.mock('../../hooks/useSyncCompletionToasts', () => ({
     financial_aid_applications: idleStatus,
     household_demographics: idleStatus,
     camper_dietary: idleStatus,
-    camper_transportation: idleStatus,
     quest_registrations: idleStatus,
     staff_applications: idleStatus,
     staff_vehicle_info: idleStatus,
@@ -56,6 +59,9 @@ vi.mock('../../hooks/useSyncCompletionToasts', () => ({
     },
     get staff() {
       return staffStatus
+    },
+    get camper_transportation() {
+      return camperTransportationStatus
     },
   }),
 }))
@@ -364,5 +370,47 @@ describe('SyncTab duplicate staff status badge (#2267)', () => {
     renderSyncTab()
 
     expect(within(staffCard()).queryByText(/dup status/)).not.toBeInTheDocument()
+  })
+})
+
+// Regression test for kindred#2356 (review follow-up): the Sync tab's persistent per-service
+// badge row rendered every other special counter (rejected, duplicate_staff_status,
+// already_processed, prod_audit_warnings, lodging_prod_audit_warnings) but not the new
+// skipped_values -- and since the fix moved camper_transportation/staff_applications off
+// Stats.Skipped entirely, their "skip" badge went permanently blank with no replacement. An
+// operator who doesn't catch the 5-second toast has no on-screen signal at all that field
+// values were discarded.
+describe('SyncTab skipped-values badge (kindred#2356)', () => {
+  afterEach(() => {
+    camperTransportationStatus = idleStatus
+  })
+
+  function transportationCard() {
+    const heading = screen.getByText('Transportation', { selector: 'div' })
+    const card = heading.closest('.flex.flex-col')
+    if (!card) throw new Error('could not find Transportation card')
+    return card as HTMLElement
+  }
+
+  it('renders a values-skipped badge when skipped_values is positive', () => {
+    camperTransportationStatus = {
+      status: 'success',
+      summary: { created: 10, updated: 0, skipped: 0, errors: 0, skipped_values: 557 },
+    }
+
+    renderSyncTab()
+
+    expect(within(transportationCard()).getByText('557 val skip')).toBeInTheDocument()
+  })
+
+  it('renders no values-skipped badge when skipped_values is zero or absent', () => {
+    camperTransportationStatus = {
+      status: 'success',
+      summary: { created: 10, updated: 0, skipped: 0, errors: 0 },
+    }
+
+    renderSyncTab()
+
+    expect(within(transportationCard()).queryByText(/val skip/)).not.toBeInTheDocument()
   })
 })
--- a/frontend/src/components/admin/SyncTab.tsx
+++ b/frontend/src/components/admin/SyncTab.tsx
@@ -283,6 +283,11 @@ export function SyncTab() {
                 {(status.summary.skipped || 0) > 0 && (
                   <span className="text-muted-foreground">{status.summary.skipped} skip</span>
                 )}
+                {(status.summary.skipped_values ?? 0) > 0 && (
+                  <span className="text-muted-foreground">
+                    {status.summary.skipped_values} val skip
+                  </span>
+                )}
                 {status.summary.errors > 0 && (
                   <span className="font-medium text-red-600 dark:text-red-400">
                     {status.summary.errors} err
@@ -859,6 +864,11 @@ export function SyncTab() {
                           {(status.summary.skipped || 0) > 0 && (
                             <span className="text-muted-foreground">
                               {status.summary.skipped} skip
+                            </span>
+                          )}
+                          {(status.summary.skipped_values ?? 0) > 0 && (
+                            <span className="text-muted-foreground">
+                              {status.summary.skipped_values} val skip
                             </span>
                           )}
                           {status.summary.errors > 0 && (
--- a/frontend/src/hooks/useSyncCompletionToasts.test.ts
+++ b/frontend/src/hooks/useSyncCompletionToasts.test.ts
@@ -1,8 +1,26 @@
 /**
  * Tests for useSyncCompletionToasts hook - sub_stats formatting
  */
-import { describe, it, expect } from 'vitest'
-import { formatStatsText } from './useSyncCompletionToasts'
+import { describe, it, expect, vi, beforeEach, type Mock } from 'vitest'
+import { renderHook } from '@testing-library/react'
+import toast from 'react-hot-toast'
+import { formatStatsText, useSyncCompletionToasts } from './useSyncCompletionToasts'
+import type { SyncStatusResponse } from './useSyncStatusAPI'
+
+vi.mock('react-hot-toast', () => ({
+  default: Object.assign(vi.fn(), {
+    success: vi.fn(),
+    error: vi.fn(),
+  }),
+}))
+
+// useSyncCompletionToasts reads its data straight from useSyncStatusAPI's return value, so
+// mocking that hook lets a test drive a running->success transition across two renders
+// without standing up a real QueryClient/pb network layer.
+let mockSyncStatus: Partial<SyncStatusResponse> | null = null
+vi.mock('./useSyncStatusAPI', () => ({
+  useSyncStatusAPI: () => ({ data: mockSyncStatus }),
+}))
 
 // Helper function to format stats (should match implementation)
 interface SubStats {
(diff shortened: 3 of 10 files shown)


adamflagg:
@coderabbitai review

adamflagg:
## Review follow-up applied in-branch (post-#2356)

All three reviewer findings were verified against the diff and fixed in-branch. No issues filed — all were M-or-smaller.

### 1. Medium/S — SyncTab's persistent badge row showed nothing for skipped_values (`frontend/src/components/admin/SyncTab.tsx`)
Confirmed: after the counter split, `Stats.Skipped` is permanently 0 for camper_transportation/staff_applications, so the old `{status.summary.skipped} skip` badge never renders and there was no replacement — an operator who misses the 5s toast had no on-screen signal at all.

Fix: added a `{status.summary.skipped_values} val skip` badge at both SyncTab card sites (`renderSyncCard` and the inlined global-card block), matching the existing `duplicate_staff_status` pattern.

New tests (`SyncTab.test.tsx`, `SyncTab skipped-values badge (kindred#2356)`):
- badge renders when `skipped_values > 0`
- badge absent when zero/absent

Mutation-check: reverted the production change, both new assertions failed as expected (`getByText('557 val skip')` — element not found); reapplied, 13/13 green.

### 2. Medium/M — no test drove the real toast render path (`frontend/src/hooks/useSyncCompletionToasts.test.ts`)
Confirmed: all three original tests called `formatStatsText` directly; the hook's "standard stats" branch that actually builds the toast string was untested.

Fix: added `useSyncCompletionToasts renders the values-skipped segment in the real toast (kindred#2356)` — three tests that mock `useSyncStatusAPI` and drive the real exported hook end-to-end (`running` → `success` transition, real `react-hot-toast` mock), asserting on `toast.success`'s actual call arguments for both camper_transportation and staff_applications.

Mutation-check: reproduced the reviewer's exact regression (reverted the toast branch to the pre-fix hand-copied `created/updated/skipped/errors` field list). Both new toast-path tests failed —
```
- "557 values skipped"
+ "Staff Applications complete: 274 created"
```
— confirming they catch what the formatStatsText-only tests missed. Reapplied, 16/16 green.

### 3. Low/XS — stale doc comments (`camper_transportation_test.go`, `staff_applications_test.go`)
Confirmed: both test doc comments still said "must increment Stats.Skipped" while the assertions below check `SkippedValues == 2` / `Skipped == 0`. Updated both comments to name `Stats.SkippedValues` and cite #2356, so a future reader doesn't "fix" the drift by reverting the assertions.

### Verification
- `go test -race ./sync/...`: PASS, 218.956s (also PASS at 282.966s in an earlier standalone run)
- Frontend: `SyncTab.test.tsx` 13/13, `useSyncCompletionToasts.test.ts` 16/16, `createSyncMutation.test.tsx` unaffected — all green
- `tsc --noEmit`, `go build ./...`, `./node_modules/.bin/eslint`, `golangci-lint run ./sync/...`, prettier — all clean on every changed file
- Full `scripts/pre-push-verify.sh --all`: everything green except `pytest (unit)` — 18 failures, all `FileNotFoundError: Could not find PocketBase database`, a pre-existing worktree environment gap (`pocketbase/pb_data/data.db` absent). Unrelated to this PR (no Python files touched); not fixed here.
- Push verified by remote ref: `git rev-list --count origin/<branch>..HEAD` == 0
- `closingIssuesReferences` confirms `Closes #2356.` registers correctly

### CodeRabbit
Requested explicitly (cross-surface diff). Review completed, covering all 10 files across both fix commits — **no actionable comments**.

Not merging — holding for GUI review per standing policy (the Sync tab badge row is a visible admin surface).

PR #2380, opened 2026-08-15
fix(sync): count and log people gated out of staff_applications and staff_vehicle_info

### LINKED ISSUE #2277: bug(sync): 22 of 296 people who complete staff onboarding have no staff row, and two derived tables silently exclude them

Verified against `main` @ `3eb44ba4` and the production snapshot. Verdict: **STALE_BODY** / NO_CODE.

> Every figure reproduces exactly - 296 answerers, 274 with a staff row, 22 without, 347 App-* values and 29 SVI values lost at the gate, 2 carrying a Unit Counselor Assignment, and the 266/8/4/1 status breakdown. There is no code to write here; it is a measurement awaiting a ruling, exactly as the body says. But the headline framing is confounded: 2026 is the only in-flight year in the table, and staff is swept by DeleteOrphans, so every prior year shows its FINAL roster while 2026 shows a moving one. The comparison that makes 22 look like 'the largest gap on record' is not like-for-like.

1. WRONG: 'pocketbase/sync/staff_applications.go:252 (loadPersonStaffMapping) and pocketbase/sync/staff_vehicle_info.go:205 (loadPersonStaffMapping)'. REPLACEMENT: ':265' and ':245', verified at 3eb44ba4.
2. WRONG: '(pocketbase/sync/staff.go:17-19 lists all four statuses; staff.go:184 calls s.DeleteOrphans after ingesting them.)'. REPLACEMENT: 'staff.go:17-30 lists all four statuses (the block now also carries the kindred#2267 precedence policy); staff.go:196 calls s.DeleteOrphans.'
3. MISLEADING: 'Per-year trend, 2019-2024 ... this is not new in 2026, but 2026 is the largest gap on record'. REPLACEMENT: 'Per-year trend ... 2026 is the largest number in this table, but the years are NOT comparable. staff is swept by DeleteOrphans (staff.go:196), so a historical year shows whichever roster CampMinder returns for it today - an applicant never hired in 2019 leaves no trace and counts 0 by construction. 2026 is the only in-flight year: its staff rows were still being created on 2026-08-06, a week before this snapshot, and already number 279 against 2025's 274. The 22 is a measurement of a moving target, and some unknown share of it will resolve as offers land. What the history shows is that this gap has always closed by the time a season settles; what it cannot show is whether it closed by hiring or by upstream purging.'
4. UNDERSTATED: '296 people answered at least one App-* staff-onboarding custom field for 2026.' REPLACEMENT: '296 people answered App-* staff-onboarding custom fields for 2026 - and the 22 without a staff row answered 11 to 26 fields each (mean 15.8), against 10 to 29 (mean 18.2) for the 274 who have one. These are substantially completed applications, not abandoned starts. Seven of the 22 held a staff row in a previous year.'

- 2026 App-* answerers, denominator = distinct persons with cm_id > 0 holding a non-empty person_custom_values row on an App-* field for 2026: 296 total, 274 with a staff row for 2026, 22 without. Reproduces the body exactly.
- Per-year answerers-with-no-staff-row (same query, year swapped): 2017 0/207, 2018 0/218, 2019 2/231, 2020 0/220, 2021 2/240, 2022 2/239, 2023 1/255, 2024 4/252, 2025 0/258, 2026 22/296. Reproduces the body's table and adds the four years it omitted, all of which are 0.
- Values lost at the gate for 2026: staff_applications 347, staff_vehicle_info 29. Reproduces exactly.
- Of the 22, those carrying a non-empty Unit Counselor Assignment for 2026: 2. Reproduces exactly.
- staff 2026 by status: active 266, resigned 8, dismissed 4, cancelled 1 (279 total). Reproduces exactly, so the body's 'dismissal is not the explanation' argument stands.
- NEW - how complete the 22's applications are, denominator = App-* fields answered per person for 2026: the 22 answered 11-26 fields (mean 15.8, 347 values); the 274 answered 10-29 (mean 18.2, 4,980 values). These are NOT abandoned single-question applications, which STRENGTHENS the body's premise - it currently says only 'answered at least one App-* field'.
- NEW - 7 of the 22 hold a staff row in some OTHER year (returning staff); 15 appear only as 2026 applicants.
- NEW - the confound: staff rows per year are 274 for 2025 and 279 for 2026, and 2025's are all stamped with one bulk re-sync timestamp while 2026's are still being created up to 2026-08-06. Because staff.go:196 sweeps orphans, a historical year's staff table is whatever CampMinder returns for it TODAY - so a 2019 applicant who was never hired shows as 0 today by construction. The near-zero history is therefore consistent with 'this happens every year and resolves or is purged upstream', not with 'this is new in 2026'.

</details>

---

## What is wrong

296 people answered at least one `App-*` staff-onboarding custom field for 2026. 22 of them have no `staff` row for 2026 at all. Two of the 22 carry a non-empty `Unit Counselor Assignment` value for 2026 — a signal CampMinder normally only populates for people already on a unit's staffing plan, not for someone who merely started an application.

This is the same shape as the staff-gate drop fixed for `staff_vehicle_info` earlier in this repair (kindred#2273), but wider. `staff_vehicle_info` and `staff_applications` both key off the identical `loadPersonStaffMapping(ctx, year)` gate (a `staff` lookup filtered to `year = <sync year>`), and both silently drop every value for a person who fails it. For 2026 that gate currently discards **347** `App-*` values in `staff_applications` and **29** SVI values in `staff_vehicle_info` — on top of whatever this PR's counters now report for the SVI side specifically.

**No names, no CampMinder IDs, no answer text below — counts and mechanisms only.**

## Measured (verified against the production snapshot, `pocketbase/pb_data/data.db`)

2026 `App-*` answerers vs. `staff` coverage:

```sql
WITH app AS (
  SELECT DISTINCT p.cm_id AS cm_id
  FROM person_custom_values pcv
  JOIN custom_field_defs d ON d.id = pcv.field_definition
  JOIN persons p ON p.id = pcv.person
  WHERE pcv.year = 2026 AND TRIM(d.name) LIKE 'App-%' AND pcv.value != '' AND p.cm_id > 0
)
SELECT COUNT(*) AS total_answerers,
  SUM(CASE WHEN EXISTS (SELECT 1 FROM staff s WHERE s.person_id = app.cm_id AND s.year = 2026)
      THEN 1 ELSE 0 END) AS has_staff_row,
  SUM(CASE WHEN EXISTS (SELECT 1 FROM staff s WHERE s.person_id = app.cm_id AND s.year = 2026)
      THEN 0 ELSE 1 END) AS no_staff_row
FROM app;
-- 296 | 274 | 22
```

Per-year trend, 2019-2024 (same query, `pcv.year` swapped in) — this is not new in 2026, but 2026 is the largest gap on record:

| year | App-\* answerers, no `staff` row |
|---|---|
| 2019 | 2 |
| 2020 | 0 |
| 2021 | 2 |
| 2022 | 2 |
| 2023 | 1 |
| 2024 | 4 |
| **2026** | **22** |

Of the 22, two carry a non-empty `Unit Counselor Assignment` value for 2026:

```sql
WITH app AS (
  SELECT DISTINCT p.cm_id AS cm_id
  FROM person_custom_values pcv
  JOIN custom_field_defs d ON d.id = pcv.field_definition
  JOIN persons p ON p.id = pcv.person
  WHERE pcv.year = 2026 AND TRIM(d.name) LIKE 'App-%' AND pcv.value != '' AND p.cm_id > 0
    AND NOT EXISTS (SELECT 1 FROM staff s WHERE s.person_id = p.cm_id AND s.year = 2026)
)
SELECT COUNT(DISTINCT p2.cm_id)
FROM person_custom_values pcv2
JOIN custom_field_defs d2 ON d2.id = pcv2.field_definition
JOIN persons p2 ON p2.id = pcv2.person
WHERE d2.name = 'Unit Counselor Assignment' AND pcv2.value != '' AND pcv2.year = 2026
  AND p2.cm_id IN (SELECT cm_id FROM app);
-- 2
```

Values lost at the identical gate, both services, 2026:

```sql
-- staff_applications: 347
SELECT COUNT(*) FROM person_custom_values pcv
JOIN custom_field_defs d ON d.id = pcv.field_definition
JOIN persons p ON p.id = pcv.person
WHERE pcv.year = 2026 AND TRIM(d.name) LIKE 'App-%' AND pcv.value != '' AND p.cm_id > 0
  AND NOT EXISTS (SELECT 1 FROM staff s WHERE s.person_id = p.cm_id AND s.year = 2026);

-- staff_vehicle_info: 29
SELECT COUNT(*) FROM person_custom_values pcv
JOIN custom_field_defs d ON d.id = pcv.field_definition
JOIN persons p ON p.id = pcv.person
WHERE pcv.year = 2026 AND (TRIM(d.name) LIKE 'SVI-%' OR TRIM(d.name) LIKE 'SVI %')
  AND pcv.value != '' AND p.cm_id > 0
  AND NOT EXISTS (SELECT 1 FROM staff s WHERE s.person_id = p.cm_id AND s.year = 2026);
```

`pocketbase/sync/staff_applications.go:252` (`loadPersonStaffMapping`) and `pocketbase/sync/staff_vehicle_info.go:205` (`loadPersonStaffMapping`) are separately-defined but structurally identical: both build `map[int]string` from `FindRecordsByFilter("staff", "year = <year>", ...)`. A person absent from `staff` for a year fails both gates the same way.

## Dismissal is not the explanation

`staff` for 2026 holds all four CampMinder statuses the sync ingests:

```sql
SELECT status, COUNT(*) FROM staff WHERE year = 2026 GROUP BY status;
-- active: 266, resigned: 8, dismissed: 4, cancelled: 1
```

(`pocketbase/sync/staff.go:17-19` lists all four statuses; `staff.go:184` calls `s.DeleteOrphans` after ingesting them.) Because the sync pulls every status, a resigned, dismissed, or cancelled staff member still has a `staff` row — absence from `staff` means CampMinder genuinely has no record for that person in that year, not that a status filter hid one. So the 22 are not people who were hired and then dismissed; they are people who filled out onboarding materials without (yet, or ever) becoming staff for the year.

## Why this is not self-healing

`staff` is synced through `BaseSyncService.DeleteOrphans` (`staff.go:184`), which deletes any `staff` row for a person CampMinder no longer returns for that year. That makes the roster a point-in-time mirror: if a person is later added to CampMinder's staff list for 2026, the next `staff` sync creates their row and both derived-table gates open retroactively — but if a person's `App-*`/SVI answers were captured, dropped, and the person is *never* added, or is added and then *removed* again, there is nothing left to distinguish "never got that far," "added and removed," or "about to be added" — **nothing in the database records which of the three happened.** The 22 rows in this snapshot could be any mix of pre-hire (offer pending), never-hired (application abandoned), or post-removal (added then deleted from `staff`), and the schema has no way to tell them apart after the fact.

## Scope

This issue is the gap itself — a roster/data-completeness question, not a Go bug. It is filed separately from the `staff_vehicle_info` transform repair (kindred#2258, kindred#2262, kindred#2268, kindred#2273) because fixing the counting and routing bugs in that PR does not, and should not, touch this gate: the gate is required (both `staff_applications` and `staff_vehicle_info` write a required relation to `staff`), so removing it converts silent drops into save errors, not saves. See kindred#2273's traps section for why the gate cannot simply be deleted.

## Fix direction (not decided here — this issue is the measurement, not the plan)

Two questions, likely separable:

1. **Should the 22 (and future years' equivalents) be visible at all before their `App-*`/SVI answers are silently dropped?** An operator running the season's staff onboarding today has no signal that people are answering set-up questions who currently don't resolve to a roster entry.
2. **Should the database record why a person left the derived tables' reach** — pre-hire, never-hired, or post-removal — rather than treating all three as indistinguishable absence? That is a schema question (a status/reason field somewhere), not a one-line fix, and needs its own decision.

Part of kindred#2257

### LINKED ISSUE #2383: fix(sync): staff_vehicle_info is the THIRD sync counting discarded values as Skipped
## What

kindred#2356 established that a sync reporting `N skipped` beside record-level `created`/`updated`/`errors`
must not be counting discarded **values**. It named **two** syncs. There is a **third**:
`pocketbase/sync/staff_vehicle_info.go` has the same defect and was not in scope.

Found while the owner was validating PR #2378 on screen — they looked at the Staff Vehicles card,
saw `200 upd, 29 skip` unchanged on both `main` and the branch, and could not tell whether the fix
had worked. It had; that sync simply was not covered.

## Measured

Both increments are inside a loop over `entries`, and an entry is **one custom-field value**, not a record:

| file | line | what it counts |
|---|---|---|
| `pocketbase/sync/staff_vehicle_info.go` | 377 | one value whose person has no `staff` row for the year |
| `pocketbase/sync/staff_vehicle_info.go` | 393 | one value whose SVI field name maps to no column |

Production snapshot, 2026: **29** SVI values are discarded at the gate for 2026 — which is exactly the
`29 skip` rendered on the card, against `200` records updated. Same shape as the
`557 skipped / 274 created` that motivated kindred#2356.

## Why it was missed

kindred#2356's body says *"Both syncs increment the same counter the same way"* and lists
`camper_transportation.go:468` and `staff_applications.go:443` — both of which use `s.Stats.Skipped += total`,
summing a `unmappedCounts` map. `staff_vehicle_info.go` reaches the same wrong unit by a **different
spelling**: two bare `s.Stats.Skipped++` sites in the entry loop. A grep for `+= total` does not find it.

That is worth recording, because kindred#2356's own warning was that these syncs are *"deliberately in step
with each other"* and that fixing one alone re-introduces divergence. Fixing two of three does the same.

## Scope

Port PR #2378's change to this sync:

- Route both value-level increments to `Stats.SkippedValues` (the counter #2378 adds).
- Leave `Stats.Skipped` for records.
- The Sync tab's per-service badge row and the completion toast then report it in the same
  unit-explicit way as the other two, with no frontend change needed beyond what #2378 already ships.

⚠️ **Check the gate site (`:377`) against kindred#2277 before moving it.** #2277 adds a counter and one
aggregated log line for **people** dropped at that same gate, and its PR (#2380) touches this file.
The two must agree on which counter gets people and which gets values — people are records
(`Stats.Skipped`), field values are not (`Stats.SkippedValues`). Sequence against #2380 rather than
running in parallel.

## Sequencing

- **Files:** `pocketbase/sync/staff_vehicle_info.go`, `pocketbase/sync/staff_vehicle_info_test.go`.
- **Predecessors:** PR #2378 (must merge first — it defines `Stats.SkippedValues`) and PR #2380
  (kindred#2277, same file).
- **Collides with:** kindred#2277.

adamflagg:
## The defect

`staff_applications.go` and `staff_vehicle_info.go` both key their `person_custom_values` extraction off `loadPersonStaffMapping(ctx, year)`, a `staff` lookup filtered to `year = <sync year>`. A person who answered the onboarding form but has no `staff` row for that year fails this gate and is dropped with a bare `continue` — no counter, no log line.

Re-measured against the production snapshot (`pocketbase/pb_data/data-prod.db`, read-only):

```sql
select count(distinct pcv.person)
from person_custom_values pcv
join custom_field_defs cfd on pcv.field_definition = cfd.id
where pcv.year = 2026 and cfd.name like 'App-%' and pcv.value != '';
```

- **297** people answered `App-*` 2026 onboarding fields (issue said 296 — a 1-person snapshot drift).
- **22** of them have no `staff` row for 2026 at all — same as the issue.
- Restricted to `App-*`-prefixed fields (the issue's original query), those 22 answered **11–26** fields each, mean **15.8** — matches the issue exactly. `isStaffApplicationField` also admits `Position Preference*` fields (no `App-` prefix), and `loadFieldDefinitions` builds its map from everything that function admits — so the shipped `gatedValues` counter isn't restricted to `App-*` and actually sums both: **347** `App-*`-only values, **388** including `Position Preference*`; **11–28** fields each including `Position Preference*`, mean **17.6**. A real 2026 run will log and report the **388/11–28/17.6** figures, not 347 — corrected here after the first draft of this body quoted the 388-inclusive range while claiming it was App-*-only and blamed a nonexistent "dedup by mapped column" difference in the issue's query.
- **2** of the 22 carry a non-empty `Unit Counselor Assignment` — matches the issue exactly.
- **7** of the 22 held a `staff` row in a previous year — matches the issue exactly.
- **29** `SVI-*` values were discarded for **6** gated people in `staff_vehicle_info.go` — matches the issue exactly. The SVI form has only 10 fields total (not the 44 App-*/Position-Preference fields the App-side form has); those 6 people's field counts run **2–10**, mean **4.8**, not the 11–26/15.8 App-side figures — an early draft's SVI comments wrongly reused the App-side numbers, corrected in this branch (see "Review findings addressed" below).

## What changed

Per the adopted default in #2277: **a counter and one aggregated log line at each gate, nothing more.**

- **`staff_applications.go`** (`loadPersonCustomValues`): the staff-row gate now tracks gated people in a `map[int]bool` instead of silently `continue`-ing. After the loop, if any people were gated:
  - `Stats.Skipped` gets `+= len(gatedPeople)` — a person gated out is exactly a record this sync would otherwise have written (per #2356's split, `Skipped` is the record-level counter).
  - `Stats.SkippedValues` gets `+= gatedValues` — the individual discarded `App-*`/`Position Preference*` answers, using the same counter this file already uses for unmapped-field discards.
  - One aggregated `slog.Warn` fires with `year`, `people`, `values` — never one per person.

- **`staff_vehicle_info.go`** (`loadPersonCustomValues`): the staff-row gate already incremented `Stats.Skipped`, but **once per discarded value, not once per gated person** — a person who answered several SVI-* fields inflated a single dropped record into that many "skips" — and emitted **no log line at all**, despite a comment claiming "it must not be silent" (kindred#2273). Both are fixed: gated people are now tracked in a set and counted once each, and one aggregated `slog.Warn` fires with `year`, `people`.

- **`pocketbase/main_test_parallelism_test.go`**: registered the two new `Sync()`-driving tests in the existing `captureSweepLogs` serial group — they swap the process-global `slog` default and cannot run with `t.Parallel()`, same as every other test already in that group.

## Deliberately NOT done

- Did not touch the gate itself. `staff` is a required relation; removing the gate converts silent drops into save errors, not saves (see kindred#2273's traps section).
- Did not add a schema column, a new table, or a UI.
- Did not touch `camper_transportation.go`.

## Tests

Both services get two new tests apiece, driving the real entry points (not a field or a helper):

- `TestLoadPersonCustomValuesCountsStaffGateDropsByPerson[NotByValue]` — drives `loadPersonCustomValues` directly with a gated person who answered multiple fields; asserts the counter lands on **distinct people**, not raw value count.
- `TestStaff{Applications,VehicleInfo}SyncLogsStaffGateDropsOnce` — drives the real `Sync()` entry point with two gated people (one with multiple answers); asserts `Stats.Skipped` (and, for applications, `Stats.SkippedValues`) and that the log contains **exactly one** occurrence of the aggregated warning string, never one per person.
- `TestLoadPersonCustomValuesCountsUnmappedFields` (`staff_vehicle_info_test.go`) — rewritten during review to also assert `len(records) == 1` and `Stats.Skipped == 0`, pinning that an unmapped-field value discard is a value-level event on a record that WAS still created, not a record-level one.

### Mutation-check evidence (all run against `go test ./sync/ -run <test>`, restored after each)

| Mutation | Test | Result |
|---|---|---|
| `staff_applications.go`: drop `gatedPeople[entry.personID] = true` | `TestLoadPersonCustomValuesCountsStaffGateDropsByPerson` | RED — `Stats.Skipped = 0, want 1`; `Stats.SkippedValues = 0, want 2` |
| `staff_applications.go`: move the `slog.Warn` inside the per-entry gate branch | `TestStaffApplicationsSyncLogsStaffGateDropsOnce` | RED — `saw 4 log lines ..., want exactly 1` |
| `staff_vehicle_info.go`: drop `gatedPeople[entry.personID] = true` | `TestLoadPersonCustomValuesCountsStaffGateDropsByPersonNotByValue` | RED — `Stats.Skipped = 0, want 1` |
| `staff_vehicle_info.go`: move the `slog.Warn` inside the per-entry gate branch | `TestStaffVehicleInfoSyncLogsStaffGateDropsOnce` | RED — `saw 5 log lines ..., want exactly 1` |
| `staff_vehicle_info.go`: revert `s.Stats.SkippedValues++` back to `s.Stats.Skipped++` at the unmapped-field site | `TestLoadPersonCustomValuesCountsUnmappedFields` | RED — `Stats.Skipped = 1, want 0`; `Stats.SkippedValues = 0, want 1` |

All five restored; full suite green afterward.

## Review findings addressed

A review pass on this PR found one code defect and documentation drift, all fixed in-branch (no follow-up issue — every finding was M-or-smaller):

- **`staff_vehicle_info.go`'s other discard site mixed units.** The pre-existing unmapped-field-value drop (a field admitted by the `SVI-`/`SVI ` prefix but routed to no column, for a person who **does** have a staff row so the record is already created) was incrementing `Stats.Skipped` per value instead of `Stats.SkippedValues`. All 10 live 2026 SVI field names currently route successfully so this fires zero times today, but the moment CampMinder publishes one new unmapped SVI field answered by, say, 200 staff, the run would create 200+ rows while reporting `skipped: 200+` — rendered on the sync admin badge/toast as 200+ dropped records when zero were dropped. Fixed to `Stats.SkippedValues++`; `TestLoadPersonCustomValuesCountsUnmappedFields` rewritten to assert `Stats.Skipped == 0` / `Stats.SkippedValues == 1` / the record is still created, confirmed RED under the reverse mutation.
- **Stale SVI production numbers in two comments.** `staff_vehicle_info.go` and its test carried "(production: 11-26 fields, mean 15.8)", copy-pasted from the App-side comment — arithmetically impossible for a 10-field form. Corrected to the measured SVI figures (2–10 fields, mean 4.8).
- **`orchestrator.go`'s `Stats.SkippedValues` doc and its two frontend mirrors (`useSyncCompletionToasts.ts`, `useSyncStatusAPI.ts`) were stale** in two ways: they still said only `camper_transportation.go`/`staff_applications.go` set the counter (now `staff_vehicle_info.go` does too, from the fix above), and they described `SkippedValues` as always belonging to a record that was still created — true for the unmapped-field flavor, false for `staff_applications.go`'s gate flavor added by this PR, where the record was never created at all. All three comments now describe both flavors and the current three-service list.
- **This PR body's own measurement paragraph was wrong.** An earlier draft claimed its `App-*`-only SQL matched the issue's 11-26/15.8, then quoted 11-28/17.6 (the `Position Preference*`-inclusive numbers) as if they were the same measurement, and invented a "the issue deduplicated by mapped column" explanation for a gap that doesn't exist — the two ranges are just different queries. Rewritten above with both queries' real numbers and a note that the shipped counter reports the larger (388-value) figure, not 347, because `isStaffApplicationField`'s field map — which `gatedValues` sums over — was never restricted to `App-*` to begin with.

## Verification

```
$ go build ./...                                    # exit=0
$ go vet ./...                                       # exit=0
$ go test ./...                                      # ok, all packages
$ go test -race ./sync/                              # ok  github.com/camp/kindred/pocketbase/sync  144.708s
$ golangci-lint run ./sync/... --new-from-rev=origin/main
0 issues.
```

This branch's base (`feature/skipped-values-counter-2356`, #2378) merged to `main` while this PR was under review; GitHub auto-retargeted this PR's base to `main` and it now gets full CI — 21/21 checks passing as of `057f256d`.

Closes #2277.

---

## Also fixes kindred#2383 — the third sync's values-vs-records counter

kindred#2383 was filed separately while this PR was in review, after the owner could not tell on screen
whether kindred#2356's counter split had worked (they were reading the Staff Vehicles card, which
kindred#2356 never covered). It names two sites in `staff_vehicle_info.go` that counted discarded
**values** through `Stats.Skipped`, the record-level counter.

This PR resolves both, and the two sites resolve in *opposite* directions, which is the part worth
reading:

- **The staff-row gate drop** is now counted **by PERSON**, not by value — `gatedPeople` collects the
  distinct people and `Stats.Skipped += len(gatedPeople)`. A person dropped at that gate is a record
  this sync would otherwise have written, so `Stats.Skipped` is its correct home. Counting it per value
  inflated a single dropped record into up to ten increments.
- **The unmapped-field drop** now increments `Stats.SkippedValues`. That one really is a value: the
  record IS created, and only one of its answers is discarded.

So the two counters end up carrying different units on purpose, and the Staff Vehicles card reports
both accurately for the first time.

Closes #2383.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -831,7 +831,7 @@ jobs:
         # REQUIRED_COLLECTIONS/REQUIRED_FIELDS happens to be an early one that
         # already fell inside the first 30 -- but kindred#1921's
         # fixture-schema-agreement step (below) walks every collection
-        # newLodgingTestApp builds, several of them late (lodging_* from
+        # newSyncTestApp builds, several of them late (lodging_* from
         # migration 1500000116+), which is what surfaced this. Pinned to
         # PocketBase's own MaxPerPage (tools/search/provider.go) rather than
         # some smaller headroom number: a perPage above MaxPerPage is
@@ -869,8 +869,8 @@ jobs:
         echo "✅ Migration smoke test passed"
 
     # kindred#1921: the sync package's fixtures build their PocketBase
-    # collections by hand (see newLodgingTestApp in
-    # pocketbase/sync/lodging_testsupport_test.go) rather than replaying
+    # collections by hand (see newSyncTestApp in
+    # pocketbase/sync/sync_testsupport_test.go) rather than replaying
     # pb_migrations, so a column a migration drops or renames can stay in the
     # fixture indefinitely -- go test ./... never notices. /tmp/collections.json
     # above is the real schema those same migrations just produced; this step
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,6 +90,7 @@ jobs:
       migrations: ${{ steps.filter.outputs.migrations }}
       frontend: ${{ steps.filter.outputs.frontend }}
       scripts: ${{ steps.filter.outputs.scripts }}
+      guardSelfTests: ${{ steps.filter.outputs.guardSelfTests }}
       docker: ${{ steps.filter.outputs.docker }}
       actions: ${{ steps.filter.outputs.actions }}
     steps:
@@ -179,6 +180,38 @@ jobs:
           scripts:
             - 'scripts/**/*.sh'
             - '.shellcheckrc'
+          # kindred#2370: scripts/dev/test-*.sh are the guard scripts' own
+          # self-tests (e.g. test-verify-migration-history.sh pins the
+          # unanchored-boundary regression from PR #2366) and, before this
+          # filter existed, ran nowhere in CI -- the `scripts` filter above
+          # already matches this directory via `scripts/**/*.sh`, but the
+          # only job reading that output is go-lint's shellcheck step
+          # (syntax, not behavior). A dedicated filter makes the
+          # guard-self-tests job's gate say what it actually depends on,
+          # instead of relying on unrelated incidental coverage from a
+          # filter meant for something else. `scripts/dev/**` (not just
+          # `*.sh`) also covers scripts/dev/lib/pb-harness.sh, the shared
+          # lib every guard here delegates its tool-absence contract to.
+          # `.github/workflows/ci.yml` is included for the same reason the
+          # `go` filter above includes it: the guard-self-tests job's own
+          # shape (which scripts it runs, how) lives in this file, so a PR
+          # that only edits the job -- like the one that first added it --
+          # must still be able to trigger it. Without this line, that PR
+          # would ship a job whose own CI run never proves it runs at all,
+          # which is the exact failure kindred#2370 is about.
+          # `pocketbase/pb_migrations/**` is included because
+          # test-verify-consolidation.sh (wired into this same job) runs
+          # end-to-end against the REAL migration tree, not a fixture --
+          # a migrations-only PR can add a schema-free tail migration
+          # (a pure data/RBAC rewrite like 1500000154) that breaks the
+          # suite's drift assertion. Without this line, that PR would not
+          # trigger guard-self-tests and would merge green; the red would
+          # then surface on some later, unrelated PR that happens to touch
+          # scripts/dev/ or ci.yml, misattributed to the wrong change.
+          guardSelfTests:
+            - 'scripts/dev/**'
+            - '.github/workflows/ci.yml'
+            - 'pocketbase/pb_migrations/**'
           docker:
             - 'docker/**'
             - 'docker-compose*.yml'
@@ -231,6 +264,62 @@ jobs:
 
     - name: Guard self-test
       run: ./scripts/dev/test-verify-no-hardcoded-lodging.sh
+
+  # kindred#2370: scripts/dev/verify-migration-history.sh gates local
+  # start_dev.sh boots and stands between a renumbered migration and a
+  # corrupted local DB, but its own 15-test self-test
+  # (test-verify-migration-history.sh) ran nowhere -- CI only wired
+  # test-verify-no-hardcoded-lodging.sh above. TEST 14/15 there pin a real
+  # defect PR #2366 introduced and fixed: an unanchored boundary regex that
+  # made the guard report silently CLEAN on a real collision, the one
+  # direction a collision predictor must never fail in.
+  #
+  # Gated on `guardSelfTests` (scripts/dev/**), not `migrations` or `go` --
+  # a change to only a script under scripts/dev/ (the guard, its suite, or
+  # the shared pb-harness.sh lib) touches none of those, so gating this job
+  # on either would leave it exactly as unwired as the absence this issue
+  # is about. See the `guardSelfTests` filter comment above.
+  #
+  # The other three self-tests found under scripts/dev/test-*.sh
+  # (test-verify-consolidation.sh, test-migration-schema-diff.sh,
+  # test-pb-harness.sh) were audited the same way and are wired here too --
+  # all three pass standalone against a scratch Go build and a scratch
+  # sqlite3 DB, with no live PocketBase server or seeded dev DB required, so
+  # none of them carry the flake risk that would make wiring them a mistake.
+  guard-self-tests:
+    name: Dev Guard Self-Tests
+    needs: detect-changes
+    if: needs.detect-changes.outputs.guardSelfTests == 'true'
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      with:
+        persist-credentials: false
+
+    # migration-history's suite needs only sqlite3/python3 (both preinstalled
+    # on ubuntu-latest), but the consolidation and pb-harness suites build a
+    # scratch pocketbase binary via pb_harness_build_binary, so Go is a real
+    # dependency of this job, not an unused setup step.
+    - name: Setup Go
+      uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
+      with:
+        go-version-file: pocketbase/go.mod
+        check-latest: true
+        cache: true
+        cache-dependency-path: pocketbase/go.sum
+
+    - name: Migration history guard self-test
+      run: ./scripts/dev/test-verify-migration-history.sh
+
+    - name: Consolidation guard self-test
+      run: ./scripts/dev/test-verify-consolidation.sh
+
+    - name: Migration schema-diff guard self-test
+      run: ./scripts/dev/test-migration-schema-diff.sh
+
+    - name: PB harness lib self-test
+      run: ./scripts/dev/test-pb-harness.sh
 
   # Python linting (parallel with other lint jobs)
   python-lint:
@@ -917,7 +1006,7 @@ jobs:
   # Summary job - gate for PR merges
   summary:
     name: CI Summary
-    needs: [pr-title-lint, detect-changes, secret-scan, lodging-guard, python-lint, go-lint, frontend-lint, docker-lint, security-lint, tests-python, tests-go, tests-migrations, tests-frontend]
+    needs: [pr-title-lint, detect-changes, secret-scan, lodging-guard, guard-self-tests, python-lint, go-lint, frontend-lint, docker-lint, security-lint, tests-python, tests-go, tests-migrations, tests-frontend]
     if: always()
     runs-on: ubuntu-latest
     steps:
@@ -927,6 +1016,7 @@ jobs:
         PR_TITLE_LINT: ${{ needs.pr-title-lint.result }}
         SECRET_SCAN: ${{ needs.secret-scan.result }}
         LODGING_GUARD: ${{ needs.lodging-guard.result }}
+        GUARD_SELF_TESTS: ${{ needs.guard-self-tests.result }}
         PYTHON_LINT: ${{ needs.python-lint.result }}
         GO_LINT: ${{ needs.go-lint.result }}
         FRONTEND_LINT: ${{ needs.frontend-lint.result }}
@@ -955,6 +1045,10 @@ jobs:
         fi
         if [ "$LODGING_GUARD" != "success" ]; then
           echo "::error::Lodging name guard failed! A camp unit name may have landed in source -- see docs/reference/lodging-registry.md."
+          exit 1
+        fi
+        if [ "$GUARD_SELF_TESTS" != "success" ] && [ "$GUARD_SELF_TESTS" != "skipped" ]; then
+          echo "::error::A scripts/dev/ guard self-test failed! See the guard-self-tests job."
           exit 1
         fi
         # Lint jobs: success or skipped (when no relevant changes)
--- a/frontend/src/components/CamperCard.test.tsx
+++ b/frontend/src/components/CamperCard.test.tsx
@@ -1,0 +1,136 @@
+/**
+ * kindred#2237 — CamperCard's right-click context menu and the kindred#2205
+ * overlay token stack.
+ *
+ * The menu is rendered by cards INSIDE the expanded `ui/FloatingQueueBadge`
+ * queue (see `FloatingUnassignedBadge`, which renders CamperCard rows straight
+ * into the badge's popover). Both surfaces listened on `document` in the
+ * bubble phase and neither stopped propagation, so a single Escape closed the
+ * context menu AND collapsed the queue it was opened from — the kindred#2205
+ * double-close, still live on the board.
+ *
+ * The menu is the overlay that sits ON TOP, so it is the one that needs the
+ * token: once it swallows the key while topmost, the badge beneath it never
+ * sees the press and needs no token of its own.
+ *
+ * Fictional data throughout.
+ */
+import { describe, it, expect, vi, beforeEach } from 'vitest'
+import { render, screen, fireEvent, waitFor } from '@testing-library/react'
+
+import CamperCard from './CamperCard'
+import { acquireOverlayToken, hasOpenModal, releaseOverlayToken } from './ui/modalStack'
+import { mockCamper } from '../test/mockData'
+import { emptyCamperSatisfaction } from '../types/satisfaction'
+
+vi.mock('../hooks/useCurrentYear', () => ({ useYear: () => 2025 }))
+
+vi.mock('@dnd-kit/sortable', () => ({
+  useSortable: () => ({
+    attributes: {},
+    listeners: {},
+    setNodeRef: vi.fn(),
+    transform: null,
+    transition: undefined,
+    isDragging: false,
+  }),
+}))
+
+vi.mock('../hooks', () => ({
+  useBunkRequestContext: () => ({
+    getSatisfiedRequestInfo: (cmId: number) => emptyCamperSatisfaction(cmId),
+  }),
+  useCamperHistoryContext: () => ({ getLastYearHistory: () => null }),
+}))
+
+vi.mock('../contexts/LockGroupContext', () => ({
+  useLockGroupContext: () => ({
+    addPendingCamper: vi.fn(),
+    removePendingCamper: vi.fn(),
+    getPendingAnimationDelay: () => 0,
+    groups: [],
+    addCamperToGroup: vi.fn(),
+    getCamperLockGroup: () => null,
+    getGroupMembers: () => [],
+    setSelectedGroupId: vi.fn(),
+    setIsLockPanelOpen: vi.fn(),
+  }),
+}))
+
+describe('CamperCard — context-menu overlay token (kindred#2237)', () => {
+  beforeEach(() => {
+    vi.clearAllMocks()
+  })
+
+  /**
+   * `handleContextMenu` defers the open by one `requestAnimationFrame` (it
+   * broadcasts `closeAllContextMenus` first and lets that settle), so every
+   * test has to wait for the menu rather than assert synchronously.
+   */
+  async function openMenu() {
+    const view = render(<CamperCard camper={mockCamper({ person_cm_id: 9001 })} isDraftMode />)
+    const card = document.querySelector('[data-camper-card], button') as HTMLElement
+    fireEvent.contextMenu(card)
+    await waitFor(() => {
+      expect(screen.getByText('View Details')).toBeInTheDocument()
+    })
+    return view
+  }
+
+  it('closes the menu on Escape when it is the only open overlay', async () => {
+    await openMenu()
+
+    fireEvent.keyDown(document, { key: 'Escape' })
+
+    await waitFor(() => {
+      expect(screen.queryByText('View Details')).not.toBeInTheDocument()
+    })
+  })
+
+  it('does NOT close once an overlay has opened on top of it', async () => {
+    await openMenu()
+
+    const topToken = acquireOverlayToken()
+    try {
+      fireEvent.keyDown(document, { key: 'Escape' })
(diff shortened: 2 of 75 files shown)


adamflagg:
## Review response — findings fixed in-branch, plus a base-branch update

**Base-branch update (materially changes this PR's CI story):** while this PR was under review, its base (`feature/skipped-values-counter-2356`, #2378) merged to `main`. GitHub auto-retargeted this PR to `main`, so it now gets **full CI — 21/21 checks passing** as of `057f256d` (rebased cleanly, no conflicts). The `closingIssuesReferences` GraphQL check now correctly returns `#2277`, so `Closes #2277.` will register on merge — no re-save needed. The "no CI, stacked" framing in the original PR body and implementer report is stale; the PR body has been rewritten to reflect this.

All four review findings verified against `pocketbase/pb_data/data-prod.db` (read-only) and the branch tree, and all four fixed in-branch — every one was M-or-smaller, so no follow-up issue filed:

1. **PR body's measurement paragraph (medium, XS).** CONFIRMED — re-ran both SQL variants: App-*-only gives 22 people / 347 values / 11-26 fields / mean 15.77 (matches the issue and the source-code comment); App-* + Position Preference gives 22 / 388 / 11-28 / mean 17.64. The shipped `gatedValues` counter sums over everything `isStaffApplicationField` admits (verified at `staff_applications.go`'s `loadFieldDefinitions`, which has no `App-`-only restriction), so a real run reports 388, not 347. Body rewritten with both numbers and an explicit note that 388 is what ships.

2. **SVI comment carried impossible App-side numbers (low, XS).** CONFIRMED — `staff_vehicle_info.go:372` and its test twin said "11-26 fields, mean 15.8," but the SVI form has only 10 distinct field names total. Measured the real gated-person distribution: 2-10 fields, mean 4.83. Both comments corrected.

3. **`Stats.SkippedValues` semantics contradiction (low, XS).** CONFIRMED — `orchestrator.go`'s doc and the frontend's `useSyncCompletionToasts.ts`/`useSyncStatusAPI.ts` mirrors all said `SkippedValues` values belong to a record "still created," but `staff_applications.go`'s new gate flavor counts values whose record was *never* created. Fixed by broadening all three comments to describe both flavors, rather than dropping the gate's `SkippedValues` line (the PR's own stated intent was for this counter to carry that evidence).

4. **`staff_vehicle_info.go`'s other, pre-existing discard site mixed units (low, S, pre-existing kindred#2273).** CONFIRMED and fixed with TDD: the unmapped-field-value site (record already created, since `rec` is instantiated before the routing check) was incrementing `Stats.Skipped` per value instead of `Stats.SkippedValues`. Currently silent — all 10 live 2026 SVI fields route — but a new unmapped SVI field answered by e.g. 200 staff would report `skipped: 200+` for zero actual dropped records. `TestLoadPersonCustomValuesCountsUnmappedFields` rewritten first (asserts `len(records)==1`, `Stats.Skipped==0`, `Stats.SkippedValues==1`), confirmed RED (`Stats.Skipped = 1, want 0`), then fixed. Mutation-checked by reverting to `Skipped++` — RED again, restored. Also updated `orchestrator.go`'s doc list (now three services set `SkippedValues`, not two) and its two frontend mirrors for the same reason.

### Local verification (post-fix, post-rebase onto main)
```
$ go build ./...                                          # exit=0
$ go vet ./...                                             # exit=0
$ go test ./...                                            # ok, all packages, 26.067s (sync pkg)
$ go test -race ./sync/                                    # ok  144.708s
$ golangci-lint run ./sync/... --new-from-rev=origin/main   # 0 issues
$ cd frontend && npx --no-install tsc --noEmit              # clean
$ ./node_modules/.bin/eslint <2 touched files>               # clean
$ ./node_modules/.bin/vitest run useSyncCompletionToasts.test.ts  # 16 passed
```
Now superseded by real CI: 21/21 GitHub Actions checks green on `057f256d`.

### GUI impact — flagging for review, not merging on my own judgment
No frontend markup or component code changed (the two touched frontend files are doc-comment-only edits). But this PR changes what **values** flow into the existing admin Sync tab completion toast/stats badge for two services: `staff_applications` and `staff_vehicle_info` runs will now surface non-zero `skipped`/`skipped_values` counts for gated people that were previously silently dropped with no counter at all, and `staff_vehicle_info`'s gate-drop count changes from an inflated per-value number to an accurate per-person one. Per the standing rule, this holds for a look before merge even though no `.tsx`/CSS changed.

**Not merging — holding for review per the GUI-impact rule.**

PR #2385, opened 2026-08-15
test(sync): widen the shared fixture builder to cover bunk_assignments/staff/bunks/bunk_plans

adamflagg:
## What

Widens the shared PocketBase test-fixture builder in `pocketbase/sync/` to cover four more
collections, and migrates the three files that were hand-building them separately over to it —
so kindred#1921's schema-agreement check (`TestLodgingTestsupportFixtureFieldsExistInProductionSchema`)
covers them automatically, with **no edit to the check itself**.

- **Renamed** `newLodgingTestApp` → `newSyncTestApp`, and its file
  `pocketbase/sync/lodging_testsupport_test.go` → `sync_testsupport_test.go`. Pure rename, its
  own commit, no behavior change (`test(sync): rename the lodging fixture builder to newSyncTestApp`).
- **Widened** `newSyncTestApp` with `bunks`, `staff`, `bunk_plans`, `bunk_assignments` (the last
  including production's post-kindred#2259 unique index from migration `1500000155`), exposed as
  four `addXCollection(t, app, ...)` helpers rather than inlined, so callers that also need other,
  out-of-scope collections can still reuse the same field definitions.
- **Migrated** the three files that hand-built these four collections separately:
  `bunk_assignments_protection_test.go`, `bunk_assignments_grain_test.go`,
  `stranded_assignment_cleanup_test.go` — one commit per file, so a wrong migration is revertible
  alone. `persons`/`camp_sessions`/`attendees`/`saved_scenarios`/`bunk_assignments_draft`/
  `lodging_*` stay hand-built in those files; only the four in-scope collections moved.

## Scope, deliberately narrowed (per the issue)

Four collections only: `bunk_assignments`, `staff`, `bunks`, `bunk_plans`. `staff_vehicle_info_test.go`
is **untouched** — its own `staff` fixture stays hand-built; kindred#2277 is editing that file in a
parallel branch and this PR does not touch it. No production `.go` file is touched.

## Issue body was stale — corrected before implementing

Per this campaign's rule, corrected `kindred#2300`'s body (edit + comment on the issue) before
starting:

1. The CI anchor had drifted a **second** time. The 2026-08-14 triage pass's own replacement
   (`ci.yml:611-613`/`:802-806`) no longer measured out — PR #2373 shifted lines after that pass
   ran. Re-derived against `main`@`41dca920`: job `tests-migrations` at `ci.yml:701`, step
   "Fixture schema agreement (kindred#1921)" at `:891`.
[...]

Issue #2300, opened 2026-08-13
chore(tests): widen the shared fixture builder so the #1921 schema-agreement check covers sync fixtures too


adamflagg:
## ⚠️ RE-CORRECTED 2026-08-14 — implementation pass (kindred#2300 PR)

The 2026-08-14 phase-C triage correction below (still visible for history) already fixed one
stale anchor but introduced/kept two more against a tree that had since moved (PR #2373 merged
between the triage pass and this implementation pass, and #2366 deleted `camper_history_test.go`
entirely). Re-verified against `main` @ `41dca920` immediately before starting this issue's PR.

### What was still wrong

1. **CI anchor, again.** The triage correction's replacement (`ci.yml:611-613` / `:802-806`) no
   longer measures out either — PR #2373 (`ci(ci): wire the migration-history guard's orphaned
   self-test into CI`) shifted line numbers. Re-derived: job `tests-migrations` ("Migration Smoke
   Test") starts at `.github/workflows/ci.yml:701` (`name:` at `:702`); the step that actually
   runs the check is "Fixture schema agreement (kindred#1921)" at `:891`, `run:` at `:895`.
2. **The "five test-local builders" survey was never resolvable, at either version.** The triage
   correction's own replacement text (`api_test.go:2620, camper_history_test.go:658,
   sync_runs_test.go:41, and two others`) cites `camper_history_test.go`, which kindred#2366
   deleted outright, and `api_test.go:2620`, which does not contain a
   `core.NewBaseCollection(<variable>)` call in either `lodging/api_test.go` or `sync/api_test.go`
   today. Re-derived by grep (`grep -rn 'core.NewBaseCollection(' --include='*_test.go' . | grep -v
   '"'`): there are **5** such call sites, not 4 (this issue's own prior guess, extrapolating "5
   minus the deleted file = 4", was also wrong — the other 4 cited identifiers never matched
   either):

   - `lodging/hooks_test.go:113` (loop variable `name`, over `lodging_assignments`/
     `lodging_assignments_draft`)
   - `sync/normalize_geographic_test.go:2361` (loop variable `name`, over `custom_field_defs`/
     `person_custom_values`/`geo_overrides`)
   - `sync/orphan_guard_test.go:768` (parameter `target`, a generic orphan-sweep test table)
   - `sync/orphan_sweep_test.go:30` (parameter `collection`, same generic pattern)
   - `sync/sync_runs_test.go:41` (constant `syncRunsCollection`)

[...]

PR #2386, opened 2026-08-15
feat(sync): wire the remaining 12 default sync jobs to DryRunnable

### LINKED ISSUE #2351: feat(sync): wire the remaining 12 default sync jobs to DryRunnable so service=all can dry-run
## What

`dry_run=true` on `/api/custom/sync/run` **without an explicit service** — i.e. the default `service=all` — always returns **HTTP 400**, because 12 of the 24 default unified jobs do not implement `DryRunnable`.

This is the deliberately-accepted remainder of kindred#2334 (PR #2346), not a regression. #2346 wired the 13 services that already carried a working `DryRun` field and an internal skip-write branch. The other 12 have neither, so wiring them is new behaviour per service rather than plumbing.

## Measured

Default job list is `GetDefaultUnifiedSyncJobs` (`pocketbase/sync/orchestrator.go`): **24 jobs** (corrected from an earlier miscount of 25 -- `sed -n '/func GetDefaultUnifiedSyncJobs/,/^}/p' pocketbase/sync/orchestrator.go | grep -oE '"[a-z_]+"' | wc -l` gives 24, not 25). `DryRunnable` implementors after #2346: **13**. Unwired: **12**.

| unwired default job |
|---|
| `session_groups` |
| `sessions` |
| `attendees` |
| `persons` |
| `bunks` |
| `bunk_plans` |
| `bunk_assignments` |
| `staff` |
| `financial_transactions` |
| `person_custom_values` |
| `household_custom_values` |
| `stranded_assignment_cleanup` |

Reproduce:

```bash
for f in pocketbase/sync/*.go; do case "$f" in *_test.go) continue;; esac
  grep -oE 'func \([a-z] \*[A-Za-z0-9_]+\) SetDryRun\(' "$f"; done | sort -u
sed -n '/func GetDefaultUnifiedSyncJobs/,/^}/p' pocketbase/sync/orchestrator.go | grep -oE '"[a-z_]+"'
```

## Why it is not urgent

The current behaviour is **safe and honest**. `UnsupportedDryRunServices` rejects up front and the 400 body names every offending service in both `error` and a machine-readable `unsupported_services` array, so an operator can immediately retry with a supported one:

```
curl -X POST ".../api/custom/sync/run?year=2025&service=family_camp_derived&dry_run=true"
```

Nothing writes on the rejected path. The catastrophic behaviour #2334 fixed — the parameter being accepted, discarded, and a real write performed while the operator believed otherwise — is gone.

## Why it is still worth doing

The campaign's ruling D14 was *"honour `dry_run` end to end … so every service reachable through the unified endpoint supports it. Not a 400."* Accepting partial coverage was an explicit owner decision (2026-08-14) to take it in two steps, with this issue as step two. Until it lands, the most natural invocation of the feature is unusable and the ruling is only half met.

## Scope

Per service, roughly what #2346 did for the other 13: add a `DryRun` field, guard every `App.Save` / `App.Delete` / flush behind it, add `SetDryRun`, add a compile-time `DryRunnable` assertion, and a test that drives the setter through the interface and proves no writes.

**Do not** simply add a no-op `SetDryRun` to clear the 400. #2346's review proved that failure mode is undetectable — satisfying `DryRunnable` is a compile-time property, so a setter that stores nothing type-checks identically to one that works, and the endpoint then answers `200 {"dry_run": true}` while the service runs wet. That is strictly worse than the current 400. `TestRealServicesSetDryRunStoresTheFlag` exists to catch exactly this; extend it rather than working around it.

## Sequencing

- **Files:** the 12 service files above, plus `pocketbase/sync/orchestrator.go` (assertions) and `pocketbase/sync/orchestrator_test.go`.
- **Predecessor:** #2346 (merged) — this builds directly on its `DryRunnable` interface.
- **Collides with:** anything touching those service files. `persons.go`/`bunk_requests.go` (#2347), `bunk_assignments.go` (#2259/#2264) were the collision risk at filing time; all three are now CLOSED, so that risk cleared. `staff.go` does **not** overlap #2277 -- #2277's actual scope is `staff_applications.go`/`staff_vehicle_info.go` only, confirmed against its own PRs (#2380). Re-check open PRs against these files before starting a slice.
- Splittable per service, so it can be taken in slices rather than as one 12-service PR.

adamflagg:
## What

`dry_run=true` on `/api/custom/sync/run` with the default `service=all` returned **HTTP 400**
because 12 of the 24 default unified jobs (corrected from the issue's stale count of 25 --
see the issue comment) did not implement `DryRunnable`. This wires all 12: `session_groups`,
`sessions`, `attendees`, `persons`, `bunks`, `bunk_plans`, `bunk_assignments`, `staff`,
`financial_transactions`, `person_custom_values`, `household_custom_values`,
`stranded_assignment_cleanup` -- exactly #2351's scope table, no more.

**This does not make the default `service=all` fully dry-run-able for the current year.**
`ResolveUnifiedSyncServices` appends `reconcile_request_lifecycle` and `bunk_requests` (and,
under `IS_DOCKER`, `process_requests`) to the 24 defaults whenever the requested year is the
current one -- none of those three were ever in #2351's scope (see the issue's own "Scope"
table), and none implement `DryRunnable`. So a current-year `service=all&dry_run=true` request
still gets HTTP 400 after this PR; only a historical replay (an explicit past `year`) actually
reaches 200 now. `TestCurrentYearDefaultSyncStillRejectsDryRun` (new, `orchestrator_test.go`)
pins the residue so it stays visible rather than reappearing as a surprise. Wiring those three
is future work, tracked by #2351's own successor note, not a gap in this PR.

## Architecture

11 of the 12 embed `BaseSyncService`. Every write those services perform routes through one of
eight call sites inside `BaseSyncService`'s shared helpers (`ProcessSimpleRecord`,
`ProcessSimpleRecordGlobal`, `ProcessCompositeRecord`, `deleteOrphans`,
`DeleteOrphansFromPreloaded`), so the fix lives mostly in one place: a `DryRun bool` field on
`BaseSyncService`, and a guard at each of those eight `App.Save`/`App.Delete` calls that skips
the write and still updates `Stats` as if it had happened.

**A blast-radius bug this uncovered and fixed along the way:** the first version of this change
gave `BaseSyncService` itself an exported `SetDryRun` method, reasoning that 7 of the 12 (the
ones with *no* write sites of their own) would then get `DryRunnable` "for free" through Go's
method promotion. That is true, but promotion doesn't stop at those 7 -- it silently hands
`DryRunnable` to **every** service that embeds `BaseSyncService`, including two outside this
issue entirely: `bunk_requests` (which `ResolveUnifiedSyncServices` appends to a *current-year*
`service=all` request, so it is genuinely reachable through the endpoint this issue is about)
and `request_processor` (whose `Sync()` delegates to a remote FastAPI call that a local bool
cannot gate at all). Both would have compiled as dry-run-safe while running fully wet -- the
exact "200 `dry_run:true`, writes anyway" failure #2334 was about, just relocated. Caught before
it shipped by a new test, `TestEmbeddingBaseSyncServiceDoesNotLeakDryRunnable`.

The fix: `BaseSyncService` keeps the `DryRun` field and the eight gated call sites, but does
**not** export a promoted `SetDryRun`. Each of the 12 wired services declares its own
three-line `SetDryRun` that sets the (possibly-promoted) field directly. A service that embeds
`BaseSyncService` but doesn't declare this method -- `bunk_requests`, `process_requests`,
`custom_field_definitions`, `financial_lookups`, `multi_workbook_export`, `divisions`,
`person_tag_definitions`, `staff_lookups` -- correctly stays unsupported and gets the honest
400. `bunk_requests` is now the test suite's negative-control exemplar in place of the two
services (`session_groups`, `persons`) that used to fill that role before this PR wired them.

Five of the 12 also have write sites of their own, outside those eight shared call sites, gated
individually in their own files:
- `attendees.go`: `logStatusChange`'s `attendee_status_history` insert.
- `persons.go`: the person upsert (`processPerson`), the household upsert
  (`processHouseholdRecord`), an attendee-relation backfill (`updateAttendeeRelations`), a
  household-relation backfill (`updatePersonHouseholdRelations`), and a household-orphan delete
  (`deleteHouseholdOrphans`) -- five independent write areas, each gated and each covered by its
  own real-app test in `persons_dryrun_test.go` (`persons` carries PII across every downstream
  join, so this is the one file in the PR where "same pattern as everywhere else" wasn't
  treated as license to skip a dedicated test).
- `person_custom_field_values.go` / `household_custom_field_values.go`: a fast-path upsert
  (`processPersonCustomFieldValue` / `processHouseholdCustomFieldValue`) that bypasses
  `ProcessSimpleRecord` for a `last_updated`-comparison shortcut.
- `stranded_assignment_cleanup.go` doesn't embed `BaseSyncService` at all -- it's a from-scratch
  `DryRun` field and `dryRun bool` parameter threaded through
  `reconcileStrandedAssignments`/`reconcileLodgingOrphans`, gating the two `app.Save` calls that
  null out a stranded draft's `bunk`/`bunk_plan` or `units`. The issue flagged this one as
  possibly having "no meaningful skip-write branch" -- on inspection it does (two clean,
  independent gate points), so it's wired rather than left out.

## What I deliberately did not do

- **No promoted `SetDryRun` on `BaseSyncService`.** Explained above -- would have silently
  exposed `bunk_requests`/`process_requests` (and six other embedders) as falsely dry-run-safe.
- **No full end-to-end `Sync()` dry-run test for the 11 CampMinder-backed services.** `Sync()`
  on all of them (except `stranded_assignment_cleanup`) makes a live HTTP call via
  `campminder.Client`, and this package has no mock CampMinder server. Instead, each wired
  write site is driven directly (its own function, not through `Sync()`) against a real
  PocketBase test app, asserting real record counts before/after -- the same production code
  `Sync()` calls, just without the network hop `Sync()` adds on top. `stranded_assignment_cleanup`
  needs no CampMinder client, so its test *does* drive the real `Sync()` end-to-end.
- **Did not touch `orchestrator.go`.** Everything needed lived in `base_sync.go`, the 12
  service files, and `orchestrator_test.go` -- no need to enter the `Stats` sub-struct region
  PR #2378 is also touching.

## Issue body correction

Filed a comment and edited #2351's body: it claimed 25 default unified jobs (24 is correct --
9 + 2 custom-value jobs + 13), and its "Collides with" list named three issues (#2347, #2259,
#2264) that are all now closed, plus `staff.go` as overlapping #2277, which #2277's own PR
(#2380) shows is false (#2277 only ever touched `staff_applications.go`/`staff_vehicle_info.go`).

## Mutation-check evidence

Two required by the task, plus every write-guard this PR added was mutation-checked the same
way (confirmed RED, then restored to GREEN):

1. **No-op setter** (`session_groups.go`): replaced `SetDryRun`'s body with `_ = dryRun`.
   `TestRealServicesSetDryRunStoresTheFlag/session_groups` failed:
   `session_groups.SetDryRun(true) left DryRun false -- the unified endpoint would report
   dry_run=true and the service would write for real`.
2. **Write-guard removed** (`base_sync.go`'s `ProcessSimpleRecord` update branch): deleted the
   `if b.DryRun { ...; return nil }` block. `TestBaseSyncServiceProcessSimpleRecordDryRunWritesNothing`
   failed: `dry run persisted a name change: got "Changed", want "Original"`.

Nine more write-guard mutations were run the same way and all went RED: `base_sync.go`'s
`DeleteOrphans`/`DeleteOrphansFromPreloaded`/`ProcessCompositeRecord` guards,
`stranded_assignment_cleanup.go`'s draft-sweep guard, `attendees.go`'s `logStatusChange` guard,
both `person_custom_field_values.go` guards (create+update), `household_custom_field_values.go`'s
update guard, and `persons.go`'s `deleteHouseholdOrphans` / `processHouseholdRecord` /
`processPerson` / `updateAttendeeRelations` guards.

**Review round found two guards this original list missed entirely** -- no test drove either,
so deleting them stayed green:
- `stranded_assignment_cleanup.go`'s `reconcileLodgingOrphans` `units`-clear guard (the lodging
  twin of the draft-sweep guard above, on a *different* write site in the same file). The only
  existing dry-run test seeded zero `lodging_assignments_draft` rows, so the loop it guards
  never ran. Added `TestStrandedAssignmentCleanupDryRunLeavesLodgingDraftUntouched` with a real
  household-orphan fixture; mutation-checked by deleting the guard -- failed with `dry run
  cleared units for a cancelled household: got [], want unchanged [...]`; restored.
- `base_sync.go`'s `ProcessSimpleRecordGlobal` create+update guards -- the file's own header
  comment claimed these were "covered by the sibling tests below," which was false; no sibling
  test exercises this helper. Added
  `TestBaseSyncServiceProcessSimpleRecordGlobalDryRunWritesNothing` (currently pins dead code --
  none of this helper's five callers are wired to `DryRunnable` yet -- so a future service that
  wires one of them doesn't inherit a silently-broken guard) and corrected the comment;
  mutation-checked by commenting out both guards in isolation -- failed with `dry run wrote 2
  rows; want 1`; restored.

Both are now covered, so "every write-guard this PR added was mutation-checked" is accurate as
of this revision.

## Tests

13 new tests across 5 new files (`base_sync_dryrun_test.go`,
`stranded_assignment_cleanup_dryrun_test.go`, `attendees_dryrun_test.go`,
`custom_field_values_dryrun_test.go`, `persons_dryrun_test.go`), plus
`TestEmbeddingBaseSyncServiceDoesNotLeakDryRunnable` in `orchestrator_test.go`, plus the
required extension of `TestRealServicesSetDryRunStoresTheFlag` /
`TestRealServicesHonorDryRunThroughUnifiedEndpoint` / `dryRunCapableRealServices` to the 12
newly-wired names. **Review round added 3 more** to the same files (no new files):
`TestStrandedAssignmentCleanupDryRunLeavesLodgingDraftUntouched` and
`TestBaseSyncServiceProcessSimpleRecordGlobalDryRunWritesNothing` (the two coverage gaps above),
plus `TestCurrentYearDefaultSyncStillRejectsDryRun` in `orchestrator_test.go` (pins the
current-year 400 residue described in "What"). 16 new tests total.

- `cd pocketbase && go build ./...` -- clean
- `go vet ./...` -- clean
- `go test ./sync/...` -- green
- `go test -race ./sync/` -- green (171s)
- `bash scripts/pre-push-verify.sh` -- green after adding two new serial-test entries to
  `main_test_parallelism_test.go`: `TestAttendeesLogStatusChangeDryRunWritesNothing` needs
  `t.Setenv` for `campminder.NewClient` (same reason four existing entries in that file already
  give), and the review-round `TestCurrentYearDefaultSyncStillRejectsDryRun` needs `t.Setenv`
  for `IS_DOCKER`.
- `gofmt -l .` -- clean

## Scope locks respected

- Did not touch `bunk_assignments_protection_test.go`, `bunk_assignments_grain_test.go`, or
  `stranded_assignment_cleanup_test.go` (owned by the parallel #2300-derived work, currently PR
  #2385) -- new fixtures for `stranded_assignment_cleanup` and `bunk_assignments` live in
  `stranded_assignment_cleanup_dryrun_test.go` and `base_sync_dryrun_test.go` instead.
- Did not touch `camper_transportation.go`, `staff_applications.go`, `staff_vehicle_info.go`.
- Did not touch `orchestrator.go`'s `Stats` sub-struct region (PR #2378).

Closes #2351.

## Summary by CodeRabbit

- **New Features**
  - Expanded dry-run support across synchronization processes, including people, households, attendees, assignments, sessions, staff, financial data, and custom fields.
  - Dry runs now report expected creates, updates, and cleanup actions without changing stored records.
  - Status history and assignment cleanup operations remain unchanged during dry runs.

- **Bug Fixes**
  - Prevented unintended database writes during dry-run synchronization.
  - Improved validation for unsupported dry-run operations.

- **Tests**
  - Added comprehensive coverage for simulated results and preserved data.

### DIFF
--- a/pocketbase/main_test_parallelism_test.go
+++ b/pocketbase/main_test_parallelism_test.go
@@ -95,6 +95,14 @@ var serialGroups = []struct {
 		tests:  []string{"TestBuildNormalizationLookupCompositeKeyDedup"},
 	},
 	{
+		// campminder.NewClient reads CAMPMINDER_PRIMARY_KEY from the environment
+		// (same reason as the block below), so building the real client this
+		// test needs for GetSeasonID() costs a t.Setenv.
+		pkg:    "sync",
+		reason: "t.Setenv: campminder.NewClient reads CAMPMINDER_PRIMARY_KEY",
+		tests:  []string{"TestAttendeesLogStatusChangeDryRunWritesNothing"},
+	},
+	{
 		// These four drive deleteOrphans, which reads the season through
 		// s.Client.GetSeasonID(). Client is a concrete *campminder.Client and
 		// campminder.NewClient reads CAMPMINDER_PRIMARY_KEY from the
--- a/pocketbase/main_test_parallelism_test.go
+++ b/pocketbase/main_test_parallelism_test.go
@@ -170,6 +170,7 @@ var serialGroups = []struct {
 			"TestLoadPersonCustomValuesNoDiscardsMeansNoWarnAppLog",
 			"TestLoadPersonCustomValuesNoDiscardsMeansNoWarnLog",
 			"TestLoadPersonCustomValuesRoutesTheFourLive2026FieldsAndDoesNotSkipThem",
+			"TestStrandedAssignmentCleanupDryRunLodgingLogReportsSimulatedSweep",
 		},
 	},
 	{
--- a/pocketbase/main_test_parallelism_test.go
+++ b/pocketbase/main_test_parallelism_test.go
@@ -91,6 +91,11 @@ var serialGroups = []struct {
 	},
 	{
 		pkg:    "sync",
+		reason: "t.Setenv: IS_DOCKER drives ResolveUnifiedSyncServices' process_requests append",
+		tests:  []string{"TestCurrentYearDefaultSyncStillRejectsDryRun"},
+	},
+	{
+		pkg:    "sync",
 		reason: "t.Setenv: API_URL points at a per-test httptest server",
 		tests:  []string{"TestBuildNormalizationLookupCompositeKeyDedup"},
 	},
--- a/pocketbase/sync/attendees.go
+++ b/pocketbase/sync/attendees.go
@@ -41,6 +41,14 @@ func (s *AttendeesSync) Name() string {
 	return "attendees"
 }
 
+// SetDryRun implements the orchestrator's DryRunnable interface (kindred#2351). Declared
+// explicitly rather than inherited by embedding BaseSyncService -- see that field's doc
+// comment on BaseSyncService for why a promoted setter is not safe. Setting it also gates
+// logStatusChange's own App.Save call below, which is outside BaseSyncService's write sites.
+func (s *AttendeesSync) SetDryRun(dryRun bool) {
+	s.DryRun = dryRun
+}
+
 // Sync performs the attendees sync
 func (s *AttendeesSync) Sync(ctx context.Context) error {
 	s.LogSyncStart("attendees")
@@ -413,6 +421,10 @@ func (s *AttendeesSync) logStatusChange(
 	}
 	if personPBID, ok := recordData["person"]; ok {
 		record.Set("person", personPBID)
+	}
+
+	if s.DryRun {
+		return nil
 	}
 
 	if err := s.App.Save(record); err != nil {
--- a/pocketbase/sync/attendees_dryrun_test.go
+++ b/pocketbase/sync/attendees_dryrun_test.go
@@ -1,0 +1,55 @@
+package sync
+
+import (
+	"testing"
+
+	"github.com/pocketbase/pocketbase/core"
+	pbtests "github.com/pocketbase/pocketbase/tests"
+
+	"github.com/camp/kindred/pocketbase/campminder"
+)
+
+// TestAttendeesLogStatusChangeDryRunWritesNothing proves logStatusChange's own
+// App.Save call -- the one write site AttendeesSync has outside
+// BaseSyncService.ProcessCompositeRecord -- is gated by DryRun (kindred#2351).
+func TestAttendeesLogStatusChangeDryRunWritesNothing(t *testing.T) {
+	// Not t.Parallel(): t.Setenv below panics if the test may run in parallel.
+	app, err := pbtests.NewTestApp()
+	if err != nil {
+		t.Fatalf("NewTestApp: %v", err)
+	}
+	t.Cleanup(app.Cleanup)
+
+	col := core.NewBaseCollection("attendee_status_history")
+	col.Fields.Add(&core.NumberField{Name: "person_id"})
+	col.Fields.Add(&core.TextField{Name: "old_status"})
+	col.Fields.Add(&core.TextField{Name: "new_status"})
+	col.Fields.Add(&core.TextField{Name: "detected_at"})
+	col.Fields.Add(&core.NumberField{Name: "year"})
+	if saveErr := app.Save(col); saveErr != nil {
+		t.Fatalf("create attendee_status_history: %v", saveErr)
+	}
+
+	// GetSeasonID is a pure getter -- no network call -- so a real *campminder.Client built
+	// from a fake key is sufficient here.
+	t.Setenv("CAMPMINDER_PRIMARY_KEY", "test-subscription-key")
+	client, err := campminder.NewClient(&campminder.Config{APIKey: "test-key", ClientID: "test-client", SeasonID: 2026})
+	if err != nil {
+		t.Fatalf("campminder.NewClient: %v", err)
+	}
+
+	s := NewAttendeesSync(app, client)
+	s.SetDryRun(true)
+
+	if callErr := s.logStatusChange(5001, 6001, "waitlisted", "enrolled", map[string]any{}); callErr != nil {
+		t.Fatalf("logStatusChange: %v", callErr)
+	}
+
+	rows, err := app.FindRecordsByFilter("attendee_status_history", "", "", 0, 0)
+	if err != nil {
+		t.Fatalf("re-query: %v", err)
+	}
+	if len(rows) != 0 {
+		t.Errorf("dry run wrote %d rows to attendee_status_history; want 0", len(rows))
+	}
+}
--- a/pocketbase/sync/base_sync.go
+++ b/pocketbase/sync/base_sync.go
@@ -57,6 +57,25 @@ type BaseSyncService struct {
 	ProcessedKeys  map[string]bool // Track processed composite keys for orphan detection
 	FieldDiffStats map[string]int  // Track which fields cause updates (for debugging)
 	Debug          bool            // Enable verbose debug logging
+	// DryRun, when true, makes every write helper below (ProcessSimpleRecord,
+	// ProcessSimpleRecordGlobal, ProcessCompositeRecord, deleteOrphans,
+	// DeleteOrphansFromPreloaded) compute its result and update Stats exactly as
+	// a normal run would, but skip the App.Save / App.Delete call that would
+	// persist it (kindred#2351). Gating lives at those eight call sites.
+	//
+	// Deliberately NOT paired with a promoted SetDryRun method here. BaseSyncService is
+	// embedded by several services outside kindred#2351's twelve -- bunk_requests (which
+	// ResolveUnifiedSyncServices appends to "all" for the current year, so it IS reachable
+	// through the unified dry_run endpoint) and request_processor (whose Sync() delegates to
+	// a remote FastAPI call this field cannot gate at all) among them. An exported SetDryRun
+	// on this type would be PROMOTED to every one of those embedders automatically, making
+	// each silently satisfy DryRunnable whether or not its own writes are actually gated --
+	// the exact "compiles, therefore looks safe, but still runs wet" trap kindred#2334 was
+	// about, discovered here by TestEmbeddingBaseSyncServiceDoesNotLeakDryRunnable. Every
+	// service this issue actually wires declares its OWN three-line SetDryRun that sets this
+	// field directly (e.g. session_groups.go); a service that does not declare one stays
+	// correctly unsupported and gets the honest 400.
+	DryRun bool
 }
 
 // NewBaseSyncService creates a new base sync service
@@ -339,6 +358,11 @@ func (b *BaseSyncService) deleteOrphans(
 
 		slog.Info("Deleting orphaned record", "entity", entityName, "name", candidate.name, "id", candidate.key)
 
+		if b.DryRun {
+			orphanCount++
+			continue
+		}
+
 		// Counted AFTER the delete lands. A failed delete leaves the row on disk,
 		// and reporting it as deleted tells an operator the opposite of the truth.
 		if err := b.App.Delete(record); err != nil {
@@ -482,6 +506,11 @@ func (b *BaseSyncService) DeleteOrphansFromPreloaded(
 		if !b.ProcessedKeys[keyStr] {
 			name := b.getRecordName(record, entityName)
 			slog.Info("Deleting orphaned record", "entity", entityName, "name", name, "key", keyStr)
+
+			if b.DryRun {
+				orphanCount++
+				continue
+			}
 
 			// Counted AFTER the delete lands. A failed delete leaves the row on disk,
 			// and reporting it as deleted tells an operator the opposite of the truth.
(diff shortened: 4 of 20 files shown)


adamflagg:
@coderabbitai review

adamflagg:
@coderabbitai review

adamflagg:
## Triage-attack review round: 4 findings, all fixed in-branch

Three findings from the internal scan, one from CodeRabbit — all confirmed real against the tree, all fixed, no issues filed (each was M-or-smaller).

### 1. [internal, HIGH] "What" implied the current-year 400 was fully fixed — it isn't
`ResolveUnifiedSyncServices` appends `reconcile_request_lifecycle` and `bunk_requests` (and, under `IS_DOCKER`, `process_requests`) to the 24 defaults whenever the requested year is the current one. Neither of the first two implements `DryRunnable`, so `service=all&dry_run=true` for the **current year** still gets HTTP 400 after this PR — only a historical replay (`year` != current) actually reaches 200 now. Neither service was ever in #2351's own "Scope" table, so this isn't a gap in the PR's work, but the body's opening paragraph read as claiming the 400 was gone.
- **Fix:** rewrote the "What" section to say so explicitly, and added `TestCurrentYearDefaultSyncStillRejectsDryRun` (`orchestrator_test.go`) pinning the residual 400 via `ResolveUnifiedSyncServices` + `UnsupportedDryRunServices` against the real orchestrator.
- Mutation-check: removed the two appends in `ResolveUnifiedSyncServices` → test failed (`got [], want [reconcile_request_lifecycle bunk_requests]`) → restored, GREEN.

### 2. [internal, MEDIUM] `reconcileLodgingOrphans`' `units`-clear guard had zero test coverage
The only existing lodging dry-run test seeded zero `lodging_assignments_draft` rows, so the guarded loop never ran — deleting the guard would have stayed green.
- **Fix:** added `TestStrandedAssignmentCleanupDryRunLeavesLodgingDraftUntouched` with a real household-orphan fixture (mirrors the wet-run test that proves the same household gets swept).
- Mutation-check: deleted the `if dryRun {...; continue}` block → failed (`dry run cleared units for a cancelled household: got [], want unchanged [...]`) → restored, GREEN.

### 3. [internal, LOW] `ProcessSimpleRecordGlobal`'s two guards had zero coverage, and the file's comment falsely claimed otherwise
`base_sync_dryrun_test.go`'s header said these were "covered by the sibling tests below" — no sibling test exercises this helper at all (none of its 5 callers are wired to `DryRunnable` yet, so the guard is currently dead code from a coverage standpoint).
- **Fix:** corrected the comment, added `TestBaseSyncServiceProcessSimpleRecordGlobalDryRunWritesNothing`.
- Mutation-check: commented out both guards (update + create) in isolation → failed (`dry run wrote 2 rows; want 1`) → restored, GREEN. Confirmed the mutation was scoped to `ProcessSimpleRecordGlobal` alone (sibling `ProcessSimpleRecord`/`ProcessCompositeRecord` tests stayed green throughout).

### 4. [CodeRabbit, on commit 106cfdff] Dry-run lodging sweep always logged `drafts_swept=0`
`reconcileLodgingOrphans`' completion log used the `writes`-only counter (real `App.Save` calls) for `drafts_swept`, so a dry run always logged 0 even when `orphaned_drafts` was nonzero — misleading for an operator previewing the sweep. The sibling bunk-sweep log already gets this right by logging `stats.Updated` instead, but copying that fix here would have double-counted: `Sync()` runs `reconcileStrandedAssignments` before `reconcileLodgingOrphans` against the same shared `*Stats`, so by the time the lodging pass logs, `stats.Updated` already carries the bunk pass's own count too.
- **Fix:** added a function-local `swept` counter (increments in both the dry-run and real-write branches, same as `stats.Updated`, but scoped to this function only) and log that instead of `writes`.
- **Test:** `TestStrandedAssignmentCleanupDryRunLodgingLogReportsSimulatedSweep` (uses `captureSweepLogs`, added to `main_test_parallelism_test.go`'s serial group since it swaps the process-global slog default). Seeds only a lodging orphan (no `bunk_plans`, so the bunk pass sweeps nothing) to isolate the two passes' counts.
- Mutation-check: reverted the log line to `writes` → failed (`dry run logged drafts_swept=0 despite one simulated lodging sweep`) → restored, GREEN.
- Re-requested CodeRabbit on the fix commit (e1b28488) — its incremental review came back clean: "No actionable comments were generated in the recent review."

### Also fixed along the way
`golangci-lint`'s `goconst` flagged `"Original"` (now 3 occurrences across the base_sync dry-run tests) — pulled into `dryRunTestOriginalName`.

### Verification
- `go build ./...`, `go vet ./...`, `gofmt -l .` — clean
- `go test ./...` — green (both rounds)
- `golangci-lint run ./sync/... ./... --new-from-rev=41dca920` — 0 issues
- `bash scripts/pre-push-verify.sh` — ALL CHECKS PASSED (both rounds)
- Both pushes verified by `git fetch && git rev-list --count origin/<branch>..HEAD` == 0
- CI Summary: SUCCESS on the final commit (e1b28488)
- CodeRabbit: real review both times (`Actionable comments posted: 1` on round 1, `No actionable comments` on round 2's incremental pass) — not a rate-limit/skip no-op

No open decision points, no GUI impact (Go-only diff), nothing held back for scope. The PR body's closing reference to #2351 stands as written — it was always scoped to exactly the 12 named jobs, which are now honestly covered end to end.

PR #2387, opened 2026-08-15
fix(frontend): delete dead role=status/aria-live AT scaffolding

### LINKED ISSUE #2379: chore(frontend): audit and delete the remaining AT-only keyboard/ARIA scaffolding
## What

Audit the remaining keyboard/ARIA scaffolding in `frontend/src/` and delete what only exists to serve
assistive technology, **if** doing so actually reduces complexity. Owner's framing, 2026-08-15:

> this tool is for users on a laptop or desktop with a mouse and keyboard, not keyboard only, and I may
> eventually support SOME pages on an iPhone size view

`frontend/CLAUDE.md` § "Accessibility — deliberately minimal" already encodes the policy. This issue is
about the code that predates it.

## Measured first, because the available win is smaller than it looks

Census of `frontend/src/`, non-test files only, 2026-08-15:

| surface | files | occurrences | verdict |
|---|---:|---:|---|
| `aria-*` | 111 | 465 | **mostly load-bearing — do not sweep** |
| `role=` | 52 | 146 | mostly load-bearing |
| `aria-label` | 90 | 230 | mostly load-bearing |
| `sr-only` | 11 | 43 | **candidate, with a trap** |
| `aria-live` / `role="status"` | 4 sites | 8 | **candidate** |
| `<div role="button">` + `tabIndex` + `onKeyDown` | 12 hits | — | **mostly already deleted** |

**Two findings that shrink this issue rather than growing it, and they are the point:**

1. **The `aria-*` bulk is test infrastructure, not accessibility.** The suite addresses the UI through the
   accessibility tree — roughly 1,721 `*ByRole` and 309 `*ByLabelText` calls across ~141 test files. Deleting
   an `aria-label` breaks the query layer far from the edit. The policy already says to *add* `aria-label` /
   `role` when it is the only handle a test can query a control by. So 465 occurrences is not 465 candidates;
   it is mostly the test harness.
2. **The `<div role="button" tabIndex={0} onKeyDown>` → `<button>` conversion is largely done.** Of the 12
   `role="button"` hits, most are **comments recording that the pattern was removed** — in
   `metrics/Css*BarChart.tsx`, `weekend/HouseholdRosterRow.tsx`, `metrics/geo/GeoDetailList.tsx` the string
   survives only in prose explaining why it no longer wraps the element. Live instances are few and need
   reading one at a time.

## What is genuinely on the table

**A. `aria-live` / `role="status"` — 4 sites.** The policy explicitly forbids adding these.

- `components/admin/lodging/UnitCapacityFields.tsx:137` — `role="status" aria-live="polite"` capacity advisory
- `components/admin/lodging/UnitAmenityFieldset.tsx:307` — same shape, sharing advisory
- `components/SocialNetworkGraph.tsx:830` — `role="status"`
- `components/FullPageSpinner.tsx:4` — `role="status"` on the spinner wrapper

The two lodging ones wrap advisory text that is **visible on screen anyway**, so the live region announces
what the user can already read. Deleting the attributes leaves the visible advisory untouched.

**B. `sr-only` — 11 files, 43 occurrences.** ⚠️ **Do not run a classname sweep.** `frontend/CLAUDE.md`
records the trap: `sr-only` on an `<input>` or `<legend>` is a **visually-hidden native control**, not AT
scaffolding — the friend-group colour swatches are real radios hidden behind styled spans, and deleting the
class makes a raw grey radio appear beside every swatch. Each of the 11 needs reading. Note also that
`aria-hidden` and `sr-only` both feed testing-library's accessible-name computation, so removals change what
`getByRole(…, { name })` matches in files you did not touch — change in batches and run the suite between.

**C. Arrow-key handling — 15 files.** Needs per-case judgement, and **most are probably correct**: arrow keys
in a combobox or listbox (`weekend/PlaceFamilyPicker.tsx`, `metrics/DrillDownModal.tsx`) are ordinary
mouse-and-keyboard control behaviour that a sighted user on a laptop uses. What the policy forbids is
**roving-tabindex and arrow-key navigation added as AT scaffolding** on things that are not controls. Only
those qualify.

## What must NOT be touched

`frontend/CLAUDE.md` already rules on these; repeating so a future sweep does not re-derive them:

- **`components/ui/modalStack.ts` is overlay-ordering correctness, not accessibility.** `isTopOverlay()` is
  what stops one Escape closing two stacked overlays, and `acquireBackgroundInert()` sets `inert`, which
  blocks **mouse clicks** through an open dialog. kindred#2205 and kindred#2237 built this deliberately;
  it stays.
- **Escape-to-close and focus *restore* serve everyone**, mouse users included.
- `ui/SortableColumnHeader.tsx`'s `aria-hidden` on the sort arrow is load-bearing — remove it and ↑/↓ folds
  into the accessible name of every sortable column across six tables.
- The four `jsx-a11y` rules in `eslint.config.js` stay on: they protect the query layer, not users.
- `components/weekend/AccessibilityFlagList.tsx` is about families' **housing** needs — wheelchair access,
  power for medical equipment, bathroom proximity. It matches every grep for "accessib" and is unrelated.

## Honest recommendation

**Do A. Do B carefully. Leave C mostly alone.** The realistic win is ~8 attributes plus a bounded `sr-only`
pass — worthwhile as hygiene and for keeping the tree consistent with the stated policy, but it is not a
large complexity reduction, and the largest single block of ARIA in the tree is deliberately staying.

If the mobile/iPhone-size work in the owner's note goes ahead, revisit then: touch targets and viewport
behaviour are a different question from AT scaffolding, and some of what looks deletable today may become
relevant for a small-screen surface.

## Sequencing

- **Files:** the 4 sites in A; the 11 files in B; `frontend/CLAUDE.md` if the policy text needs sharpening.
- **Predecessors:** none. kindred#2205 and kindred#2237 have both merged.
- **Collides with:** any in-flight frontend PR touching `components/weekend/` or `components/admin/lodging/`.
  Sequence against whichever is open.

adamflagg:
## Summary

Deletes the remaining `aria-live`/`role="status"` AT-announcement scaffolding
and one genuine `sr-only` duplicate, per owner ruling: `frontend/CLAUDE.md`
§ "Accessibility — deliberately minimal" — three users, mouse and keyboard,
no AT and none planned. Where the tree contradicted the issue body, this PR
follows the tree and says so below.

## Part A — `role="status"`/`aria-live`

Deleted, visible text left untouched and verified with a new test each:

- `components/admin/lodging/UnitCapacityFields.tsx` — the "Capacity advisory"
  `role="status" aria-live="polite"` wrapper, plus its now-obsolete
  live-region rationale comment.
- `components/admin/lodging/UnitAmenityFieldset.tsx` — the "Sharing advisory"
  wrapper, same treatment.
- `components/SocialNetworkGraph.tsx` — `role="status"` on the graph's
  degrade banner (line moved to 841 from the issue's 830; four PRs landed on
  `main` since the issue was filed).

**Kept, deliberately deviating from the issue: `components/FullPageSpinner.tsx:4`.**
The issue named this line for deletion, but the tree disagrees with the
issue: this file's own regression test (added in kindred#2348, and still
present) documents `role="status"` as the deliberate query handle
`ProtectedRoute.test.tsx`, `AdminRoute.test.ts`, and
`RequirePermission.test.ts` use for "still loading" — three test files, four
assertions, all of them the *only* handle those tests have for this state.
This is exactly the "test's only query handle" case the same issue's own
MUST-NOT-TOUCH list protects. Added a comment on the component recording why,
so a future sweep doesn't re-open it.

## Part B — the 11 files still carrying `sr-only`

Read all 11. Only **4 files carry an actual `sr-only` className** — the
other 7 hits are prose comments (`Tooltip.tsx`, `FamilyCard.tsx`,
`LodgingBoard.tsx`, `LodgingMap.tsx`, plus the two lodging-advisory files
already touched in Part A) narrating *removed* `sr-only` text from prior PRs
(kindred#2219, kindred#2348) — nothing to delete there.

Of the 4 real sites (8 individual `className="sr-only"` occurrences):

| File | Kept/Deleted | Why |
|---|---|---|
| `admin/lodging/BedInventoryEditor.tsx` | **Deleted** | A standalone `<label className="sr-only" htmlFor="bed-type-picker">`, fully redundant with the adjacent `<select aria-label="Add a bed type">` — the `aria-label` alone already satisfies `getByLabelText('Add a bed type')` (7 call sites in `LodgingUnitForm.test.tsx`), confirmed by a new test and a mutation-check (temporarily restored the label, watched the new test fail, removed it again). |
| `MergeRequestsModal.tsx` | Kept | A real native radio hidden behind a styled card `<label>` — `frontend/CLAUDE.md`'s own trap note (`sr-only` on `<input>` is a visually-hidden native control). Deleting the class puts a raw grey radio next to the card. |
| `weekend/FriendGroupActionBar.tsx` (×3: input, name span, legend) | Kept | Same native-radio pattern as `MergeRequestsModal`, plus a `<legend>` — the trap note groups `<legend>` with `<input>` explicitly. |
| `weekend/WeekendFriendGroups.tsx` (×3: input, name span, legend) | Kept | Same pattern, second copy (not shared code). The name span here is also `WeekendFriendGroups.test.tsx`'s literal query handle: `within(card).getByRole('radio', { name: 'Indigo' })`. |

Every kept site now carries a one-line "KEPT (kindred#2379): ..." comment
explaining why, so the next sweep doesn't have to re-derive it.

**Net: 1 of 11 files had real scaffolding to delete.** The other 10 either
had nothing but prose mentioning a long-since-deleted announcement, or held
real native-control markup / test-query handles the issue's own policy
protects. The achievable win here was smaller than the issue's "11 files"
framing suggested — reported honestly rather than manufacturing scope.

## Part C — arrow-key handling

Narrowed the issue's ~15-file grep (which matched lucide `Arrow*` icon
imports and unrelated `tabIndex` hits) down to real
`onKeyDown` + `Arrow*` key handlers: 3 files.

- `weekend/PlaceFamilyPicker.tsx` — a real combobox/listbox typeahead
  (owner ruling, kindred#2080). Arrow-key navigation here is ordinary
  desktop keyboard behavior for a search-select, explicitly built for
  "pointer or keyboard" per its own doc comment — not AT scaffolding.
- `pipeline-debug/PipelineBatchList.tsx` — its `Arrow` grep hits are lucide
  sort-icon imports (`ArrowDown`/`ArrowUp`/`ArrowUpDown`), not keyboard
  handling. Its `onKeyDown` is Enter/Space on a clickable virtualized row —
  not arrow-key navigation, out of this issue's Part C scope.
- `BunkRequestRow.tsx` — same Enter/Space-on-clickable-div shape, no
  arrow-key handling. This *is* the `<div role="button" tabIndex={0}
  onKeyDown={…}>` → `<button>` simplification `frontend/CLAUDE.md`
  separately prefers, but that's a distinct, larger change (event
  propagation, native button styling) this issue didn't ask for — left
  alone rather than expanding scope.

**No roving-tabindex or arrow-key-navigation-as-AT-scaffionding found.**
Part C is a no-op in this PR, as the issue anticipated it might be.

## Verification

- `cd frontend && ./node_modules/.bin/vitest run` — 450 files, 6881 tests,
  all passing.
- `bash scripts/pre-push-verify.sh` — prettier/eslint/tsc/vitest all PASS.
- New tests added: a `role="status"`/`aria-live` absence check on the
  capacity advisory, a source-grep absence check on `SocialNetworkGraph`'s
  degrade banner, a visible-text-on-mount test replacing the deleted
  live-region assertions, and a query-handle test for the bed-type picker
  (mutation-checked).

Closes #2379.

### DIFF
--- a/frontend/src/components/FullPageSpinner.tsx
+++ b/frontend/src/components/FullPageSpinner.tsx
@@ -1,5 +1,10 @@
 import { Loader2 } from 'lucide-react'
 
+// KEPT — deliberately NOT deleted by kindred#2379, despite that issue naming
+// this exact line: `role="status"` is not an AT announcement here (no AT
+// users, per frontend/CLAUDE.md), it is the query handle `ProtectedRoute`,
+// `AdminRoute`, and `RequirePermission`'s tests use for "still loading" — see
+// this file's own regression test (kindred#2348) and its comment.
 export const FullPageSpinner = () => (
   <div className="flex min-h-screen items-center justify-center" role="status">
     <Loader2 className="text-primary h-12 w-12 animate-spin" />
--- a/frontend/src/components/MergeRequestsModal.tsx
+++ b/frontend/src/components/MergeRequestsModal.tsx
@@ -189,6 +189,11 @@ export default function MergeRequestsModal({
                     : 'border-border hover:border-primary/50'
                 }`}
               >
+                {/* KEPT (kindred#2379): a real native radio, not AT
+                    scaffolding — `sr-only` hides its default browser
+                    rendering while the card `<label>` stays the visible,
+                    clickable control. Deleting the class puts a raw radio
+                    button next to the card. */}
                 <input
                   type="radio"
                   name="targetRequest"
--- a/frontend/src/components/SocialNetworkGraph.test.ts
+++ b/frontend/src/components/SocialNetworkGraph.test.ts
@@ -112,4 +112,10 @@ describe('SocialNetworkGraph header layout — slim single row', () => {
     // EdgeFilters import should be removed since the component is no longer used
     expect(source).not.toMatch(/\bEdgeFilters\b/)
   })
+
+  it('has no aria-live/role="status" scaffolding on the degrade banner (kindred#2379)', () => {
+    // Nobody here uses AT (frontend/CLAUDE.md "Accessibility — deliberately
+    // minimal") — the banner's visible text is the whole story.
+    expect(source).not.toContain('role="status"')
+  })
 })
--- a/frontend/src/components/SocialNetworkGraph.test.ts
+++ b/frontend/src/components/SocialNetworkGraph.test.ts
@@ -117,5 +117,6 @@ describe('SocialNetworkGraph header layout — slim single row', () => {
     // Nobody here uses AT (frontend/CLAUDE.md "Accessibility — deliberately
     // minimal") — the banner's visible text is the whole story.
     expect(source).not.toContain('role="status"')
+    expect(source).not.toContain('aria-live')
   })
 })
--- a/frontend/src/components/SocialNetworkGraph.tsx
+++ b/frontend/src/components/SocialNetworkGraph.tsx
@@ -837,10 +837,7 @@ export default function SocialNetworkGraph({ sessionCmId }: SocialNetworkGraphPr
             </div>
 
             {degradeBanner && (
-              <div
-                role="status"
-                className="border-border flex items-center gap-2 border-b bg-amber-50 px-4 py-2 text-sm text-amber-800 dark:bg-amber-950/30 dark:text-amber-200"
-              >
+              <div className="border-border flex items-center gap-2 border-b bg-amber-50 px-4 py-2 text-sm text-amber-800 dark:bg-amber-950/30 dark:text-amber-200">
                 <span aria-hidden>⚠</span>
                 {degradeBanner}
               </div>
--- a/frontend/src/components/admin/lodging/BedInventoryEditor.tsx
+++ b/frontend/src/components/admin/lodging/BedInventoryEditor.tsx
@@ -95,9 +95,9 @@ export function BedInventoryEditor({ beds, onChange }: BedInventoryEditorProps) 
       {/* In the same wrap as the chips, so adding a bed reads as extending the
           set rather than operating a separate control below it. */}
       <li className="inline-flex items-center gap-1.5">
-        <label className="sr-only" htmlFor="bed-type-picker">
-          Add a bed type
-        </label>
+        {/* No standalone <label> — redundant with the select's own
+            aria-label below, which is already the query handle tests use
+            (kindred#2379). */}
         <select
           id="bed-type-picker"
           aria-label="Add a bed type"
--- a/frontend/src/components/admin/lodging/LodgingUnitForm.test.tsx
+++ b/frontend/src/components/admin/lodging/LodgingUnitForm.test.tsx
@@ -627,6 +627,15 @@ describe('LodgingUnitForm — the parent picker reflects a live allocation chang
 })
 
 describe('LodgingUnitForm — beds', () => {
+  it('names the bed-type picker with aria-label alone, no separate sr-only label (kindred#2379)', () => {
+    const { container } = render(
+      <LodgingUnitForm areas={AREAS} units={[]} year={2026} onSaved={vi.fn()} onCancel={vi.fn()} />
+    )
+
+    expect(container.querySelector('label[for="bed-type-picker"]')).toBeNull()
+    expect(screen.getByLabelText('Add a bed type')).toBeInTheDocument()
+  })
+
   it('shows the suggested occupancy from the bed inventory', async () => {
     const user = userEvent.setup()
     render(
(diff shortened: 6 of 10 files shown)


adamflagg:
## Triage close-out

Reviewed the reviewer's one finding, verified it against the current tree, and fixed it in-branch.

**Finding (confirmed, fixed):** `SocialNetworkGraph.test.ts`'s guard test `'has no aria-live/role="status" scaffolding on the degrade banner (kindred#2379)'` only asserted the absence of `role="status"`, not `aria-live`, despite the title claiming both.

Mutation-check transcript:
1. Re-added `aria-live="polite"` to the degrade banner's `<div>` (`SocialNetworkGraph.tsx:840`) — ran the test with the **original** assertion (only `role="status"` checked): `Tests 13 passed (13)`. Confirmed the gap is real — the regression sails through green.
2. Reverted the mutation. Added `expect(source).not.toContain('aria-live')` to the test.
3. Re-ran against clean code: `13 passed (13)` — green as expected.
4. Re-applied the `aria-live="polite"` mutation: `1 failed | 12 passed (13)`, failing on the new assertion. Confirmed RED.
5. Reverted the mutation again, re-ran: `13 passed (13)` — GREEN, test file is the only diff.

No other findings — everything else in the reviewer's report was independently re-verified (FullPageSpinner keep, sr-only file/occurrence counts, both friend-group mutation checks, Part C conclusion) and checked out as claimed; the 4 discarded items (comment-count-off-by-one, a squash-message typo, a note about #2379's own body being uncorrected, and a "kept for consistency" span) are cosmetic/non-actionable and were left as-is.

Full frontend suite still green (450 files / 6881 tests) and `scripts/pre-push-verify.sh` all-passing after the fix.

PR #2388, opened 2026-08-15
fix(sync): give wrapper sites a typed error to classify their own rejections

### LINKED ISSUE #2292: refactor(sync): wrapper sites cannot classify their own failures — typed errors so Rejected can reach them
## ⚠️ CORRECTED 2026-08-14 (second pass) — one Scope site no longer exists

`pocketbase/sync/households.go` was **DELETED** by #2290 (merged today): the whole `HouseholdsSync`
service was unreachable dead code — never registered, never constructed, not reachable by service-name
string. Its only live symbol, `serviceNameHouseholds`, was relocated to `persons.go` and renamed
`householdsCollection`.

**So the `ProcessSimpleRecord` census below is five counting sites, not six.** Drop
`households.go:169`; the remaining five are `bunks.go`, `financial_transactions.go`,
`session_groups.go`, `staff.go` and `sessions.go`. The mixed-callee count (5) is unchanged; only the
call-site total moves from 10 to 9.

**Every other line number in this body predates ~18 PRs merged on 2026-08-14** and should be
re-derived rather than trusted. The argument stands; the anchors have moved.

---

> The argument is sound and the sequencing blocker is gone - #2295 is CLOSED and its orphan-sweep guard shipped in #2299, so this can start now. But the Scope list is wrong in both directions: five of the entries it names are NOT mixed, and the real census is 5 mixed callees reached by 10 counting call sites. The campaign plan's two specific claims both check out - Stats.Rejected is blocked rather than declined, and orchestrator.go:300-302 still says 'Nothing writes this counter yet' while 32 sites write it. That stale comment is on main today and needs an owner.

1. WRONG (Sequencing): 'this issue depends on kindred#2295. There is no point routing a failure to Rejected while a rejection still costs the record its row.' REPLACEMENT: 'kindred#2295 is CLOSED and its orphan-sweep guard shipped in PR #2299 (a0d88f8c) - see BaseSyncService.skipSweepForRejections at base_sync.go:181-220 and the guard.Rejected accounting at :286. 32 sites now write Stats.Rejected. This issue is unblocked and can start.'
2. WRONG (What shipped instead): 'Nothing has been reclassified yet... #2293 ships the machinery only... with every site still on Stats.Errors.' REPLACEMENT: 'That was true of #2293. Reclassification then landed in #2299: 32 sites across 13 files now increment Stats.Rejected, and the sweep guard makes it safe. What has NOT moved is the ten call sites this issue is about, and rejection_sites_test.go:44-52 records exactly why - they stay loud until this issue gives them typed errors.'
3. WRONG (Scope list). Current text names 'attendees.go (processEnrollment; not processAttendee), bunk_assignments.go (processAssignment), bunk_plans.go (processBunkPlan), bunk_requests.go (processRow), persons.go (processPerson, processHouseholdRecord), household_custom_field_values.go and person_custom_field_values.go (syncXCustomFieldValues)' plus 'ProcessSimpleRecord / ProcessSimpleRecordGlobal'. REPLACEMENT, re-derived against 3eb44ba4: 'Genuinely mixed - 5 callees, 10 counting call sites: processEnrollment (attendees.go:281, rejection :289; call site :203, Errors++ :222); processAssignment (bunk_assignments.go:417, rejections :424/:429/:434/:439; call site :228, Errors++ :230); processRow (bunk_requests.go:204, rejections :208/:213; call site :140, Errors++ :142); processPerson (persons.go:421, rejection :460; call site :329, Errors++ :331); ProcessSimpleRecord (base_sync.go:614, rejections :624/:634) reached by six counting sites - bunks.go:109, financial_transactions.go:149, households.go:169, session_groups.go:103, staff.go:173, sessions.go:251. NOT mixed, remove from scope: processBunkPlan (bunk_plans.go:304, only createBunkPlan's error at :402); processHouseholdRecord (persons.go:1424, only App.Save and collection lookup); syncPersonCustomFieldValues / syncHouseholdCustomFieldValues (:294 / :289 - their inner processXCustomFieldValue now counts rejections in place as Rejected++ and returns nil, shipped in #2320, so only API-fetch and collection errors remain); ProcessSimpleRecordGlobal (base_sync.go:725, no transform rejections at all). processAttendee (attendees.go:172, call site :121) stays out for the reason already given - it is a pure rejection and needs no sentinel.'
4. WRONG (Consequence to expect): 'until kindred#2295 lands, one malformed upstream record reaching any site - wrapper or not - fails that service's run... Note this is currently broader than wrapper sites alone.' REPLACEMENT: 'As of #2299 this is narrowed back to the ten call sites listed in Scope. Everywhere else a malformed record now increments Stats.Rejected, which is warn-only and skips the collection's orphan sweep. At the ten remaining sites a malformed upstream record still fails the run, because applyCompletionStatus (orchestrator.go:368-382) fails on any non-zero Errors. That is the intended interim behaviour and the first occurrence should be triaged, not suppressed.'
5. WRONG (second issue comment, 2026-08-13T18:12:49Z): 'PR #2299's census file (rejection_sites_test.go) carries a deferral comment naming only processAttendee and ProcessSimpleRecord. processPerson is being added to it in that PR.' REPLACEMENT: it was NOT added. rejection_sites_test.go:50-51 on main still reads 'wrapper sites whose callee returns both classes (processAttendee, ProcessSimpleRecord)'. That comment is wrong on two counts and should be fixed with this issue: processAttendee does not return both classes (this issue's own first correction established that), and processPerson, processAssignment and processRow are missing.

- Stats.Rejected++ sites on main, non-test: 32, across 13 files (staff_lookups 6, financial_lookups 4, sessions 3, bunks 3, session_groups 2, person_tag_definitions 2, person_custom_field_values 2, household_custom_field_values 2, financial_transactions 2, divisions 2, custom_field_definitions 2, staff 1, persons 1). Command: grep -rn 'Rejected++' pocketbase/ --include=*.go | grep -v _test.go
- Stats.Errors++ sites on main, non-test: 68 across 29 files. Command: grep -rn 'Errors++' pocketbase/ --include=*.go | grep -v _test.go
- WRAPPER SITE CENSUS, which is what this issue's Scope list should say: 5 genuinely mixed callees (processEnrollment, processAssignment, processRow, processPerson, ProcessSimpleRecord) reached by 10 counting call sites (4 bespoke + 6 through the shared helper). Denominator: every Errors++ site whose immediately preceding line is a call to a function that returns BOTH a transform rejection and an App.Save error, read function-by-function.
- 5 entries in the current Scope list are NOT mixed: processAttendee, processBunkPlan, processHouseholdRecord, the two syncXCustomFieldValues, and ProcessSimpleRecordGlobal. Of these, processAttendee is a pure rejection (needs no sentinel, can move straight to Rejected) and the other four are pure infrastructure (stay on Errors permanently).

</details>

---

kindred#2284 split `Stats.Errors` into two counters with opposite tolerances: `Errors` for infrastructure failures (zero tolerance, fails the run) and `Rejected` for per-record transform failures (warn-only for its first season). Classifying the 84 sites to implement that split surfaced a category the issue's model does not have.

## The premise that does not hold everywhere

#2284 assumes each `Stats.Errors++` site is *either* infrastructure *or* transform, and that reading the preceding lines tells you which. For roughly 52 sites that is exactly right — the line above is a bare `transformXToPB` or a bare `App.Save`, and the classification is unambiguous.

For the **wrapper** sites it is not, because the function being called returns both classes of error through one return value:

```go
// attendees.go — processEnrollment both validates upstream data AND writes to SQLite
for _, enrollmentData := range sessionStatuses {
    // ...
    if err := s.processEnrollment(personCMID, enrollment, existingAttendees); err != nil {
        slog.Error("Error processing enrollment", "person_cm_id", personCMID, "error", err)
        s.Stats.Errors++
    }
}
```

`processEnrollment` returns `fmt.Errorf("invalid or missing SessionID")` — unambiguously a rejected upstream record — and *also* returns the result of `ProcessCompositeRecord`, which performs the record's `App.Save`. One counter bump at that call site cannot distinguish a locked database from a malformed record.

> **Correction (2026-08-13):** an earlier revision of this body used `processAttendee` as the example here. That was wrong, and the distinction matters for whoever implements this. `processAttendee` returns **only** `fmt.Errorf("invalid or missing PersonID")` or `nil` — it counts and swallows `processEnrollment`'s errors itself, in its own loop, and then returns `nil` regardless. So the outer call site (`attendees.go:113`) is **unambiguously a rejection** and needs no typed error at all; it can move straight to `Rejected` once the sweep guard lands. The genuinely mixed site is the inner one shown above. Targeting the outer wrapper would add a sentinel where none is needed while leaving the real ambiguity in place.

This is not confined to bespoke wrappers. The shared helper has the same shape:

```go
// base_sync.go — ProcessSimpleRecord, alongside its App.Save path
return fmt.Errorf("recordData missing required 'year' field")
return fmt.Errorf("'year' field must be numeric, got %T", yearValue)
```

So the sites reached through `ProcessSimpleRecord` / `ProcessSimpleRecordGlobal` inherit it too.

## What shipped instead

**Nothing has been reclassified yet.** An earlier revision of this section said the escalation PR reclassified 27 unambiguous sites. It attempted to, then reverted them: CodeRabbit found that a rejected record skips `TrackProcessedKey` via the same `continue` that bumps the counter, so `DeleteOrphans` deletes that record's existing row in the same run. Moving a site to a warn-only counter would have made that deletion *quiet*.

So kindred#2293 shipped the escalation machinery only — `Stats.Rejected` as a declared field, `applyCompletionStatus`, the `SubStats` sum, and the Sync tab badge — with **every** site left on `Stats.Errors`. Reclassification moved to kindred#2295, bundled with the orphan-sweep guard that makes it safe. The two are one semantic unit: rejecting a record must stop deleting its row before any site is allowed to reject.

That leaves the wrapper problem this issue describes exactly where it was, and it is still the right reasoning for wrapper sites specifically. Failing loud on a malformed upstream record is noisy but recoverable and *informative*; the alternative it replaces is a silent green checkmark, which is the defect the whole kindred#2257 campaign is about. It also mirrors the campaign's logic for `Rejected` being warn-only: nothing has ever persisted sync stats, so there is no evidence to classify these from. Failing loud generates that evidence — the first run a wrapper site turns red tells us which one, on what data, and whether it deserves `Rejected`.

⚠️ **Consequence to expect:** until kindred#2295 lands, one malformed upstream record reaching *any* site — wrapper or not — fails that service's run. That is the intended behaviour for now, but it is a real change from today's silence and the first occurrence should be triaged, not reflexively suppressed. Note this is currently broader than wrapper sites alone, because the reclassification was reverted.

## Fix direction

Give the wrappers a way to say which kind of failure they had, then let the call site pick the counter. A sentinel or typed error is the smallest change that works:

```go
var errRejectedRecord = errors.New("record rejected")

// in the wrapper:
return fmt.Errorf("%w: invalid or missing SessionID", errRejectedRecord)

// at the call site:
if errors.Is(err, errRejectedRecord) {
    s.Stats.Rejected++
} else {
    s.Stats.Errors++
}
```

Do this per wrapper, with a test per site proving both branches route to the counter they should. Prefer that over pushing the counter increment down into the wrapper — the call site is where the loop's `continue` already lives, and moving the counting away from it splits one decision across two places.

**Sequencing:** this issue depends on kindred#2295. There is no point routing a failure to `Rejected` while a rejection still costs the record its row.

**Do not** do this as one sweeping change. The wrappers do not share an error vocabulary today, and a uniform codemod would be exactly the kind of large diff with no defect behind it that kindred#2257 warns against for the already-safe sites.

## Scope

Sites to work through, all in `pocketbase/sync/`:

- bespoke wrappers — `attendees.go` (**processEnrollment**; *not* `processAttendee`, see the correction above), `bunk_assignments.go` (processAssignment), `bunk_plans.go` (processBunkPlan), `bunk_requests.go` (processRow), `persons.go` (processPerson, processHouseholdRecord), `household_custom_field_values.go` and `person_custom_field_values.go` (syncXCustomFieldValues)
- shared helper — `base_sync.go`'s `ProcessSimpleRecord` / `ProcessSimpleRecordGlobal` and every site reaching them

Re-derive the list against `main` before starting rather than trusting this one; the campaign has already had two issue bodies go stale against a moving tree.

Part of kindred#2257

adamflagg:
## What

kindred#2284 split `Stats.Errors` (infrastructure, zero tolerance) from `Stats.Rejected`
(per-record data quality, warn-only). Five wrapper functions couldn't participate: their
callee returns both classes of failure through one `error`, so the call site's single
counter bump could not tell a locked database from a malformed record.

This gives each of the five a shared sentinel, `errRejectedRecord` (`base_sync.go`), and
wraps it on the wrapper's own pre-write validation paths. Every `App.Save` /
`FindCollectionByNameOrId` failure inside the wrapper stays unwrapped, so it is still
`Errors` by default -- a wrapper that grows a new failure path is safe unless it
deliberately opts a check into the sentinel. Call sites branch with
`errors.Is(err, errRejectedRecord)`.

**All 5 mixed callees and all 9 call sites are done:**

- `processEnrollment` (attendees.go) -- 1 call site, in `processAttendee`
- `processAssignment` (bunk_assignments.go) -- 1 call site, in `Sync`
- `processRow` (bunk_requests.go) -- 1 call site, in `RunSync`
- `processPerson` (persons.go) -- 1 call site, in `processBatchPersons`
- `ProcessSimpleRecord` (base_sync.go) -- 5 call sites: bunks.go (`Sync`),
  financial_transactions.go (`SyncForYear`), session_groups.go (`Sync`), staff.go
  (`syncStaff`), sessions.go (`Sync`)

(The issue's census said "10 counting call sites" for these five; this PR's own earlier
drafts of this section also said 10 in two places despite the reconciliation paragraph
right here saying 9 -- both were wrong. It's 9: 4 singular + 5 for `ProcessSimpleRecord`.)

`rejection_sites_test.go`'s stale comment is corrected: it previously said the mixed
wrappers were `processAttendee, ProcessSimpleRecord`, which was wrong on both counts --
`processAttendee` does not return both classes (it counts and swallows
`processEnrollment`'s errors itself and is a pure rejection needing no sentinel), and
`processPerson`, `processAssignment`, `processRow` were missing entirely.

## What is, and is NOT, wrapped as a rejection

Wrapped (`errRejectedRecord`) -- the wrapper's own pre-write validation of raw
per-record data, matching the issue's worked example:

- `processEnrollment`: missing/invalid `SessionID`
- `processAssignment`: missing `PersonID` / `SessionID` / `BunkID` / `BunkPlanID`
- `processRow`: missing/invalid `PersonID` in the CSV row
- `processPerson`: missing person `ID`
- `ProcessSimpleRecord`: missing/non-numeric `year` in `recordData`

Left unwrapped (stays `Errors`) -- deliberately, not by omission:

- Every `App.Save` / `App.Delete` / `FindCollectionByNameOrId` failure in all five wrappers
  and in the sibling helper `ProcessCompositeRecord` (used by `processEnrollment` and
  `processAssignment`, not itself one of the five). These are local SQLite operations,
  the census's existing zero-tolerance rule.
- `PopulateRelations`' "required relation not found" (`attendees.go`,
  `bunk_assignments.go`). Both call sites reach it only after already confirming the
  referenced session/person/bunk exists via a `valid*CMIDs` check, so a miss here means
  an internal inconsistency between that check and the live `LookupRelation` lookup, not
  routine upstream data quality -- closer to an invariant violation than a rejection.
- `ProcessSimpleRecord`'s two `year`-validation errors are wrapped for correctness, but
  are defensive rather than currently reachable: all five call sites build `year` locally
  from `s.Client.GetSeasonID()`, never from raw upstream data, before calling in. Typed
  correctly regardless -- a function's error classification isn't scoped to what its
  current callers happen to pass.

## Verification

- Failing test first for every wrapper (compile failure on the undefined sentinel, then a
  real assertion failure), then implementation, then green. `pocketbase/sync/rejection_wrapper_test.go`
  is new, 12 tests: two tests per wrapper driving the wrapper function itself -- the real
  entry point, not a helper -- with (a) malformed input, asserting
  `errors.Is(err, errRejectedRecord)`, and (b) valid input against a `failingSaveApp` that
  forces `App.Save` to fail, asserting the opposite. Two wrappers get a third test:
  `processRow` for its separately-validated invalid-vs-missing `PersonID` cases, and
  `ProcessSimpleRecord` for its non-numeric-year branch.
- Mutation-checked all 5: reverted each sentinel wrap in turn, confirmed the matching test
  went RED with the expected message, restored, confirmed GREEN.
- `rejection_sites_test.go`'s structural AST census (`TestPerRecordRejectionsUseTheRejectedCounter`,
  `TestNoUnexpectedSiteUsesTheRejectedCounter`) scans every literal `.Errors++`/`.Rejected++`
  in the package and classifies it by the `slog` call directly above it. All 9 new call
  sites are literal `if/else` blocks (no shared helper -- a helper would make the AST
  scanner report the bump as unclassifiable) with new `"Rejected ..."` messages added to
  `expectedRejectSites()`. This proves each call site's counter bump sits under the right
  log line -- but not that the `errors.Is` condition sending control to it has the right
  polarity (see "Review fixes" below for why that's a separate, and real, gap).
- `go build ./...`, `go vet ./...`, `gofmt -l`: clean.
- `golangci-lint run ./...` (CI-only, not in pre-push, run directly since this diff touches
  ten files): caught two real issues from the new imports -- persons.go had a pre-existing
  local `errors := 0` counter that the new `errors` package import now shadowed (renamed to
  `errCount`), and the test file's `failingSaveApp.Save` delegate returned the pocketbase
  interface's error unwrapped (`wrapcheck`; now `fmt.Errorf("delegating Save: %w", err)`,
  matching the existing `flakyCollectionApp` pattern in
  `bunk_assignments_protection_test.go`). Both fixed; 0 issues now.
- `go test ./...`: full suite green, including `TestSyncAndLodgingTestsRunInParallel` /
  `TestSerialTestExemptionsAreAllStillNeeded` (package main) -- the 4 new tests that build a
  real `*campminder.Client` via `t.Setenv` were added to the existing
  `"t.Setenv: campminder.NewClient reads CAMPMINDER_PRIMARY_KEY"` serial group in
  `main_test_parallelism_test.go`, alphabetically; the other 8 run parallel.
- `go test -race ./sync/`: clean.

## Safety precondition (kindred#2295, confirmed before relying on it)

`BaseSyncService.skipSweepForRejections` (base_sync.go ~252) and the `guard.Rejected`
accounting (~332) are present on `main` -- a rejected record no longer costs its row to
the orphan sweep for every sweep that routes through `DeleteOrphans` /
`DeleteOrphansGuarded`, which is what makes routing genuine data-quality failures to
`Stats.Rejected` safe for those sweeps.

`BunkRequestsSync`'s two sweeps, `purgeOrphanedRequests` and `purgeZombieBRs`, do not --
they predate #2292 and are hand-rolled rather than routed through that machinery, so they
never picked up the `Rejected` guard for free. Before this PR's `processRow` change, a
malformed CSV row (e.g. an Excel-mangled `PersonID` cell) bumped `Stats.Errors`, which
`applyCompletionStatus` fails the whole run on -- so an operator saw the deletion that
followed. Reclassifying that same row onto `Stats.Rejected` (warn-only, deliberately
absent from `applyCompletionStatus`) removed that visibility: the rejected row's person
never reaches `s.csvPersonIDs`, so `purgeOrphanedRequests` reads their still-current,
still-valid `original_bunk_requests` rows as orphaned and deletes them, and the run
reports success. `orphan_guard.go`'s own doc predicted exactly this failure mode for a
"future reclassification into one of [the hand-rolled sweeps]." Fixed here by gating
`purgeOrphanedRequests` on `s.skipSweepForRejections("bunk_requests", nil)`, the same
check the guarded sweeps get automatically; `purgeZombieBRs` is already gated
transitively, since `RunSync` only calls it when `purgeOrphanedRequests` returned a
non-nil person set. Pinned by two new tests in `bunk_requests_rejection_sweep_test.go`:
one seeding a real, current OBR row and asserting it survives a run with
`Stats.Rejected > 0`, and a regression guard proving a clean run's genuine orphan is
still purged. Mutation-checked: reverting the guard reproduces the deletion and the skip
test fails with the exact expected message; restored, both tests green.

A second, independent gap: the AST census above (`rejection_sites_test.go`) proves a
`"Rejected ..."` log line sits above a `Stats.Rejected++` bump, but never reads the
`if errors.Is(err, errRejectedRecord)` condition that sends control there -- so a
polarity swap (`if errors.Is(...)` flipped to `if !errors.Is(...)`) moves both branches'
bodies together and the census stays green reading whichever line still sits above the
bump. Measured: inverting the condition at `bunks.go`'s call site alone left
`go test ./sync/` green with only this PR's existing tests. New file
`sentinel_branch_polarity_test.go` adds `TestSentinelBranchesRouteToTheRightCounter`,
which walks every `errors.Is(_, errRejectedRecord)` branch in the package via AST and
asserts the true side bumps `Rejected` and the false side bumps `Errors` (or vice versa
under negation) -- independent of message text, so it also catches a swap at
`attendees.go`, `persons.go` and `bunk_requests.go`, all three verified the same way.

Closes #2292.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved synchronization reporting by distinguishing rejected or invalid records from processing failures.
  * Rejected records now appear as warnings and are counted separately, while operational failures remain errors.
  * Added validation handling for missing or malformed enrollment, assignment, request, person, session, staff, and transaction data.
  * Prevented orphan-request cleanup when rejected records make the imported data incomplete.
* **Tests**
  * Added coverage to verify rejection tracking, failure classification, and safe orphan cleanup.

### DIFF
--- a/pocketbase/main_test_parallelism_test.go
+++ b/pocketbase/main_test_parallelism_test.go
@@ -105,7 +105,13 @@ var serialGroups = []struct {
 		// test needs for GetSeasonID() costs a t.Setenv.
 		pkg:    "sync",
 		reason: "t.Setenv: campminder.NewClient reads CAMPMINDER_PRIMARY_KEY",
-		tests:  []string{"TestAttendeesLogStatusChangeDryRunWritesNothing"},
+		tests: []string{
+			"TestAttendeesLogStatusChangeDryRunWritesNothing",
+			"TestProcessAssignment_MissingPersonID_IsRejected",
+			"TestProcessAssignment_SaveFailure_IsInfraError",
+			"TestProcessEnrollment_MissingSessionID_IsRejected",
+			"TestProcessEnrollment_SaveFailure_IsInfraError",
+		},
 	},
 	{
 		// These four drive deleteOrphans, which reads the season through
--- a/pocketbase/sync/attendees.go
+++ b/pocketbase/sync/attendees.go
@@ -3,6 +3,7 @@ package sync
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log/slog"
 	"strconv"
@@ -209,25 +210,34 @@ func (s *AttendeesSync) processAttendee(
 		}
 
 		if err := s.processEnrollment(personCMID, enrollment, existingAttendees); err != nil {
-			// duplicateSessions is keyed by SessionID and holds every entry that shares a
-			// SessionID with another entry in this attendee's array — first, second, or
-			// later; it doesn't identify which one is "the duplicate". So this only tells
-			// us the group fact: this SessionID appears more than once. A (person, session)
-			// key collision — overwriting the first entry on an idempotent upsert, or, on a
-			// first-ever sync before existingAttendees has an entry for the key, taking the
-			// create branch and hitting idx_attendees_unique — is one plausible cause of
-			// this failure, but not the only one; this entry could equally be failing for a
-			// reason unrelated to the collision. Tag it so the possible collision is legible
-			// in the logs rather than the error reading as an unexplained local DB fault.
-			if sessionIDFloat, ok := enrollment["SessionID"].(float64); ok &&
+			// processEnrollment now distinguishes its two failure classes with
+			// errRejectedRecord (kindred#2292): a malformed enrollment (e.g. a
+			// missing SessionID) never reaches App.Save, so the idx_attendees_unique
+			// collision diagnostic below -- which is specifically about a DB write
+			// colliding -- only applies to the non-rejection branch.
+			if errors.Is(err, errRejectedRecord) {
+				slog.Warn("Rejected enrollment", "person_cm_id", personCMID, "error", err)
+				s.Stats.Rejected++
+			} else if sessionIDFloat, ok := enrollment["SessionID"].(float64); ok &&
 				duplicateSessions[int(sessionIDFloat)] != nil {
+				// duplicateSessions is keyed by SessionID and holds every entry that shares a
+				// SessionID with another entry in this attendee's array — first, second, or
+				// later; it doesn't identify which one is "the duplicate". So this only tells
+				// us the group fact: this SessionID appears more than once. A (person, session)
+				// key collision — overwriting the first entry on an idempotent upsert, or, on a
+				// first-ever sync before existingAttendees has an entry for the key, taking the
+				// create branch and hitting idx_attendees_unique — is one plausible cause of
+				// this failure, but not the only one; this entry could equally be failing for a
+				// reason unrelated to the collision. Tag it so the possible collision is legible
+				// in the logs rather than the error reading as an unexplained local DB fault.
 				slog.Error("Error processing enrollment (this SessionID appears more than "+
 					"once in this attendee's SessionProgramStatus — see kindred#2263)",
 					"person_cm_id", personCMID, "session_cm_id", int(sessionIDFloat), "error", err)
+				s.Stats.Errors++
 			} else {
 				slog.Error("Error processing enrollment", "person_cm_id", personCMID, "error", err)
+				s.Stats.Errors++
 			}
-			s.Stats.Errors++
 		}
 	}
 
@@ -294,7 +304,7 @@ func (s *AttendeesSync) processEnrollment(
 	// Extract session ID
 	sessionID, ok := enrollment["SessionID"].(float64)
 	if !ok {
-		return fmt.Errorf("invalid or missing SessionID")
+		return fmt.Errorf("%w: invalid or missing SessionID", errRejectedRecord)
 	}
 	sessionCMID := int(sessionID)
 
--- a/pocketbase/sync/base_sync.go
+++ b/pocketbase/sync/base_sync.go
@@ -4,6 +4,7 @@ package sync
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"log/slog"
 	"reflect"
@@ -14,6 +15,20 @@ import (
 	"github.com/camp/kindred/pocketbase/campminder"
 	"github.com/pocketbase/pocketbase/core"
 )
+
+// errRejectedRecord is the sentinel kindred#2292 gives to wrapper functions whose
+// callee can fail two different ways -- a per-record transform rejection
+// (malformed or missing upstream data) or an infrastructure failure (a local
+// App.Save/App.Delete that did not complete) -- through the same `error` return.
+//
+// A wrapper marks a rejection by wrapping this sentinel: `fmt.Errorf("%w: reason",
+// errRejectedRecord)`. Its call site then branches on `errors.Is(err,
+// errRejectedRecord)` to choose Stats.Rejected (warn-only, kindred#2284/#2295) over
+// Stats.Errors (zero tolerance). An error that reaches the call site WITHOUT this
+// sentinel is always infrastructure -- that is the default, not an opt-in -- so a
+// wrapper that adds a new failure path is safe by construction unless it
+// deliberately wraps errRejectedRecord.
+var errRejectedRecord = errors.New("record rejected")
 
 // RelationConfig defines configuration for populating a relation field
 type RelationConfig struct {
@@ -669,10 +684,17 @@ func (b *BaseSyncService) ProcessSimpleRecord(
 	existingRecords map[any]*core.Record,
 	compareFields []string, // Optional: specific fields to check for updates
 ) error {
-	// Extract year from recordData - required for year isolation
+	// Extract year from recordData - required for year isolation. A missing or
+	// malformed year means recordData itself is malformed -- a rejection, not an
+	// infrastructure failure -- so it's wrapped in errRejectedRecord for the call
+	// site (kindred#2292). Every one of ProcessSimpleRecord's five callers builds
+	// "year" locally from s.Client.GetSeasonID(), never from raw upstream data, so
+	// this path is defensive today rather than reachable; it is typed correctly
+	// regardless, for the same reason a function's error return is not scoped to
+	// what its current callers happen to pass.
 	yearValue, ok := recordData["year"]
 	if !ok {
-		return fmt.Errorf("recordData missing required 'year' field")
+		return fmt.Errorf("%w: recordData missing required 'year' field", errRejectedRecord)
 	}
 
 	var year int
@@ -682,7 +704,7 @@ func (b *BaseSyncService) ProcessSimpleRecord(
 	case int:
 		year = v
 	default:
-		return fmt.Errorf("'year' field must be numeric, got %T", yearValue)
+		return fmt.Errorf("%w: 'year' field must be numeric, got %T", errRejectedRecord, yearValue)
 	}
 
 	// Use composite key for lookup to ensure year isolation
@@ -913,7 +935,12 @@ func (b *BaseSyncService) ProcessCompositeRecord(
 	existingRecords map[string]*core.Record,
 	skipFields []string, // Fields to skip during comparison (like "year" for idempotency)
 ) error {
-	// Extract year from recordData - required for year isolation
+	// Extract year from recordData - required for year isolation. Deliberately
+	// NOT wrapped in errRejectedRecord: ProcessCompositeRecord is not one of
+	// kindred#2292's five named mixed wrappers (processEnrollment and
+	// processAssignment are, and both call this as their final write step), so
+	// this stays out of scope for this issue and keeps its pre-existing
+	// classification (Errors).
 	yearValue, ok := recordData["year"]
 	if !ok {
 		return fmt.Errorf("recordData missing required 'year' field")
--- a/pocketbase/sync/bunk_assignments.go
+++ b/pocketbase/sync/bunk_assignments.go
@@ -3,6 +3,7 @@ package sync
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log/slog"
 
@@ -215,8 +216,13 @@ func (s *BunkAssignmentsSync) Sync(ctx context.Context) error {
 				assignmentData["SessionID"] = float64(sessionID)
 
 				if err := s.processAssignment(assignmentData, existingAssignments); err != nil {
-					slog.Error("Error processing assignment", "error", err)
-					s.Stats.Errors++
+					if errors.Is(err, errRejectedRecord) {
+						slog.Warn("Rejected assignment", "error", err)
+						s.Stats.Rejected++
+					} else {
+						slog.Error("Error processing assignment", "error", err)
+						s.Stats.Errors++
+					}
 				}
 			}
 		}
@@ -565,25 +571,26 @@ func (s *BunkAssignmentsSync) processAssignment(
 	assignmentData map[string]any,
 	existingAssignments map[string]*core.Record,
 ) error {
-	// Extract required fields
+	// Extract required fields. A missing field here is malformed upstream data,
+	// not an infrastructure failure -- kindred#2292.
 	personID, ok := assignmentData["PersonID"].(float64)
 	if !ok {
-		return fmt.Errorf("missing PersonID")
+		return fmt.Errorf("%w: missing PersonID", errRejectedRecord)
 	}
 
 	sessionID, ok := assignmentData["SessionID"].(float64)
 	if !ok {
-		return fmt.Errorf("missing SessionID")
+		return fmt.Errorf("%w: missing SessionID", errRejectedRecord)
 	}
 
 	bunkID, ok := assignmentData["BunkID"].(float64)
 	if !ok {
-		return fmt.Errorf("missing BunkID")
+		return fmt.Errorf("%w: missing BunkID", errRejectedRecord)
 	}
 
 	bunkPlanID, ok := assignmentData["BunkPlanID"].(float64)
 	if !ok {
-		return fmt.Errorf("missing BunkPlanID")
+		return fmt.Errorf("%w: missing BunkPlanID", errRejectedRecord)
 	}
 
 	personCMID := int(personID)
--- a/pocketbase/sync/bunk_requests.go
+++ b/pocketbase/sync/bunk_requests.go
@@ -326,6 +326,18 @@ func findOrphanedPersonIDs(csvPersonIDs map[int]bool, existingOBRPersonIDs []int
 // Called after CSV sync to clean up data from campers who have cancelled/unenrolled.
 // Returns the set of OBR person cm_ids (post-purge) for use by the zombie BR sweep.
 func (s *BunkRequestsSync) purgeOrphanedRequests(year int) (map[int]bool, error) {
+	// This sweep is hand-rolled rather than routed through
+	// BaseSyncService.DeleteOrphansGuarded, so it does not pick up
+	// OrphanSweepGuard's Rejected arm for free -- orphan_guard.go's own doc
+	// warns this is exactly the gap a hand-rolled sweep falls into. A row
+	// this run's processRow rejected (kindred#2292) never reaches
+	// s.csvPersonIDs, which makes its still-current requester look identical
+	// to one who genuinely cancelled. Skip explicitly rather than purge on an
+	// incomplete set (kindred#2295's precondition, applied by hand here).
+	if s.skipSweepForRejections("bunk_requests", nil) {
+		return nil, nil
+	}
+
 	if len(s.csvPersonIDs) == 0 {
 		slog.Info("No CSV persons tracked, skipping orphan purge")
 		return nil, nil
--- a/pocketbase/sync/bunk_requests.go
+++ b/pocketbase/sync/bunk_requests.go
@@ -5,6 +5,7 @@ import (
 	"crypto/md5" //nolint:gosec // G501: MD5 used for change detection, not security
 	"encoding/csv"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"log/slog"
 	"os"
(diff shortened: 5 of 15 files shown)


adamflagg:
@coderabbitai review

adamflagg:
## Review findings — all fixed in-branch, nothing filed

Verified all three reviewer findings against the tree directly (not the PR body) before acting. All three are M-or-smaller, so per policy they were fixed here rather than filed.

**1. HIGH/S — `purgeOrphanedRequests` silently purged a rejected row's still-current data.**
Confirmed real: `BunkRequestsSync`'s two sweeps (`purgeOrphanedRequests`, `purgeZombieBRs`) predate #2292 and are hand-rolled rather than routed through `BaseSyncService.DeleteOrphansGuarded`, so they never picked up `OrphanSweepGuard`'s `Rejected` arm — exactly the gap `orphan_guard.go`'s own doc warns a "future reclassification into one of [the hand-rolled sweeps]" would hit. Before this PR's `processRow` change, a malformed CSV row bumped `Stats.Errors`, which fails the whole run via `applyCompletionStatus`, so an operator saw the deletion that followed. After #2292 reclassified that row onto warn-only `Stats.Rejected`, the person's real `original_bunk_requests` rows got read as orphaned (their id never reached `s.csvPersonIDs`) and purged, with the run now reporting success.
Fix: gate `purgeOrphanedRequests` on `s.skipSweepForRejections("bunk_requests", nil)`, matching the guarded sweeps (`purgeZombieBRs` is already gated transitively — `RunSync` only calls it when `purgeOrphanedRequests` returns a non-nil set). New tests in `bunk_requests_rejection_sweep_test.go`: `TestPurgeOrphanedRequests_SkipsWhenRunHadRejections` (seeds a real, current OBR row, asserts it survives a run with `Stats.Rejected=1`) and `TestPurgeOrphanedRequests_StillPurgesWithoutRejections` (regression guard — a clean run's genuine orphan is still purged). Mutation-checked: removing the guard reproduces the exact `got 0 OBRs, want 1` failure; restored, both green.

**2. MEDIUM/M — nothing pinned the `errors.Is` branch polarity at the 9 new call sites.**
Confirmed real, independently: inverted `bunks.go`'s `if errors.Is(err, errRejectedRecord)` to `if !errors.Is(...)` and ran `go test ./sync/` — clean pass, no failures, on the code as it stood before this fix. The existing AST census (`rejection_sites_test.go`) only reads the `slog` message directly above each counter bump; a polarity swap moves the log line and the bump together, so the census can't see it. Confirmed the reviewer's simultaneous-inversion claim too: flipping `attendees.go`, `bunk_requests.go` and `persons.go` at once also left the full suite green.
Fix: new `sentinel_branch_polarity_test.go` / `TestSentinelBranchesRouteToTheRightCounter` walks every `errors.Is(_, errRejectedRecord)` `if` in the package via AST and asserts the true side bumps `Rejected` and the false side bumps `Errors` (handling `attendees.go`'s `else if` collision-diagnostic chain, and negation). Mutation-checked against all 4 call sites (bunks.go alone, then attendees/bunk_requests/persons together) — each reproduces the exact "does not bump Stats.X" failure; restored, green.

**3. LOW/XS — PR body said "10 call sites" while its own reconciliation paragraph and the tree said 9, and undercounted the new tests by one.**
Confirmed: the diff adds exactly 9 call sites and `processRow` also has a 3rd test (invalid vs. missing `PersonID`), which the original "5 wrappers, 12 tests, ProcessSimpleRecord gets a third" phrasing didn't add up to. Since the PR body is the permanent squash commit message, fixed both counts directly in the body (`gh pr edit --body-file`) rather than leaving the correction in a comment — this is the same class of stale-identifier issue CLAUDE.md's "A Wrong Issue Body Gets Fixed, Not Worked Around" section calls out, applied to a PR body instead of an issue body. Also expanded the "Safety precondition" section to document finding #1's gap and fix, since that's new information about this PR's actual safety story that belongs in the permanent record.

## Verification after fixes
`go build ./...`, `go vet ./...`, `gofmt -l .`, `golangci-lint run ./sync/...` (0 issues), `go test ./...` (all packages), `go test -race ./sync/` (233s, clean), and the parallelism guard (`TestSerialTestExemptionsAreAllStillNeeded`, `TestSyncAndLodgingTestsRunInParallel`) all pass. No `t.Setenv` or process-global swap was added, so `main_test_parallelism_test.go`'s `serialGroups` needed no changes.

`gh api graphql` confirms `closingIssuesReferences` already resolves to #2292 — no body edit was needed for that.

Requested `@coderabbitai review` explicitly (the automatic check on this PR was a "Review skipped — auto reviews disabled" no-op notice rendering green, not a real review).

adamflagg:
## CodeRabbit's "Merge Risk: Moderate" note — investigated, not holding

CodeRabbit's walkthrough flagged: *"Malformed bunk assignments may still bypass rejected-record accounting, which can leave valid existing assignments eligible for orphan cleanup and deletion; rejection logs also include the raw PersonID value."*

Traced both halves against the actual code rather than taking the note at face value:

**Orphan-cleanup claim — not accurate.** `BunkAssignmentsSync.deleteOrphans` (bunk_assignments.go:832) already routes through `BaseSyncService.DeleteOrphansGuarded`, and that shared path (`base_sync.go` `deleteOrphans`, ~line 329) unconditionally does `guard.Rejected = b.Stats.Rejected` before the delete loop and returns early via `skipSweepForRejections` whenever `Rejected > 0` — the comment right above it is explicit that this "can never reach the delete loop past this point," independent of the collapse-ratio Check. `processAssignment`'s four sentinel checks (missing PersonID/SessionID/BunkID/BunkPlanID, bunk_assignments.go:576-592) all `return` before `TrackProcessedCompositeKey`, and `s.Stats.Rejected++` fires at the call site (line 221) before `deleteOrphans` runs later in `Sync`. This is the same mechanism this PR's own PR body already confirmed present for the three bespoke rejecting services, `bunk_assignments.go` included. This is exactly the class of gap finding #1 in the earlier review caught for `bunk_requests.go` — but `bunk_assignments.go` doesn't have it, because unlike `bunk_requests.go`'s two hand-rolled sweeps, it was already routed through the guarded machinery before this PR.

**Raw-PersonID-in-logs claim — real, but pre-existing and out of this PR's diff.** `processRow`'s `"invalid PersonID: %s"` message (bunk_requests.go) is unchanged text from `main` (confirmed via `git show main:pocketbase/sync/bunk_requests.go`) — this PR only wraps it with `errRejectedRecord`, it does not add the value to the log or change what gets logged. Logging `person_cm_id` (a CampMinder record ID, not a name) in server-side `slog` output is an established, deliberate pattern at dozens of other reject sites in this same package (`attendees.go`, `financial_lookups.go`, `staff_lookups.go`, etc.) — it isn't something #2292 introduces, and it isn't in the diff this PR is responsible for. Not filing: it's neither new nor large enough to justify an issue on its own, and litigating an existing codebase-wide logging convention is outside this PR's scope.

Not holding the PR on either half.

PR #2389, opened 2026-08-15
fix(sync): compare before writing in staff_applications, staff_vehicle_info, camper_transportation

### LINKED ISSUE #2384: bug(sync): three syncs rewrite every record every run — staff_applications, staff_vehicle_info, camper_transportation
## ⚠️ CORRECTED 2026-08-15 — it is THREE syncs, not two

The owner reproduced the same non-idempotency on **`camper_transportation`** while validating PR #2378:

> yeah so i cant validate w/transpo, also not idempotent

Verified: `pocketbase/sync/camper_transportation.go` has a bare `updated++` at `:811` and **no
comparison helper anywhere in the file** — no `needsUpdate`, no `fieldEquals`, no
`compareRecordNeedsUpdate`. Same defect, same shape, third instance.

So the affected set is **`staff_applications`, `staff_vehicle_info`, and `camper_transportation`**, and
the fix must cover all three. That matters here more than usual: `camper_transportation` and
`staff_applications` are the two syncs kindred#2356 called *"deliberately in step with each other"*, and
kindred#2383 already records that fixing two of three re-introduces the divergence the template exists
to prevent. Treat all three together.

---

## What

Three sync services **rewrite every existing record on every run**, reporting them all as `updated`
even when nothing changed. Found by the owner while trying to validate PR #2378:

> the staff app sync appears to not be idempotent and always returns 274 upd, for example, per run on
> both main dev and the worktree. staff vehicles on both also shows 200 upd, 29 skip

Reproduced: consecutive runs against an unchanged upstream report the same non-zero `updated` count
every time.

## Why

`upsertRecords` (`pocketbase/sync/staff_applications.go`, the loop from ~:960) fetches the existing
record, calls `record.Set(...)` for **every** field unconditionally, saves, and increments `updated`.
**There is no comparison against the stored values at any point.** `staff_vehicle_info.go` and
`camper_transportation.go` have the same shape.

## This is a gap, not the house style — 8 of the sync services already do it correctly

Services carrying a change comparison today:

`base_sync.go` · `family_camp_derived.go` · `normalize_geographic.go` · `persons.go` ·
`camper_dietary.go` · `financial_aid_applications.go` · `household_demographics.go` · `staff_skills.go`

And a **shared helper already exists**: `compareRecordNeedsUpdate(existing, newData, compareFields)` at
`pocketbase/sync/base_sync.go:1020`. `staff_skills.go` — a direct sibling of these, same custom-field
ingest shape — has its own `recordNeedsUpdate` at `:656`. `camper_dietary.go` likewise has one at
`:496`, which is notable because `camper_transportation.go` is its closest structural twin and has none.

## Why it matters

1. **`updated` is not a signal.** An operator cannot tell a run that changed nothing from one that
   changed everything, which is precisely the "silent green checkmark" problem the kindred#2257
   campaign is about. **It also blocked on-screen validation of PR #2378 twice** — the owner could not
   confirm the counter split on either affected sync, because every run looks identical.
2. **Every row's `updated` timestamp churns**, so `updated` cannot be used to find genuinely-changed
   rows, and any future consumer that watches it sees hundreds of changes a run.
3. **Wasted writes** — on the order of 474 unnecessary `Save()` calls per full sync across
   `staff_applications` and `staff_vehicle_info` alone, before counting `camper_transportation`.

## Scope

Adopt the existing comparison in all three services rather than writing a fourth one:

- Prefer `compareRecordNeedsUpdate` (`base_sync.go:1020`) if its `compareFields` shape fits; otherwise
  mirror the closest sibling — `staff_skills.go:656` for the two staff syncs, `camper_dietary.go:496`
  for `camper_transportation`.
- Skip the `Save()` when nothing changed, and do **not** count it as `updated`.
- ✅ **An unchanged record counts as `Stats.Skipped`. Settled 2026-08-15 by the house pattern, not a
  guess** — both services that already do this comparison agree: `camper_dietary.go:581`
  (`skipped++` with a "Skipping unchanged dietary record" debug log) and `staff_skills.go:267`
  (`s.Stats.Skipped++` in the `else` of its `recordNeedsUpdate` branch). Follow them.

  This is now coherent in a way it was not a day ago: kindred#2356 (merged) moved discarded field
  values onto `Stats.SkippedValues`, which frees `Stats.Skipped` to mean records — and a record that
  needed no write is a record. It also directly answers the symptom that started this issue: a no-op
  run will report `0 updated, N skipped` instead of `N updated`, so an operator can tell the two apart
  at a glance.

## Test

Run each sync twice against a fixed fixture and assert the second run reports `updated == 0`. Drive the
real `Sync()` entry point — asserting on a helper proves nothing here. Mutation-check by removing the
comparison and confirming the second-run assertion goes red.

## Relationship to kindred#2301

kindred#2301 surveys duplicated `fieldEquals`-style helpers and recommends extracting one. This issue is
the opposite problem — three services with **no** helper at all — and does not need #2301 to land first.
If #2301 does land first, adopt whatever it extracts.

## Sequencing

- **Files:** `pocketbase/sync/staff_applications.go`, `pocketbase/sync/staff_vehicle_info.go`,
  `pocketbase/sync/camper_transportation.go`, and their tests.
- **Predecessors:** PR #2378 (merged). PR #2380 (kindred#2277) touches the two staff files — sequence
  after it. kindred#2383 touches `staff_vehicle_info.go` and `camper_transportation.go` — same lane.
- **Collides with:** kindred#2277, kindred#2383.

adamflagg:
## The defect

`staff_applications.go`, `staff_vehicle_info.go` and `camper_transportation.go` all fetched each
existing record, called `record.Set(...)` for every field unconditionally, saved, and incremented
`Stats.Updated` — with no comparison against stored values anywhere in any of the three. A run
against completely unchanged CampMinder data rewrote and reported every row as "updated," every
single time.

Re-measured against the production snapshot (`pocketbase/pb_data/data-prod.db`, read-only) for the
2026 season:

```sql
select year, count(*) from staff_applications group by year order by year;
select year, count(*) from staff_vehicle_info group by year order by year;
select year, count(*) from camper_transportation group by year order by year;
```

- `staff_applications`: **274** rows for 2026 — matches the issue's figure.
- `staff_vehicle_info`: **200** rows for 2026 — matches the issue's figure.
- `camper_transportation`: **1,667** rows for 2026 — the issue said "likewise" without a number;
  this is the real one. Before this fix a no-op nightly sync on this table alone rewrote all 1,667
  rows.

## What changed

Reused the existing shared helper, per the spec — no fourth comparison implementation:

- **`base_sync.go`**: `compareRecordNeedsUpdate(existing, newData, compareFields)` (already used by
  `staff_skills.go` and `camper_dietary.go`) is now also the comparison for these three.
- **`staff_applications.go`**: added `staffApplicationsCompareFields` (44 App-*/Position-Preference
  columns plus `staff`, excluding the `person_id`/`year` key), a `recordNeedsUpdate` delegate, and
  rebuilt `upsertRecords` around a `data map[string]any` so the same map both drives the comparison
  and the write. `upsertRecords` now returns `(created, updated, skipped, errCount)`.
- **`staff_vehicle_info.go`**: same shape — `staffVehicleInfoCompareFields` (10 SVI-* columns plus
  `staff`), same `data`-map refactor, same 4-value return.
- **`camper_transportation.go`**: same shape — `camperTransportationCompareFields` (13 BUS-* columns
  plus `attendee` and `used_legacy_fields`, excluding the `person_id`/`session_id`/`year` key).

**An unchanged record now counts as `Stats.Skipped`**, matching both existing comparing siblings
(`camper_dietary.go` and `staff_skills.go`) — this is the settled decision from the spec, not
re-litigated here. Consequence, stated plainly, and corrected from an earlier draft of this body
that undercounted the two staff syncs (see PR review): `Stats.Skipped` in `staff_applications.go`
and `staff_vehicle_info.go` is `+=`'d twice per run — once for the kindred#2277 staff-row gate
(`gatedPeople`) and once for this PR's no-write count (`upsertSkipped`) — so a no-op run flips
from `274 updated` to `0 updated, 296 skipped` for staff applications (274 no-write + 22 gated,
both re-measured against the production snapshot) and from `200 updated` to `0 updated, 206
skipped` for staff vehicles (200 + 6 gated). `camper_transportation.go` has no second
`Stats.Skipped` writer, so it is the one place the naive `1,667 updated` -> `0 updated, 1,667
skipped` holds exactly. That is the point either way: a run that changed nothing is now
distinguishable from one that rewrote everything.

**`Stats.Skipped` carries two record-level meanings in the two staff syncs**, and this PR is not the
first thing to add a second meaning to it — `staff_applications.go`/`staff_vehicle_info.go` already
gained a staff-row-gate meaning under kindred#2277 (merged as this PR was rebasing onto it, see
below). Both meanings count whole records, so the unit stays consistent, and a comment sits at each
site saying so. `camper_transportation.go` has no gate-drop counter, so `Stats.Skipped` carries
exactly one meaning there.

## Rebase note

PR #2380 (kindred#2277 + kindred#2383) merged to `main` while this branch was in progress and
touches the same two files (`staff_applications.go`, `staff_vehicle_info.go`) — it restructured the
staff-row-gate skip counting from a bare `continue`/`s.Stats.Skipped++` into a `gatedPeople` set
counted once per person. Rebased onto `origin/main` after it merged; the one resulting conflict (in
`staff_vehicle_info.go`, both branches touching the same gate `continue` block) was resolved by
keeping #2380's `gatedPeople` structure and updating this PR's comment to describe the two Skipped
meanings against the new code shape rather than the old one.

## Deliberately NOT done

- Did not write a fourth comparison helper — reused `compareRecordNeedsUpdate` per the spec.
- Did not touch `camper_dietary.go` or `staff_skills.go` (already idempotent) beyond reading them as
  the reuse targets.
- Did not touch the staff-row gate itself, RBAC, or any other sync service.

## Tests

Per service, one test drives the real `Sync()` entry point twice against a fixed fixture (not
`upsertRecords` directly — spec is explicit that driving the helper "proves nothing"):
`TestStaffApplicationsSyncSecondRunIsIdempotent`,
`TestStaffVehicleInfoSyncSecondRunIsIdempotent`,
`TestCamperTransportationSyncSecondRunIsIdempotent`. Each builds on the shared `newSyncTestApp`
fixture (custom_field_defs, persons, person_custom_values, staff, attendees) plus a local collection
builder carrying every column the comparison checks — a column missing from that builder would make
every field read back as `nil`, which would make the assertion pass for the wrong reason (nothing
comparable) instead of the right one.

First run: `Created=1, Updated=0, Errors=0`. Second run against the identical fixture:
`Created=0, Updated=0, Errors=0, Skipped=1`.

### Mutation-check evidence (`go test ./sync/ -run <test>`, restored after each)

| File | Mutation | Test | Result |
|---|---|---|---|
| `camper_transportation.go` | replaced the `recordNeedsUpdate` branch with `if false` | `TestCamperTransportationSyncSecondRunIsIdempotent` | RED — `created=0 updated=1 errors=0, want ... updated=0` |
| `staff_vehicle_info.go` | same | `TestStaffVehicleInfoSyncSecondRunIsIdempotent` | RED — same shape |
| `staff_applications.go` | same | `TestStaffApplicationsSyncSecondRunIsIdempotent` | RED — same shape |

All three restored; full suite green afterward.

**Review follow-up (kindred#2384 PR review):** the three tests above only pinned the skip
direction — nothing asserted that a genuinely changed record still gets written, so a mutant that
makes `recordNeedsUpdate` always return `false` (i.e. never update again) passed the full suite.
Added a third run to each test that edits the seeded `person_custom_values` row via a new
`updatePersonCustomValue` helper and asserts `Updated=1` plus the new value on the re-fetched
record. Mutation-checked the same way — `if false` in place of each `recordNeedsUpdate` body — and
confirmed RED before restoring:

| File | Mutation | Test | Result |
|---|---|---|---|
| `staff_applications.go` | `recordNeedsUpdate` -> `return false` | third run of `TestStaffApplicationsSyncSecondRunIsIdempotent` | RED — `created=0 updated=0 errors=0, want ... updated=1` |
| `staff_vehicle_info.go` | same | third run of `TestStaffVehicleInfoSyncSecondRunIsIdempotent` | RED — same shape |
| `camper_transportation.go` | same | third run of `TestCamperTransportationSyncSecondRunIsIdempotent` | RED — same shape |

All three restored; `go test ./sync/ -run 'IsIdempotent$'` green afterward.

## Verification

```
$ go build ./...                # exit=0
$ go vet ./...                  # exit=0
$ go test ./...                 # ok, all packages
$ go test -race ./sync/         # ok  github.com/camp/kindred/pocketbase/sync  238.123s
$ go test . -run 'Parallel|Serial|Exempt'   # PASS (my three new tests call t.Parallel()
                                             # and touch no process globals, so no
                                             # serialGroups entry needed)
```

Pushed and rebased onto `origin/main` (post-#2380); remote ref confirmed moved via
`git fetch -q && git rev-list --count origin/feature/sync-idempotency-2384..HEAD` = `0`.

## GUI impact

These counts render in the admin Sync tab toast and the per-service badges
(`useSyncCompletionToasts.ts`, `useSyncStatusAPI.ts`). After this PR, a no-op nightly run on a
season with no CampMinder changes will show these three services' badges reading **0 updated** with
a nonzero **skipped** count, instead of the previous behavior of reporting every stored row as
"updated" on every run regardless of whether anything actually changed. Concretely, for the 2026
season as re-measured against the production snapshot: staff applications shows **0 updated, 296
skipped** (not 274 — see the corrected figure above), staff vehicles **0 updated, 206 skipped**
(not 200), and camper transportation **0 updated, 1,667 skipped**.

Closes #2384.

### DIFF
--- a/pocketbase/sync/camper_transportation.go
+++ b/pocketbase/sync/camper_transportation.go
@@ -234,9 +234,13 @@ func (s *CamperTransportationSync) Sync(ctx context.Context) error {
 	slog.Info("Loaded existing records", "count", len(existingRecords))
 
 	// Step 5: Upsert records
-	created, updated, errors := s.upsertRecords(ctx, records, existingRecords, year)
+	created, updated, skipped, errors := s.upsertRecords(ctx, records, existingRecords, year)
 	s.Stats.Created = created
 	s.Stats.Updated = updated
+	// Stats.Skipped counts records that needed no write (kindred#2384). This
+	// service has no field-level staff-gate skip like the two staff syncs, so
+	// unlike them Skipped carries exactly one meaning here.
+	s.Stats.Skipped += skipped
 	s.Stats.Errors = errors
 
 	// Step 6: Delete orphans
@@ -706,6 +710,32 @@ func makeTransportationKey(personID, sessionID, year int) string {
 	return fmt.Sprintf("%d:%d|%d", personID, sessionID, year)
 }
 
+// camperTransportationCompareFields lists the fields to compare for
+// idempotency checks (kindred#2384). Excludes the unique key fields
+// (person_id, session_id, year) since loadExistingRecords already matched on
+// them, and PocketBase-managed fields. Includes `attendee` even though it is
+// not part of the key -- unlike person_id/session_id/year, it can change
+// independently of the key if the attendee mapping is rebuilt.
+var camperTransportationCompareFields = []string{
+	"attendee",
+	colToCampMethod, colFromCampMethod,
+	colDropoffName, colDropoffPhone, colDropoffRelationship,
+	colPickupName, colPickupPhone, colPickupRelationship,
+	colAltPickup1Name, colAltPickup1Phone, colAltPickup1Relationship,
+	colAltPickup2Name, colAltPickup2Phone,
+	"used_legacy_fields",
+}
+
+// recordNeedsUpdate checks if any compared field differs between existing
+// record and new data. Uses compareFields (inclusion list): only the listed
+// fields are checked for changes. Delegates to the shared
+// compareRecordNeedsUpdate in base_sync.go.
+func (s *CamperTransportationSync) recordNeedsUpdate(
+	existing *core.Record, newData map[string]any, compareFields []string,
+) bool {
+	return compareRecordNeedsUpdate(existing, newData, compareFields)
+}
+
 // loadExistingRecords loads existing camper_transportation records for a year
 func (s *CamperTransportationSync) loadExistingRecords(ctx context.Context, year int) (map[string]string, error) {
 	result := make(map[string]string) // compositeKey -> PB ID
@@ -748,22 +778,44 @@ func (s *CamperTransportationSync) upsertRecords(
 	records map[string]*camperTransportationRecord,
 	existingRecords map[string]string,
 	year int,
-) (created, updated, errors int) {
+) (created, updated, skipped, errors int) {
 	col, err := s.App.FindCollectionByNameOrId("camper_transportation")
 	if err != nil {
 		slog.Error("Error finding camper_transportation collection", "error", err)
-		return 0, 0, len(records)
+		return 0, 0, 0, len(records)
 	}
 
 	for _, rec := range records {
 		select {
 		case <-ctx.Done():
-			return created, updated, errors
+			return created, updated, skipped, errors
 		default:
 		}
 
 		key := makeTransportationKey(rec.personID, rec.sessionID, year)
 		existingID, exists := existingRecords[key]
+
+		// Build data map for comparison and for the eventual write.
+		data := map[string]any{
+			"attendee":                rec.attendeeID,
+			"person_id":               rec.personID,
+			"session_id":              rec.sessionID,
+			"year":                    rec.year,
+			colToCampMethod:           rec.toCampMethod,
+			colFromCampMethod:         rec.fromCampMethod,
+			colDropoffName:            rec.dropoffName,
+			colDropoffPhone:           rec.dropoffPhone,
+			colDropoffRelationship:    rec.dropoffRelation,
+			colPickupName:             rec.pickupName,
+			colPickupPhone:            rec.pickupPhone,
+			colPickupRelationship:     rec.pickupRelation,
+			colAltPickup1Name:         rec.altPickup1Name,
+			colAltPickup1Phone:        rec.altPickup1Phone,
+			colAltPickup1Relationship: rec.altPickup1Rel,
+			colAltPickup2Name:         rec.altPickup2Name,
+			colAltPickup2Phone:        rec.altPickup2Phone,
+			"used_legacy_fields":      rec.usedLegacyFields,
+		}
 
 		var record *core.Record
 		if exists {
@@ -773,29 +825,23 @@ func (s *CamperTransportationSync) upsertRecords(
 				errors++
 				continue
 			}
+
+			// Check if update is actually needed (kindred#2384). An unchanged
+			// record counts as a skip, not an update -- see the file-level
+			// comment on Stats.Skipped in Sync() for why that unit is safe here.
+			if !s.recordNeedsUpdate(record, data, camperTransportationCompareFields) {
+				s.DebugLog("Skipping unchanged transportation record",
+					"person_id", rec.personID, "session_id", rec.sessionID, "year", year)
+				skipped++
+				continue
+			}
 		} else {
 			record = core.NewRecord(col)
 		}
 
-		// Set all fields
-		record.Set("attendee", rec.attendeeID)
-		record.Set("person_id", rec.personID)
-		record.Set("session_id", rec.sessionID)
-		record.Set("year", rec.year)
-		record.Set("to_camp_method", rec.toCampMethod)
-		record.Set("from_camp_method", rec.fromCampMethod)
-		record.Set("dropoff_name", rec.dropoffName)
-		record.Set("dropoff_phone", rec.dropoffPhone)
-		record.Set("dropoff_relationship", rec.dropoffRelation)
-		record.Set("pickup_name", rec.pickupName)
-		record.Set("pickup_phone", rec.pickupPhone)
-		record.Set("pickup_relationship", rec.pickupRelation)
-		record.Set("alt_pickup_1_name", rec.altPickup1Name)
-		record.Set("alt_pickup_1_phone", rec.altPickup1Phone)
-		record.Set("alt_pickup_1_relationship", rec.altPickup1Rel)
-		record.Set("alt_pickup_2_name", rec.altPickup2Name)
-		record.Set("alt_pickup_2_phone", rec.altPickup2Phone)
-		record.Set("used_legacy_fields", rec.usedLegacyFields)
+		for field, value := range data {
+			record.Set(field, value)
+		}
 
 		if err := s.App.Save(record); err != nil {
 			slog.Error("Error saving camper_transportation record",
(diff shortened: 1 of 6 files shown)


adamflagg:
Closing out the review on this PR.

**Fixed in-branch (both findings were M-or-smaller):**

1. **High/S — mutation coverage gap.** The three `*IsIdempotent` tests only pinned the skip
   direction: replacing `recordNeedsUpdate` with `return false` in all three services made the
   sync never write an update to an existing row again, and `go test ./sync/ -count=1` still came
   back green. Added a third run to each test (`updatePersonCustomValue` helper edits the seeded
   `person_custom_values` row) that asserts `Updated=1` and the new value on the re-fetched
   record. Mutation-checked the same way and confirmed RED before restoring:
   - `staff_applications.go`: `created=0 updated=0 errors=0, want ... updated=1`
   - `staff_vehicle_info.go`: same shape
   - `camper_transportation.go`: same shape

   All three restored; `go test ./sync/ -run 'IsIdempotent$'` green afterward.

2. **Medium/XS — PR body numbers wrong.** `Stats.Skipped` in `staff_applications.go` and
   `staff_vehicle_info.go` is `+=`'d twice per run (the kindred#2277 staff-row gate plus this PR's
   no-write count), so a no-op run actually reports **296** skipped for staff applications and
   **206** for staff vehicles, not 274/200 as originally written. Re-verified against the
   production snapshot: 274 no-write + 22 gated = 296; 200 + 6 gated = 206.
   `camper_transportation.go` has no second `Stats.Skipped` writer, so its 1,667 figure was
   already correct. Corrected the PR body (it's the permanent squash commit message) in the
   "What changed" and "GUI impact" sections.

**Not filed** — both findings landed inside this PR at M-or-smaller size, no new issue needed.

Gates re-run after the fix: `go build ./...`, `go vet ./...`, `go test ./sync/ -count=1`,
`go test . -run 'Parallel|Serial|Exempt'` (no new serialGroups entry needed — the new assertions
call no process globals), `scripts/pre-push-verify.sh`. Pushed and remote ref confirmed moved.

PR #2395, opened 2026-08-16
chore(scripts): move worktrees in-repo to .worktrees/

adamflagg:
Worktrees now live at `.worktrees/<feature>/` inside the repo instead of a sibling `../<repo>-worktrees/` directory.

## Why

The harness pins the session working directory to the project root and reverts any `cd` that resolves outside it. Measured twice:

```
$ cd ~/kindred-worktrees/staff-gate-visibility-2277
/home/adam/kindred-worktrees/staff-gate-visibility-2277
exit=0
Shell cwd was reset to /home/adam/kindred
```

Corroborated across 25 session transcripts: 13,229 recorded `cwd` values, none outside `~/kindred`.

Two consequences. The statusline reads `workspace.current_dir`, so it reported `~/kindred (main)` — clean and green — no matter which feature branch was being edited. And every worktree-directed command had to re-pay its full path through `git -C` or an absolute path.

A symlink does not fix this: the clamp resolves `realpath`. Probed with the pre-existing `docs/camp` symlink, whose logical path is inside the repo and physical path outside — still reset. So the directories had to move for real.

Verified after the change, running `new.sh` from a *different* worktree:

```
call 1: cd ~/kindred/.worktrees/e2e-probe   → no reset
call 2: cwd at start of new call: ~/kindred/.worktrees/e2e-probe   ← persisted

statusline: ~/kindred/.worktrees/e2e-probe (feature/e2e-probe)
```

## What changed

- **`WORKTREES_DIR`** in `new.sh`, `cleanup.sh`, `list.sh` and the `.lefthook.yml` post-merge hook → `$MAIN_REPO/.worktrees`.
- **`MAIN_REPO` derivation** switched from `git rev-parse --show-toplevel` to the parent of `--git-common-dir`. `--show-toplevel` names the *current* worktree when run from inside one; now that worktrees nest under the main repo, that spelling would have produced `.worktrees/<a>/.worktrees/<b>`. Confirmed fixed by creating a worktree from a sibling worktree and watching it land in the main repo.
- **`.gitignore` / `.dockerignore` / new `.ignore`.** The `.gitignore` entry is load-bearing beyond git: the `grep` shim runs with `--hidden`, so the leading dot does *not* keep worktrees out of search results — `--ignore-files` reading that entry is what does. Measured with a planted worktree copy: `rg`, `rg --hidden` and bare `grep -r` all returned 1 hit, `rg -uu` returned 2 (documented, opt-in), and `git status` showed nothing.
- **`git clean` guard** in `.claude/hooks/worktree-guard.sh`, with a new self-test.

## The git clean guard

Gitignoring the worktrees puts them in range of `git clean`. The guard is narrow because the boundaries were measured against a gitignored nested repo, not assumed:

| Command | Result |
|---|---|
| `git clean -dff` | survives — without `-x`/`-X`, ignored files are untouched |
| `git clean -xdf` | survives — git prints "Skipping repository" itself |
| `git clean -nxdff` | survives — dry run deletes nothing |
| `git clean -xdff` | **DELETED** — "Removing .worktrees/" |
| `git clean -Xdff` | **DELETED** — `-X` is ignored-files-only, so it hits too |

So it blocks exactly: an ignored-file flag, **and** force given twice, **and** not a dry run. Blocking wider would flag commands git already refuses to execute, which trains the reader to bypass the guard.

`scripts/dev/test-worktree-guard.sh` pins all 21 cases including flag reorderings (`-ffdx`), long forms (`--force --force`) and chaining. Written first and confirmed red (8 failures) before the guard existed — which is how the tokenizer bug surfaced: `read` reports failure at EOF on a line with no trailing newline, so the final token, where the flags sit, was being silently dropped.

Wired into CI's `guard-self-tests` job, and `.claude/hooks/**` added to the `guardSelfTests` path filter so editing the hook alone triggers the suite that pins it — the unwired-guard shape #2370 was about.

## Not migrated

Existing worktrees were drained rather than moved. `new.sh` links private config with `ln -sfr`, so each worktree carries six *relative* symlinks; the new location is one level deeper and `git worktree move` does not rewrite symlink targets. All six would have resolved to `~/kindred/kindred-local/…` and dangled, silently costing branding, `CLAUDE.local.md`, vault config, vite overrides and the lodging registry PocketBase reads at boot. Freshly created worktrees get correct relative links automatically — verified on the probe.

## Verification

- `scripts/dev/test-worktree-guard.sh` — 21/21
- `shellcheck --severity=warning` on all five changed scripts — clean
- `markdownlint-cli2`, YAML parse of both workflow files — clean
- `scripts/pre-push-verify.sh` — pass
- End-to-end: `new.sh` → correct path, cwd persists, five of six private-config symlinks resolve (the sixth, `sheets_sharing.local.json`, has no target in `kindred-local` and dangles identically in `main` — pre-existing, untouched here), `.git` pointer correct; `cleanup.sh` and `list.sh` both operate on the new location

## Summary by CodeRabbit

* **New Features**
  * Worktrees are now consistently created and managed in the repository’s `.worktrees/` directory, including when working from nested worktrees.
  * Unsafe cleanup commands that could delete ignored worktree files are now blocked, with safer alternatives provided.

* **Documentation**
  * Updated development and Git workflow guidance to explain the new worktree layout, search behavior, and cleanup precautions.

* **Tests**
  * Added automated coverage for worktree safety protections and integrated it into CI.

### DIFF
--- a/.claude/hooks/worktree-guard.sh
+++ b/.claude/hooks/worktree-guard.sh
@@ -1,11 +1,14 @@
 #!/usr/bin/env bash
-# PreToolUse hook for Bash — blocks direct `git worktree add` calls.
-# Must use ./scripts/worktree/new.sh which handles port allocation,
-# branch naming, DB seeding, and config symlinks.
+# PreToolUse hook for Bash. Blocks two commands that silently break the
+# worktree layout. Self-tested by scripts/dev/test-worktree-guard.sh.
 #
 # Exit codes:
 #   0 = allow the tool call
 #   2 = block (Claude Code surfaces stderr as the denial reason)
+#
+# The hook only sees commands Claude issues through the Bash tool. Scripts
+# that shell out internally -- including new.sh's own `git worktree add` --
+# do not pass through it, and a developer's own terminal is unaffected.
 
 set -u
 
@@ -15,18 +18,16 @@ if command -v jq >/dev/null 2>&1; then
   command=$(printf '%s' "$input" | jq -r '.tool_input.command // ""' 2>/dev/null)
 else
   # Fallback: scan the raw JSON payload. Less precise than parsing
-  # .tool_input.command — may catch the literal string in other fields — but
+  # .tool_input.command -- may catch the literal string in other fields -- but
   # preserves enforcement. Failing open would leave worktree creation
   # unguarded on hosts without jq.
   command=$input
 fi
 
+# ── 1. Direct `git worktree add` ────────────────────────────────────────
 # Match `git ... worktree add`, allowing intervening flags like `-C <dir>` or
 # `--git-dir=…` that would otherwise bypass the guard. Shell separators
 # (`;`, `&`, `|`) bound the match so chained commands still trigger.
-# The wrapper script itself shells out to `git worktree add` internally, but
-# those subprocesses don't go through the Bash tool, so the hook only sees
-# Claude's direct invocations.
 if printf '%s' "$command" | grep -qE '\bgit\b[^|;&]*\bworktree[[:space:]]+add\b'; then
   cat >&2 <<'EOF'
 [worktree-guard] BLOCKED: direct `git worktree add` is not allowed in this project.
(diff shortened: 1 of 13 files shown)


adamflagg:
@coderabbitai review

Issue #2392, opened 2026-08-15
question(lodging): lodging_units covers 2026 only, blocking historical name resolution and any session-grained backfill


adamflagg:
### 1. ✅ The owner does NOT need `bunking.manage`. There is an admin bypass.

This body says the endpoints are *"both gated on `bunking.manage`"*, which is true and reads as a
blocker for an account that holds admin instead. **It is not.** `sync.RequirePermission`
(`pocketbase/sync/api.go:60`) short-circuits before the permission check:

```go
if e.Auth.GetBool("is_admin") { return handler(e) }
```

✅ Verified 2026-08-27 against the production snapshot: `users` carries both `is_admin` and
`cached_permissions`, and **exactly one account has `is_admin = 1`.** That account clears every call
this operator task needs — the four roll-forward previews/applies *and* `/api/custom/sync/run`, which
goes through the same helper. **The run is not blocked on a permission grant.**

### 2. Step 1's mechanism is being REMOVED, and the outcome is unchanged

Step 1 records kindred#2399's `carriedFields(col, backward)` flag as the shipped prerequisite.
**kindred#2500's 2026-08-27 ruling removes that parameter entirely** — `is_confirmed` is excluded
unconditionally, on a roll in *any* direction, because `backward` had exactly one consumer and this was
it.

⇒ **Nothing here breaks, and step 1 becomes true by construction.** Whether #2500 has shipped or not,
a backward roll lands historical rows `is_confirmed = false`, which is what this task requires. Worth a
post-roll spot check either way: `SELECT COUNT(*) FROM lodging_units WHERE year = 2025 AND is_confirmed = 1;`
must be **0**.

### 3. ⚠️ The "known and accepted caveat" is sharper than when it was written

The caveat records that rolled-back rows carry 2026's amenities, and that *"who lived where"* is safe
because it reads the unit's **name**. That still holds. But it was written when an unconfirmed row's
amenities rendered `unknown`, which made the caveat self-limiting.

**kindred#2526's 2026-08-27 face-value ruling removes that cushion:** `is_confirmed` no longer gates
render, so a backfilled 2023 row asserts **2026's amenities as 2023 fact** rather than declining to
answer. It remains a footnote — the historical surfaces read `cabin_name` through
`HousingNameResolver`, not amenities — but ⛔ **do not build anything that reads historical amenities on
the strength of this backfill.**

### Unchanged and still true

[...]

PR #2399, opened 2026-08-17
fix(pb): stop lodging roll-forward from carrying is_confirmed across seasons

adamflagg:
## Defect

`notCarried` in `pocketbase/lodging/rollforward.go` denied only `year`, `area`
and `parent_unit`. Every other column -- including `is_confirmed` -- rode
along verbatim through `carriedFields`, because a new column is carried by
default (deny-list, not allow-list) unless it is explicitly excluded.

`is_confirmed` is a permanent staff attestation, not a per-season flag: #2029
made it deliberately carry forward on a roll ("Confirmation carries
forward... a yearly re-confirm is unnecessary",
`docs/reference/lodging-registry.md:368-373`; `SeasonRollForwardPanel.tsx`'s
own doc comment says the same). That is correct for the FORWARD direction --
the only one the Season panel exposes today (`fromYear = currentYear-1,
toYear = currentYear`). It is wrong for a BACKWARD roll -- the mechanism
kindred#2392 proposes for backfilling `lodging_units` into 2022-2025 -- which
would stamp `is_confirmed = true` onto historical rows nobody has ever
checked against a physical cabin, in a field the docs describe as permanent.

## Fix

`carriedFields` now takes a `backward` flag (`to < from`, computed once per
pass in `copyAreas`/`copyUnits`) and excludes `is_confirmed` **only** on a
backward roll. A forward roll -- today's only real usage -- still carries
`is_confirmed`, matching #2029's design and every doc/comment that describes
it; a backward roll starts every carried unit unconfirmed, closing the gap
kindred#2392 raised.

An earlier version of this PR added `is_confirmed` to `notCarried`
unconditionally, which is direction-blind and would have silently
un-confirmed every already-verified cabin on the very next ordinary season
advance -- reversing #2029's design rather than fixing the backward-roll gap.
Caught and corrected during independent review; kindred#2392's ruled plan
(step 1) has been corrected to match.

Test-first: `TestApplyRollForwardCarriesIsConfirmedOnAForwardRoll` (red
against the unconditional-exclusion version, mutation-checked) pins that a
forward roll still carries `is_confirmed`.
`TestApplyRollForwardDoesNotCarryIsConfirmedOnABackwardRoll` pins that a
backward roll does not. Both replace the single direction-blind test the PR
originally shipped with.

`go build ./... && go test ./lodging/...` (full package), `go vet`,
`gofmt -l`, and `golangci-lint run` are all clean on the two touched files.

- **Does not run the roll.** Rolling 2026 -> 2025 -> ... -> 2022 to backfill
  the historical registry is an operator task against production, called out
  explicitly in kindred#2392, and is not performed by this PR.
- **Does not add a direction guard** to `PreviewRollForward`/`ApplyRollForward`
  rejecting `from > to` -- kindred#2392 asks for that to be verified with a
  `Preview` call before the real roll runs, which is part of the operator
  task, not this code change. (This PR's `backward` flag makes carry-forward
  behavior correct either way; it does not restrict which directions are
  callable.)
- **Does not touch `pocketbase/lodging/registry.go`** -- owned by a sibling
  PR (kindred#2390).
- **Does not close kindred#2392.** That issue's plan has three more steps
  (direction-guard verification, the actual roll, and the per-year sync run)
  that are explicitly operator tasks, not agent work.

Related to kindred#2392 (step 1 of its ruled plan).

### DIFF
--- a/pocketbase/lodging/rollforward.go
+++ b/pocketbase/lodging/rollforward.go
@@ -27,9 +27,17 @@ import (
 //	year                   — set explicitly to the target
 //	area / parent_unit     — relations, re-resolved into the target year by the
 //	                         two later passes
+//	is_confirmed           — asserts a human physically walked THIS season's
+//	                         cabin (docs/reference/lodging-registry.md:377), not
+//	                         a fact that outlives the season it was checked in.
+//	                         Carrying it would let a roll-forward -- including a
+//	                         BACKWARD one onto a season nobody has walked -- stamp
+//	                         is_confirmed = true on a row nobody confirmed
+//	                         (kindred#2392).
 var notCarried = map[string]bool{
 	"id": true, "created": true, "updated": true,
 	"year": true, "area": true, "parent_unit": true,
+	"is_confirmed": true,
 }
 
 // carriedFields returns every field on the collection that should travel to the
--- a/pocketbase/lodging/rollforward.go
+++ b/pocketbase/lodging/rollforward.go
@@ -8,9 +8,10 @@ import (
 	"github.com/pocketbase/pocketbase/core"
 )
 
-// Columns copied verbatim onto the new season's row. `code` is here and is
-// NEVER regenerated: a building renamed next season keeps its code, and the
-// code is the only thing linking its years together.
+// Columns copied verbatim onto the new season's row, on EVERY roll regardless
+// of direction. `code` is here and is NEVER regenerated: a building renamed
+// next season keeps its code, and the code is the only thing linking its
+// years together.
 //
 // `id`, `created` and `updated` are excluded — the new row is a new record.
 // `area` and `parent_unit` are excluded because they are relations that must be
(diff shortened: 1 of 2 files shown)


adamflagg:
## Independent review — findings and fix

**Scope check:** diff is exactly `pocketbase/lodging/rollforward.go` +
`rollforward_test.go`, matching the item (`#2392` step 1, "add `is_confirmed`
to `notCarried`"). No scope creep, no silent narrowing. `closingIssuesReferences`
is correctly empty — this PR deliberately closes nothing.

### Fixed in-branch (L, in code this PR wrote)

The original fix added `is_confirmed` to `notCarried` **unconditionally**.
`notCarried` has no notion of direction, so that also strips `is_confirmed`
on a **forward** roll — the *only* direction the Season panel exposes today
(`fromYear = currentYear-1, toYear = currentYear`; there is no UI for a
backward roll). That directly reverses #2029's deliberate design:

- `docs/reference/lodging-registry.md:368-373` — "Is confirmation ephemeral?
  No. A confirmed unit carries that state across seasons by roll-forward...
  asserts *permanently*... A yearly re-confirm is unnecessary."
- `SeasonRollForwardPanel.tsx`'s own doc comment — "Values, `is_confirmed`
  and `code` all carry forward... there is no exclusion step."
- The original #2029 commit message — "Confirmation carries forward."

Left as shipped, the very next ordinary season advance (which happens every
year, via the existing admin panel) would have silently un-confirmed every
already-verified cabin, forcing staff to re-walk buildings nothing changed
about — the opposite of the stated product intent.

Fix: `carriedFields` now takes a `backward` flag (`to < from`, computed once
per pass in `copyAreas`/`copyUnits`) and excludes `is_confirmed` **only** on
a backward roll. Forward rolls (today's only real usage) keep carrying
`is_confirmed`, matching #2029; backward rolls (the historical-backfill
mechanism kindred#2392 proposes) start every carried unit unconfirmed,
closing the actual gap the issue raised.

**Mutation-check transcript** — reverting `backward := to < from` to an
unconditional `backward := true` (i.e. restoring the original PR's blanket
exclusion) turns the forward-carry test red and leaves the backward test
green, confirming the tests pin the direction, not just "something changed":

```
=== NAME  TestApplyRollForwardCarriesIsConfirmedOnAForwardRoll
    rollforward_test.go:128: is_confirmed did not carry forward on a FORWARD
    roll; want true -- confirmation is a permanent staff attestation per
    #2029, not a per-season flag
--- FAIL: TestApplyRollForwardCarriesIsConfirmedOnAForwardRoll (1.07s)
--- PASS: TestApplyRollForwardDoesNotCarryIsConfirmedOnABackwardRoll (1.12s)
```

Restored, both pass; `go build ./...`, `go vet ./lodging/...`, `gofmt -l`,
`golangci-lint run`, and the full `lodging` package (`go test ./lodging/...`,
all ~90 tests) are all clean.

Corrected the PR body (it's the squash-merge commit message) and
kindred#2392's ruled-plan step 1 in place, with a comment on the issue
explaining the correction, per this repo's "a wrong issue body gets fixed,
not worked around" policy.

### Verified, no changes needed

- No real names / CampMinder IDs anywhere in the diff, PR body, or tests.
- Nothing under `docs/plans/`.
- No GUI-impact files (`frontend/src/**/*.{ts,tsx,css}`) touched — Go-only
  change, no owner eyeball needed on that front.
- CI Summary green on the pushed commit; push confirmed landed
  (`git rev-list --count origin/<branch>..HEAD` = 0).

### Filed out of scope

Nothing. Everything found was in code this PR wrote, so it was fixed here
rather than filed.

PR #2400, opened 2026-08-17
refactor(frontend): drop the areas drawer's unusable Map centre inputs

### LINKED ISSUE #2397: refactor(frontend): drop the areas drawer's unusable Map centre inputs — there is no map to place them against
## The problem

`LodgingAreasDrawer` exposes each area's map centroid as **two bare number inputs** and renders **no map at
all** — `grep -c 'MapBaseLayer\|svg\|canvas\|img' LodgingAreasDrawer.tsx` returns **0**.

`frontend/src/components/admin/lodging/LodgingAreasDrawer.tsx:198-215`:

```tsx
<span className="text-muted-foreground">Map centre</span>
<input className={`${FIELD} w-20`} type="number" step="0.001" min={0} max={1}
       defaultValue={area.map_x} aria-label={`${area.name} map X`}
       onBlur={(e) => void saveCentroid(area, 'map_x', e.target)} />
```

Owner, 2026-08-16:

> i'm just editing map centre labeled X and Y coordinates with no frame of reference - useless, tbh, we
> should just only show the area names there anyway

Typing `0.4389` into a normalised coordinate with nothing to compare it against is not an editing affordance.
The same file's sibling component made this exact judgement already: `UnitMapPositionField`'s header records
that PR #2024 deleted the equivalent number inputs for **units** because *"Typing `0.4389` was never how a
pin gets placed anyway — dragging it onto the cabin is (kindred#2013)."* Areas never got the follow-through.

## Ask

**Remove the two `Map centre` inputs from the drawer.** Keep name and order. The drawer's own header already
describes its scope as *"Eight rows of name, order and map centroid"* — that becomes name and order.

## What must NOT be lost

- **`map_x`/`map_y` stay on the collection.** They are read elsewhere; this is a change to one editing
  surface, not a schema change. The drawer's header notes the centroid "is not decorative. A later phase
  renders the actual camp map" — that remains true.
- **`saveCentroid` may still be needed.** Confirm no other caller before deleting it; if it is now unused,
  delete it in the same PR rather than leaving a dead export.
- Areas whose centroid is currently unset must stay unset. Removing the inputs must not write anything.

## If a real editor is wanted later

The pattern to copy is `UnitMapPositionField` — a gated drag on a rendered map, saving on pointer-up. That is
being asked for at the unit level on the session map in #2396. An area centroid editor would be the same
shape and is explicitly **not** in scope here; this issue only removes a control that cannot be used.

## Anchors

- `frontend/src/components/admin/lodging/LodgingAreasDrawer.tsx:198-215` — the two inputs
- `frontend/src/components/admin/lodging/LodgingAreasDrawer.tsx:4-9` — the header describing the drawer's scope
- `frontend/src/components/admin/lodging/UnitMapPositionField.tsx` — the precedent, and the reasoning for it

_Filed from the phase-D triage pass, on the owner's ruling. Split out of #2214._

adamflagg:
## The defect

`LodgingAreasDrawer` exposed each area's map centroid (`map_x`/`map_y`) as two bare number
inputs with nothing to compare them against — the file renders no map at all
(`grep -c 'MapBaseLayer\|svg\|canvas\|img'` on it returns `0`). Typing `0.4389` into a normalised
coordinate was never a usable editing affordance. The sibling component, `UnitMapPositionField`,
made this exact call for units already in kindred#2013/#2024 ("Typing `0.4389` was never how a
pin gets placed anyway — dragging it onto the cabin is"); areas never got the follow-through.

## The fix

Removed the two "Map centre" number inputs from `LodgingAreasDrawer.tsx` (`map X` / `map Y`).
Kept the name field, the order (move up/down) controls, and delete. `saveCentroid` — the onBlur
handler that wrote `map_x`/`map_y` from those inputs — had no other caller anywhere in the tree,
so it's deleted in this same PR rather than left as dead code, along with its now-unused
`LodgingAreaRecord` import.

`map_x`/`map_y` stay on the `lodging_areas` collection unchanged; no migration. They're still
read elsewhere (e.g. `mapModel.ts`, `boardLayout.ts`). This removes one editing surface, not the
fields — a later phase can add a real drag-to-place editor on a rendered map, the same shape
`UnitMapPositionField` already uses for units.

The other write paths in this file (`updateLodgingArea` for rename/reorder, `createLodgingArea`
for new areas) never touched `map_x`/`map_y` and still don't — removing the inputs introduces no
new write of `0`/`0` or `null` over an area's existing (possibly unset) centroid.

## Tests

- New: asserts the "Map centre" inputs (`aria-label` `"<name> map X"` / `"<name> map Y"`) and the
  "Map centre" label text are absent from the rendered drawer.
- New: renders an area whose centroid is unset (`map_x: 0, map_y: 0` — PocketBase's convention
  for "unset", not `null`), edits its name, and asserts the resulting `updateLodgingArea` payload
  carries neither `map_x` nor `map_y`.
- Removed: the three tests that exercised the deleted inputs directly (typing/clearing/rejecting
  a centroid edit) — the behavior they pinned no longer exists.
- Mutation-checked: reintroducing a `map_x`/`map_y` write on the rename path turned the new
  "leaves an unset centroid unset" test red before the fix was restored, confirming it actually
  pins the no-write guarantee rather than passing vacuously.

- No PocketBase migration. `map_x`/`map_y` are untouched on the collection.
- No new centroid editor (drag-on-map). That is explicitly out of scope per the issue and is the
  same shape as the unit-level map editor being tracked separately in kindred#2396.
- No change to `UnitMapPositionField.tsx` or any other file. Its header comment references
  `LodgingAreasDrawer`'s former `saveCentroid` by name as a design precedent; that comment is now
  slightly stale (the function it names no longer exists) but the file is out of this PR's scope
  and the surrounding claim ("same save-on-interaction shape") is still accurate.

Closes #2397.

### DIFF
--- a/frontend/src/components/admin/lodging/LodgingAreasDrawer.test.tsx
+++ b/frontend/src/components/admin/lodging/LodgingAreasDrawer.test.tsx
@@ -18,6 +18,20 @@ const AREAS = [
   { id: 'a2', name: 'South Zone', code: 'SOUTH', map_x: 0.6, map_y: 0.7, sort_order: 3 },
 ]
 
+// PocketBase stores an unset centroid as 0, never null (see mapModel.ts's
+// hasCoordinates). This is the shape #2397's "must stay unset" guarantee is
+// about. Kept out of the shared AREAS fixture — several tests above assert
+// on exact sort_order arithmetic derived from AREAS, and a third row would
+// shift every one of them.
+const UNSET_CENTROID_AREA = {
+  id: 'a3',
+  name: 'Unset Zone',
+  code: 'UNSET',
+  map_x: 0,
+  map_y: 0,
+  sort_order: 5,
+}
+
 // Module-scoped so `vi.mock` can close over them, so call records and resolved
 // values outlive the test that set them unless cleared here.
 beforeEach(() => {
@@ -151,22 +165,42 @@ describe('LodgingAreasDrawer', () => {
     expect(screen.queryByRole('button', { name: 'Move North Zone up' })).not.toBeInTheDocument()
   })
 
-  it('saves a map centroid, which the later map view renders from', async () => {
-    updateLodgingArea.mockResolvedValue({})
-    const user = userEvent.setup()
+  it('renders no Map centre inputs — there is no map to place them against (#2397)', async () => {
     render(<LodgingAreasDrawer open onClose={vi.fn()} />, { wrapper })
     await waitFor(() => {
       expect(screen.getByDisplayValue('North Zone')).toBeInTheDocument()
     })
 
-    const x = screen.getByLabelText('North Zone map X')
-    await user.clear(x)
-    await user.type(x, '0.45')
+    expect(screen.queryByLabelText('North Zone map X')).not.toBeInTheDocument()
+    expect(screen.queryByLabelText('North Zone map Y')).not.toBeInTheDocument()
+    expect(screen.queryByText('Map centre')).not.toBeInTheDocument()
+  })
+
+  it('leaves an unset centroid unset when a different field is edited', async () => {
+    // The centroid inputs are gone, so nothing in this drawer should ever
+    // put map_x/map_y on a write again. An area whose centroid is unset
+    // (map_x: 0, map_y: 0 — PocketBase's convention, not null) must stay
+    // that way rather than getting overwritten as a side effect of editing
+    // its name.
+    listLodgingAreas.mockResolvedValue([...AREAS, UNSET_CENTROID_AREA])
+    updateLodgingArea.mockResolvedValue({})
+    const user = userEvent.setup()
+    render(<LodgingAreasDrawer open onClose={vi.fn()} />, { wrapper })
+    await waitFor(() => {
+      expect(screen.getByDisplayValue('Unset Zone')).toBeInTheDocument()
+    })
+
+    const name = screen.getByLabelText('Unset Zone name')
+    await user.clear(name)
+    await user.type(name, 'Renamed Zone')
     await user.tab()
 
     await waitFor(() => {
-      expect(updateLodgingArea).toHaveBeenCalledWith('a1', { map_x: 0.45 })
+      expect(updateLodgingArea).toHaveBeenCalledWith('a3', { name: 'Renamed Zone' })
     })
+    const [, payload] = updateLodgingArea.mock.calls[0] as [string, Record<string, unknown>]
+    expect(payload).not.toHaveProperty('map_x')
+    expect(payload).not.toHaveProperty('map_y')
   })
 
   it('derives a new area sort_order from the highest in use, not the list length', async () => {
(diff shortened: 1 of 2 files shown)


adamflagg:
## Independent review

**Scope check against #2397:** matches — the two "Map centre" number inputs are removed, name/order/delete are kept, `map_x`/`map_y` stay on the `lodging_areas` collection with no migration, and `saveCentroid` is deleted only after confirming it had no other caller in the tree. No scope creep, no silent narrowing.

**Mutation-check on the new "leaves an unset centroid unset" test:** re-added a `map_x: 0` write to the rename handler's `updateLodgingArea` call in `LodgingAreasDrawer.tsx`. Result: RED — `expect(updateLodgingArea).toHaveBeenCalledWith('a3', { name: 'Renamed Zone' })` failed because the actual call carried `map_x` too. Reverted, suite back to 17/17 green. The test pins the no-write guarantee for real, not vacuously — confirms the PR body's own mutation-check claim.

**Fixed in this branch:** `LodgingAreasDrawer.test.tsx`'s "deleting an area" describe block still had the old comment "one click from a numeric input" — a direct leftover of the exact inputs this PR removes. The `.tsx` file's parallel comment was updated in the same PR; the test file's copy was missed. Trimmed it to match. Small, in a file this PR already touches, no behavior change — tests/lint/tsc still green after.

**Considered and left alone:** `UnitMapPositionField.tsx`'s header comment names the now-deleted `saveCentroid` as a design precedent. Verified: that file is genuinely untouched by this PR, and the surrounding claim ("same save-on-interaction shape") is still true — only the specific function name is stale. Agreed with the author's call to leave it out of scope here rather than reach into an unrelated file's docs; a one-line rename doesn't carry a decision, so it can go in a follow-up.

**GUI impact:** `LodgingAreasDrawer.tsx` is a visual change (removes two number inputs from the areas drawer) — flagging for a look before merge per policy, not because anything looks wrong.

**CI:** green (CI Summary success) both before and after the fix-up commit; re-verified after pushing.

No real names, no PocketBase/CampMinder IDs, nothing under `docs/plans/` in the diff.

adamflagg:
Collapsed each area row per owner feedback ("wasting vertical space, delete should fit next to the name and arrows on one line"): Delete used to render in its own block below the name/arrows row. It's now a sibling in the same flex row as the name input and the reorder chevrons — one row per area instead of two, no wrapper div left behind.

Style-matched to this file's own existing grammar rather than introducing anything new: Delete keeps the `ACTION_LINK` treatment it already had, just moved up into the row it belongs in. No new colors, icons, or tokens.

TDD: extended `LodgingAreasDrawer.test.tsx` with a test pinning that the delete button and the move-down arrow share a parent element with the name input (i.e. one row, not two) — verified it failed against the old two-row markup before making the change.

`./scripts/pre-push-verify.sh` is green (prettier/eslint/tsc/vitest).

PR #2401, opened 2026-08-17
feat(frontend): flag single-parent families on the family card

### LINKED ISSUE #2254: feat(frontend): order a family card children youngest-first, and flag single-parent families
## ⚠️ CORRECTED 2026-08-16 — phase D triage. The deferral sentence is refuted. This is RUNNABLE NOW.

**⛔ Strike this from the body: _"prefer resolving #2072 first rather than adding a chip the gutter would then
absorb."_**

#2072's own scoping ruling says the glyph gutter replaces **only the two chips it duplicates** —
`Private bathroom` and `Power`. `Whole building`, the warn chips and the share chip **stay as words**. A
one-adult / single-parent chip is not a housing-need flag, so the per-need gutter can never absorb it.

The precedent is already in the file: the muted `Near another family` chip at `FamilyCard.tsx:522` survives
the gutter under identical reasoning.

So this issue does **not** depend on #2072 — which matters, because #2072 is HELD on a staff conversation with
no date, and chaining behind it would park a decision-free S item indefinitely.

**Classification: SCOPED, S, not UI-gated.** Ordering children youngest-first is a sort, and the single-parent
flag reuses the existing muted-chip grammar rather than introducing a visual channel.

⚠️ Half 1 has already shipped — a PR here closes this issue on **half 2** only.

⚠️ **File collision:** `FamilyCard.tsx` (+ its test) is shared with #2072, #2212, #2275 and #2332. Whichever
runs first locks the file.

---

> Both halves confirmed as the brief describes: youngest-first ordering shipped in #2319 (`044eeac1`) and the single-parent affordance was deferred. The shipped `youngestFirst` never mutates — it partitions into two fresh arrays and returns `[...known, ...unknown]`, sorting only the local `known` array, so the shared `children_oldest_first` copy is untouched. The remaining half's own ADDITIONS banner is wrong twice: the `attendingAdults` anchor moved, and its caution that under-transcription will mislabel families is falsified by an independent field.

1. WRONG (in the ADDITIONS banner): "`attendingAdults` and `is_attending_adult_name` already ship (kindred#2085, closed) and are called at `FamilyCard.tsx:255`" → REPLACEMENT: "`attendingAdults` and `isAttendingAdultName` already ship (kindred#2085, closed) and are called at `FamilyCard.tsx:304` (`computeAttendingAdults`, aliased at the import on `:65`). Note the frontend helper is `isAttendingAdultName` in `householdIdentity.ts:46`; `is_attending_adult_name` is the separate PYTHON helper at `api/constants/lodging.py:42`. `attendingAdults` is grain-gated (`householdIdentity.ts:61`: `if (party.grain !== 'household') return []`), so `.length === 1` cannot false-positive a person-grain party."
2. WRONG (in the ADDITIONS banner): "⚠️ **One caution on the single-parent flag that is not in the body.** It is derived from the adult roster, and `docs/reference/family-camp-field-provenance.md` §6 shows that roster under-transcribes: ~45% of households naming additional adults end up with no slot 3-5 at all. A household whose second adult was never transcribed will be flagged single-parent **wrongly**." → REPLACEMENT: "⚠️ **This caution was tested against production on 2026-08-13 and does NOT hold for the single-adult case.** Of the 42 rostered-2026 family households with exactly one named attending adult, 41 independently answer `Total Adults-FC` (cm_id 209430) = **1**, and the 42nd gives no numeric answer at all. **Zero** claim two or more. The under-transcription in §6 is real but runs the other way: the only disagreements between the two signals are `family_camp_adults` holding MORE names than `Total Adults-FC` reports (10 households at n=2 that answer 1, 6 at n=3 that answer 2). So `attendingAdults(party).length === 1` is corroborated, not suspect. Still use `attendingAdults` — it is the attendance-scoped, placeholder-filtered predicate — and still do not count raw `family_camp_adults` rows."
3. WRONG (in the ADDITIONS banner): "**Blast radius is 7 of 265 multi-child rostered-2026 households (2.6%)**, not \"every family card\"." → REPLACEMENT: "Two different blast radii, do not conflate them. The SORT's is small (it only reorders cards with more than one child whose ages differ) and shipped in #2319. **The SINGLE-PARENT FLAG's is 42 of 381 rostered-2026 family households — 11.0%** (re-measured 2026-08-13). That is the density figure the placement decision has to survive, and it is additive to #2072's 20.7% needs-union on the same card."
4. WRONG: "## 1. Order the children youngest-first" (whole section, now shipped) → REPLACEMENT: prefix the section with "> ### ✅ SHIPPED 2026-08-13 in PR #2319 (`044eeac1`). Do not rebuild. `youngestFirst` (`FamilyCard.tsx:203-211`) partitions into fresh `known`/`unknown` arrays and returns `[...known, ...unknown]`, so it NEVER mutates the shared `children_oldest_first` copy — verified. Unknown age (`null`) trails in original relative order rather than coercing to 0. The truncation consequence was taken deliberately: the OLDEST children now clip first."

- 42 of 381 rostered-2026 FAMILY households (11.0%) have exactly one named attending adult — the true blast radius of the single-parent flag. Denominator = distinct households with a status_id=2 attendee in a 2026 `session_type='family'` session. SQL: WITH rostered AS (SELECT DISTINCT h.id hid FROM attendees a JOIN camp_sessions s ON s.id=a.session JOIN persons p ON p.id=a.person JOIN households h ON h.id=p.household WHERE a.year=2026 AND a.status_id=2 AND s.session_type='family'), named AS (SELECT fa.household hid, TRIM(CASE WHEN TRIM(fa.name)<>'' THEN fa.name ELSE TRIM(fa.first_name)||' '||TRIM(fa.last_name) END) dn FROM family_camp_adults fa WHERE fa.year=2026), adults AS (SELECT hid, COUNT(*) n FROM named WHERE dn<>'' AND LOWER(dn) NOT IN ('na','n/a','none','-','0','no') GROUP BY hid) SELECT ... → 381 rostered, 42 one-adult, 320 two-adult, 19 three-plus, 0 with none named
- The under-transcription caution is FALSIFIED: of those 42 single-adult households, 41 independently answer `Total Adults-FC` (cm_id 209430) = 1 and 1 gives no numeric answer. ZERO claim 2+. Denominator = the 42. Same CTEs plus: tot AS (SELECT p.household hid, MAX(CAST(v.value AS INTEGER)) ta FROM person_custom_values v JOIN custom_field_defs d ON d.id=v.field_definition JOIN persons p ON p.id=v.person WHERE d.cm_id=209430 AND v.year=2026 AND v.value GLOB '[0-9]*' GROUP BY p.household) → 1|1|41 and 1|(no answer)|1
- Full agreement matrix (named adults vs `Total Adults-FC`, 2026 rostered family households, denominator 381): n=1 → 41 agree / 1 no answer; n=2 → 309 agree, 10 say 1, 1 says 4; n=3 → 6 agree, 6 say 2, 1 says 4; n=4 → 3 agree, 1 says 5, 1 says 2; n=5 → 1 agrees. The only disagreements are OVER-counts in `family_camp_adults`, never under-counts, which is the opposite of the direction the body's caution warns about.
- `family_camp_adults` 2026 name-column population by slot (denominator = 840 rows): adult_number 1 → 429 rows, 398 `name`, 425 `first_name`, 0 `last_name`; 2 → 384/347/371/0; 3 → 20/20/0/0; 4 → 6/6/0/0; 5 → 1/1/0/0. `last_name` is empty for 100% of 2026, confirming `name` is the column of record. SQL: SELECT adult_number, COUNT(*), SUM(TRIM(name)<>''), SUM(TRIM(first_name)<>''), SUM(TRIM(last_name)<>'') FROM family_camp_adults WHERE year=2026 GROUP BY adult_number;

</details>

---

> ## ⚠️ ADDITIONS 2026-08-13 — verdict is VALID, every anchor in this body resolves
>
> **`hard_predecessors: []` — this is NOT blocked by kindred#1943.** `attendingAdults` and
> `is_attending_adult_name` already ship (kindred#2085, closed) and are called at
> `FamilyCard.tsx:255`, so single-parent detection is `attendingAdults(party).length === 1` today.
>
> **Blast radius is 7 of 265 multi-child rostered-2026 households (2.6%)**, not "every family card".
>
> **It is not silent** — `test_children_sort_oldest_first_by_the_raw_float` (`:481`) fails on the first
> run.
>
> **Keep the null-age warning.** It survives the `or None` conversion, because
> `(a.age ?? 0) - (b.age ?? 0)` coerces null to 0. Deleting it invites the bug back.
>
> **Sort display-only, inside `ChildList`** — never mutate the shared `children_oldest_first`, which
> other surfaces read.
>
> ⚠️ **One caution on the single-parent flag that is not in the body.** It is derived from the adult
> roster, and `docs/reference/family-camp-field-provenance.md` §6 shows that roster under-transcribes:
> ~45% of households naming additional adults end up with no slot 3-5 at all. A household whose second
> adult was never transcribed will be flagged single-parent **wrongly**. Use `attendingAdults`, which
> is the attendance-scoped predicate, and do not fall back to counting `family_camp_adults` rows.
>
> *(An earlier proposed correction was rejected: truncated mid-token, misdescribed
> `_household_sort_name` — which returns the first child with a non-blank `last_name`, not
> `children[0]` — overstated blast radius ~38x, and deleted the live null-age warning.)*

---

Two asks from the same report, both about reading a family card at a glance while assigning cabins.

## 1. Order the children youngest-first

The card renders a family's children as one `Name (age) · Name (age)` run, in **payload order** — `ChildList` maps `children` as given, with no sort (`FamilyCard.tsx:203-220`). Staff want **youngest first**, because the youngest child is what drives the housing decision (crib, ground floor, proximity).

> ### ⚠️ Do not sort on `PartyChild.age` naively — `0.0` is a sentinel
>
> `age` is `yy.mm`, and **`0.0` means *unknown age*, not *newborn***. This is the same trap recorded on kindred#2212, where it is the reason 24 households are affected rather than 25.
>
> A plain ascending sort therefore puts **every unknown-age child at the front of every card** — the exact opposite of the intent, and wrong in a way that looks right. Unknown ages need an explicit bucket, and a decision about whether they lead or trail.

**This interacts with the truncation.** The children run is `truncate`d (`FamilyCard.tsx:277`), so on a family with several children the sort decides *which children are visible at all*, not merely their order. Youngest-first plus truncation means the oldest children silently vanish from the card. Worth deciding deliberately rather than discovering it.

## 2. Make single-parent families obvious

Single-parent families "definitely affect housing" and are not currently distinguishable at a glance. The adults line already renders and already truncates (`FamilyCard.tsx:330-340`), so the count is present but not salient.

Worth settling first:

- **What counts as single-parent.** One adult *attending* is not the same as one adult *on file* — see kindred#1943 and kindred#2085: a family-camp adult is not a person in this system, and `party_size` counts listed rather than attending adults (kindred#1925). Deriving this from the wrong field mislabels real families.
- **Where it goes.** This is a per-family flag on a card whose chip row is already crowded, which is the same question kindred#2072's glyph gutter is held on. If the gutter lands, this is a glyph; if it does not, this needs its own answer. **Prefer resolving #2072 first** rather than adding a chip the gutter would then absorb.

## Anchors

- `frontend/src/components/weekend/FamilyCard.tsx:203-220` — `ChildList`, payload order
- `frontend/src/components/weekend/FamilyCard.tsx:277` — the truncating children line
- `frontend/src/components/weekend/FamilyCard.tsx:330-340` — the adults line

*Reported via kindred-feedback#108 — the private original, which holds the reporter, the page it was filed from, the app version, and the screenshot.*

adamflagg:
## Context

Item #2254 half 2. Half 1 (youngest-first child ordering) already shipped
in #2319 (`044eeac1`) and is untouched by this PR — only the single-parent
flag is new here.

Staff read the family card while deciding cabin placement. A single-parent
family "definitely affects housing," but the card gave no way to see it at
a glance: the adults line already renders and truncates, so the count was
present but not salient.

## What changed

`FamilyCard.tsx` gains a muted `Single parent` chip, shown when exactly one
attending adult is on the household party.

- **Derivation**: `attendingAdults(party).length === 1`, reusing the same
  household-filtered, placeholder-stripped adult list `FamilyCardIdentity`
  already computes for the grey adults line — not `party.adults.length` or
  `party_size`, both of which count listed-but-not-attending adults
  (kindred#1925/#2046) and would false-positive a two-parent household
  where one adult never RSVP'd for this session. `attendingAdults` is also
  grain-gated to household parties, so an adult weekend guest (person
  grain, which IS its own identity) is never flagged.
- **Style**: reuses the existing muted `Near another family` chip's exact
  classes (`tone="muted"`) — no new chip color, token, or icon.

## Why the deferral in the issue no longer applies

The issue originally said to prefer resolving #2072 first, in case its
glyph gutter would absorb this chip. The 2026-08-16 triage pass struck that
sentence: #2072's own scoping ruling replaces only the two NEED chips
(`Private bathroom`/`Power`) with the gutter — `Whole building`, the warn
chips, and the share chips (including the `Near another family` precedent
this PR copies) all stay as words. A single-parent flag is not a housing-need
chip, so the gutter can never absorb it, and this PR does not touch, open,
or depend on #2072/#2212/#2275/#2332.

## Deliberately not done

- No change to the already-shipped youngest-first sort (half 1).
- No gutter/glyph integration — out of scope per the struck deferral above.
- No new chip tone, color, or icon — the ask was to reuse the existing
  muted grammar, not invent a new visual channel.

## Testing

`frontend/src/components/weekend/FamilyCard.test.tsx` — TDD: 3 new
assertions (renders the chip, matches the `Near another family` chip's
classes exactly, filters out placeholder adult slots) failed red before
the implementation; 4 more (two adults, zero named adults, adult weekend
guest) were true from the start and stay green as regression guards.

- `vitest run src/components/weekend/FamilyCard.test.tsx` — 73/73 passed
- `vitest run` (full frontend suite) — 450 files / 6887 tests passed
- `tsc --noEmit` — clean
- `eslint src/components/weekend/FamilyCard.tsx src/components/weekend/FamilyCard.test.tsx` — 0 errors (2 pre-existing warnings from the already-shipped sort code, untouched by this diff)
- `scripts/pre-push-verify.sh` — ALL CHECKS PASSED (prettier, eslint, tsc, vitest)

Closes #2254.

### DIFF
--- a/frontend/src/components/weekend/FamilyCard.test.tsx
+++ b/frontend/src/components/weekend/FamilyCard.test.tsx
@@ -644,6 +644,97 @@ describe('FamilyCard — what it shows', () => {
     render(<FamilyCard party={party()} onOpen={vi.fn()} />)
     expect(screen.queryByText('Whole building')).not.toBeInTheDocument()
   })
+
+  describe('single-parent flag (kindred#2254 half 2)', () => {
+    // #2072's own scoping ruling replaces only the two need chips
+    // (`Private bathroom`/`Power`) with the glyph gutter — `Whole building`,
+    // the warn chips and the share chips, including this one, stay as
+    // words. So this reuses the muted `Near another family` chip's exact
+    // grammar rather than inventing a new visual channel, and does not wait
+    // on #2072.
+    it('flags a household with exactly one attending adult', () => {
+      // The default fixture already has one adult (`Emma Johnson`).
+      render(<FamilyCard party={party()} onOpen={vi.fn()} />)
+      expect(screen.getByText('Single parent')).toBeInTheDocument()
+    })
+
+    it('matches the muted "Near another family" chip\'s classes exactly — no new tone invented', () => {
+      render(
+        <FamilyCard
+          party={party({
+            share: {
+              preference: 'yes_share',
+              proximity: ['near'],
+              request_text: '',
+              needs_resolution: false,
+            },
+          })}
+          onOpen={vi.fn()}
+        />
+      )
+      const singleParentChip = screen.getByText('Single parent')
+      const nearChip = screen.getByText('Near another family')
+      expect(singleParentChip.className).toBe(nearChip.className)
+    })
+
+    it('says nothing when two adults are attending', () => {
+      render(
+        <FamilyCard
+          party={party({
+            adults: [
+              { adult_number: 1, display_name: 'Emma Johnson', relationship: 'Mother' },
+              { adult_number: 2, display_name: 'Liam Johnson', relationship: 'Father' },
+            ],
+          })}
+          onOpen={vi.fn()}
+        />
+      )
+      expect(screen.queryByText('Single parent')).not.toBeInTheDocument()
+    })
+
+    // A placeholder slot ("NA", "-", ...) is not an attending adult
+    // (`isAttendingAdultName`, kindred#1925) — a household with ONE real
+    // adult and four placeholder slots is still one attending adult, and a
+    // household with ZERO real adults is a data gap, not a single parent.
+    it('does not count a placeholder slot as a second parent', () => {
+      render(
+        <FamilyCard
+          party={party({
+            adults: [
+              { adult_number: 1, display_name: 'Emma Johnson', relationship: 'Mother' },
+              { adult_number: 2, display_name: 'NA', relationship: '' },
+            ],
+          })}
+          onOpen={vi.fn()}
+        />
+      )
+      expect(screen.getByText('Single parent')).toBeInTheDocument()
+    })
+
+    it('says nothing when no adult is named at all — a data gap, not a single parent', () => {
+      render(<FamilyCard party={party({ adults: [] })} onOpen={vi.fn()} />)
+      expect(screen.queryByText('Single parent')).not.toBeInTheDocument()
+    })
+
+    // Person-grain parties (adult weekend guests) ARE their own identity —
+    // there is no separate "attending adults" list to be short one of.
+    // `attendingAdults` already returns `[]` for this grain, so the
+    // household-composition question does not apply.
+    it('never flags an adult weekend guest (person grain) as a single parent', () => {
+      render(
+        <FamilyCard
+          party={party({
+            grain: 'person',
+            display_name: 'Priya Patel',
+            adults: [{ adult_number: 1, display_name: 'Priya Patel' }],
+            children: [],
+          })}
+          onOpen={vi.fn()}
+        />
+      )
+      expect(screen.queryByText('Single parent')).not.toBeInTheDocument()
+    })
+  })
 })
 
 describe('FamilyCard — spec §3.8, what must stay off it', () => {
--- a/frontend/src/components/weekend/FamilyCard.tsx
+++ b/frontend/src/components/weekend/FamilyCard.tsx
@@ -478,6 +478,14 @@ function FamilyCardChips({
   const wantsToShare = proximity.includes('with') || proximity.includes('similar_ages')
   const wantsNear = proximity.includes('near')
   const conflictDetail = answersConflictDetail(party.share)
+  // kindred#2254 half 2: single-parent, derived from the ATTENDING adult
+  // list, not `party.adults.length` or `party_size` -- both of those count
+  // listed-but-not-attending adults (kindred#1925/#2046) and would
+  // false-positive a two-parent household where one adult never RSVP'd for
+  // this session. `computeAttendingAdults` already grain-gates to household
+  // parties (an adult weekend guest IS its own identity, not a household of
+  // one), so no separate `isHousehold` check is needed here.
+  const isSingleParent = computeAttendingAdults(party).length === 1
 
   return (
     <span className="flex flex-wrap gap-1">
@@ -520,6 +528,11 @@ function FamilyCardChips({
       {/* NEAR and WITH are different requests: NEAR is satisfied by map
             distance between units, WITH by putting both in one room. */}
       {wantsNear && <Chip label="Near another family" tone="muted" />}
+      {/* #2072's own scoping ruling replaces only the two NEED chips
+            (`Private bathroom`/`Power`) with its glyph gutter -- this is not
+            a housing-need flag, so it stays a word, reusing this same
+            `Near another family` muted grammar rather than a new one. */}
+      {isSingleParent && <Chip label="Single parent" tone="muted" />}
 
       {party.is_returning === true && (
         <span className="text-forest-700 dark:text-forest-300 inline-flex items-center gap-0.5 text-xs font-semibold">


adamflagg:
## Independent review — PR #2401 (item #2254 half 2)

**Scope check**: diff is exactly two hunks, both purely additive (+13/-0 `FamilyCard.tsx`, +91/-0 test file). No touch to the shipped youngest-first sort (half 1, #2319) and no touch to #2072/#2212/#2275/#2332's shared file. Matches the issue's half-2 scope exactly — not more, not less.

**Derivation verified against source**: `attendingAdults()` in `householdIdentity.ts` grain-gates to `party.grain === 'household'` (returns `[]` for person-grain) and filters via `isAttendingAdultName`, which strips the placeholder tokens (`na`, `n/a`, `none`, `-`, `0`, `no`). The PR's claim that this differs from `party.adults.length`/`party_size` (which count listed-but-not-attending adults) checks out against the code, not just the comment.

**Chip reuse verified**: both `Near another family` and `Single parent` use `tone="muted"`, no `icon`/`title`, so they resolve to the identical `CHIP_TONE.muted` class string — confirmed by reading `Chip`'s implementation, not just trusting the test.

**Mutation-check** (own verification, not just re-running the author's suite):
1. Set `isSingleParent = false` → 3 of the new tests went RED (chip-renders, exact-class-match, placeholder-filtering tests all failed as expected). Restored, 73/73 green again.
2. Swapped the derivation for `(party.adults ?? []).length === 1` (the exact wrong-field mistake the code comment warns against) → 2 tests went RED, including "never flags an adult weekend guest" (false positive on person-grain). Restored, 73/73 green again.

Both mutations pin real defects; the tests are not vacuous.

**CI**: `CI Summary` (the required check) is COMPLETED/SUCCESS, confirmed via `statusCheckRollup`, not inferred from an absent row. All 15 non-skipped checks passed. Also independently ran `vitest run FamilyCard.test.tsx` (73/73), `eslint` (0 errors; the 3 warnings present are all pre-existing, unchanged by this diff — confirmed against `main`), and `tsc --noEmit` (clean) locally.

**closingIssuesReferences**: `[2254]` only, via GraphQL — correct, since half 1 already shipped and this PR completes the issue's remaining scope.

**PII / privacy**: no CampMinder IDs, no docs/plans/ files, only fictional names already on the approved list (Emma Johnson, Liam Johnson, Priya Patel).

**GUI impact**: `frontend/src/components/weekend/FamilyCard.tsx` is a visual change (new chip on a staff-facing card) — flagging for owner eyeball per policy, holding rather than merging.

No findings. Nothing fixed, nothing filed — the diff is clean as submitted.

Issue #2394, opened 2026-08-16
bug(sync): every persons sync fails — orphan deletes hit a unique-index collision on blanked relations


adamflagg:
## ⚠️ CORRECTED 2026-08-16 — phase D triage. RE-SCOPED. Read this before the body below.

Verified against `main` @ `a946565a` and the **production** snapshot `pocketbase/pb_data/data-prod.db`.
The body's symptom and both its headline measurements reproduce **exactly**. What it gets wrong is *why*
the delete happens, and therefore what the fix is.

### 1. ⛔ THE DELETE ITSELF IS THE BUG. The unique index is not the defect — it is the safety net.

The body treats the uniqueness violation as the problem and proposes cascade / pre-delete / refuse-and-queue
as the fix. All three answer "how should this delete behave". **The delete should not be happening at all.**

Root cause, traced end to end:

1. `gatherPersonIDs` builds the sync's ID list from **attendees + staff** (`persons.go:128-197`). Staff are
   fetched deliberately, so `staff.person` can be populated — `:1699-1702` says so.
2. `transformPersonToPB` returns `nil, nil` for anyone whose CampMinder record has **no `CamperDetails`
   block** (`persons.go:598-608`) — i.e. someone who has never been a camper.
3. `processPerson` sees the nil, bumps `skippedStaff` and `Stats.Skipped`, and **returns before
   `TrackProcessedKey`** (`:477-482` versus `:511`).
4. `deleteOrphans` treats every untracked `(cm_id, year)` in `persons` as an orphan (`:1136-1160`).

**So the orphans are staff who have never been campers.** The sync fetches them on purpose and then
deletes them for not being campers. 26 of them, every night.

Production, 2026: **299** `persons` rows carry `is_camper = 0`, and **277 of them are referenced by
`staff.person`**. The 26 are the subset with no `CamperDetails` at all.

**Fix the index alone and you make this worse.** The collision is currently the only thing *stopping* the
wrong delete. Remove it and the sync starts successfully deleting ~26 staff person-records a night,
blanking `staff.person` as it goes. `staff` already carries exactly **1** blanked `person` — one that got
through.

**Owner ruling 2026-08-16: do BOTH.**
- **(a) Track staff-only persons as processed** so they never become orphan candidates — the actual fix,
  at `persons.go:477-482`.
- **(b) Harden the collision-prone relations anyway**, so a genuinely-correct future orphan delete behaves
  properly instead of erroring. This is defence in depth, not the fix.

[...]

Issue #2307, opened 2026-08-13
chore(config): standardise on "enrollment" — 38 one-L "enrolment" uses remain, including two user-visible strings


adamflagg:
## ⚠️ CORRECTED 2026-08-17 — phase D execution. Recounted on `main` @ `a946565a`: **47 lines / 48 occurrences across 17 files**, not 32.

The 2026-08-16 correction below asserts 32. That does not reproduce. Measured, denominator stated:

```
grep -rIn --binary-files=without-match -i 'enrolment' . | grep -v node_modules | grep -v '^\./\.git/'
```

- **47 lines / 48 occurrences across 17 files.** `lodging_assignments_sync.go:859` carries two on one line, which is the 47-vs-48 gap.
- Split: **38 code lines / 39 occurrences across 16 files** (`*.go`, `*.py`, `*.ts`, `*.tsx`, `*.js`) plus **9 doc lines in one file** (`docs/architecture/sync-layer.md`).
- The 2026-08-14 correction's figures (36 code lines / 37 occurrences) were right for base `3eb44ba4`. Two lines have been added since: **`pocketbase/sync/camper_transportation.go` is now 4, not 2** — `1649a9e4` (#2389) added two more while rewriting that file's header.

**`docs/architecture/sync-layer.md` had to be in scope, and the "A PR for this closes #2307 only" line below is superseded.** The heading at `sync-layer.md:180` is `## Reading Derived Informational Tables (Active-Enrolment Filtering)`, and both `pocketbase/sync/quest_registrations.go:43` and `pocketbase/sync/camper_dietary.go:49` already cite it with the **two-L** spelling. Leaving the heading alone would have kept two live cross-references pointing at a heading that does not exist under that name.

**Batched with #2398 as agreed** — the file sets overlap and two PRs would have conflicted.

---

## ⚠️ CORRECTED 2026-08-16 — phase D triage. Recounted: 32, not 38. And it must run LAST.

Verified against `main` @ `a946565a`.

### The count is 32, not 38

`grep -rn "enrolment"` across `.go`, `.py`, `.ts`, `.tsx` (excluding `node_modules`) returns **32**
occurrences. Recount before acting — counts drift, and this one already has.

### ⚠️ Two traps before renaming anything

1. **Separate the user-visible strings from the internal ones.** A user-visible wording change deserves the
   owner's eye; internal renames do not. Enumerate which of the 32 are rendered.
[...]

PR #2402, opened 2026-08-17
fix(sync): gate deleteOrphans on SyncSuccessful in staff_applications and staff_vehicle_info

### LINKED ISSUE #2301: refactor(sync): evaluate whether the four standalone sync services should embed BaseSyncService

adamflagg:
## Defect

`staff_applications.go` and `staff_vehicle_info.go` each declared a `SyncSuccessful` field but never read it in their own `deleteOrphans`, and each only set it to `true` **after** the orphan sweep had already run. An orphan sweep that runs after a PARTIAL fetch deletes rows the feed simply did not return — `OrphanSweepGuard`'s collapse check alone does not catch that case, only a total-or-near-total collapse.

Verified against the tree before the fix: `deleteOrphans` in both files had no `SyncSuccessful` check at all, and it was only set `true` post-sweep.

## Fix

Mirrors the four already-correct services (`camper_dietary.go`, `camper_transportation.go`, `quest_registrations.go`, `household_demographics.go`):

- `Sync()` now sets `SyncSuccessful = len(records) > 0` immediately after extraction, before the sweep runs (instead of `true` unconditionally in dry-run, and `true` again after a successful sweep).
- `deleteOrphans` now returns `(0, nil)` and logs a skip when `SyncSuccessful` is false, before consulting `OrphanSweepGuard`.

Also corrects `orphan_guard.go`'s doc comment from "ten services" to **nine** — `camper_history` was removed entirely in #2366. Verified against the tree: the nine non-`BaseSyncService`-embedding services that construct their own `OrphanSweepGuard` are `camper_dietary`, `camper_transportation`, `family_camp_derived`, `household_demographics`, `normalize_geographic`, `quest_registrations`, `staff_skills`, `staff_applications`, and `staff_vehicle_info` (`family_camp_derived` builds three, one per derived table, so it's still counted once). This matches the issue's own 2026-08-16 correction — no further edit to the issue body was needed.

Does **not** perform the `BaseSyncService` embedding refactor #2301 originally proposed. The survey referenced in the issue is done and the owner-approved answer is "do not embed" — a speculative refactor with hedged benefits is exactly what the standing "fewer issues, not more" instruction excludes. `base_sync.go`, `persons.go`, and `orchestrator.go` are untouched; they belong to sibling work in flight on this same file set (`orphan_guard.go` is shared with #2394 and #2307).

## Tests

`TestStaffApplicationsDeleteOrphansSkipsWhenSyncNotSuccessful` and `TestDeleteOrphansSkipsWhenSyncNotSuccessful` (staff_vehicle_info) were written first and verified failing against the unfixed code — `deleted = 1, want 0`, exit 1 — proving `deleteOrphans` previously ignored `SyncSuccessful` and swept an orphan row even with the field left at its zero value.

The four existing tests that call `deleteOrphans` directly (bypassing `Sync()`) needed `SyncSuccessful = true` set explicitly once the gate was added, mirroring the pattern `camper_dietary_test.go` already established. This was confirmed necessary, not cosmetic: running them before the update showed all four failing (`expected an error ... got nil`, `deleted = 0, want 1`), because they now hit the new gate before reaching `OrphanSweepGuard`.

## Verification

- `go build ./...` — pass
- `go test ./sync/...` (no `-race`) — pass, 96s
- `golangci-lint run` — 0 issues
- `gofmt -l` on all changed files — clean
- `go test -race ./sync/...` scoped to the affected test families (`TestStaffApplications*`, `TestDeleteOrphans*`, `TestStaffVehicleInfo*`, `TestOrphanSweepGuard*`) — 26 passed, 0 failed, 34s

Note: a full unscoped `pre-push-verify.sh` run (`go test -race ./...`) hit the test binary's 10-minute wall-clock timeout mid-run on this host. At the time, 3-4 sibling agents were each running their own full `go test -race ./...` on the same machine concurrently (confirmed via `ps aux`), and the goroutines still parked when it timed out were in files this PR does not touch (`orchestrator_test.go`, `family_camp_registration_text_test.go`, `base_sync_test.go`). The scoped `-race` run above targets every test in the affected families and passed cleanly in 34s, which is the evidence this PR's change is race-safe; the full-suite timeout looks like host contention, not a regression.

Closes #2301.

### DIFF
--- a/pocketbase/sync/orphan_guard.go
+++ b/pocketbase/sync/orphan_guard.go
@@ -51,13 +51,14 @@ const (
 // fifteenth, every guarded service now fills in this struct.
 //
 // The COLLAPSE arm reaches every sweep in the package. BaseSyncService fills this
-// struct in for the sweeps it owns, and the ten services that do not embed it --
-// camper_dietary, camper_history, camper_transportation, family_camp_derived,
-// household_demographics, normalize_geographic, quest_registrations, staff_skills,
-// staff_applications and staff_vehicle_info -- each construct one inside their own
-// deleteOrphans (kindred#2280, kindred#2296). family_camp_derived is the tenth and
-// the last to arrive: it performs THREE sweeps rather than one, so it builds three
-// guards, one per derived table.
+// struct in for the sweeps it owns, and the nine services that do not embed it --
+// camper_dietary, camper_transportation, family_camp_derived, household_demographics,
+// normalize_geographic, quest_registrations, staff_skills, staff_applications and
+// staff_vehicle_info -- each construct one inside their own deleteOrphans
+// (kindred#2280, kindred#2296). camper_history was the tenth until it was removed
+// entirely in kindred#2366. family_camp_derived is the ninth and the last to arrive:
+// it performs THREE sweeps rather than one, so it builds three guards, one per
+// derived table.
 //
 // The REJECTION arm is narrower, and that is the part to watch. Only BaseSyncService
 // fills in Rejected, so for those ten SkipReason and RejectionsExplainShortfall can
--- a/pocketbase/sync/orphan_guard.go
+++ b/pocketbase/sync/orphan_guard.go
@@ -61,7 +61,7 @@ const (
 // derived table.
 //
 // The REJECTION arm is narrower, and that is the part to watch. Only BaseSyncService
-// fills in Rejected, so for those ten SkipReason and RejectionsExplainShortfall can
+// fills in Rejected, so for those nine SkipReason and RejectionsExplainShortfall can
 // never fire -- they build the guard without it. That is harmless today because none
 // of them counts a rejection, and nothing pins it: a future reclassification into one
 // of those files gets the collapse guard but no rejection protection and no warning.
--- a/pocketbase/sync/staff_applications.go
+++ b/pocketbase/sync/staff_applications.go
@@ -186,12 +186,21 @@ func (s *StaffApplicationsSync) Sync(ctx context.Context) error {
 	}
 	slog.Info("Extracted staff application records", "count", len(records))
 
+	// The extraction finished without error, so len(records) is now a fact about
+	// the SOURCE rather than about whether this run worked. Gate the sweep on it:
+	// a year in which nobody has an application yet is a legitimately empty
+	// upstream, not a collapse, and refusing there wedged the table -- a refused
+	// sweep never clears the rows, so `existing` stayed high and every later run
+	// refused again. This mirrors camper_dietary.go / camper_transportation.go /
+	// quest_registrations.go / household_demographics.go (kindred#2283,
+	// kindred#2301): deleteOrphans reads this instead of running unconditionally.
+	s.SyncSuccessful = len(records) > 0
+
 	if s.DryRun {
 		slog.Info("Dry run mode - extracted but not writing",
 			"records", len(records),
 		)
 		s.Stats.Created = len(records)
-		s.SyncSuccessful = true
 		return nil
 	}
 
@@ -236,7 +245,6 @@ func (s *StaffApplicationsSync) Sync(ctx context.Context) error {
 		return fmt.Errorf("orphan sweep refused: %w", err)
 	}
 
-	s.SyncSuccessful = true
 	slog.Info("Staff applications extraction completed",
 		"year", year,
 		"created", s.Stats.Created,
@@ -1168,6 +1176,17 @@ func (s *StaffApplicationsSync) deleteOrphans(
 	existingRecords map[string]string,
 	year int,
 ) (int, error) {
+	// An orphan sweep that runs after a PARTIAL fetch deletes rows the feed
+	// simply did not return. Sync() sets SyncSuccessful from the size of this
+	// run's extraction, so a year nobody has applied for yet skips the sweep
+	// and succeeds rather than refusing forever (kindred#2301). The guard below
+	// still owns the case that matters: a source that came back SHORT.
+	if !s.SyncSuccessful {
+		slog.Info("Skipping orphan deletion: the source returned no rows for this year",
+			"entity", "staff_applications", "year", year)
+		return 0, nil
+	}
+
 	guard := OrphanSweepGuard{
 		Entity:   "staff_applications",
 		Year:     year,
--- a/pocketbase/sync/staff_applications_test.go
+++ b/pocketbase/sync/staff_applications_test.go
@@ -423,8 +423,14 @@ func seedStaffApplication(t *testing.T, app core.App, cmID, year int) string {
 // Gap 2. staff_applications builds its computed set from the SAME
 // loadPersonStaffMapping(ctx, year) gate as staff_vehicle_info, so it has the
 // identical year-wipe path: an empty staff mapping makes every value fail the
-// gate, the computed set comes back empty, and the unguarded sweep deletes the
-// whole year and then sets SyncSuccessful = true.
+// gate, the computed set comes back empty, and (pre-kindred#2279) the
+// unguarded sweep deleted the whole year and reported success.
+//
+// NOTE since kindred#2301: this pins OrphanSweepGuard's CONTRACT, not a state
+// Sync() can now produce on its own -- Sync() sets SyncSuccessful from
+// len(records), so an empty computed set skips the sweep before the guard is
+// even consulted. SyncSuccessful is set explicitly here because this drives
+// deleteOrphans directly rather than through Sync().
 func TestStaffApplicationsDeleteOrphansRefusesEmptyComputedSet(t *testing.T) {
 	t.Parallel()
 	app := newStaffApplicationsTestApp(t)
@@ -432,6 +438,7 @@ func TestStaffApplicationsDeleteOrphansRefusesEmptyComputedSet(t *testing.T) {
 
 	s := NewStaffApplicationsSync(app)
 	s.Year = 2026
+	s.SyncSuccessful = true
 
 	existing := map[string]string{makeStaffAppKey(1001, 2026): recID}
 	deleted, err := s.deleteOrphans(context.Background(),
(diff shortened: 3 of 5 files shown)


adamflagg:
## Independent review — #2402

**Scope check:** diff matches issue #2301's phase-D-corrected body exactly — the two MECHANICAL items (SyncSuccessful gate in `staff_applications.go`/`staff_vehicle_info.go`, and the "ten"→"nine" doc-comment fix in `orphan_guard.go`) and nothing more. The `BaseSyncService` embedding refactor was correctly skipped per the owner-approved "do not embed" answer already on the issue. No scope creep, no silent narrowing.

**Verified against the tree, not just the diff:**
- `Sync()` in both files now resets `SyncSuccessful = false` at the top, sets it to `len(records) > 0` right after extraction, and `deleteOrphans` reads it before the sweep — this is an exact mirror of the already-correct pattern in `camper_dietary.go` (line-for-line comment structure included).
- The nine-service count in the COLLAPSE-arm paragraph checks out: grepped every listed service and confirmed none embeds `BaseSyncService`; `camper_history.go` is indeed gone (removed in #2366, confirmed via `git log`).
- `go build ./...`, `go vet ./sync/...`, `golangci-lint run ./sync/...` (0 issues), `gofmt -l` (clean), and the full `go test ./sync/...` (non-race, 57s) all pass locally on the pushed HEAD.
- Pre-push-hook claim checked against `.lefthook.yml`: `pre-push` → `go-build` runs only `go build .`, no `go test`. The PR body's claim that the gating hook doesn't run the full suite is accurate.

**Mutation-check (pins the defect, not just green-on-green):** removed the `!s.SyncSuccessful` early-return block from both `deleteOrphans` methods, ran `TestStaffApplicationsDeleteOrphansSkipsWhenSyncNotSuccessful` and `TestDeleteOrphansSkipsWhenSyncNotSuccessful` — both went RED (`deleted = 1, want 0`). Restored the files, confirmed clean diff and a passing rebuild. The tests pin the fix.

**Fixed in-branch (one finding, in code this PR touched):** the "ten"→"nine" doc-comment correction in `orphan_guard.go` was incomplete — a second occurrence three lines below the one the author fixed ("...so for those **ten** SkipReason and RejectionsExplainShortfall can never fire...", in the REJECTION-arm paragraph) still said "ten" against the same nine-service set. Changed to "nine" in `9b5e0a5d`, re-ran build/vet/lint/tests, pushed, confirmed `git rev-list --count origin/...HEAD` = 0.

**CI:** "CI Summary" is COMPLETED/SUCCESS at HEAD `9b5e0a5d` (not absent, not a stale run against a prior commit).

**closingIssuesReferences:** `{2301}` — correct, matches the issue this PR is meant to close.

**GUI impact:** none — diff touches only `pocketbase/sync/*.go`.

**PII/real-name scan:** clean — test data uses the existing fictional numeric CM-ID convention (1001/1002/2026), no real names or production IDs anywhere in the diff.

No other findings above the review floor.

PR #2403, opened 2026-08-17
fix(sync): adopt SkippedValues at the last two mechanism-C sites

adamflagg:
## Defect

`quest_registrations.go` and `household_demographics.go` each had one more
"mechanism-C" silent drop from kindred#2257's inventory: a custom-field
VALUE admitted by a prefix filter but missing a case in the exact-match
column mapper was discarded with no counter and no log line.

- `mapQuestFieldToRecord` (quest_registrations.go): `isQuestRegistrationField`
  admits by prefix (`Quest-`/`Q-`/`Quest `), but `MapQuestFieldToColumn` is
  a finite 45-case exact-match switch — any admitted field outside that
  switch vanished silently.
- `mapPersonFieldToRecord` (household_demographics.go): `IsHHField` admits
  by the `HH-` prefix, but `MapHHFieldToColumn` is a finite 17-case
  exact-match switch — same gap.

Both are the exact shape already fixed at `camper_transportation.go`
(kindred#2356), `staff_applications.go`, and `staff_vehicle_info.go`
(kindred#2277) — a prefix-admission filter strictly wider than the
exact-match mapping table underneath it.

## Fix

Both mapping functions (`mapQuestFieldToRecord`, `mapPersonFieldToRecord`)
now return the resolved column name, or `""` when unmapped. Their callers
(`aggregateQuestEntries` / `aggregateToRows`) count discard events by field
name, fold the total into `Stats.SkippedValues`, and emit one aggregated
`slog.Warn` per run — the same shape `camper_transportation.go` and
`staff_applications.go` use, not a per-value log line.

`aggregateQuestEntries` stays a pure, non-method function (its own comment
already commits to that, for testability — see `TestAggregateQuestEntriesIsOrderIndependent`
etc.), so it returns the discard counts as a second named value rather
than reaching into `Stats` directly; the caller, `loadPersonCustomValues`,
owns `Stats` and does the folding.

`orchestrator.go`'s `Stats.SkippedValues` doc comment (the only change in
that file — a doc comment, no logic) is updated so its "nothing but those
services increments this" sentence names all five adopters instead of
three.

## Verified unreachable, left untouched

Per the issue's explicit instruction, and confirmed against the tree
before touching anything:

- `household_demographics.go`'s `mapHouseholdFieldToRecord` has its own
  `column == ""` early return, but its admission filter
  (`isHouseholdDemographicsField`'s non-`HH-` branch: `Synagogue`,
  `Center`, `Custody Issues`, `Board`) and `MapHouseholdFieldToColumn`'s
  4-case switch are the identical 4-name set — that branch is structurally
  unreachable.
- `camper_dietary.go`'s equivalent (`isCamperDietaryField`'s 5 admitted
  `Family Medical-*` names vs. `MapDietaryFieldToColumn`'s 5-case switch)
  is the same identity — also unreachable.

Neither file is touched by this PR.

## What this deliberately did not do

- Did not touch `camper_dietary.go`, `persons.go`, `base_sync.go`,
  `orphan_guard.go`, `staff_applications.go`, or `staff_vehicle_info.go`.
- Did not add the BUS-*/App-* style "known-retired vs. unrecognized"
  field classification `camper_transportation.go`/`staff_applications.go`
  carry — that split rests on domain research (which CampMinder fields
  are retired-but-expected) this item didn't do for Quest-*/HH-* fields.
  The aggregated warning here logs the raw per-field discard counts
  instead; a follow-up can add classification if a real unmapped field
  ever shows up.
- Did not add `t.Parallel()`-incompatible log-content assertions —
  `captureSweepLogs` (used by the three existing adopters' tests) swaps
  slog's process-global default handler and requires a serial-test
  exemption entry in `pocketbase/main_test_parallelism_test.go`, a file
  outside this item's ownership list. Both new tests assert on
  `Stats.SkippedValues` directly instead and run with `t.Parallel()`.
- This retires zero issue numbers. The two sites fixed here were never
  filed as children of kindred#2257 — it stays open as the umbrella
  tracking issue.

## Testing

- New failing-first Go tests (verified red via a real build failure
  before the signature changes landed, then green after):
  `TestQuestLoadPersonCustomValuesCountsUnmappedFields`,
  `TestAggregateQuestEntriesReturnsUnmappedCounts`,
  `TestAggregateToRowsCountsUnmappedPersonFields`,
  `TestMapPersonFieldToRecordReturnsTheResolvedColumn`.
- Mutation-checked: neutering the `Stats.SkippedValues +=` increment in
  both files turned the two DB-backed tests red (confirmed, then
  restored), so they pin real behavior rather than passing vacuously.
- `go build ./...`, `go vet ./...`, `go test ./sync/...` (full package,
  135s and again 92s), and `golangci-lint run --config ../.golangci.yml
  ./sync/...` (0 issues) all pass on the final diff.

Refs kindred#2257.

### DIFF
--- a/pocketbase/sync/household_demographics.go
+++ b/pocketbase/sync/household_demographics.go
@@ -562,9 +562,31 @@ func (s *HouseholdDemographicsSync) aggregateToRows(
 	}
 
 	// Process person custom values (HH- fields) -> _summer columns
+	//
+	// unmappedCounts tracks discard events: an HH- field accepted by IsHHField
+	// (the prefix test) that has no case in MapHHFieldToColumn (kindred#2257,
+	// adopting the mechanism shipped for camper_transportation.go and
+	// staff_vehicle_info.go at kindred#2356/#2277).
+	unmappedCounts := make(map[string]int)
 	for _, v := range personValues {
 		rec := getRecord(v)
-		s.mapPersonFieldToRecord(rec, v.fieldName, v.value)
+		if column := s.mapPersonFieldToRecord(rec, v.fieldName, v.value); column == "" {
+			unmappedCounts[v.fieldName]++
+		}
+	}
+	if len(unmappedCounts) > 0 {
+		total := 0
+		for _, n := range unmappedCounts {
+			total += n
+		}
+		s.Stats.SkippedValues += total
+
+		// One aggregated warning per run, not one per discarded value.
+		slog.Warn("Household demographics: discarding values for unmapped HH-* fields",
+			"year", year,
+			"discard_events", total,
+			"unmapped_fields", unmappedCounts,
+		)
 	}
 
 	// Process household custom values -> _family columns
@@ -623,11 +645,16 @@ func (s *HouseholdDemographicsSync) setColumn(rec *householdDemographicsRecord, 
 // The four boolean arms are unchanged. They are logical ORs over the
 // contributing values, never first-wins, so they were never part of the defect;
 // at this grain they OR over one camper's answers instead of the household's.
-func (s *HouseholdDemographicsSync) mapPersonFieldToRecord(rec *householdDemographicsRecord, fieldName, value string) {
+// It returns the resolved column name, or "" if fieldName has no case in
+// MapHHFieldToColumn -- the caller (aggregateToRows) uses this to count
+// discarded values (kindred#2257).
+func (s *HouseholdDemographicsSync) mapPersonFieldToRecord(
+	rec *householdDemographicsRecord, fieldName, value string,
+) string {
 	// Use MapHHFieldToColumn for the field mapping
 	column := MapHHFieldToColumn(fieldName)
 	if column == "" {
-		return // Unknown field
+		return "" // Unknown field
 	}
 
 	switch column {
@@ -677,6 +704,8 @@ func (s *HouseholdDemographicsSync) mapPersonFieldToRecord(rec *householdDemogra
 	case "form_filler":
 		s.setColumn(rec, &rec.formFiller, column, value)
 	}
+
+	return column
 }
 
 // mapHouseholdFieldToRecord maps a household custom field onto the
--- a/pocketbase/sync/household_demographics_test.go
+++ b/pocketbase/sync/household_demographics_test.go
@@ -407,6 +407,62 @@ func TestAggregateKeepsEveryPersonAnswer(t *testing.T) {
 	}
 }
 
+// TestAggregateToRowsCountsUnmappedPersonFields pins the kindred#2257
+// mechanism-C fix in this file: an HH- field admitted by IsHHField (the
+// prefix test) but missing a case in MapHHFieldToColumn used to be discarded
+// by mapPersonFieldToRecord's `if column == "" { return }` with no counter
+// and no log line -- the same silent drop already fixed at
+// camper_transportation.go and staff_vehicle_info.go (kindred#2356/#2277).
+// The record itself is still created from the routed field on the same row --
+// this is a VALUE discard, not a record one, so it must land on
+// Stats.SkippedValues, not Stats.Skipped.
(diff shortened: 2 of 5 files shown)


adamflagg:
## Independent review — PR #2403 (kindred#2257 phase D)

**Verdict: clean.** No changes made in-branch; nothing filed.

### What I verified

- **Scope matches the claim exactly.** Diff touches only `household_demographics.go`(+test), `quest_registrations.go`(+test), and `orchestrator.go` (doc comment only, confirmed no logic change). No other file in the "did not touch" list was touched.
- **Reachability claims checked against the tree, not taken on faith:**
  - `quest_registrations.go`: `isQuestRegistrationField` is a genuine 3-prefix match (`Quest-`/`Q-`/`Quest `) against `MapQuestFieldToColumn`'s 45-case exact-match switch — strictly wider, so the site is reachable. Confirmed.
  - `household_demographics.go` (`mapPersonFieldToRecord`): `IsHHField` prefix vs. `MapHHFieldToColumn`'s 17-case switch — same shape, reachable. Confirmed.
  - `household_demographics.go` (`mapHouseholdFieldToRecord`, left untouched): `isHouseholdDemographicsField`'s non-HH branch admits exactly `{Synagogue, Center, Custody Issues, Board}`, and `MapHouseholdFieldToColumn`'s case set is the identical 4 names — `column == ""` is provably unreachable. Correctly left alone.
  - `camper_dietary.go` (left untouched): `isCamperDietaryField`'s 5 admitted `Family Medical-*` names are identical to `MapDietaryFieldToColumn`'s 5-case switch — also provably unreachable. Correctly left alone.
- **`closingIssuesReferences` is empty** via GraphQL — correct, matches the issue's explicit "this retires ZERO issue numbers."
- **No GUI-impact files** (pure Go, `pocketbase/sync/`).
- **No PII / real names / CampMinder IDs** in the diff.
- **CI Summary: SUCCESS** (confirmed via `statusCheckRollup`, not just the rollup dashboard — status COMPLETED, conclusion SUCCESS). All 18 checks green.
- **golangci-lint (`--config ../.golangci.yml ./sync/...`): 0 issues.** `go build ./...`, `go vet ./...`, and the full `go test ./sync/...` package (44.6s) all pass locally, matching the PR's own claims.
- **Mutation-checked the actual fix**, not just re-run the tests: neutered the `Stats.SkippedValues += total` line in both `quest_registrations.go` and `household_demographics.go` (replaced with `_ = total`), rebuilt, and re-ran the four new tests.
  - `TestAggregateToRowsCountsUnmappedPersonFields` → **RED**: `Stats.SkippedValues = 0, want 1`
  - `TestQuestLoadPersonCustomValuesCountsUnmappedFields` → **RED**: `Stats.SkippedValues = 0, want 1`
  - `TestMapPersonFieldToRecordReturnsTheResolvedColumn` / `TestAggregateQuestEntriesReturnsUnmappedCounts` stayed green, as expected — they pin the pure-function return value, not `Stats`, and the mutation only touched the Stats-folding step.
  - Restored both files (`git diff` clean afterward) and reran the full suite green.
- **Doc-comment accuracy**: `orchestrator.go`'s updated "five adopters" list is correct — grepped for `Stats.SkippedValues` writes and found exactly 5 files (`camper_transportation.go`, `staff_applications.go`, `staff_vehicle_info.go` [via `++`, not `+=` — the grep pattern had to account for both forms], `quest_registrations.go`, `household_demographics.go`).
- **No double-counting risk**: both `loadPersonCustomValues`/`aggregateToRows` call sites run exactly once per `Sync()`, and the discard loops increment once per unmapped value — `total` is an exact count.
- CodeRabbit was skipped per the author's stated judgment (5-file, single-surface Go diff adopting an established pattern) — reasonable, not requested by me either.

### Disposition
Nothing to fix, nothing to file. PR is ready as-is.

Issue #2339, opened 2026-08-14
chore(lodging): a two-unit alias is double-counted by any consumer that expands member_units


adamflagg:
## What happens

Six rows in `lodging_unit_aliases` map one alias string to **two** units. Any consumer that expands
`member_units` and credits an occupancy to every member turns *one household per unit* into an
apparent share in both.

The six (described by shape, not by name, to keep this body clean):

- two room pairs in one multi-room area, each written as a single "`<room> 1and2`" string
- the two downstairs rooms of one shared building, under two different alias spellings
- the two halves of one house in that same multi-room area
- the two halves of one village unit, numbered `<n>a` / `<n>b`

*(Unit names are deliberately generalised here — the real ones live in the private registry. Resolve them via `lodging_unit_aliases` when working this issue.)*

`lodging_unit_aliases` has 187 rows; 8 alias strings resolve to more than one unit.

## It already produced a wrong answer

A shareability analysis resolved `family_camp_registrations.cabin_assignment` through the alias table
and counted a "share" wherever two households landed on the same unit in the same weekend. Eight
cells were miscounted this way. One area's apparent 18 multi-household occupancies collapsed to
**2** once the split case was excluded — the other 16 were two households allocated one room each,
which is the normal use of those rooms and the opposite of a share.

The rule that was missing: **a string covering N units with H households is only evidence of
over-occupancy when `H > N`.** At `H <= N` the households fit one per unit and the assignment is
ambiguous rather than shared.

## Fix direction

This is small and mostly documentation, but it should live where the next reader will hit it:

- A doc comment on `lodging_unit_aliases` (and in `docs/reference/lodging-registry.md`) stating the
  `H > N` rule explicitly, with the split case named.
- If any shipped consumer expands `member_units` for an occupancy count, it needs the same guard —
  audit them rather than assume. The board's own conflict warning is the one to check first.
- Consider whether a multi-unit alias should carry a flag saying "these are separate locations that
  are routinely allocated independently", since that is what makes the split legitimate rather than
  a data-entry accident.

## Acceptance

- [ ] The `H > N` rule is written at the table and in the registry reference.
- [ ] Every consumer that expands `member_units` is listed here, with a verdict on whether it needs
      the guard.
- [ ] A test for any consumer that gets the guard, using a two-unit alias with exactly two
      households — the case that currently reads as a share.

PR #2404, opened 2026-08-17
fix(frontend): judge merge ambiguity on rooms, so a container is not a false share

### LINKED ISSUE #2371: fix(frontend): two families on a container get flagged 'did not request sharing' while the alias-route equivalent is guarded
## ⚠️ CORRECTED 2026-08-16 — phase D triage. RE-SCOPED by owner ruling.

**⛔ The redistribution-on-split ask below is RETIRED.** Owner ruling 2026-08-15/16: this issue is now the
**consistency defect only** — two container-named parties trip "did not request sharing" while the
alias-route equivalent is guarded. Everything below about redistributing occupancy when a merge splits is
history, not scope. **Classification: SCOPED, M, decision-free.**

**The anchor in the body is CURRENT and correct.** `frontend/src/components/weekend/boardLayout.ts:557-562`:

```ts
function mergeGroupSignature(party: RosterPartyRow): string | null {
  const codes = occupiedCodes(party)
  // A SINGLE named code (including a container) is one confirmed claim, not a
  // multi-unit alias resolution, and must never enter this grouping — see
  // `occupiedCodes`'s own doc comment for why a container name stays as itself.
  if (codes.length < 2) return null
```

**The asymmetry, stated precisely.** `mergeGroupSignature` returns `null` for any party naming fewer than
two codes, so a party naming a **single container** never enters the ambiguous-merge grouping.
`isAmbiguousMergePair` (`:649-654`) therefore returns `false` for two container-named parties, they get
mutually attributed by `overlappingPartyKeys`, and both are chipped. Two parties naming the same **multi-code
alias set** do enter the grouping, satisfy the `H <= N` guard, and are correctly skipped.

Both come out of `occupiedLeafCodes`, which expands a named container to the rooms beneath it — so two
families on one container both expand to the same leaves and "overlap" as an artefact of that expansion,
not because either asked to share.

**Acceptance must include a test for two CONTAINER-named parties.** #2339's suite is entirely multi-code,
which is exactly why this was missed. The routine's own doc comment at `:588-597` records the same class of
bug returning "one level down" once already — that is the precedent for pinning it.

Closes #2371. Closes nothing else.

---

## What

When a building container holding two families is **split**, both families end up credited with **both** leaves. They should be distributed **one family per leaf**, with staff free to swap them around afterwards.

## Why this surfaced now

kindred#2339 (PR #2364) fixed the *reading* side of this: `overlappingPartyKeys` no longer treats two parties that each name the same multi-leaf set as sharing every room with each other. That stopped a false conflict warning, and — because `consentFlag` derives from overlap by design — stopped a false "did not request sharing" line and a false shared-state chip alongside it.

But the underlying **placement** is still wrong. Both families genuinely hold `["<leaf-a>","<leaf-b>"]` in `lodging_assignments.units`; the read path now merely declines to draw a bad conclusion from it. Verified on a 2026 fixture: two households in one two-leaf house each carried the identical two-code set.

The governing rule is already written down — `docs/architecture/lodging-occupancy.md:37-39`:

> An extended family spanning two or more registrations may occupy one house together, **each registration in its own room**. This is not sharing a unit; it is several parties distributed across a house, and it is legitimate.

"Each registration in its own room" is exactly what the split should produce, and does not.

## ⚠️ This changes behaviour the existing split/merge work deliberately chose

The split/merge feature is well built and its current behaviour may be intentional for cases this issue has not considered — a party genuinely too large for one leaf, an odd party count against an even leaf count, a container with three or more leaves, a leaf already occupied by a third party, or a split of a container whose parties came from an alias rather than from a real placement.

**Do not change the split path from this issue alone.** Investigate the use cases first and record what each should do. In particular:

- What happens when parties outnumber leaves, or leaves outnumber parties?
- Does a party that legitimately needs two rooms still get them?
- Is the swap affordance staff would rely on afterwards already there, or does it need building?
- Does the ingest-side alias expansion (`placementFor`, which writes the full member set by design and is documented as deliberately unjudged) need to stay unjudged, with the split being the only place that decides?

## Sequencing

- **Files (expected):** the split/merge path under `frontend/src/components/weekend/`, `dragPlacement.ts`, and whatever writes `lodging_assignments.units` for a split. `pocketbase/sync/lodging_assignments_sync.go`'s `placementFor` is likely **out** of scope — it records what CampMinder holds and the architecture doc says the ingest must not judge occupancy.
- **Predecessors:** kindred#2339 (merged) fixed the read side and is what makes this visible rather than hidden behind a false warning.
- **Related:** kindred#2358 (a widened `bunk_assignments` row still resolves to the first session) is the same family of defect — a placement that names more than it should — in the summer path.
- **Needs a decision before code.** Treat as DECISION, not SCOPED.

adamflagg:
Two families placed on one building container tripped the amber "did not
request sharing" flag, while the exact same claim written as an alias — both
families naming the building's two rooms explicitly — was correctly left
alone. One placement, two different verdicts, decided by nothing but which
spelling the ingest happened to use.

## The defect

`overlappingPartyKeys` measures overlap in ROOMS: `occupiedLeafCodes` expands
every named container down to the rooms beneath it, so two families on one
two-room house both resolve to `{r1, r2}` and intersect on both. That is an
artefact of the expansion, not a request to share.

The `H <= N` guard that exists to excuse exactly that (H households claiming
the same N-room set is only a real double-booking once H > N) was measuring
something else. `mergeGroupSignature` keyed the group on the codes a placement
NAMED and bailed below two of them, so a single container scored N = 1 and
never entered the grouping at all. The alias route satisfied the guard; the
container route could not reach it.

Measured on the two-container fixture before the fix:

- `overlappingPartyKeys` returned **2** party keys where it should return 0
- `buildBoard().flaggedCount` was **1** where it should be 0

`#2339`'s suite, which introduced the guard, is entirely multi-code — six
tests, not one of them on a container. That is why this survived it.

## The fix

Key the group on the EXPANDED leaf set rather than on the named codes, and
compute that expansion ONCE (`leavesByParty`) so the guard and the overlap it
guards read the same value. The guard and the thing it guards now answer the
same question.

Behaviour that is deliberately unchanged, and now pinned:

- **Two families on a ONE-room container still flag.** N = 1, H = 2, so this
  is the H > N arm of the same rule — a container is not a licence to
  co-locate.
- **Three families on a two-room container still flag**, all three.
- **A container family against a family confirmed in one of its rooms still
  flags.** The single-room family is not in the same room-set group, so the
  pair between them was never ambiguous — the same asymmetry the alias route
  already had.
- Every existing alias assertion is untouched and green: a plain leaf, an
  alias pair, an alias trio, and the two board-level cases.

`boardLayout.ts`'s doc comment already recorded this bug class returning "one
level down" once, when the slot-level overlap fix left `FamilyCard`'s
`sharedSlot` re-deriving overlap in its own words. This fix does not move it
again: it changes the ONE routine both `consentFlag` and `sharedSlot` read,
rather than adding a container special case beside it.

## Deliberately not done

- **The redistribution-on-split ask is retired and is not built here.** Both
  families still genuinely hold both leaves in `lodging_assignments.units`;
  this changes only what the board concludes from that, not the placement.
- Nothing in the split/merge path, `dragPlacement.ts`, or
  `lodging_assignments_sync.go` is touched.
- No appearance change. This is logic only — no styling, no markup.

## Testing

TDD. Three assertions written first and confirmed red (`2` where `0` was
expected, twice; `flaggedCount` `1` where `0` was expected), then the fix. The
four guard assertions that were green on first write were mutation-checked
against two broken builds — a guard that ignores `H <= N` (kills three of
them) and a guard that drops the same-group requirement (kills the fourth).

Gates: `tsc --noEmit` 0, `vitest run src/components/weekend` 0 (1195 tests),
`eslint src --max-warnings=386` 0, `prettier --check` 0,
`scripts/pre-push-verify.sh` 0.

Closes #2371.


adamflagg:
## Independent review — PR #2404 (kindred#2371)

**Scope check.** Diff matches the corrected issue exactly: `mergeGroupSignature` re-keyed from named codes to `occupiedLeafCodes`, `leavesByParty` computed once and shared by the grouping and the overlap accumulation. Confirmed only the two owned files were touched in the original diff, and the retired redistribution-on-split ask was genuinely not built — nothing under the split/merge path, `dragPlacement.ts`, or `lodging_assignments_sync.go` changed.

**CI.** Green: 17/17 checks after my push (was 15/15 before). `closingIssuesReferences` = `[2371]` via GraphQL, matching "Closes #2371. Closes nothing else." — no fix needed there.

**"One routine" claim, verified.** `LodgingUnitCard.tsx:367` calls `overlappingPartyKeys(parties, units)` directly and feeds `sharedSlot`; `unitBadges.ts` consumes the already-computed `overlappingParties` count rather than re-deriving overlap independently. So `consentFlag` and the card's `sharedSlot` chip and the sharing badge are all downstream of the one function this PR changed — no fourth silent re-derivation left standing.

**Mutation-check on the new tests** (`overlappingPartyKeys — a CONTAINER is ambiguous on the same terms as an alias`): reverted `mergeGroupSignature` to key on `occupiedCodes(party)` (named codes) instead of the expanded `leavesByParty` set, restoring the pre-fix behavior. Result: exactly 3 of the new container tests went RED (the H==N container-vs-container case, the container-vs-explicit-room-set case, and the board-level `flaggedCount` case), while all 101 other tests — including every pre-existing alias assertion — stayed green:

```
FAIL  ... > does not flag two households each named at the SAME two-room container (H == N)
- Expected 0, + Received 2
FAIL  ... > treats a container name and its explicit room set as the SAME claim
- Expected 0, + Received 2
FAIL  ... > raises no amber flag on the BOARD for two households on one two-room container
- Expected 0, + Received 1
Test Files  1 failed (1)
     Tests  3 failed | 101 passed (104)
```

Restored the fix, confirmed back to 104/104 green and no diff against origin. The tests pin the defect.

**Independent full-suite verification** (not just the author's numbers): `vitest run src/components/weekend` → 1195/1195 passed, `tsc --noEmit` clean, `eslint` clean on the changed file, `markdownlint-cli2` clean.

**No real names / CampMinder IDs** in the diff or tests — `household_cm_id` values are synthetic (500001–500203 range), facility names ("Aspen House", room codes) are fictional test fixtures. No `docs/plans/` files touched.

**Fixed in-branch (pre-existing, small, no product judgement):** `docs/reference/lodging-registry.md`'s kindred#2339 section described the `H <= N` guard purely in terms of an "alias" — narrower than what the code has done since this PR, since `overlappingPartyKeys` now reaches a directly-named multi-room container on the same terms. Added a short clarifying paragraph rather than rewriting the existing argument. (The author's own note also named `docs/architecture/lodging-occupancy.md` as needing the same fix — checked it and that file never actually states the `H <= N` rule at all, so there was nothing to correct there; the one real instance was in `lodging-registry.md`.)

**Holding, not clean.** Two reasons, neither a defect:
1. `boardLayout.ts` is a GUI-impact file by the mechanical check, and this is not a false positive from the broad regex — the fix changes real on-screen output: the amber "did not request sharing" chip no longer renders for a container-named pair that previously got flagged. No JSX/CSS changed, but the visible result did.
2. The author flagged a genuine judgement call, pinned by "treats a container name and its explicit room set as the SAME claim": keying the guard on expanded rooms means a family named at a container and a family naming that container's rooms explicitly now group together as one ambiguous claim, where before only the alias-spelled version did. I agree with the reasoning (same "one household per room, staff haven't said which" state the rule exists for) and it's correctly tested, but it is a product-scope call worth the owner's eyes before merge, not a code defect for me to file or fix around.

Not merged, no issue closed. Orchestrator decides.

PR #2405, opened 2026-08-17
fix(sync): normalise family camp adult date_of_birth and relationship_to_camper

adamflagg:
`processAdults` stored `date_of_birth` and `relationship_to_camper` exactly as the
parent typed them, so two enrolled siblings answering for the same adult slot
"disagreed" whenever they had merely typed the same answer two different ways.
Measured against the production snapshot, all years, over
`(household, year, adult_slot)` groups: **1,124 of 9,388 DOB groups** and
**315 of 5,751 relationship groups** held two or more distinct values — and over
half of that was formatting and case, not disagreement.

This normalises both columns. It does **not** change the merge and does **not**
change the record grain; both remain kindred#2275's open subject.

## The date parser

`normalizeDateOfBirth` rewrites an answer into the single canonical form
`YYYY-MM-DD`. Accepted shapes, every one of which occurs in production:

| shape | example | answers |
|---|---|---|
| `M/D/YYYY` | `9/2/1979` | 10,418 |
| `M/D/YY` | `9/2/79` | 2,042 |
| `M-D-YYYY` | `09-02-1979` | 384 |
| `MMDDYYYY` | `10221978` | 369 |
| `M-D-YY` | `9-2-79` | 147 |
| `M.D.YY` | `2.13.80` | 49 |
| `M.D.YYYY` | `9.2.1979` | 45 |
| `MMDDYY` | `090279` | 35 |
| `YYYY-M-D` | `1979-09-02` | 1 |
| month name | `October 28, 1981`, `Oct 6, 1981` | in the tail |
| day-first month name | `28 Nov 1967`, `9-Oct-1974` | in the tail |
| space separated | `03 16 1976` | in the tail |
| mixed separators | `05-02/1972` | in the tail |

**Two-digit year century rule, stated explicitly:** `YY >= 30` is `19YY`,
`YY < 30` is `20YY`. The pivot is at 30 because it lands in a gap that is empty
in production — the two-digit years actually stored are 01-24 (52 answers,
children's birthdays typed into an adult field) and 43-99 (2,188 answers), with
25-42 absent — so no value present today can be misclassified by it.

**It normalises, it does not discard.** A value the parser cannot read is
returned unchanged, never blanked. **Residual after parsing: 162 of 13,823
answers (1.17%)** — `11/13`, `6274`, `1974`, `None`, `na`, `N/A`. Each is a real
answer someone typed; emptying them would be data loss dressed as a cleanup.
It also does not judge plausibility: a mistyped year (`2/13/2986`) is a
well-formed date and is canonicalised like any other, because rejecting it would
put it straight back in the junk bucket the two earlier passes at this issue
wrongly inflated.

## The relationship normaliser

`normalizeRelationshipToCamper` folds case on a single all-letters token
(`mother`/`MOther`/`MOTHER` → `Mother`) and the two synonym pairs
Mom↔Mother and Dad↔Father. Nothing else.

- **`Mother` and `Father` are never folded into each other.** 92 of the residual
  groups are exactly that pair, and they are two children naming two different
  people into one adult slot. That is the signal, not noise.
- Synonym lookup is **exact-match on the whole value, never a substring**, which
  is what keeps the small step-parent tail (~21 answers: `Step Father`,
  `step mother`, `Stepmom`, `Dad/Stepdad`) from being rewritten into
  `Mother`/`Father`. A substring implementation is mutation-tested against.
- Free text keeps the capitalisation the parent typed; only a single
  all-letters token is re-cased. Re-casing free text too would collapse 2 more
  groups out of 315 and rewrite 100 more stored values — not a trade worth
  making.

## Residual divergence — the actual deliverable

Re-measured with the shipped Go normalisers replayed over every stored value in
`pocketbase/pb_data/data-prod.db` (all years, no roster filter):

| column | slots | divergent before | divergent after | collapsed |
|---|---|---|---|---|
| `date_of_birth` | 9,388 | 1,124 | **541** | 583 (52%) |
| `relationship_to_camper` | 5,751 | 315 | **169** | 146 (46%) |
| **combined** | 9,629 union slots | **1,439** | **710** | 729 (51%) |

2026 alone: DOB 94 → 43, relationship 17 → 8. 2025: DOB 126 → 61,
relationship 27 → 17.

What survives, and why it is real:

- **`date_of_birth`, 541** — 360 different birth YEAR, 105 same year with a
  different month/day, 70 unparseable on one side, 6 month/day transposed.
- **`relationship_to_camper`, 169** — 92 `Father | Mother`, 11
  `Father | Parent`, 11 `Mother | Parent`, and a 55-group tail of genuinely
  different relations and typos.

So the grain decision now rests on **710 residual divergences over 9,629
(household, year, adult_slot) slots — 7.4%**, of which the hard identity core
(different birth year, plus Mother-vs-Father) is 452 slots, 4.7%. The issue's
§5 estimate of "~259 of 315" relationship divergences collapsing under
normalisation was optimistic: the measured figure is 146 of 315, because it was
extrapolated from the nine largest pair shapes and the tail is mostly genuine.

## Operational note

The first sync after this ships rewrites `date_of_birth` on essentially every
`family_camp_adults` row (13,660 of 13,823 source answers change textual form)
and `relationship_to_camper` on 3,107 of 6,969. These are updates, not deletes,
so the orphan-sweep ratio guard is not involved. Both normalisers are
idempotent — pinned by test — so the run after that one rewrites nothing.

## Deliberately NOT done

- **No grain change and no migration.** Full fan-out to
  `(household, year, person_id, adult_number)` is an open owner decision that
  this re-measure exists to inform.
- **No merge-policy change.** First-non-empty-wins over load order still decides
  which sibling's answer is stored, for every attribute except `email`. A
  recency or validity tie-break for DOB would be a policy call, not a
  normalisation.
- **No plausibility validation, no blanking of junk, no deletion of rows.**
- **No change to `adult_number` 3-5**, which carry no attributes at all.
- **No backfill of the 1,181 person-partition Adult-name values** that
  `processAdults` reads from nothing.

No issue is retired here. kindred#2275 stays open, now holding a measured
residual instead of an estimate.

### DIFF
--- a/pocketbase/sync/family_camp_derived.go
+++ b/pocketbase/sync/family_camp_derived.go
@@ -10,6 +10,7 @@ import (
 	"strconv"
 	"strings"
 	"time"
+	"unicode"
 
 	"github.com/pocketbase/pocketbase/core"
 )
@@ -761,6 +762,18 @@ func (s *FamilyCampDerivedSync) loadPersonCustomValues(
 // this per-attribute merge policy, and that is kindred#2275's subject.
 // Today's behavior for those fields is pinned by test instead of changed on
 // a guess.
+//
+// NORMALISED, separately from the merge (kindred#2275 phase D, owner ruling
+// 2026-08-16): date_of_birth and relationship_to_camper are rewritten into a
+// canonical form BEFORE they are stored -- see normalizeDateOfBirth and
+// normalizeRelationshipToCamper. This is not a merge policy and does not
+// change which sibling wins; it removes the disagreements that were only ever
+// two spellings of one answer. Measured against the production snapshot, all
+// years: date_of_birth divergence falls from 1,124 (household, year, adult
+// slot) groups to 541, and relationship_to_camper from 315 to 169. The 710
+// that survive are real -- 360 different birth YEARS and 92 Mother-vs-Father
+// slot collisions -- and that residual is what the grain decision on
+// kindred#2275 now rests on.
 func (s *FamilyCampDerivedSync) processAdults(
 	householdValues []customValueEntry, personValues []customValueEntry,
 ) []*adultData {
@@ -829,9 +842,13 @@ func (s *FamilyCampDerivedSync) processAdults(
 		case strings.Contains(v.fieldName, "Gender") && adult.gender == "":
 			adult.gender = v.value
 		case strings.Contains(v.fieldName, "DOB") && adult.dateOfBirth == "":
-			adult.dateOfBirth = v.value
+			// Normalised, not merged: see normalizeDateOfBirth. Which sibling
+			// wins is still load order (kindred#2275); what changes is that
+			// the winner is stored in one canonical form, so two siblings who
+			// typed the SAME birthday two ways no longer read as a conflict.
+			adult.dateOfBirth = normalizeDateOfBirth(v.value)
 		case strings.Contains(v.fieldName, "Relationship") && adult.relationship == "":
-			adult.relationship = v.value
+			adult.relationship = normalizeRelationshipToCamper(v.value)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


adamflagg:
## Independent review — PR #2405

**Verdict: clean.** No code changes made in this pass.

### Scope check
Diff is exactly the two files claimed: `pocketbase/sync/family_camp_derived.go` and its test file, +554/-2. Confirmed against `main` — no grain change, no migration file, no merge-policy switch touched. `closingIssuesReferences` is empty via GraphQL, which is correct: this PR does not close kindred#2275 (the grain decision stays open), and the issue body was in fact edited one minute after the PR opened to record the measured residual (710/9,629, 7.4%) in place of the earlier estimate — consistent with the "fix the issue body" rule.

### CI
"CI Summary" is SUCCESS and actually ran real work, not just skips: Go Lint, Go Tests (3 shards), Migration Smoke Test all SUCCESS. CodeQL clean. No GUI-impact files (`gh pr diff --name-only` matches nothing under `frontend/src/**/*.{ts,tsx,css}`), so no owner-eyeball hold applies.

### Mutation-check transcript
Two mutations run against the working tree, each restored via a saved backup before re-running the suite:

**Mutation 1 — bypass normalization at the call sites** (`adult.dateOfBirth = v.value` / `adult.relationship = v.value` instead of the normalizer calls):
```
--- FAIL: TestProcessAdultsNormalisesSiblingDateFormats (3 subtests)
    got date_of_birth "9/2/1979", want canonical "1979-09-02"
--- FAIL: TestProcessAdultsNormalisesRelationshipCase (3 of 4 subtests)
    got relationship "mother", want "Mother"
```
RED, as expected. Restored, `git status --short` clean, `go build ./...` + `go test ./sync/...` green again.

**Mutation 2 — collapse the Father/Mother synonym pair** (pointed both `"dad"` and `"father"` at `"Mother"` in `relationshipSynonyms`):
```
--- FAIL: TestNormalizeRelationshipToCamperKeepsDistinctAnswersDistinct
    Mother and Father must never normalise to the same value
```
RED — the `t.Fatal` guard at the bottom of that test caught it. Restored, tree clean.

Both tests pin the defects they claim to pin.

### Additional probes (not part of the required checklist, but the date parser is the highest-risk surface in the diff)
Threw a standalone probe test at `normalizeDateOfBirth` covering leap-year validity (`Feb 29, 1980` valid / `Feb 29, 1981` invalid), invalid-calendar rejection (`2/30/1980`), the two-digit-year pivot boundary (`1-1-30`), an out-of-range month with no day/month swap attempted (`13/1/1980` stays unparsed rather than being silently reinterpreted as day-first), and an ISO string with an invalid month. All behaved as documented — invalid dates come back unchanged, no cross-pattern misrouting between the ISO/numeric/digits regexes. Removed the probe file afterward; it's not part of the branch.

Checked whether `date_of_birth` from `family_camp_adults` feeds any downstream logic (age calc, solver, etc.) that a parse ambiguity could silently corrupt — it doesn't; every consumer found is display/passthrough (`api/services/lodging_repository.py`, `frontend/src/types/pocketbase-types.ts`). So the inherent M/D-vs-D/M ambiguity for any US-convention date normalizer is a real but low-stakes property, already disclosed in the PR body's "does not judge plausibility" section, not a defect.

Verified the "second sync is a no-op" operational claim against the actual write path: `adultNeedsUpdate` (`family_camp_derived.go:1892`) compares `date_of_birth` and `relationship_to_camper` field-by-field before writing, so the one-time rewrite-then-stable behavior described in the PR body holds.

### PII / names
No real names or CampMinder IDs in the diff or PR body. `Emma Johnson`, `Liam Garcia`, and `Amy Johnson` are all on the approved fictional set (`tests/CLAUDE.md`). Nothing under `docs/plans/` touched.

### Disagreed with / filed
Nothing. No findings survived that were worth a fix or a filed issue at this review's floor.

Branch confirmed in sync with `origin/feature/phase-d-fc-normalise` (0 ahead, 0 behind) before and after this review.

PR #2406, opened 2026-08-17
fix(sync): the persons orphan sweep deleted staff who were never campers

### LINKED ISSUE #2394: bug(sync): every persons sync fails — orphan deletes hit a unique-index collision on blanked relations

adamflagg:
The `persons` sync has failed every night in its recorded history. `sync_runs`
shows `errors_count = 26, deleted_count = 0` on 2026-08-14, -15 and -16 — 26
orphan deletes, all of them failing, while created/updated moved normally
(4/85, 1/111, 1/103). Each failure was a uniqueness violation raised by a
`DELETE`, which is not an error a delete should ever be able to produce.

## The defect: the delete should never have been attempted

`getPersonIDsFromStaff` pulls staff person IDs into the run **on purpose**, so
`staff.person` can be populated. `transformPersonToPB` then returns `nil` for
anyone whose CampMinder record carries no `CamperDetails` block — someone who
has never been a camper — and `processPerson` returned at that `nil` *before*
`TrackProcessedKey`. `deleteOrphans` reads every untracked `(cm_id, year)` in
`persons` as "CampMinder no longer returns this person", so the sync fetched
these people deliberately and then nominated them for deletion for not being
campers.

The uniqueness error was the symptom, not the defect. `person_custom_values.person`
is `cascadeDelete: false`, so PocketBase blanks the relation and re-saves the
dependent rows instead of deleting them; those rows are covered by
`UNIQUE (year, person, field_definition)`, so the second person blanked for the
same `(year, field_definition)` collides on `('', year, def)`, the re-save is
rejected, and that rejection aborts the parent delete. **The collision was the
only thing stopping a wrong delete from succeeding.**

Measured on the production snapshot:

| | |
|---|---:|
| `persons` rows for 2026 with `is_camper = 0` | 299 |
| ...of those, referenced by `staff.person` | 277 |
| `person_custom_values` rows already blanked | 152 (116 in 2023, 36 in 2026) |
| `staff` rows already blanked | 1 |
| `person_custom_values` rows belonging to a non-camper 2026 person | 12,269 |

## The fix

**(a) `processPerson` now tracks the key in the `pbData == nil` branch.** A
person the run fetched from CampMinder is by definition not absent from
CampMinder, which is the only thing an orphan sweep is entitled to conclude.
`skippedStaff` and `Stats.Skipped` are deliberately unchanged — no row was
written, and that is what "skipped" counts. Skipped and orphaned are different
facts, and that branch only ever meant the first one.

**(b) Migration `1500000158` flips `person_custom_values.person` to
`cascadeDelete: true`**, so a genuinely-correct *future* orphan delete cascades
instead of blanking-and-colliding. A custom-field value is meaningless without
the person it describes. This is defence in depth, not the fix, and it lands in
the same change as (a) on purpose: shipping it alone would have removed the
safety net and let the sync start *successfully* deleting ~26 staff
person-records a night.

**(c) Both orphan sweeps are abandoned when the CampMinder staff feed cannot be
read.** (a) only protects a staff-only person while the run actually *sees*
them, and `gatherPersonIDs` swallows a `getPersonIDsFromStaff` failure —
`slog.Warn(...); staffPersonIDs = nil` — then carries on with attendees only.
On such a run every staff-only person is absent from `ProcessedKeys` and reads
as an orphan again. Neither existing guard catches it: `OrphanSweepGuard.Check`
compares a computed set still holding ~90% of the rows on disk against a 50%
floor, and `skipSweepForRejections` reads `Stats.Rejected`, which a swallowed
fetch error never bumps. That mattered less while the relation
blanked-and-collided, because the collision aborted most of those deletes —
which is precisely why the skip has to land in the same change as (b) rather
than after it.

`deleteHouseholdOrphans` has the identical exposure and is fixed alongside:
`extractUniqueHouseholds` reads `Households` straight off the CampMinder
payload with no reference to `CamperDetails`, and `processBatchPersons` calls it
for every person in the batch, staff included, so a household reachable only
through staff vanishes from the computed set on the same run. Its existing "no
guard needed" note is about *rejections*, which its tracking position above the
transform genuinely does cover; a missing upstream feed removes keys wherever
the tracking sits. Both skip rather than error, because
`updateRelationsAndCleanup` turns a returned error into `Stats.Errors++` and a
deliberate skip is not an operational error — the same verdict `SkipReason`
already gives a rejection.

## Deliberately not done

- **The 152 already-blanked `person_custom_values` rows are left in place.**
  That table has no `person_id` column — only the `person` relation — so once
  blanked there is no surviving pointer to whom the answer belonged. Re-linking
  is impossible and deleting is destructive; this repo does not run destructive
  data migrations. They stay as inert residue.
- **The other five collision-prone relations to `persons` are not flipped** —
  `staff (year, person)`, `financial_aid_applications (person, year)`,
  `bunk_assignments (year, person, session)`,
  `bunk_assignments_draft (year, session, person, scenario)` and
  `normalized_mappings (person, session, category)`. Deleting a staff record or
  a financial-aid application because a person row went away is a product
  decision nobody has made, and their collision is currently what makes such a
  delete fail *loudly*; converting that into a silently severed row would be a
  downgrade.
- **`financial_transactions` is untouched.** Its 124,088 blank-`person` rows are
  household-level by design — 122,824 carry a `household`, and only 1,264 have
  both keys blank.
- **The 75 blanked `attendees` rows are not re-linked.** `person_id` survives on
  all of them and 16 still resolve against `persons`, so recovery is possible,
  but it is a separate piece of work.
- **`base_sync.go` is unchanged.** Its counting was already right: it bumps
  `Stats.Errors` per failed `App.Delete` and only reaches `orphanCount++` on
  success, which is exactly why 26 errors showed up against 0 deletions.

## Verification

- New `TestStaffOnlyPersonSurvivesOrphanSweep` drives the real path against a
  real `persons` collection — two campers processed, one staff-only person
  fetched and written off, then the real `deleteOrphans`. Before the fix it
  failed with the production log line verbatim
  (`Deleting orphaned record entity=person ... id=2001|2026`); after it, the row
  survives and the tracked campers still do too. Three rows sits under
  `OrphanSweepMinRows`, so the collapse guard's ratio arm cannot be what saves
  the row.
- New `TestProcessPersonTracksStaffOnlyPersonAsProcessed` pins the tracking
  itself and pins that the skip accounting did not move.
- New `TestOrphanSweepSkippedWhenStaffFetchFailed` and
  `TestHouseholdOrphanSweepSkippedWhenStaffFetchFailed` pin (c). Each is a pair:
  with the flag set the untracked row survives, and with it clear a genuine
  orphan is still swept, so neither case can pass by the sweep doing nothing.
- `go build ./...`, `go test ./...` and `golangci-lint run` all clean.
- The migration was boot-checked against a **fresh** database with the real
  PocketBase binary (`scripts/dev/lib/pb-harness.sh`): 118 JS migrations apply,
  the server reaches `/api/health`, and `person_custom_values.person` reads
  `cascadeDelete: true` while `field_definition` correctly stays `false`.

Closes #2394.

## Summary by CodeRabbit

- **Bug Fixes**
  - Related custom person data is now automatically removed when its associated person is deleted.
  - Staff-only people are preserved during synchronization, even when camper details are unavailable.
  - Person and household records are no longer mistakenly removed when staff data cannot be retrieved.
  - Genuine orphaned records continue to be cleaned up when synchronization data is complete.
- **Tests**
  - Added coverage for staff-only records, failed staff retrieval, and orphan cleanup scenarios.

### DIFF
--- a/pocketbase/pb_migrations/1500000158_person_custom_values_cascade.js
+++ b/pocketbase/pb_migrations/1500000158_person_custom_values_cascade.js
@@ -1,0 +1,80 @@
+/// <reference path="../pb_data/types.d.ts" />
+/**
+ * Migration: person_custom_values.person cascades instead of blanking.
+ *
+ * kindred#2394, part (b). Part (a) is the fix -- persons.go now tracks a
+ * staff-only person as processed, so the nightly sweep stops nominating 26
+ * people it deliberately fetched. This migration is defence in depth: it makes
+ * a GENUINELY correct orphan delete behave properly instead of erroring.
+ *
+ * What went wrong. `person` is declared cascadeDelete: false
+ * (1500000028_person_custom_values.js), so deleting a person does not delete
+ * their custom values -- PocketBase CLEARS the relation and re-saves the rows.
+ * Those rows are covered by
+ *
+ *   CREATE UNIQUE INDEX idx_person_cf_vals_unique
+ *     ON person_custom_values (year, person, field_definition)
+ *
+ * so the moment a SECOND person's rows are blanked for the same
+ * (year, field_definition), the blanked rows collide on ('', year, def) and the
+ * re-save is rejected. That rejection aborts the parent delete and surfaces as
+ * `field_definition: Value must be unique; person: Value must be unique; year:
+ * Value must be unique.` on a DELETE, which is not an error a DELETE should
+ * ever produce.
+ *
+ * A custom-field value is meaningless without the person it describes, so
+ * cascading is the honest answer here: the row goes with its person rather than
+ * surviving as an answer belonging to nobody.
+ *
+ * ⛔ The 152 rows already blanked in production are LEFT IN PLACE, deliberately.
+ * This table has no person_id column -- only the `person` relation -- so once
+ * blanked there is no surviving pointer to who the answer belonged to and they
+ * cannot be re-linked. Deleting them is the only other option, and this repo
+ * does not run destructive data migrations. They stay as inert residue.
+ * (Production snapshot: 152 rows, 116 in year 2023 and 36 in 2026.)
+ *
+ * ⛔ Only this relation is changed. Five other relations to `persons` also sit
+ * inside a unique index and would also collide when blanked --
+ * staff (year, person), financial_aid_applications (person, year),
+ * bunk_assignments (year, person, session),
+ * bunk_assignments_draft (year, session, person, scenario) and
+ * normalized_mappings (person, session, category). They are NOT flipped, for
+ * two reasons. First, deleting a staff record or a financial-aid application
+ * because a person row went away is a product decision nobody has made, and it
+ * is a very different proposition from deleting a custom-field answer -- the
+ * issue says so explicitly. Second, their collision is currently the only thing
+ * that makes such a delete fail loudly; converting that into a silent severed
+ * row would be a downgrade, not a hardening. person_custom_values is the one
+ * collection that both fails today and has an unambiguous answer.
+ *
+ * Idempotent: setting cascadeDelete to a value it already holds is a no-op, so
+ * a re-run (see the renumbering trap in docs/reference/pocketbase-migrations.md)
+ * converges rather than diverging. No guard, though: a missing `person` relation
+ * is the one state this migration must not shrug at, since clearing that field's
+ * blanking behaviour is its whole purpose.
+ */
+
+migrate(
+  (app) => {
+    const col = app.findCollectionByNameOrId('person_custom_values');
+
+    const rel = col.fields.getByName('person');
+    if (!rel) {
+      throw new Error('person_custom_values: expected an existing "person" relation field');
+    }
+    rel.cascadeDelete = true;
+
+    app.save(col);
+  },
+  (app) => {
+    const col = app.findCollectionByNameOrId('person_custom_values');
+
+    const rel = col.fields.getByName('person');
+    if (!rel) {
+      throw new Error('person_custom_values: expected an existing "person" relation field');
+    }
+    rel.cascadeDelete = false;
+
+    app.save(col);
+  }
+);
--- a/pocketbase/sync/persons.go
+++ b/pocketbase/sync/persons.go
@@ -57,9 +57,6 @@ type personHouseholdIDs struct {
 type gatherPersonIDsResult struct {
 	personIDs    []int        // All unique person IDs (from attendees + staff)
 	camperIDsSet map[int]bool // IDs from attendees (true campers)
-	// staffFetchFailed is true when the CampMinder staff feed could not be read
-	// and the run fell back to attendees only.
-	staffFetchFailed bool
 }
 
 // NewPersonsSync creates a new persons sync service
@@ -139,7 +136,6 @@ func (s *PersonsSync) Sync(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
-	s.staffFetchFailed = gatherResult.staffFetchFailed
 	if len(gatherResult.personIDs) == 0 {
 		slog.Info("No attendees or staff found, skipping persons sync", "year", year)
 		s.SyncSuccessful = true
@@ -199,11 +195,13 @@ func (s *PersonsSync) gatherPersonIDs(year int) (*gatherPersonIDsResult, error) 
 	}
 
 	staffPersonIDs, err := s.getPersonIDsFromStaff()
-	staffFetchFailed := false
 	if err != nil {
 		slog.Warn("Error getting staff person IDs, continuing with attendees only", "error", err)
 		staffPersonIDs = nil
-		staffFetchFailed = true
+		// Set on the service here rather than returned and copied across by the
+		// caller: one writer, one reader, and no assignment a later refactor can
+		// drop without the compiler noticing. Sync clears it per run.
+		s.staffFetchFailed = true
 	}
 
 	personIDs := s.mergePersonIDs(attendeePersonIDs, staffPersonIDs)
@@ -214,9 +212,8 @@ func (s *PersonsSync) gatherPersonIDs(year int) (*gatherPersonIDsResult, error) 
 		"year", year)
 
 	return &gatherPersonIDsResult{
-		personIDs:        personIDs,
-		camperIDsSet:     camperIDsSet,
-		staffFetchFailed: staffFetchFailed,
+		personIDs:    personIDs,
+		camperIDsSet: camperIDsSet,
 	}, nil
 }
 
@@ -1163,6 +1160,29 @@ func (s *PersonsSync) printDataQualitySummary() {
 }
 
 // deleteOrphans deletes persons that exist in PocketBase but weren't processed from CampMinder
+// skipSweepForStaffFetchFailure reports whether a sweep must be abandoned because
+// this run could not read the CampMinder staff feed, and logs why when it must.
+//
+// The consequence differs per sweep -- persons loses the staff-only people
+// themselves, households loses whatever is reachable only through them -- so the
+// caller supplies that half of the message; everything else is shared.
+//
+// It warns rather than returning an error on purpose. updateRelationsAndCleanup
+// turns a deleteOrphans error into Stats.Errors++, and a deliberate skip is not an
+// operational failure. That is the same verdict OrphanSweepGuard.SkipReason gives a
+// rejection, and it carries the same accepted cost, stated there: a run that keeps
+// skipping lets genuine orphans accumulate, and the signal to go fix the upstream is
+// the warning repeating night after night rather than a separate alerting path.
+func (s *PersonsSync) skipSweepForStaffFetchFailure(entity, consequence string, year int) bool {
+	if !s.staffFetchFailed {
+		return false
+	}
+
+	slog.Warn("Orphan sweep skipped: the CampMinder staff feed could not be read, so "+consequence,
+		"entity", entity, "year", year)
+	return true
+}
+
 func (s *PersonsSync) deleteOrphans(year int) error {
 	// A swallowed staff-feed failure makes the computed set known-incomplete, so
 	// abandon the sweep rather than narrow it (kindred#2394). gatherPersonIDs
@@ -1179,11 +1199,9 @@ func (s *PersonsSync) deleteOrphans(year int) error {
 	// people" -- so it skips without failing the run: updateRelationsAndCleanup
 	// turns a returned error into Stats.Errors++, and a deliberate skip is not an
 	// operational error.
-	if s.staffFetchFailed {
-		slog.Warn("Orphan sweep skipped: the CampMinder staff feed could not be read, "+
-			"so every staff-only person is missing from the computed set and their stored "+
-			"rows would be read as orphans",
-			"entity", "persons", "year", year)
+	if s.skipSweepForStaffFetchFailure("persons",
+		"every staff-only person is missing from the computed set and their stored rows "+
+			"would be read as orphans", year) {
 		return nil
 	}
 
@@ -1718,10 +1736,8 @@ func (s *PersonsSync) updatePersonHouseholdRelations(
 // feed failed, and this sweep would delete it. Hence the one guard this function
 // does take.
 func (s *PersonsSync) deleteHouseholdOrphans(year int, processedIDs map[int]bool) error {
-	if s.staffFetchFailed {
-		slog.Warn("Household orphan sweep skipped: the CampMinder staff feed could not be read, "+
-			"so households reachable only through staff are missing from the computed set",
-			"entity", "households", "year", year)
+	if s.skipSweepForStaffFetchFailure("households",
+		"households reachable only through staff are missing from the computed set", year) {
 		return nil
 	}
 
--- a/pocketbase/sync/persons.go
+++ b/pocketbase/sync/persons.go
@@ -35,6 +35,11 @@ type PersonsSync struct {
 	missingDataStats map[string]int
 	skippedStaff     int
 
+	// staffFetchFailed records that this run could not read the CampMinder staff
+	// feed, so its computed set is known-incomplete and the orphan sweep must not
+	// run. Set from gatherPersonIDs, read by deleteOrphans. See kindred#2394.
+	staffFetchFailed bool
+
 	// Sub-entity stats for combined sync (households)
 	// Note: person_tags removed - tags are now a multi-select relation on persons
 	householdStats *Stats
@@ -52,6 +57,9 @@ type personHouseholdIDs struct {
 type gatherPersonIDsResult struct {
 	personIDs    []int        // All unique person IDs (from attendees + staff)
 	camperIDsSet map[int]bool // IDs from attendees (true campers)
+	// staffFetchFailed is true when the CampMinder staff feed could not be read
+	// and the run fell back to attendees only.
+	staffFetchFailed bool
 }
 
 // NewPersonsSync creates a new persons sync service
@@ -121,6 +129,7 @@ func (s *PersonsSync) Sync(ctx context.Context) error {
 	s.LogSyncStart("persons (combined: persons + households)")
 	s.Stats = Stats{}
 	s.SyncSuccessful = false
+	s.staffFetchFailed = false
 	s.ClearProcessedKeys()
 
 	year := s.Client.GetSeasonID()
@@ -130,6 +139,7 @@ func (s *PersonsSync) Sync(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+	s.staffFetchFailed = gatherResult.staffFetchFailed
 	if len(gatherResult.personIDs) == 0 {
 		slog.Info("No attendees or staff found, skipping persons sync", "year", year)
 		s.SyncSuccessful = true
@@ -189,9 +199,11 @@ func (s *PersonsSync) gatherPersonIDs(year int) (*gatherPersonIDsResult, error) 
 	}
 
 	staffPersonIDs, err := s.getPersonIDsFromStaff()
+	staffFetchFailed := false
 	if err != nil {
 		slog.Warn("Error getting staff person IDs, continuing with attendees only", "error", err)
 		staffPersonIDs = nil
+		staffFetchFailed = true
 	}
 
 	personIDs := s.mergePersonIDs(attendeePersonIDs, staffPersonIDs)
@@ -202,8 +214,9 @@ func (s *PersonsSync) gatherPersonIDs(year int) (*gatherPersonIDsResult, error) 
 		"year", year)
 
 	return &gatherPersonIDsResult{
-		personIDs:    personIDs,
-		camperIDsSet: camperIDsSet,
+		personIDs:        personIDs,
+		camperIDsSet:     camperIDsSet,
+		staffFetchFailed: staffFetchFailed,
 	}, nil
 }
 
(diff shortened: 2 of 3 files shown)


adamflagg:
@coderabbitai review

adamflagg:
## Independent review — #2394

Reviewed at score floor 60. I did not write this PR. Scope matches the item: part (a) tracking + part (b) the cascade migration, nothing more. `closingIssuesReferences` reads `[2394]` from the API, which is correct. No frontend files, so nothing needs the owner's eyes on visual grounds. No real names or CampMinder IDs in the diff, tests or body; nothing under `docs/plans/`. Migration `1500000158` is the next free number with no duplicate prefixes.

### What I verified rather than took on trust

**The migration.** Booted a throwaway DB with the real PocketBase binary via `scripts/dev/lib/pb-harness.sh` and read the collection back out of `_collections`:

```
RELATION person: cascadeDelete=True collectionId=col_persons
RELATION field_definition: cascadeDelete=False collectionId=col_custom_field_defs
INDEXES: ["CREATE UNIQUE INDEX `idx_person_cf_vals_unique` ON `person_custom_values` (`year`, `person`, `field_definition`)", ...]
```

So the flip lands, `field_definition` correctly stays `false`, and the unique index is untouched — the trap the item warned about (quietly adding a partial index and removing the safety net) was not walked into.

**The author's tests.** Mutation-checked by deleting the `TrackProcessedKey` call from the `pbData == nil` branch:

```
--- FAIL: TestProcessPersonTracksStaffOnlyPersonAsProcessed
    ProcessedKeys is missing "2001|2026" -- a person this run fetched from CampMinder
    on purpose reads as absent from it, and deleteOrphans deletes their row
--- FAIL: TestStaffOnlyPersonSurvivesOrphanSweep
    INFO Deleting orphaned record entity=person name="Noah Martinez" id=2001|2026
    INFO Deleted orphaned records entity=person count=1
    the staff-only person's row was swept as an orphan (sql: no rows in result set)
```

Both go red and reproduce the production log line verbatim. They pin the defect. Restored, green again.

### Finding, fixed in-branch

**The tracking only protects a staff-only person while the run actually sees them, and one code path guarantees it does not.**

`gatherPersonIDs` swallows a `getPersonIDsFromStaff` failure — `slog.Warn(...); staffPersonIDs = nil` — and carries on with attendees only. On such a run every staff-only person is absent from `ProcessedKeys` and reads as an orphan again, which is precisely the defect part (a) exists to prevent.

Neither existing guard catches it, and I checked both rather than assuming:

- `OrphanSweepGuard.Check` compares a computed set still holding roughly 90% of the rows on disk against a 50% floor. Nowhere near firing.
- `skipSweepForRejections` reads `Stats.Rejected`, which a swallowed fetch error never bumps.
- `SyncSuccessful` is set true on the first non-empty batch of the attendee-only ID list, so it does not gate this either.

This matters *because of* part (b). While the relation blanked-and-collided, the unique index aborted most of those deletes — the PR body itself calls that collision "the only thing stopping a wrong delete from succeeding". Flipping it to cascade removes that net, so on a staff-feed-failure run the deletes would now succeed silently and take the custom values with them. The skip therefore belongs in the same change as the cascade, not after it.

`deleteHouseholdOrphans`, three lines below in `updateRelationsAndCleanup`, has the identical exposure. `extractUniqueHouseholds` reads `Households` straight off the CampMinder payload with no reference to `CamperDetails`, and `processBatchPersons` calls it for every person in the batch, staff included — so a household reachable only through staff vanishes from the computed set on the same run. Its "no guard needed" note is sound but is about *rejections*, which its tracking position above the transform genuinely does cover; a missing upstream feed removes keys wherever the tracking sits. Fixed alongside, and I extended that comment rather than contradicting it.

Both skip rather than error, matching the verdict `SkipReason` already gives a rejection: `updateRelationsAndCleanup` turns a returned error into `Stats.Errors++`, and a deliberate skip is not an operational failure.

Mutation-check on the new guard (helper forced to return false):

```
--- FAIL: TestOrphanSweepSkippedWhenStaffFetchFailed/staff_fetch_failed:_the_sweep_is_abandoned,_not_narrowed
    row swept = true, want false (lookup err: sql: no rows in result set)
--- FAIL: TestHouseholdOrphanSweepSkippedWhenStaffFetchFailed/staff_fetch_failed:_the_sweep_is_abandoned
    household swept = true, want false (lookup err: sql: no rows in result set)
--- PASS: TestOrphanSweepSkippedWhenStaffFetchFailed/staff_fetch_succeeded:_a_genuine_orphan_is_still_swept
--- PASS: TestHouseholdOrphanSweepSkippedWhenStaffFetchFailed/staff_fetch_succeeded:_a_genuine_orphan_is_still_swept
```

Each test is a pair on purpose: the second case is what stops the first from passing vacuously by way of a sweep that does nothing at all.

**PR body corrected.** This repo squash-merges with `PR_BODY` as the commit message (confirmed: `squash_merge_commit_message: PR_BODY`), so the body is the permanent record and its claims get reviewed as code. It asserted "With (a) in place, a person carrying a `staff` row for the year is always tracked, so they can no longer become an orphan candidate anyway" — false in exactly the failure mode where the removed safety net mattered. That sentence is gone and the fix is written up as part (c). Still exactly one closing keyword.

### CodeRabbit

Requested and a real review landed with two nitpicks. Worth flagging that `verify-pr-state.sh coderabbit` reported `NOT VERIFIED ... unrecognised notice` for it — this review is a `COMMENTED` review whose body opens with a nitpick block rather than `**Actionable comments posted: N**`, so the script's matcher missed a genuine review. Read the review off the API instead of trusting either signal.

**Taken — the propagation path was untested.** Fair, and I fixed it structurally rather than with a test. The flag was returned on `gatherPersonIDsResult` and copied onto the service by `Sync`, which is a wiring step a refactor could drop silently; a test could not catch it either, since the CampMinder base URL is a package constant and a client whose staff call fails cannot be built without real network. `gatherPersonIDs` is already a `*PersonsSync` method, so it now sets the field where the failure is observed. One writer, one reader, no copy to forget, and a smaller diff.

**Taken — one shared helper.** The two skip blocks now go through `skipSweepForStaffFetchFailure`, with the per-sweep half of the message passed in.

**Declined — "add a summary-visible skip signal and alerting".** Verified against the code rather than waved off: this is the same behaviour `OrphanSweepGuard.SkipReason` deliberately has for a rejection skip, and its doc states the reasoning — the signal is the warning repeating run after run, and there is no separate alerting path by design. Adding one means reopening the shared `Stats` / `sync_runs` contract every guarded sweep reports through, and `base_sync.go` / `orphan_guard.go` / `orchestrator.go` are out of scope for this PR. A package-wide, documented decision, not a defect here.

### Disagreements and things I deliberately left

- **The 152 already-blanked `person_custom_values` rows stay.** Confirmed the table carries no `person_id` column, only the relation, so re-linking really is impossible and the only alternative is a destructive migration. Correct call, correctly stated in the migration header.
- **The other five collision-prone relations stay `cascadeDelete: false`.** I checked `staff`'s `idx_staff_person_year (year, person)` exists as claimed. Leaving them is right: their collision is still a loud failure, and cascading a staff record or an aid application off a person delete is an unmade product decision.
- **`base_sync.go`, `orphan_guard.go`, `orchestrator.go` untouched.** I found no reason to touch them; the counting really is right, which is why 26 errors showed against 0 deletions.
- Nothing filed as out of scope — the one finding was small enough to fix here, which is where it belonged.

Nothing discarded as refuted: the single hypothesis I raised held up under checking. Gates re-run on the final head: `gofmt`, `go vet`, `go build ./...`, `golangci-lint run ./sync/...` (0 issues), full `go test ./...` green, and CI Summary green with zero failing checks.

Not merged, no issue closed by hand.

adamflagg:
@coderabbitai review

adamflagg:
**Response to CodeRabbit's two blocking concerns. Both verified against the tree; both refuted.**

### 1. "Concurrent sync runs can race on the shared failure state"

Refuted on two independent grounds.

**The orchestrator forbids it.** `MarkSyncRunning` (`orchestrator.go:1075`) returns `sync already in progress: %s` when `IsRunning(syncType)` is true. There is no path that starts a second `persons` run while one is live.

**And `staffFetchFailed` is exactly as concurrency-safe as the state this package already runs on.** `Sync()` resets `s.Stats = Stats{}` at `persons.go:122`; `ProcessedKeys` and `SyncSuccessful` are the same shape — per-instance mutable state, reset at the top of the run, on a service instance registered once and reused. If two same-type runs could overlap, every counter in `pocketbase/sync/` would already be corrupting itself, and the orphan-sweep guard would be reading a half-built `ProcessedKeys` — a far larger failure than this flag. The new field introduces no sharing that was not already load-bearing.

A genuine `sync.Mutex` on service state would be a package-wide change to a pre-existing structural property, not a fix to this PR. It is not in scope here, and this PR is not the thing that made it true.

### 2. "The silent skip provides no operator-visible indication"

Factually incorrect — the skip is not silent. `skipSweepForStaffFetchFailure` calls:

```go
slog.Warn("Orphan sweep skipped: the CampMinder staff feed could not be read, so "+consequence,
    "entity", entity, "year", year)
```

once per affected sweep, naming the entity, the year and the specific consequence. That is deliberately a warning rather than a returned error, because `updateRelationsAndCleanup` turns a returned error into `Stats.Errors++`, and `Stats.Errors` is zero-tolerance and fails the run (`orchestrator.go:368-381`, shipped in #2284). A deliberate skip is not an infrastructure failure, and counting it as one would reintroduce the exact red-every-night condition this PR exists to remove. This matches the verdict `OrphanSweepGuard.SkipReason` already gives a rejection, and carries the same accepted cost, stated there: a run that keeps skipping lets genuine orphans accumulate, and the signal is the warning repeating night after night.

### On the two ❓ rows in CodeRabbit's objectives table

- *"no explicit handling for `attendees`, `staff`, financial transactions, or bunk assignments"* — correct, and deliberate. The migration header says so at length: those relations are **not** flipped because deleting a staff record or a financial-aid application when a person row goes away is a product decision nobody has made, and because their collision is currently the only thing making such a delete fail loudly. Converting that into a silently severed row would be a downgrade, not a hardening. `person_custom_values` is the one collection that both fails today and has an unambiguous answer.
- *"the summaries do not verify full sync completion"* — the tests pin the mechanism, and the review booted a real PocketBase binary against a throwaway DB to confirm migration 1500000158 lands `person.cascadeDelete = true`, leaves `field_definition` false, and leaves `idx_person_cf_vals_unique` intact. End-to-end completion needs a live CampMinder credential and is an operator verification after deploy, not a unit test.

**Net: no code change from this review.** Logged rather than actioned, per the campaign's finding-disposition rules.

Issue #2275, opened 2026-08-12
bug(sync): family camp adult attributes resolve by load order, losing 1,151 dates of birth


adamflagg:
## ⚠️ CORRECTED 2026-08-21 — the "no change to any row today" reassurance above does not hold

Re-measured against the production snapshot `pocketbase/pb_data/data-prod.db` by replaying the
real `normalizeDateOfBirth` (not a raw-string diff — a raw-string diff is dominated by formatting
variance the normaliser already collapses, and overstates disagreement by roughly 2x).

**This does not change the fix direction.** First-wins with a CampMinder-id tiebreak is what's
implemented, in `pocketbase/sync/family_camp_derived.go` (`processAdults` sorts the person-values
slice by `personCMID` before merging; `mergeFirstNonEmpty` itself is unchanged). The correction is
narrower: the claim that switching the tiebreak leaves today's stored winners untouched is wrong.

Of the 544 `Family Camp DOB 1`/`DOB 2` `(household, year, adult_slot)` groups that still genuinely
disagree after normalisation (541 last recorded 2026-08-17; +3 drift), switching the ordering from
`person_custom_values.id` to `persons.cm_id` changes the stored winner in **261 of 544 (48%)** — not
zero. The change IS durable (a resync can no longer flip it, which is the property the ruling was
actually chasing), but it is not a no-op on today's data.

**This is not staff-visible.** Per this issue's own accounting a few sections down, `date_of_birth`
"has zero readers anywhere" in the current UI — nothing renders it, so this data-value flip does
not change anything a staff member can see or any staff-facing count. Recorded so it is not
mis-cited as evidence of zero impact for a future change to a column that IS read.

**Headline recount, same snapshot, raw (unnormalised) figures:** 1,159 DOB answers lost across
1,132 colliding household-years, vs. the 1,151 / 1,124 last recorded 2026-08-19 — drift of +8,
consistent with ongoing 2026 registration. Re-derive before citing either number again; this issue
already carries a standing warning that counts here drift.

---

## ⛔ RULING 2026-08-18 — THE BADGE IS DROPPED. Do not build it.

**This supersedes the section below headed *"What REMAINS — the badge and the tooltip"*, which is the
single most dangerous paragraph in this issue: it tells a fresh implementer that a board badge plus
tooltip is the whole of the open work. The owner killed that badge on 2026-08-18. It was recorded as a
comment and never reached this body.**

### Why it was dropped

[...]

Issue #2263, opened 2026-08-12
bug(sync): attendee enrollments collapse to one row per person/session, dropping the program dimension


adamflagg:
## ⚠️ CORRECTED 2026-08-16 — phase D triage. Closes nothing; no code ships.

Verified against `main` @ `a946565a`. The premise holds — attendee enrollments do collapse to one row per
person/session, dropping the program dimension.

**But this issue's own close condition is satisfied by evidence rather than by a PR**, so **no code ships and
it retires no other number.** Before scheduling any work here, measure how many person/session pairs actually
carry more than one distinct program in production. If that count is small, the correctness argument stands but
the priority does not — say which you are acting on.

⚠️ Use `pocketbase/pb_data/data-prod.db` **by name**. The sibling `data.db` is the local dev copy and every
figure in the 2026-08-14 correction blocks on this backlog came from it.

---

## ⚠️ CORRECTED 2026-08-14 — phase C triage

Verified against `main` @ `3eb44ba4` and the production snapshot. Verdict: **VALID** / NO_CODE.

> Confirmed on both counts: the duplicate-SessionID counter is on main (`AttendeesSync.DuplicateSessionEnrollments`, shipped in ff8f8342 / PR #2315) and the issue is currently OPEN — it was closed at 03:22:58Z on 2026-08-14 by a closing keyword naming this issue that appeared inside PR #2315's prose explaining it was *not* closing, and reopened 78 seconds later. Close condition, one line: it closes once the shipped counter has reported over a full real season and the answer (multi-program sessions exist, or do not) is recorded — no code closes it today. Every file:line anchor in the body shifted by ~80-130 lines when #2315 inserted the counter.

### Corrections to the body below

[...]

PR #2421, opened 2026-08-17
feat(sync): record discarded family-camp adult answers in attribute_conflicts

adamflagg:
Adds one additive JSON column, `family_camp_adults.attribute_conflicts`, and has
`processAdults` write the answers its merge discards into it — keyed by the column they
were destined for, so a reader can say *which* attribute was answered twice and *what*
the other answer was.

```json
{"date_of_birth": ["1981-09-02"], "relationship_to_camper": ["Mother"]}
```

The winning answer stays in its own column; this map holds only the losers. NULL when a
slot's answers agreed.

## The grain change was DECLINED, and this PR deliberately does not touch the grain triple

A 2026-08-15 ruling on #2275 said to re-key `family_camp_adults` to camper grain. The owner
**reversed** that on 2026-08-17 in favour of this column (Option B), after correcting the
framing the campaign had been carrying:

> the divergent adult rows are **not two children reporting on their parents**. They are the
> same parent filling in the family-camp section of a per-camper form, once per camper, on a
> form where that section should have been skipped after the first child.

So a divergence is one person being less careful the second time — a **data-entry artifact of
form design**, not a grain problem. That reading is written into the migration comment and into
`adultData.conflicts`, because getting it wrong is what sends the next reader back toward the
re-key that was just declined.

Nothing here moves the grain triple: not the write key (`familyCampAdultKey`), not
`TrackProcessedCompositeKey`, not `idx_fc_adults_unique`. Verified against a scratch DB built
from `pb_migrations/` alone — the unique index is still `(household, year, adult_number)` and
row counts are unchanged. Fanning out would have inflated `party_size` (`beds` in
`api/services/lodging_roster_service.py` counts `family_camp_adults` rows) and left 9,143
historical rows permanently at the old grain.

## The merge policy is unchanged

First-non-empty-wins still decides which answer is stored, for every attribute, and
`preferEmail` (#1945) still breaks the email tie on well-formedness. `mergeFirstNonEmpty`
returns exactly what the previous code returned; the only change is that the discarded
value is kept instead of vanishing. Pinned by a test that asserts the stored value in both
load orders, and by the existing first-loaded-sibling tests, which are untouched.

Proved rather than asserted: replaying `processAdults` over every row of
`pocketbase/pb_data/data-prod.db`, all years, and hashing the eight stored attributes of all
9,789 resulting slots gives the **same fingerprint** (`24d14af8cf6baba38b43b537`) before and
after every change in this PR, including the spelling guard below.

## Two spellings of one answer must stay silent

The #2405 normalisers run **before** the comparison, so the 583 `date_of_birth` and 146
`relationship_to_camper` divergences that were only ever two spellings of one answer
(`09-02-1979` vs `9/2/1979`; `mother` vs `Mom`) never reach it.

The free-text columns have no normaliser, and that gap was loud. Measured on the snapshot,
**189 of 1,429 lit slots — and 32 of 2026's 124, a quarter of the year staff actually look
at — differed from the winner in nothing but letter case**: `Amy Johnson` vs `amy johnson`,
`Amy@example.com` vs `amy@example.com`. Email was the worst column at 111 of 325, `first_name`
118 of 791.

`sameAnswer` folds case and collapses whitespace runs **at the comparison only**. It gates
what is recorded and nothing else — which is what the identical fingerprint above certifies.
It is deliberately narrow: no punctuation stripping, no token reordering, no nickname folding.
`Amy Johnson` vs `Amy R Johnson` is still two answers, because it is.

The email arm needs the guard on **both** branches, and that is the non-obvious half: a leading
space fails `emailFormatPattern`, so `" a@x.com"` is *displaced* by `"a@x.com"` and was being
recorded as conflicting with itself.

Normalisation itself is still not changed — nothing stored is re-cased, and `Mother`/`Father`
are not folded into each other.

## Scale, re-measured — and it is larger than the ruling's figure

The ruling quoted 46 of 788 slots in 2026 (5.8%) and 619 of 9,629 all-years (6.4%). That was
measured on the residual of the **two normalised columns only**. The column as specified is
`{column: [other values]}`, so this implementation records every person-sourced attribute that
merges — which is a bigger population. Re-measured by replaying this branch's `processAdults`
over every row of `pocketbase/pb_data/data-prod.db`, all years, no roster filter:

| scope | slots | ≥1 conflict | of which DOB/relationship |
|---|---|---|---|
| 2026 | 834 | **92 (11.0%)** | 45 |
| all years | 9,789 | **1,240 (12.7%)** | 610 |

**The DOB/relationship subset lands one row under the ruling's 2026 figure (45 vs 46) and
within nine all-years (610 vs 619).** The one 2026 row is a `relationship_to_camper` answer
whose two spellings differ only in capitalisation, which `sameAnswer` now silences; the
all-years gap is mostly the denominator — the ruling counted `(household, year, adult_slot)`
groups present in `person_custom_values`, this counts the `family_camp_adults` rows those
groups actually produce, which the nameless-row admission filter (#1946) trims.

Per column, all years, slots lighting up on that column:

| column | recorded | before the spelling guard | independently measured "answers lost" in #2275's body |
|---|---|---|---|
| `first_name` | 673 | 791 | 791 colliding groups |
| `date_of_birth` | 536 | 536 | 541 post-normalisation residual |
| `gender` | 507 | 507 | 511 |
| `email` | 214 | 325 | 326 |
| `relationship_to_camper` | 166 | 168 | 169 |
| `pronouns` | 75 | 75 | 76 |
| `last_name` | 11 | 15 | 15 |

The middle column is the corroboration: **before folding spellings, every column lands within a
handful of the figure #2275 derived independently by SQL**, which is what shows this records
exactly the answers the merge was already discarding. The drop from the middle column to the
first is entirely capitalisation, and it is concentrated in the three columns that have no
normaliser. `date_of_birth`, `gender` and `pronouns` are unchanged by folding — there was no
case noise in them at all.

`first_name` is still the biggest contributor and is worth knowing about: it is a **mislabeled
full name** (773 of 788 2026 values contain a space), so its remaining conflicts are things like
`Amy Johnson` vs `Amy R Johnson`. If the badge should be narrowed to the two normalised
columns, that is a one-line change to the arms in `processAdults`.

## Bundled rider — a false comment in `sync/attendees.go`, closes nothing

```go
// Note: CampMinder attendee ID exists in enrollment["ID"] but we don't need it in PocketBase
```

**There is no `ID` key.** Measured across the entire season: 2,315 attendees, 2,904
`SessionProgramStatus` entries, **0** carrying one — the evidence recorded when #2263 closed on
2026-08-17. The conclusion ("we don't need it") was fine; the premise would send someone hunting
for a stable per-enrollment identifier down a dead end. That absence is load-bearing — the vendor
exposes no identity below `(PersonID, SessionID)`, so the composite key could not be refined even
if someone wanted to. Comment-only: `processEnrollment` and every key are untouched. This closes
no issue.

## Not in scope

- **No badge or tooltip.** This PR is the column plus the writer. The frontend surfacing is
  the remaining half of Option B and would make this a GUI PR.
- **No backfill.** `family_camp_derived` recomputes every row it touches, so the column fills
  on the next sync run and reads correctly as "not yet computed" until then.
- **No change to what is stored, to the merge, or to the grain.**

## Tests

TDD — every assertion below was written and watched fail before the implementation existed,
and the ones that could have passed vacuously were mutation-checked (drop the value sort;
record a conflict on agreeing answers; drop the dedup guard; stop collapsing PocketBase's
`"null"`; drop the `adultNeedsUpdate` clause; drop `sameAnswer` from each of its three call
sites; drop `maxSize` from the migration — each turns the corresponding test red).

- `TestProcessAdultsRecordsResidualAttributeConflicts` — silent when the answers agree, silent
  when normalisation collapses them, silent on gap-fill, silent when the difference is only
  case or whitespace; recorded (with the losing value, keyed by column) when they genuinely
  differ; duplicate losers deduped; multiple losers sorted.
- `TestProcessAdultsConflictsLeaveTheMergedValueUnchanged` — both load orders: the stored value
  is still the first-non-empty winner, and reversing the input reverses winner *and* loser.
- `TestProcessAdultsEmailConflictRecordsTheDisplacedValue` — when `preferEmail` displaces the
  first-loaded answer, the **displaced** value is the conflict, not the candidate.
- `TestProcessAdultsEmailCaseVariantIsNotAConflict` — one address in two capitalisations is one
  answer, in both load orders, and the untrimmed spelling `preferEmail` displaces does not get
  recorded as conflicting with itself.
- `TestAdultConflictsJSONIsCanonical` — stable rendering, so the compare-before-write does not
  rewrite conflicted rows on every run.
- `TestUpsertAdultsPersistsAttributeConflicts` — real record round trip: written on create, an
  identical re-run skips, and a resolved conflict clears the column.
- `TestAdultsCollectionHasAttributeConflictsColumn` — the Go writer is inert without the
  migration, and the migration keeps its `maxSize: 50000` rather than falling back to
  PocketBase's 1 MB json default.

Gates: `pre-push-verify.sh` all green (go build, golangci-lint, full Go suite, prettier,
eslint, tsc, vitest, migration headers, `options:{}` anti-pattern scan, pb-js-lint). Migration
applied, reverted and re-applied against a scratch DB; `maxSize: 50000` verified present in the
live schema rather than silently defaulted. A pre-migration row whose column is SQL NULL reads
back as "no conflicts" and is **not** rewritten, so the first sync after deploy touches only
rows that genuinely changed. `frontend/src/types/pocketbase-types.ts` regenerated from that
scratch DB — +5/-2 lines, no formatting churn, and the table has no frontend consumers today.

**Deliberately closes nothing.** kindred#2275 stays OPEN to carry the remaining half of Option B —
the badge and tooltip that read this column. Owner ruling, 2026-08-17: keep **all seven** columns
recording for now ("more data doesn't hurt while we're still landing the GUI and tweaking the
backend"), so the wide reading in this PR is the intended scope and the narrowing option is
declined. Do not add a closing keyword here until the badge ships.

### DIFF
--- a/docs/reference/tables.md
+++ b/docs/reference/tables.md
@@ -652,6 +652,7 @@ Family camp adult attendees extracted from custom values.
 | `gender` | text | Gender |
 | `date_of_birth` | text | Date of birth |
 | `relationship_to_camper` | text | Relationship |
+| `attribute_conflicts` | json | Answers the merge DISCARDED, as `{column: [other values]}`. NULL when the slot's answers agreed. The winner stays in its own column; this holds only losers, and only where they differ by more than case or whitespace (kindred#2275) |
 
 **Unique**: `(household, year, adult_number)`
 
--- a/frontend/src/types/pocketbase-types.ts
+++ b/frontend/src/types/pocketbase-types.ts
@@ -611,8 +611,9 @@ export type EnrollmentSnapshotsRecord = {
   year: number
 }
 
-export type FamilyCampAdultsRecord = {
+export type FamilyCampAdultsRecord<Tattribute_conflicts = unknown> = {
   adult_number: number
+  attribute_conflicts?: null | Tattribute_conflicts
   created: IsoAutoDateString
   date_of_birth?: string
   email?: string
@@ -1788,7 +1789,9 @@ export type DivisionsResponse<Texpand = unknown> = Required<DivisionsRecord> &
   BaseSystemFields<Texpand>
 export type EnrollmentSnapshotsResponse<Texpand = unknown> = Required<EnrollmentSnapshotsRecord> &
   BaseSystemFields<Texpand>
-export type FamilyCampAdultsResponse<Texpand = unknown> = Required<FamilyCampAdultsRecord> &
+export type FamilyCampAdultsResponse<Tattribute_conflicts = unknown, Texpand = unknown> = Required<
+  FamilyCampAdultsRecord<Tattribute_conflicts>
+> &
   BaseSystemFields<Texpand>
 export type FamilyCampMedicalResponse<Texpand = unknown> = Required<FamilyCampMedicalRecord> &
   BaseSystemFields<Texpand>
--- a/pocketbase/pb_migrations/1500000160_family_camp_adults_attribute_conflicts.js
+++ b/pocketbase/pb_migrations/1500000160_family_camp_adults_attribute_conflicts.js
@@ -1,0 +1,80 @@
+/// <reference path="../pb_data/types.d.ts" />
+/**
+ * Migration: add `attribute_conflicts` to family_camp_adults. kindred#2275.
+ *
+ * ONE ADDITIVE JSON COLUMN. The row grain is UNCHANGED and stays
+ * (household, year, adult_number) -- see "what this is not" below.
+ *
+ * Shape: {column: [other values]} -- the family_camp_adults column that
+ * received more than one answer, mapped to the answers the merge DISCARDED.
+ * The answer that won is still in the column itself, so a reader shows the
+ * column value and this map's values side by side. NULL when the slot's
+ * answers agreed, which is ~94% of rows.
+ *
+ *   {"date_of_birth":["1981-09-02"],"relationship_to_camper":["Mother"]}
+ *
+ * WHY ONE ADULT SLOT RECEIVES TWO ANSWERS -- read this before proposing a
+ * re-grain, because the obvious reading is wrong. It is NOT two children
+ * reporting on their parents. CampMinder asks the family-camp questions on a
+ * per-CAMPER form that covers all of a household's summer and family sessions,
+ * so one parent fills the same family-camp section once per child, on a form
+ * where that section should have been skipped after the first. A divergence is
+ * therefore one person being less careful the second or third time -- a
+ * data-entry artifact of form design, not two independent observers disagreeing
+ * about a fact, and not evidence that the row is keyed at the wrong grain.
+ *
+ * WHAT THIS IS NOT: the camper re-grain. It was proposed (a 2026-08-15 ruling),
+ * then DECLINED by the owner on 2026-08-17 in favour of this column, on the
+ * reading above. So NOTHING here touches the grain triple -- not the write key
+ * in pocketbase/sync/family_camp_derived.go, not TrackProcessedCompositeKey,
+ * not idx_fc_adults_unique. The merge policy is also unchanged:
+ * first-non-empty-wins still decides which answer is stored, and preferEmail
+ * (kindred#1945) still breaks the email tie on well-formedness.
+ *
+ * ONLY THE RESIDUAL LIGHTS UP. The kindred#2405 normalisers run first, so the
+ * 583 date_of_birth and 146 relationship_to_camper divergences that were only
+ * ever two spellings of one answer (`09-02-1979` vs `9/2/1979`; `mother` vs
+ * `Mom`) collapse before the comparison and record nothing.
+ *
+ * NO DATA BACKFILL. The column is written by the family_camp_derived sync,
+ * which recomputes every row it touches; until that runs the column is NULL,
+ * which reads correctly as "not yet computed".
+ *
+ * maxSize 50000: the map holds at most the seven person-sourced columns, each
+ * with at most four discarded answers (five campers is the observed maximum on
+ * one household form), each bounded by its own column's width -- the widest is
+ * `email` at 500 characters. Roughly 14 KB worst case, and PocketBase REJECTS
+ * an over-cap write rather than truncating it, so the headroom is deliberate.
+ *
+ * PocketBase v0.23 syntax: fields.add(new Field({...})) with properties
+ * DIRECT, not inside an options{} wrapper -- a plain object, or a wrapped
+ * property, is silently ignored. The add is guarded so a re-run (or a
+ * re-applied migration after consolidation) is a no-op.
+ */
+
+migrate(
+  (app) => {
+    const col = app.findCollectionByNameOrId('family_camp_adults');
+
+    if (!col.fields.getByName('attribute_conflicts')) {
+      col.fields.add(
+        new Field({
+          type: 'json',
+          name: 'attribute_conflicts',
+          required: false,
+          presentable: false,
+          maxSize: 50000,
+        })
+      );
+    }
+
+    app.save(col);
+  },
+  (app) => {
+    const col = app.findCollectionByNameOrId('family_camp_adults');
+
+    col.fields.removeByName('attribute_conflicts');
+
+    app.save(col);
+  }
+);
--- a/pocketbase/pb_migrations/1500000160_family_camp_adults_attribute_conflicts.js
+++ b/pocketbase/pb_migrations/1500000160_family_camp_adults_attribute_conflicts.js
@@ -9,7 +9,12 @@
  * received more than one answer, mapped to the answers the merge DISCARDED.
  * The answer that won is still in the column itself, so a reader shows the
  * column value and this map's values side by side. NULL when the slot's
- * answers agreed, which is ~94% of rows.
+ * answers agreed. Measured by replaying processAdults over data-prod.db:
+ * 1,240 of 9,789 slots carry a map all years (12.7%), 92 of 834 in 2026
+ * (11.0%) -- so ~87% of rows are NULL, not the ~94% an earlier draft of this
+ * comment carried. That 94% was the residual of the two NORMALISED columns
+ * only; this column records every person-sourced attribute that merges, which
+ * is a larger population.
  *
  *   {"date_of_birth":["1981-09-02"],"relationship_to_camper":["Mother"]}
  *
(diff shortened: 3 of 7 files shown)


adamflagg:
## Independent review — 4 findings, all fixed in-branch, nothing filed

Reviewed the full `main...HEAD` diff without having written it. Every numeric claim in the
body was re-derived independently before being trusted; the method is at the bottom.

### Fixed in-branch

**1. (medium) A spelling was being recorded as a conflicting answer.** `mergeFirstNonEmpty`
and the email arm compared with `==`, so any pair that differed only in letter case or
whitespace was written into `attribute_conflicts`. Replaying `processAdults` over
`data-prod.db`, all years: **189 of the 1,429 lit slots — and 32 of 2026's 124, over a
quarter of the year staff will actually look at — were nothing but capitalisation.** Worst
columns: `email` 111 of 325, `first_name` 118 of 791, `last_name` 4 of 15.

That contradicts the PR's own governing principle ("two spellings of one answer stay
silent"), which it had applied to the two normalised columns and to nothing else. `email`
is the least arguable: a mail domain is case-insensitive by RFC 1035 and every provider
treats the local part that way, so `Amy@example.com` vs `amy@example.com` is not a second
answer.

Fixed with `sameAnswer` (fold case, collapse whitespace runs) gating **only** what is
recorded. The non-obvious half is that the email arm needs it on **both** branches: a
leading space fails `emailFormatPattern`, so `" a@x.com"` is *displaced* by `"a@x.com"` and
was being recorded as conflicting with itself. Deliberately narrow — no punctuation
stripping, no token reordering, no nickname folding; `Amy Johnson` vs `Amy R Johnson` is
still two answers.

Recorded conflicts drop 1,429 → 1,240 all years, 124 → 92 in 2026. Body tables updated.

**2. (low) The migration comment's "~94% of rows" was wrong.** The real figure for this
column is ~87% (~89% for 2026). 94% was the residual of the two *normalised* columns only —
exactly the mistake the PR body's own "Scale, re-measured" section exists to correct, left
uncorrected one file over. Rewritten with the measured figures and a note saying where the
94% came from, so nobody re-derives it.

**3. (low) `TestAdultsCollectionHasAttributeConflictsColumn` did not pin `maxSize`.** The
body claims the 50000 cap was verified in the live schema, but nothing stopped a later edit
from dropping it — PB v0.23 then silently applies its 1 MB json default and the cap the
column's comment justifies exists nowhere. Added `maxSize: 50000` to the assertion list.

**4. (low) `docs/reference/tables.md` still listed the old column set** for
`family_camp_adults`. That table is what a reader checks before writing a query. Added the
column. Committed with the formatter hook off on purpose: prettier is not a markdown gate in
this repo, and letting the pre-commit hook touch that file would have dragged 958 lines of
unrelated table realignment into the diff. markdownlint is clean.

### Refuted — checked, not defects

- **Removing the `&& adult.x == ""` guards changed the switch's fall-through semantics.**
  True in principle: an arm that no longer fails can no longer fall through to a later one.
  Enumerated all 1,270 `custom_field_defs` rows through `isFamilyCampField` + `extraFieldCMIDs`
  + `extractAdultNumberFromField`; **no admitted field name matches two arms**, so no
  behaviour changes. The stored-value fingerprint below independently confirms it.
- **The `-Other` follow-up fields feed their parent question's column.**
  `FAM Camp Adult 1-Pronouns other`, `FAM CAMP Adult 1/2 Gender-Other` and
  `Family Camp Adult 2-Pronouns Other` are all admitted and all route to the same column as
  the question they elaborate — so they would land as false conflicts. **Zero rows in
  `person_custom_values` across all ten years**, and the mis-routing predates this PR (they
  could already win the merged column). Not filed: no observable effect, and filing it would
  be a speculative-refactor issue.
- **PII into the synthetic CD seed.** The column carries names, emails and birthdays, but
  `family_camp_adults` is on `scan_leaks.py`'s `DROP_LIST_TABLES` and is emptied wholesale.
- **Churn on the first sync after deploy.** Probed a row whose column is genuinely SQL NULL
  (migration applied, never written): `GetString` renders `"null"`, `storedAttributeConflicts`
  collapses it to `""`, and `adultNeedsUpdate` returns **false**. The first run touches only
  rows that genuinely changed.
- **`Set(col, nil)` does not really store SQL NULL.** It does — verified by reading
  `attribute_conflicts IS NULL` straight out of the test DB. The comment is accurate.
- **The `attendees.go` rider is unsupported.** It is supported: 2,315 attendees / 2,904
  `SessionProgramStatus` entries / 0 carrying an `ID` key is verbatim the evidence recorded
  when #2263 closed on 2026-08-17.
- **New column breaks a consumer.** No frontend consumers of `FamilyCampAdults*`;
  `fetch_family_camp_adults` uses `getattr` with no `fields=` restriction and
  `lodging_roster_service` projects to `PartyAdult`, which does not carry it. Migration
  number 1500000160 is above `origin/main`'s highest (159) and claimed by no other branch.

### Filed

Nothing. Every finding was in code this PR wrote or in one file, so all four were fixed here.

### Verification

**Independent replay.** Rebuilt the field-admission map, the person→household mapping and
both custom-value loads straight from `pocketbase/pb_data/data-prod.db`, then called this
branch's `processAdults` per year. Reproduced the pre-review body **exactly** — 2026: 834
slots / 124 lit / 46 DOB-or-relationship; all years: 9,789 / 1,429 / 612; and every per-column
figure (`first_name` 791, `date_of_birth` 536, `gender` 507, `email` 325,
`relationship_to_camper` 168, `pronouns` 75, `last_name` 15). Also confirmed the body's 9,143
historical rows against the table itself. The measurements were sound; it was the recording
rule that needed narrowing.

**The fix does not change the merge.** SHA-256 over the eight stored attributes of all 9,789
replayed slots, sorted: `24d14af8cf6baba38b43b537` **before and after** — byte-identical.

**Mutation checks** (implementation broken, suite run, restored):

| mutation | result |
|---|---|
| drop `slices.Sort(values)` in `conflictsJSON` | RED — `{"gender":["Nonbinary","F"]}` |
| drop the dedup guard in `noteConflict` | RED — `{"gender":["F","F"]}` |
| record a conflict on agreeing answers | RED — `{"relationship_to_camper":["Mother"]}` vs none |
| stop collapsing PocketBase's `"null"` | RED — `resolved row still carries attribute_conflicts = null` |
| drop the `adultNeedsUpdate` clause | RED — `clearing a conflict must count as a change` |
| drop the displaced-email `noteConflict` | RED |
| drop the `else if` email `noteConflict` | RED |
| **drop `sameAnswer` from `mergeFirstNonEmpty`** | RED (new) |
| **drop `sameAnswer` from the email true branch** | RED (new) |
| **drop `sameAnswer` from the email `else if`** | RED (new) |
| **drop `maxSize: 50000` from the migration** | RED (new) |
| write `""` instead of `nil` in `setAttributeConflicts` | **GREEN — unpinned.** Left alone: it round-trips identically through `storedAttributeConflicts`, so there is no behaviour to pin |

New tests written and watched fail before the fix: three cases in
`TestProcessAdultsRecordsResidualAttributeConflicts` (case-only, whitespace-only, and case
folding on a non-free-text column) and `TestProcessAdultsEmailCaseVariantIsNotAConflict`
(both load orders plus the displaced-untrimmed case).

`bash scripts/pre-push-verify.sh` green on the full run. Push confirmed by the remote ref, not
by the tool's verdict. Not merged, and CodeRabbit not tagged.

adamflagg:
@/tmp/claude-1000/-home-adam-kindred/27c7bc7d-e1c2-482e-876c-21fefc7be78a/scratchpad/2421-hold.md

Issue #2382, opened 2026-08-15
feat(lodging): write-ins must be scenario-scoped, not only session-scoped


adamflagg:
## ⚠️ CORRECTED 2026-08-16 — phase D triage. Model changed, and the open sub-question is now RULED.

Verified against `main` @ `a946565a`.

### 1. The fix is a SPLIT, not a draft twin

`family_available` conflates two different questions and only one of them is scenario-scoped:

| question | value | scenario-scoped? |
|---|---|---|
| staff↔family **ROLE** — "this staff cabin is released to families" | `true` | **No.** A release names no occupant |
| **OCCUPANCY** — "a family/staff member is in this room" | `false` | **Yes** |

A draft twin of the conflated table carries the whole row, so it would carry the ROLE too. Classify every
read and write of `family_available` as ROLE or OCCUPANCY first — **that classification is the migration
design.**

### 2. ⛔ OWNER RULING 2026-08-16 — a new scenario DOES inherit the live board's write-ins

The open question was: once write-ins are scenario-scoped, a fresh scenario starts empty (today all 21 rows
are shared by every scenario). **Ruling: copy them automatically in BOTH seed paths** — `copy_from_mirror`
and `copy_scenario_to_scenario`.

The reason this is not merely a convenience: without it, **#2247's placement gate would permit dropping a
family into a room the live board records as occupied.** That is a failure mode this split would *create*,
not one it inherits. Precedent is already in the tree — `copy_scenario_to_scenario` faces the same choice for
`lodging_slot_merges` and answers it by copying the source scenario's own rows while deliberately not
copying the weekend-level tier (`lodging_write_service.py:533-548`).

### 3. #2381 does NOT dissolve into this. Confirmed, contrary to an earlier hypothesis.

It was hypothesised that once occupancy is its own table, N write-ins under a container becomes a natural
one-to-many and #2381 disappears. **It does not.** The single-cover arity survives a storage change because
it lives in four type signatures a new table does not touch: `write_in_covers -> dict`,
`LodgingUnitSummary.write_in`, `coveringWriteIn`, and `availabilityAction.unitId`. Sequence #2382 first
anyway — it is the larger change and shares files — but do not expect #2381 to vanish.

Closes #2382. Does **not** close #2381.

---

## What

Write-ins are stored per **(unit, session, year)** and are therefore shared by every scenario for that
weekend. They need to be **scenario-scoped**.

Owner, 2026-08-15:

[...]

Issue #2381, opened 2026-08-15
bug(frontend): write-ins do not merge/split like assignments — a merged building shows only one


adamflagg:
## ⚠️ CORRECTED / RE-SCOPED 2026-08-17 — this is a BEHAVIOUR-PARITY bug, not a display bug

**What changed and why.** The body below previously framed this as a *display gap* (a merged building
resolves one write-in and hides the rest) with an *open design question* about the CLEAR affordance. The
owner read the reporter's original feedback (`kindred-feedback#115`) directly and restated the intent: the
complaint is not "the card is missing names", it is that **write-ins do not merge and split the way ordinary
assignments do.** Regular placements survive a merge/split round-trip with their leaf placement intact;
write-ins do not. That reframing makes the arity change the *mechanism* rather than the goal, and it adds two
parts the old body did not have (drop the auto footer note; per-write-in removal).

The clear-affordance question is **answered** and is no longer open — see part 4.

Everything below this block that was still sound is kept: the four-type-signature analysis, the file anchors,
the "does not dissolve into #2382" finding, and the what-is-NOT-wrong list. Only the scope and the framing
were wrong.

Also folded in: **`kindred-feedback#106`** ("Clear write-in, make that more clear") — same reporter, same
defect surfacing a second time from the other end. It is not a separate wording issue; it is part 4 here.

---

## Goal

**A write-in must merge and split exactly the way an assignment does.** N leaf write-ins under a merged
container show as N; splitting the container restores each to its own room; neither operation loses, moves or
rewrites a row.

That parity already holds for assignments, and it holds *because merge writes nothing to them*.
Verified on `main`:

- `LodgingBoard.tsx:312` (the `mergeUnit` callback) → `setCombined(parentUnit.unit_id, parentUnit.name, true)`
  at `:316` → `POST` slot-merge → `api/services/lodging_write_service.py:949` (`set_slot_merge`) writes **one
  `lodging_slot_merges` row and nothing else** — it has no delete branch and touches no other table.
  `lodging_assignments.units` is never rewritten by a merge or a split.
- The board then rolls leaf assignments up to whichever level draws (`coveredCodes` at
  `frontend/src/components/weekend/unitLevel.ts:38`, `drawnUnits` at `:231`), which is why a merged card can
  already show several disjoint-room parties at once: `BoardSlot.parties` is a LIST
  (`frontend/src/components/weekend/boardLayout.ts:491-495`), and `overlappingPartyKeys`
[...]

PR #2422, opened 2026-08-17
feat(pb): add the lodging write-in tables, dark (#2382, 1 of 4)

adamflagg:
`lodging_availability.family_available` answers two unrelated questions through one boolean, and only one of them is scenario-scoped:

| stored | means | question | scenario-scoped? |
|---|---|---|---|
| `true` on a `staff_default` unit | a staff cabin OPENED to families this weekend | **ROLE** | **No** — an operational fact ("we're moving staff to X for weekend Y") |
| `false` | somebody is in it — the write-in | **OCCUPANCY** | **Yes** |
| no row | the unit's own role decides | ROLE (default) | n/a |

Owner ruling, 2026-08-15 (#2382): the ROLE half is **not** scenario-scoped and stays exactly where it is; the OCCUPANCY half is, because not every write-in is non-rostered staff — some are paper registrations for families arriving with no children, and that is a modelling choice belonging to the scenario that made it.

So the fix is a **split**, not a draft twin of the conflated table. This PR lands the receiving half of that split and **nothing else**.

## What this PR does

- **`pocketbase/pb_migrations/1500000161_lodging_write_ins.js`** — creates `lodging_write_ins` and `lodging_write_ins_draft`. The occupancy fields mirror `lodging_availability`'s live schema exactly (`unit`, `session`, `session_cm_id`, `year`, `occupant_name`, `note`), so PR 2's backfill is a straight copy with no per-column judgement. The draft adds `scenario` declared the way `lodging_assignments_draft` declares it — required, `cascadeDelete: true`, into `saved_scenarios` — which makes "the live board is never overridable by a scenario" a schema guarantee and makes deleting a scenario take its write-ins with it. Both `name:` values inside `new Collection({...})` are string **literals**, not the file's `LIVE` / `DRAFT` constants: `scripts/dev/verify-migration-history.sh`'s CHECK 2 — the load-bearing half of #2245's renumber detector, since history-sync erases the `_migrations` evidence CHECK 1 reads — finds the collections a migration CREATEs by regexing `name: "..."` out of the up arm, and a constant reads as no collection at all.
- **`pocketbase/lodging/hooks.go`** — registers both collections in `yearScopedRefs` and in `yearImmutableRefs[lodging_units]`, following `collectionSlotMerges` exactly. Both are single `unit` relations, so the filter is a plain `unit = {:id}` with no `.id` join.
- **`api/services/lodging_repository.py`** + **`api/constants/collections.py`** — the live and draft CRUD, following the `lodging_assignments`/`_draft` pattern already in that file.
- **`frontend/src/types/pocketbase-types.ts`** — regenerated against a migrations-only scratch database. Pure addition, 35 lines, no churn.

## What this PR deliberately does NOT do

**After this PR the application behaves identically. That is the acceptance criterion.**

- **No data migration.** The 21 production rows stay in `lodging_availability`. No backfill, no copy, no delete.
- **No reader is switched.** `write_in_covers` (`lodging_roster_service.py:469`), `LodgingUnitSummary.write_in`, `coveringWriteIn`, `set_availability`, `lodgingApi.ts` and `UnitAvailabilityControl.tsx` are untouched and still resolve write-ins through `family_available`. The new tables are empty and nothing reads them.
- `copy_from_mirror`, `copy_scenario_to_scenario`, `writeIn.ts`, `unitBadges.ts` and `boardLayout.ts` (owned by in-flight #2404) are untouched.

## This does not close anything

⛔ **#2382 stays OPEN until PR 4.** This is PR **1 of 4** and closes no issue. The remaining three:

- **PR 2** — backfill the 21 `family_available = false` rows into `lodging_write_ins`, and switch `write_in_covers` / `set_availability` to read and write them.
- **PR 3** — seed a scenario's write-ins in both copy paths (`copy_from_mirror`, `copy_scenario_to_scenario`), per the owner ruling of 2026-08-16; without it #2247's placement gate would permit dropping a family into a room the live board records as occupied.
- **PR 4** — move the frontend off `family_available_override === false` (`writeIn.ts`, `unitBadges.ts`, #2093's forest open-tint proxy), leave `lodging_availability` holding only the ROLE override, and take #2382 to done.

#2381 is **not** dissolved by this. The single-cover arity survives a storage change because it lives in four type signatures a new table does not touch, so the unique index here deliberately mirrors today's one-row-per-`(unit, weekend)` shape rather than widening on its behalf.

## Why this is not a revert of 1500000135

`1500000135_lodging_availability_no_scenario.js` deleted the scenario column and named the twin it was declining to build:

> Availability is a fact about the WEEKEND, not about the plan: a burst pipe closes a cabin in every scenario for that weekend, so there is nothing for a scenario to disagree about.

That reasoning was correct and still is — for availability, which is exactly why `lodging_availability` keeps its no-scenario shape here and is left alone. What changed is what the table *stores*: #2228 turned it into the named-global-occupant store and 1500000148 moved the historical notes into `occupant_name`. A write-in is not a burst pipe, so "nothing for a scenario to disagree about" stopped being true for the occupancy half. The migration header says all of this and cites the ruling rather than reading as a revert.

⛔ A nullable `scenario` sentinel column was explicitly rejected. `api/services/lodging_repository.py` records why verbatim: `lodging_assignments` dropped its scenario column because it *"was dead weight that invited a `scenario != \"\"` write rule instead of a draft table."* The live+draft pair is this repository's established pattern.

## Measurement

Re-verified against `pocketbase/pb_data/data-prod.db` (opened by name — the sibling `data.db` is the dev copy):

```
SELECT family_available, count(*) FROM lodging_availability GROUP BY family_available;
0|21
```

All 21 rows are OCCUPANCY. **Zero are ROLE** — the role half of the column has never been written. So there is no risky partition and no ambiguous row for PR 2 to adjudicate. All 21 carry a non-empty `occupant_name`, and all 21 are distinct on `(session_cm_id, year, unit)`, so the unique index below holds over them.

## Tests

Written failing first, then implemented.

**Go** (`pocketbase/lodging/hooks_test.go`, +5 tests) — the guards are exercised against the running engine, not asserted against the maps:

- `TestGuardUnitYearRejectsACrossYearWriteIn` / `…DraftWriteIn` — a 2027 write-in naming a 2026 unit is refused, live and draft.
- `TestGuardUnitYearAllowsAWriteInSharingItsUnitsYear` — positive control, so a guard that refused every write to these tables would not pass the two above.
- `TestGuardYearImmutableRejectsAUnitYearChangeWithAWriteInDependent` / `…DraftWriteInDependent` — re-seasoning a unit with a dependent write-in is refused.
- `newWriteInsTestApp` mirrors `newSlotMergesTestApp`, adding both collections *after* `setupCollections`, so `TestGuardUnitYearSkipsAnAbsentCollection` keeps a genuinely absent collection to pin.

Mutation-checked, reading the exit code:

- dropping `collectionWriteInsDraft` from `yearScopedRefs` → RED (the guard test **and** `TestYearRefMapsAgreeOnEveryRelation`, which enforces the two maps stay symmetric);
- replacing either filter with one that compiles but matches zero rows (`unit = {:id} && year = 9999`) → RED on both year-immutable tests. That is the silent failure mode `dependentRef`'s doc comment warns about, and it is caught.

**Python** (`tests/unit/api/services/test_lodging_repository.py`, `tests/unit/api/test_lodging_constants.py`) — the live read carries no scenario predicate; the draft read REPLACES rather than overlays (matching #1974's rule for placements, **not** `lodging_slot_merges`' weekend-level union, because a write-in is an occupancy and not a fact about the weekend); both lookups key their own unique index; every client-supplied string reaching a filter is escaped. All four new reads were added to the two `session_cm_id` lists (keyed-on-the-CampMinder-id, never-compared-to-a-string); the stable-sort list took the two paginated `fetch_*` reads only, because that list holds `fetch_*` reads and no `find_*` — `find_availability_override` and `find_slot_merge` are absent from it too.

## Gates

| Gate | Result |
|---|---|
| `bash scripts/pre-push-verify.sh` | ALL CHECKS PASSED |
| `cd pocketbase && go test ./...` | pass (12 packages) |
| `uv run pytest tests/` | 6367 passed, 23 skipped |
| `golangci-lint run` | 0 issues |
| `scripts/dev/verify-lodging-schema.sh` | OK (120 js migrations applied) |
| `scripts/dev/verify-migration-history.sh` | OK |
| `scripts/dev/verify-no-hardcoded-lodging.sh` | OK |
| `migrate up` → `migrate down 1` → `migrate up` on a scratch DB | clean both ways; re-apply with the collections already present is a no-op |
| field caps read back off the migrated DB | `max: 2000` on both text columns, not the 5000 default — no `options` wrapper trap |

**No GUI impact.** The only file changed under `frontend/` is the generated types file.

### The shared gate that made the migration explain itself

`scripts/pre-push-verify.sh`'s `options: {` scan filtered only full-line `//` comments, so a JSDoc header that *quotes* the anti-pattern failed the check that enforces it. This PR first worked around it by wording the note backwards. That is fixed here instead: the filter now skips `*` and `/*` continuation lines as well, and the migration header states the rule the way 1500000132, 1500000135 and 1500000146 already state it. A line whose first non-space character is `*` or `/` cannot be executable JS carrying a field wrapper, so nothing real is hidden — verified in both directions (the three existing headers yield zero residual hits; a real `{ type: "text", name: "x", options: { max: 100 } }` is still reported).

### Review fixes folded in

- The `new Collection({...})` names became string literals, so `verify-migration-history.sh` CHECK 2 can see them. Measured on a scratch DB holding both tables with the `1500000161` row deleted from `_migrations` — the state a renumber leaves — the detector went from **exit 0 with no output** (silent false CLEAN) to exit 1 naming both collections. That detector is also what makes the up arm's skip-if-exists guard defensible, since `docs/reference/pocketbase-migrations.md` warns an idempotent CREATE otherwise leaves a divergent schema "with nothing left to reveal it".
- The down arm ran `app.delete(...)` **inside** the `try`, so a genuine delete failure was caught by the same `catch` that means "already absent" and `migrate down` would report a revert it never performed. The lookup stays in the `try`; the delete moved out, per that document's "Error Handling in Data Migrations".

### DIFF
--- a/api/constants/collections.py
+++ b/api/constants/collections.py
@@ -71,6 +71,26 @@ LODGING_ASSIGNMENTS_DRAFT = "lodging_assignments_draft"
 # plan.
 LODGING_SLOT_MERGES = "lodging_slot_merges"
 
+# Write-in OCCUPANCY -- who is in a room that the roster does not otherwise
+# know about (1500000161, kindred#2382). Split out of `lodging_availability`,
+# which conflated two unrelated questions in one boolean: `family_available`
+# true on a staff cabin is a staff<->family ROLE override for the weekend, and
+# false is an occupancy. The owner ruled (2026-08-15) that the ROLE is NOT
+# scenario-scoped -- it is an operational fact, "we're moving staff to X for
+# weekend Y" -- while an occupancy IS, because not every write-in is
+# non-rostered staff: some are paper registrations for families arriving with
+# no children, and that is a modelling choice belonging to the scenario that
+# made it.
+#
+# So availability keeps only the role half, and occupancy gets a live+draft
+# pair beside LODGING_ASSIGNMENTS / LODGING_ASSIGNMENTS_DRAFT -- the same split
+# between record and plan, and for the same reason. A nullable `scenario`
+# sentinel was explicitly rejected: it is the shape lodging_assignments dropped
+# because it "was dead weight that invited a `scenario != \"\"` write rule
+# instead of a draft table."
+LODGING_WRITE_INS = "lodging_write_ins"
+LODGING_WRITE_INS_DRAFT = "lodging_write_ins_draft"
+
 # The single work queue for cabin strings ingest could not resolve. Owned and
 # solely written by the ingest layer; the admin UI reads it filtered to
 # kind = "unresolved_alias". Deliberately NOT a second surfaces-only collection.
--- a/api/services/lodging_repository.py
+++ b/api/services/lodging_repository.py
@@ -49,6 +49,8 @@ from api.constants.collections import (
     LODGING_SESSION_STATUS,
     LODGING_SLOT_MERGES,
     LODGING_UNITS,
+    LODGING_WRITE_INS,
+    LODGING_WRITE_INS_DRAFT,
 )
 from api.constants.filters import ACTIVE_ENROLLED_FILTER
 from api.dependencies import lodging_cache
@@ -358,6 +360,69 @@ class LodgingRepository:
             LODGING_SLOT_MERGES,
             query_params={
                 "filter": (f"session_cm_id = {session_cm_id} && year = {year} && {scenario_clause}"),
+                "sort": STABLE_SORT,
+            },
+        )
+
+    async def fetch_write_ins(self, year: int, session_cm_id: int) -> list[Any]:
+        """The LIVE board's write-ins for one weekend (kindred#2382).
+
+        A write-in is an occupancy the roster does not otherwise know about:
+        non-rostered weekend staff, or a paper registration for a family
+        arriving with no children. It used to be stored as
+        `lodging_availability.family_available = false`, where it shared a
+        column with the staff<->family ROLE override and inherited that
+        column's session-only scope.
+
+        NO SCENARIO PREDICATE, and that is not the old shape coming back. The
+        live board is a scope in its OWN RIGHT -- the owner's second
+        requirement on 2026-08-15, "we need to allow write ins to happen in
+        campminder prod, not just scenarios, for staff to properly evaluate
+        the board" -- so these rows live in their own table with no scenario
+        column at all, exactly as `lodging_assignments` does. A scenario's
+        write-ins come from `fetch_draft_write_ins` and REPLACE these; there
+        is no overlay.
+
+        NOTHING READS THIS YET. PR 1 of kindred#2382 lands the tables and this
+        CRUD dark; `write_in_covers` still resolves write-ins out of
+        `lodging_availability` until PR 2 switches it.
+        """
+        return await self._page(
+            LODGING_WRITE_INS,
+            query_params={
+                "filter": f"session_cm_id = {session_cm_id} && year = {year}",
+                "sort": STABLE_SORT,
+            },
+        )
+
+    async def fetch_draft_write_ins(self, year: int, session_cm_id: int, scenario_id: str) -> list[Any]:
+        """One scenario's write-ins for one session -- ALL of them.
+
+        REPLACE, not overlay, matching `fetch_draft_assignments` under
+        kindred#1974: these rows are the scenario's whole set of write-ins, so
+        a unit with no row here holds no write-in in this scenario, whatever
+        the live board says. A scenario is seeded by an explicit copy (both
+        seed paths, by owner ruling 2026-08-16) rather than by rendering the
+        live board through the gaps.
+
+        `fetch_slot_merges` is the near neighbour that does the opposite --
+        it unions the weekend-level tier in -- and the difference is what the
+        row records. A draw level is a fact about the WEEKEND that no sync
+        writes, so a shared tier costs nothing. A write-in is an occupancy,
+        the same kind of fact as a placement, so it follows the placement
+        rule.
+
+        `scenario_id` is client-supplied and escaped, for the reason
+        `fetch_draft_assignments` spells out. The weekend is named by
+        `session_cm_id`, a number, so there is no literal for an injected `||`
+        to close.
+        """
+        return await self._page(
+            LODGING_WRITE_INS_DRAFT,
+            query_params={
+                "filter": (
+                    f'session_cm_id = {session_cm_id} && year = {year} && scenario = "{pb_escape(scenario_id)}"'
+                ),
                 "sort": STABLE_SORT,
             },
         )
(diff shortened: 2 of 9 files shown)


adamflagg:
## Independent review — 3 findings, all fixed in-branch, nothing filed

Reviewed at a score floor of 60 against `main`. Two findings are in code this PR wrote; one is a pre-existing shared-gate defect that this PR had to work around, fixed here because it is one file and carries no product judgement. Nothing was filed as an issue.

### Fixed

**1. The migration was invisible to the renumber detector (silent false CLEAN)** — `pocketbase/pb_migrations/1500000161_lodging_write_ins.js`

The two `new Collection({...})` calls passed `name: LIVE` / `name: DRAFT`. `scripts/dev/verify-migration-history.sh`'s CHECK 2 finds the collections a migration CREATEs by regexing `name: "..."` / `name: '...'` out of the up arm, so a constant reads as *no collection at all*. That script's own header calls CHECK 2 the load-bearing half of #2245, because history-sync erases the `_migrations` evidence CHECK 1 reads — and this branch already lives at 1500000161 because 1500000160 was taken, which is the exact situation the detector exists for.

Measured on a scratch DB holding both tables with the `1500000161` row deleted from `_migrations` (the state a renumber leaves):

```
# before
$ verify-migration-history.sh --db <scratch> --migrations-dir pocketbase/pb_migrations
EXIT=0                                   <- no output at all

# after
COLLECTION ALREADY EXISTS — an unapplied migration would re-create it.
  migration: 1500000161_lodging_write_ins.js
  collection: lodging_write_ins
... and the same block for lodging_write_ins_draft
EXIT=1
```

This is also what makes the up arm's skip-if-exists guard defensible rather than a doc violation — see "refuted" below.

**2. The down arm swallowed a real delete failure** — same file

```js
try {
  app.delete(app.findCollectionByNameOrId(name));   // mutation INSIDE the try
} catch {
  // Already absent.
}
```

`docs/reference/pocketbase-migrations.md` § "Error Handling in Data Migrations" is explicit that the mutating op stays *outside* the catch. Inside it, a genuine failure on the delete — a relation into these tables from a later migration, a permission fault, the DB gone away — is reported as "already absent" and `migrate down` claims a revert it never performed. The loop shape makes it worse: nothing runs after the catch for that iteration, which is the `continue`-shaped case that document singles out. The lookup stays in the `try`; the delete moved out behind an `if`.

**3. `scripts/pre-push-verify.sh`'s `options: {` scan failed on comments quoting the rule** (pre-existing)

The scan filtered only full-line `//` comments, so a JSDoc header that *quotes* the anti-pattern failed the check that enforces it. This is not hypothetical for one file: **1500000132, 1500000135 and 1500000146 all carry the sentence today** and would go red on any diff that touched them, and it is why this PR's own header worded the rule backwards and spent a paragraph explaining a lint quirk instead of the schema. The filter now skips `*` and `/*` continuation lines as well, and the migration header is restored to the canonical wording. Verified both ways — the three existing headers yield zero residual hits, and a real `{ type: "text", name: "x", options: { max: 100 } }` is still reported.

**PR body** (it is the permanent squash commit message here): corrected "the four new reads were added to the three existing hand-maintained parametrized lists" — all four went into the two `session_cm_id` lists, but the stable-sort list took only the two paginated `fetch_*` reads, because that list holds no `find_*` at all (`find_availability_override` and `find_slot_merge` are absent from it too). The "not fixed here" section was rewritten now that it is fixed, and the literal-name requirement recorded.

### Refuted

- **"The skip-if-exists CREATE guard violates `docs/reference/pocketbase-migrations.md`."** It does deviate, and the deviation is deliberate and documented in the file. The document's actual objection is that an idempotent CREATE leaves "a divergent schema in place permanently **with nothing left to reveal it**" — and the thing that reveals it is CHECK 2, which finding 1 restores. Prod and CI apply from scratch, so the residual exposure is a local dev DB, which now gets a loud diagnosis. Left as the author wrote it.
- **The production measurement.** Re-ran against `pocketbase/pb_data/data-prod.db` (by name, not the sibling dev `data.db`): `SELECT family_available, count(*) ... → 0|21`. Zero ROLE rows. Claim holds.
- **"The occupancy fields don't really mirror `lodging_availability`."** They do, field for field, checked against the live schema and then against a fresh boot: `unit` cascadeDelete **true**, `session` cascadeDelete **false**, `session_cm_id` required/min 1/max null/onlyInt, `year` 2010–2100 onlyInt, `occupant_name` and `note` max **2000** (not the 5000 `options` trap). Relations resolve to `lodging_units` / `camp_sessions`, and the draft's `scenario` to `saved_scenarios`.
- **"`scenario` is declared differently from `lodging_assignments_draft`."** Identical: relation, required, `cascadeDelete: true`, into `saved_scenarios`. Read off the live schema.
- **"`guardUnitDelete` should count write-ins too."** No. `countAssignments` exists for the *cascadeDelete: false* hazard, where PocketBase strips the id and re-saves the row unvalidated. `unit` on both write-in tables is cascadeDelete **true**, exactly like `lodging_availability` — the row goes cleanly, nothing is silently corrupted.
- **"`verify-lodging-schema.sh` should assert the new tables."** `lodging_slot_merges` is not asserted there either; the precedent is consistent, and I pinned the schema by fresh boot instead (120 js migrations applied, both collections, all seven indexes, caps read back at 2000).
- Below the floor, left alone: `TestWriteInWrites` pins the collection name but not which SDK method was called (a `MagicMock` cannot tell `.create` from `.delete`) — that is what its docstring claims and more than the sibling tables have; and `occupancyFields`' "PocketBase mutates the field objects it is handed" rationale is unverified (v0.23 derives field ids deterministically), but the practice is harmless either way.

### Mutation checks

The two Go guard tests genuinely pin, verified by breaking the implementation and reading the failure:

```
# yearImmutableRefs live filter -> "unit = {:id} && year = 9999"
--- FAIL: TestGuardYearImmutableRejectsAUnitYearChangeWithAWriteInDependent
    hooks_test.go:1949: re-seasoned a unit with a dependent write-in row; want a refusal

# drop collectionWriteInsDraft from yearScopedRefs
--- FAIL: TestYearRefMapsAgreeOnEveryRelation
    yearImmutableRefs[lodging_units] lists lodging_write_ins_draft as a dependent, but
    yearScopedRefs[lodging_write_ins_draft] declares no relation into lodging_units
--- FAIL: TestGuardUnitYearRejectsACrossYearDraftWriteIn
```

Both restored afterwards; `go test ./lodging/` is green.

### Gates

`bash scripts/pre-push-verify.sh` → ALL CHECKS PASSED. `go test ./lodging/` green. `pytest tests/unit/api/services/test_lodging_repository.py tests/unit/api/test_lodging_constants.py` → 140 passed. Scratch DB re-verified after the edits: `up → down 1 → up` clean in both directions, `_migrations` row restored, all seven indexes present. Push confirmed by the remote ref.

Not merged, and CodeRabbit not tagged. Worth noting the green **CodeRabbit** check on this PR reads `Review skipped: automatic reviews are disabled` — it is a no-op notice, not a review.

PR #2423, opened 2026-08-17
feat(api,frontend): split request text by source field and by child

### LINKED ISSUE #2330: feat(frontend,sync): denote which form each of a family's bunk-request signals came from
## ⚠️ CORRECTED 2026-08-16 — phase D triage. Read this block before the 2026-08-14 one below it.

Verdict: **VALID / DECISION (UI-gated)**, scope materially wider than the body below describes.

### 1. ⛔ The field set is FIVE, not three. Owner ruling 2026-08-16.

The body below is correct that **three** fields feed `family_camp_registrations.request_text`. It is
wrong to treat that as the design surface. Staff require **all five free-text bunk-request fields** on
the panel, across **two lanes**:

| # | CampMinder field | id | lane | author | rostered 2026 households |
|---|---|---|---|---|---|
| 1 | `COVID-19 Bunking Requests` | 206286 | family-camp form | parent | **205** |
| 2 | `Share Bunk With` (`bunk_request_form`) | — | bunking CSV | parent | **104** |
| 3 | `Shared-request` | 274133 | family-camp form | parent | **100** |
| 4 | `BunkingNotes Notes` (`bunking_notes`) | — | bunking CSV | **staff** | **28** |
| 5 | `Internal Bunk Notes` (`internal_notes`) | — | bunking CSV | **staff** | **8** |

A sixth candidate exists and is **not** in the ruled set: `Do Not Share Bunk With`
(`staff_not_bunk_with`), 3 rostered households. Same code path; include it only if asked.

**`socialize_with` (`RetParent-Socializewithbest`) is NOT free text and must be excluded.** It has
exactly two distinct values in 2026, both 40 characters, and `frontend/src/utils/requestBucket.ts:18`
already classes it `IMMATERIAL_SOURCE_FIELDS`.

**Two of the five are staff-authored, not parent-authored** — `BunkingNotes Notes` and `Internal Bunk
Notes`. All 34 `BunkingNotes` values end in an inline staff signature and timestamp
(`<staff name> (Mon  D YYYY  H:MMam)`); no other field does this (0 of 248, 0 of 158, 0 of 134, 0 of 8).
They must not render in the same amber treatment as a family's own request.

### 2. Denominator and scope corrections

The body below denominates on `family_camp_registrations` rows and reports **67 of 251 (27%)** as the
consolidation case. That figure **reproduces exactly at its own scope** (three family-camp fields,
registration denominator) and is not wrong — it is too narrow. At the ruled five-field scope, against
**rostered** households:

- **382** households rostered into a 2026 family session (`status_id = 2`, 8 sessions)
- **270 (71%)** carry at least one free-text bunk-request signal
- **128 of 270 (47%)** carry text in **two or more distinct source fields** — the real consolidation case
- **32** rostered households carry text ONLY in the bunking-CSV lane, so their request text is
  **invisible on every weekend surface today** (the API reads only `family_camp_registrations`)

### 3. Layout load — what the design must survive

Blocks per household after exact-lowercase dedup on (field, text), five-field set:

| blocks | households |
|---|---|
| 1 | **116** (43%) |
| 2 | **87** (32%) |
| 3 | **38** (14%) |
| 4 | **22** (8%) |
| 5 | **6** (2%) |
| 7 | **1** |

**29 households render four or more blocks.** Largest household totals **1,109 characters**. The panel
is **416px** wide (`FamilyDetailsPanel.tsx:366`, `w-[26rem]`) and already stacks Placement, Party,
Share request, Housing needs and the journey card above this section.

Max children answering ONE field is **3**. The worst-case layout is driven by field count × sibling
divergence, **not** family size: the 7-block household has only two children, and the single 4-child
household renders 2 blocks.

Sibling divergence, five-field set: 103 households have ≥2 children answering the same field, across
131 (household, field) groups; **83 of 131 (63%) hold divergent text**, 48 are exact duplicates.
Divergence is real, not punctuation noise — stripping all punctuation collapses only 2 more groups, and
75 of the 83 differ within their first 20 characters. `Share Bunk With` diverges near-universally
(29 groups, 28 divergent) because siblings name their own friends; `BunkingNotes Notes` never does
(6 groups, 0 divergent).

### 4. #2256 is FOLDED IN, not a predecessor. It is closed as superseded by this issue.

The body below sequences #2256 ahead of this work ("has to land first or be folded in"). **Folded in.**
`bunking_notes` is one of the five fields above, rendered by the same component through the same API
model, so a separate PR for one field would be re-done immediately.

Measured, which settles the disagreement that made this look like a blocker: the invisible-household
count is driven almost entirely by **`bunk_request_form`** (104 rostered households, 32 with no
family-camp text at all), **not** by `bunking_notes` (28 households, 7 with no family-camp text). So
this issue delivers 104 households of value regardless of `bunking_notes` coverage.

**Data-coverage caveat, carried forward from #2256 — a caveat, not a blocker.** The CampMinder report
that sources the bunking CSV is scoped to summer enrolment, so `bunking_notes` reaches roughly 28 of
382 rostered households until that report is widened. Owner ruling 2026-08-16: *"sourcing fresh data is
irrelevant from showing the data on the modal … we can widen the report itself to capture more session
enrolled campers and see what data we have after."* Build for all five fields; render nothing where a
field is empty.

### 5. The `'; '` join is genuinely irreversible — this cannot be a frontend-only fix

`lodging_requests.go:401-413` dedupes by lower-cased value, then `strings.Join(a.textParts, "; ")`.
**10 of 422 non-blank 2026 request values (2.4%) contain `'; '` themselves** (37 of 2,161 across all
years), so no client-side split is possible. The split must happen in Go/API. `api/schemas/lodging.py:324-343`
already claims this and the claim is TRUE as measured.

Splitting by field costs almost no duplication: only **4** rostered households hold the same exact text
in two different fields (today's per-household dedup at `lodging_requests.go:398-406` collapses those,
and a per-field split would re-expose them); 9 more share their first 20 characters across fields.

### 6. Confirmed from the 2026-08-14 block below

`request_source_field` **never names a text field** — it holds `FAM CAMP-Share Cabins` (415 rows) or
`FAM CAMP-Shared Cabin` (10) for 2026, both share-**gate** fields. Restricted to rows with non-blank
`request_text`, the 244/5/2 split reproduces exactly. Anything reading it as text provenance is wrong.

One figure in the block below needs correcting: `share_answers_conflict` is true on 16 of 464 rows,
which is **6.9% of the 232 form-sourced answerers**, not the 7.5% asserted in
`api/schemas/lodging.py`'s `ShareRequestSummary` comment.

⚠️ **Every figure in a 2026-08-14 correction block on any issue was measured against
`pocketbase/pb_data/data.db`, which is the LOCAL DEV copy, not production.** The two databases hold
identical row counts on the family-camp tables, which is why nobody noticed. Figures in THIS block are
from `data-prod.db` by name.

### 7. ⏸ Still gated: what these fields are CALLED on screen, and the layout

**Not decided.** A mockup was built 2026-08-16 against the four real household shapes with three
candidate layouts (always-expanded labelled stack / one chip per source field opening on click /
two groups splitting family-authored from staff-authored). The owner has not yet ruled.

Also undecided and needed before implementation: **display names for the five fields.**
`COVID-19 Bunking Requests` carries 205 of 270 households and has plainly been repurposed as the
general bunking-request field years after COVID; `Shared-request` and `Share Bunk With` are
near-indistinguishable to a reader and come from two different forms. Rendering the raw CampMinder
names would confuse staff. **Do not invent labels.**

Whatever wins must compose with **#2255's ruling** for this same panel: a YES renders a chip naming the
thing, a NO renders nothing, a null renders nothing, clicking a chip opens that pair's text, all start
collapsed, and there is no "nothing applicable" text anywhere.

---

> The three-field join and its cm_ids are exactly right and still current. Two claims are wrong: `eligibility_source`/`answers_conflict` ARE already rendered — `answersConflictDetail` (boardLayout.ts:369) names which form won and FamilyCard.tsx:480 uses it — just not on the panel this issue is filed against; and the #2250 warning is stale, that bug closed and shipped. On the central question: which form a free-text signal came from is NOT recoverable from the derived column (the joiner concatenates with '; ' and keeps no field boundary), but it IS fully recoverable from person_custom_values, and it only actually matters for the 67 of 251 2026 households (27%) whose text comes from two different fields.

1. WRONG: "**Neither is rendered.** `ShareRequestPanel` draws `preference`, `preference_raw`, `proximity` and `request_text` and nothing else". REPLACE WITH: "Both are already rendered — on the BOARD CARD. `answersConflictDetail` (boardLayout.ts:369-388) branches on `eligibility_source` and produces 'the Family Camp form said X — staff use the form's answer', and FamilyCard.tsx:480 consumes it. What is missing is the same treatment on `ShareRequestPanel`, which FamilyDetailsPanel.tsx:301 and HouseholdRosterRow.tsx:286 both render and which draws only preference, preference_raw, proximity and request_text. So step 1 is 'reuse `answersConflictDetail` on the panel', NOT 'display two fields nothing shows' — and it must not invent a second wording."
2. STALE: "⚠️ **#2250 is a live bug on the disagreement signal** — the 'Answers disagree' chip hides its explanation in `sr-only` text. Fix it or coordinate with it". REPLACE WITH: "#2250 is CLOSED and shipped (a5f90380, PR #2313) — the chip now carries a real tooltip. The warning stands in a different form: reuse `answersConflictDetail` rather than adding a second disagreement affordance beside it."
3. ADD, as the finding on recoverability: "For the FREE-TEXT half the source is destroyed in the derived column and only there. `lodging_requests.go:400-413` appends the three values to `a.textParts` and joins them with '; ', keeping no field boundary; `request_source_field` does NOT help — it is set only on the gate branch (:389, :397-398) and names the share-gate field, which is why its stored values are 'FAM CAMP-Share Cabins' (244) and 'FAM CAMP-Shared Cabin' (5) rather than any of the three text fields. The raw per-field values ARE still in `person_custom_values`, so step 3 is a real, buildable change rather than an archaeology problem. Scope note: it only bites where a household answered two of the three — 67 of 251 households with 2026 request text (27%); 184 have a single source, and the three fields barely overlap in time (206286 spans 2021-2026, 240598 only 2024-2025, 274133 only 2026), so year alone already answers 'which form' for the other 73%."

- request_text is derived for 2026 ONLY: 251 of 464 2026 registrations carry non-blank request_text; every earlier year is 0. SQL: SELECT year,request_source_field,COUNT(*) FROM family_camp_registrations WHERE request_text<>'' GROUP BY 1,2 -> 2026|FAM CAMP-Share Cabins|244, 2026|FAM CAMP-Shared Cabin|5, 2026||2
- share_eligibility_source, 2026 only, denominator 464 registrations: form=232, registration=193, none=39. Every year 2017-2025 is blank. SQL: SELECT year,share_eligibility_source,COUNT(*) FROM family_camp_registrations GROUP BY 1,2
- share_answers_conflict is true on 16 of 464 2026 registrations (3.4%) and 0 in every earlier year. That is the SAME 16 boardLayout.ts:326-329 calls 'stale, can only be equal or higher after #2269' — this snapshot still shows 16, so either the snapshot predates the #2269 recompute or the count genuinely did not move. SQL: SELECT year,share_answers_conflict,COUNT(*) FROM family_camp_registrations GROUP BY 1,2
- Multi-source free text, from the RAW values (which survive): households with non-blank values in 2 of the 3 text fields — 2026: 67 of 251 (27%); 2025: 84 of 365 (23%); 2024: 45 of 320 (14%). No household has all three. SQL: WITH defs AS (SELECT id,cm_id FROM custom_field_defs WHERE cm_id IN (274133,240598,206286)), v AS (SELECT pcv.year yr,p.household hh,d.cm_id cm FROM person_custom_values pcv JOIN defs d ON d.id=pcv.field_definition JOIN persons p ON p.id=pcv.person WHERE trim(pcv.value)<>'') SELECT yr,nfields,COUNT(*) FROM (SELECT yr,hh,COUNT(DISTINCT cm) nfields FROM v GROUP BY 1,2) GROUP BY 1,2
- The three text fields are live in different eras, which is why 'which form' is mostly answerable by year alone: 206286 appears 2021-2026 (144/260/284/356/412/249 person rows), 240598 only 2024-2025 (112/171), 274133 only 2026 (173). SQL: SELECT pcv.year,d.cm_id,COUNT(*) FROM person_custom_values pcv JOIN custom_field_defs d ON d.id=pcv.field_definition WHERE d.cm_id IN (274133,240598,206286) AND trim(pcv.value)<>'' GROUP BY 1,2
- Cross-check that the 2026 join is sound: 184 single-source + 67 two-source = 251, which equals the 251 rows with non-blank request_text. The join returns something, so the zero-row trap does not apply.

</details>

---

Ask: on the family details panel, **visually separate and denote where each bunk-request signal came from** — the bunking form fields (Share Bunk With, internal notes, bunking notes), the registration form's generic share choice, and the Family Camp information sheet's own choice-plus-detail fields. What that then means for the board (agreement/disagreement chip, detail, icon) is explicitly deferred to a later conversation.

## The API pre-joins the exact thing this asks to separate

`ShareRequestPanel.tsx:17` states the current contract outright: the verbatim request text *"arrives pre-joined from three source fields and is never split."* Those three are, from `pocketbase/sync/lodging_fields.go:150-155`:

| CampMinder field | cm_id |
|---|---|
| `Shared-request` | 274133 |
| `FAM CAMP-Share Comments` | 240598 |
| `COVID-19 Bunking Requests` | 206286 |

They are concatenated into `family_camp_registrations.request_text` in Go and passed through unchanged by the API (`api/services/lodging_roster_service.py:1729`). **So this is not a frontend-only change** — the source is destroyed before the panel ever sees it.

## What provenance the wire already carries

For the *structured* half, `ShareRequestSummary` already has it:

- `eligibility_source: 'form' | 'registration' | 'none'` — which of the two share questions produced the verdict
- `answers_conflict: boolean` — that the two disagree

**Neither is rendered.** `ShareRequestPanel` draws `preference`, `preference_raw`, `proximity` and `request_text` and nothing else, so the panel shows the *verdict* while hiding *which form gave it*. That is the cheapest half of this issue: display two fields already on the wire, no backend work.

The two structured sources are `FAM CAMP-Share Cabins` (240877, registration, early) and `FAM CAMP-Shared Cabin` (263379, the Family Camp information form, later in season). Staff treat the later form as authoritative and registration as a fallback, which is what `eligibility_source` records — see `docs/reference/family-camp-field-provenance.md`.

## Summer already solved "which form said this" — model it

`utils/sourceFromField.ts:15-22` is the authoritative 6→2 projection of `source_field` onto **family vs staff**: `bunk_request_form` and `socialize_with` → family; `staff_not_bunk_with`, `bunking_notes`, `internal_notes`, `manual` → staff.

Those names are the summer bunking form's own columns (`pocketbase/sync/bunk_requests.go:26-32` — *Share Bunk With*, *Do Not Share Bunk With*, *Internal Bunk Notes*, *BunkingNotes Notes*, *RetParent-Socializewithbest*), i.e. exactly the fields this request names. Summer badges the distinction on the camper panel (`CamperDetailsPanel.tsx:936`, `isStaffBadge`). Weekend has no equivalent, so per CLAUDE.md §4 the weekend treatment should be summer's badge grammar rather than a new one.

⚠️ **But the two surfaces are not fed the same way, and that matters here.** Summer's per-source rows arrive from the CSV into `original_bunk_requests` with `source_field` preserved per row. The weekend's arrive as CampMinder custom fields joined into one string — and **a weekend family's bunking-form notes are not on the weekend panel at all.** That gap is **#2256** (*show the "BunkingNotes Notes" field in the family details panel*), which has to land first or be folded in: "denote the source" of a field the panel does not yet show is not implementable.

## Sequencing

1. **Render `eligibility_source` and `answers_conflict`** in `ShareRequestPanel`. No backend change. ⚠️ **#2250 is a live bug on the disagreement signal** — the "Answers disagree" chip hides its explanation in `sr-only` text. Fix it or coordinate with it; do not add a second disagreement affordance beside it.
2. **#2256** — get the bunking-form note onto the panel at all.
3. **Split `request_text`** into its three parts, or carry a parallel per-source structure, so the verbatim text can say which form it came from. Go + API + types.
4. **Only then** decide the board-level treatment. Deferred by the reporter on purpose — do not invent it in a PR.

## Anchors

- `frontend/src/components/weekend/ShareRequestPanel.tsx:17`, `:47-56` — the "never split" contract and what is drawn
- `pocketbase/sync/lodging_fields.go:150-155` — the three fields joined into `request_text`
- `api/services/lodging_roster_service.py:1729` — the API passthrough
- `frontend/src/utils/sourceFromField.ts:15-22` — summer's family/staff projection
- `pocketbase/sync/bunk_requests.go:26-32` — the five bunking-form columns this request names
- `frontend/src/components/CamperDetailsPanel.tsx:936` — summer's source badge in use
- `docs/reference/family-camp-field-provenance.md` — provenance of the two share questions

*Reported via kindred-feedback#114 — the private original, which holds the reporter, the page it was filed from, and the app version.*

adamflagg:
## The defect

`family_camp_registrations.request_text` collapses several distinct free-text
source fields into one string with a `'; '` join and keeps no field boundary.
Staff opening a family's modal could see the sentences but not which form — or
which child — produced any of them. Both dimensions were destroyed by the same
join, and neither is recoverable downstream: **10 of 422 non-blank 2026 request
values contain `'; '` themselves**, so no client-side split was ever possible.

Worse, one whole lane never arrived at all. Three of the five ruled fields come
from the summer bunking CSV into `original_bunk_requests`, and the roster read
only `family_camp_registrations` — so **32 households rostered into a 2026
family session had request text that appeared on no weekend surface**.

## What ships

`ShareRequestSummary` carries a new `request_blocks` alongside the unchanged
`request_text`: one block per source field, one entry per distinct answer,
every contributing child named.

`ShareRequestPanel` renders them per the 2026-08-17 owner ruling — a hybrid of
mockup options A and B:

- **Blocks start expanded.** Nothing is hidden behind a click on first render,
  because a scanning eye must not miss request text.
- **Every block is collapsible.** The affordance is a *collapse* control, not a
  default-closed state, so a staff member facing the heaviest household can
  fold blocks away.
- **Both splits render when expanded** — which source field an answer came
  from, and which child said it. Recovering only one would leave the other
  still lost.
- **Labels are the original CampMinder field names, verbatim, unless the owner
  named one.** The rule was ruled deliberately — *"call them the original
  fieldnames for now until staff can weigh in after it's live"* — and staff
  have since weighed in, over two passes of the owner's 2026-08-17 review of
  the live panel. Three of the six source fields now **display** under a
  different name, through a `DISPLAY_LABELS` map keyed on the source field:

  | source field (identity) | displays as | why |
  |---|---|---|
  | `BunkingNotes Notes` | **Bunking Notes** | reads as a typo |
  | `COVID-19 Bunking Requests` | **Reg Form Bunk Notes** | misnamed at source — 205 rostered households of ordinary bunking requests, nothing to do with COVID |
  | `Shared-request` | **Fam Info Form Bunk Notes** | named after the form staff meet it on |

  The other three — `Internal Bunk Notes`, `Share Bunk With` and
  `FAM CAMP-Share Comments` — keep the raw CampMinder spelling, because the
  owner did not name them and **silence is not a rename**. That is the rule the
  next reader gets, in the map's own docblock: *verbatim unless the owner named
  it*.

  ⚠️ `Shared-request` (cm_id 274133) is the **current** 2026 family-camp form
  field, 112 rostered households. It is not its lookalike
  `FAM CAMP-Share Comments` (cm_id 240598), which has 171 values in 2025, 112
  in 2024 and **zero** in 2026 — renaming that one instead would have left the
  field staff actually read still called `Shared-request`.

  **Only the display moves.** The source-field identity that
  `REQUEST_TEXT_SOURCES`, the `original_bunk_requests.field` slug, the Go
  ingest map, the collapse key and `data-source-field` all key on is untouched
  for all three, and a test asserts exactly that on the attribute the rest of
  the system reads. No label outside that map was "improved".
- **Every block renders on the same amber rail**, family-authored or
  staff-authored. The two staff fields briefly shipped on a grey one so an
  internal note could not read as a family's own ask; the owner's 2026-08-17
  review reversed that and standardised on amber, which also removes a
  weekend-only divergence from summer's camper panel. **The authorship
  distinction itself is retained** — it is what gates the two staff blocks
  server-side (below), so it still travels on the wire and still renders as
  `data-authorship`; it simply is not painted a colour for now.
- **The amber blockquote is inherited, not reinvented** — the same treatment
  `CamperDetailsPanel` uses for parent request text, so a request reads the
  same wherever staff meet one. `break-words` is the one addition: the longest
  single 2026 answer is 680 characters in a 416px panel, and an unbroken token
  would otherwise force a horizontal scroll.
- **A source field with no text renders nothing at all** — no "nothing
  applicable" clutter, composing with #2255's chip ruling in this same modal.

`request_text` stays on the wire (`HouseholdRosterTable` still reads it) and is
the panel's fallback if blocks ever come back empty against a non-empty join:
losing a family's ask would be worse than losing its provenance.
`request_text` and the FAMILY-authored blocks stay **ungated**;
`MedicalNarrative` stays **gated** — not converged. The two staff-authored
blocks are gated too, for the reason below.

## ⛔ The sixth candidate field is EXCLUDED, deliberately

`Do Not Share Bunk With` (`staff_not_bunk_with`, **3** rostered 2026
households) travels this same code path and the mockup offered it, but the
2026-08-17 ruling did not name it. **Treating silence as a yes would be a
guess**, so it is excluded — at the read, not just at the render, so it cannot
appear by accident. Adding it later is one row in `REQUEST_TEXT_SOURCES` plus
one entry in `BUNKING_CSV_REQUEST_TEXT_FIELDS`.

`RetParent-Socializewithbest` (`socialize_with`, 107 rostered households) is
excluded for a different reason: it is **not free text** — two distinct values
in 2026, both 40 characters — and `requestBucket.ts` already classes it
immaterial.

`FAM CAMP-Share Comments` **is** carried, though it is 0 households for 2026:
it is one of the three fields the Go ingest already joins into `request_text`
(live 2024-2025), so leaving it out would have made the split lose text the
joined column shows today.

## Re-verified against production

Every figure below is from `pocketbase/pb_data/data-prod.db`, opened by name.
Denominator is the **382** households rostered into one of 2026's eight family
sessions (`status_id = 2`).

| source field | lane | author | households |
|---|---|---|---|
| `COVID-19 Bunking Requests` | family-camp form | parent | 205 |
| `Share Bunk With` | bunking CSV | parent | 104 |
| `Shared-request` | family-camp form | parent | 100 |
| `BunkingNotes Notes` | bunking CSV | **staff** | 28 |
| `Internal Bunk Notes` | bunking CSV | **staff** | 8 |
| `FAM CAMP-Share Comments` | family-camp form | parent | 0 |
| ~~`Do Not Share Bunk With`~~ | bunking CSV | **staff** | 3 — excluded |
| ~~`RetParent-Socializewithbest`~~ | bunking CSV | parent | 107 — not free text |

**270 of 382 (71%)** carry at least one signal; **128 of 270 (47%)** carry text
in two or more distinct source fields — the real consolidation case.

The issue's "blocks per household" table counts **entries** (distinct
`(field, text)` pairs), and it reproduces exactly: 116 / 87 / 38 / 22 / 6 / 1
at 1-7, so **29 households render four or more entries**. Counted by **source
field**, which is what a block is here, the distribution is 142 / 90 / 29 / 9 —
so the ceiling is **four blocks**, and the seven-entry household spreads those
seven across them. Longest single answer **680** characters; heaviest household
**1,109** characters total; median answer **44**. Panel is **416px**
(`FamilyDetailsPanel.tsx`, `w-[26rem]`).

Sibling divergence is real and is why entries are not collapsed to one winner:
**48 of 131** `(household, field)` sibling groups hold exactly duplicate text
(one parent's answer copied onto each child's record — those collapse to one
entry naming both), and the other **83 genuinely disagree**.

## Where the split happens, and why not in Go

The issue sanctions "Go/API". This does it in the **API**, and deliberately
does not touch `pocketbase/sync/lodging_requests.go`:

- Three of the five fields live in `original_bunk_requests`, which the
  family-camp derived sync does not read. Reaching them from Go would mean a
  new cross-sync dependency, a new JSON column, and a migration.
- The Go collapse's `isRequestField` gate is load-bearing for
  `request_last_updated`; widening it to admit CSV fields would change a
  derived column nobody asked to change.
- Nothing is re-derived in Python. The gate, the NEAR/WITH modes and
  `request_text` are still read from their derived columns and still never
  re-parsed — the new read returns free text unjoined and normalises nothing.

Two `@cached_by_year` reads were added (422 + 1,262 rows on 2026 — one page
for the family-camp lane, two for the bunking-CSV lane, which is not narrowed
to family-camp people and so comes back over a PAGE_SIZE of 1,000), absent
from `build_summary`'s TaskGroup for the same reason last year's cabins are:
no `WeekendSummaryEntry` carries a party.

Both PocketBase filters were verified against a live server before pushing —
the nested `field_definition.cm_id = …` relation filter returns 459 rows and
both expands resolve, so the silent-zero-rows trap does not apply.

## ⛔ The two staff-authored blocks need `bunking.manage`

The family-authored blocks and `request_text` are **ungated**, unchanged. The
two staff-authored ones are **not**, and that is a deliberate narrowing added
during review rather than an extra gate invented for its own sake:
`BunkingNotes Notes` and `Internal Bunk Notes` are `original_bunk_requests`
rows, and that table's own PocketBase listRule is `bunking.manage` while every
other API route serving its raw `content` (`api/routers/debug.py`) requires an
admin. `/api/lodging/roster` takes any authenticated user, so emitting them
unconditionally would have put internal staff commentary about named families
in front of **8 of the 13 production users** who can read it nowhere else —
against what `api/routers/lodging.py` already states ("a thing here is either
behind `bunking.manage` or it is not … So are internal notes").

`build_roster` takes `include_staff_notes`, **defaulting to withholding**, so
a caller that forgets it shows less than it could and never more than it may.
The router resolves it from the caller's permission (`_may_read_staff_notes`);
admin passes, exactly as `require_permission` lets it through. Nothing is
hidden client-side — a block that arrives is a block staff may read.

**This gate is the reason `authorship` survives the amber standardisation.**
Now that nothing paints with it, the field looks unused from the panel alone;
deleting it would hand staff notes back to the 8 of 13 production users without
`bunking.manage`. The rail comment in `ShareRequestPanel.tsx` says so, so the
next reader does not re-derive it.

## Tests

59 new tests, written failing first and mutation-checked where any went green
on first write.

- `tests/unit/api/services/test_lodging_rules.py` — the registry: the ruled
  five plus dormant `FAM CAMP-Share Comments`, family-before-staff order,
  every storage key maps to exactly one label, both exclusions, and an
  unregistered label sorting last and reading `staff` rather than `family`.
- `tests/unit/api/services/test_lodging_repository.py` — both lanes' filters
  and expands, neither excluded field ever requested, blank answers and
  household-less persons dropped, cached by year.
- `tests/unit/api/services/test_lodging_roster_service.py` — per-field split,
  per-child sub-labels, identical sibling text collapsing to one entry naming
  both, staff authorship, no blocks for a silent household, no cross-household
  leak, the excluded field dropped even if a row arrives anyway, and the lander
  not paying for the read.
- `frontend/src/components/weekend/ShareRequestPanel.test.tsx` — the existing
  15 tests are preserved (two rationale comments were corrected where the
  ruling superseded them); 23 added for the split, expanded-by-default,
  collapse and reopen, a fold NOT outliving the household it was made on, one
  amber rail for both lanes, the three relabelled fields displaying as
  `Bunking Notes` / `Reg Form Bunk Notes` / `Fam Info Form Bunk Notes` while
  every `data-source-field` identity stays raw, `Internal Bunk Notes`,
  `Share Bunk With` and `FAM CAMP-Share Comments` staying verbatim, empty rendering
  nothing, the 1,100-char answer wrapping inside 416px, four blocks with seven
  entries, and the fallback.
- `tests/unit/api/routers/test_lodging_endpoints.py` — the staff-note gate
  reaching the service, per caller: nothing, `bunking.manage`, admin.
- `frontend/src/components/weekend/HouseholdRosterTable.test.tsx` — the
  Requests column survives a household whose only text is in a block.

Gates: `bash scripts/pre-push-verify.sh` — ALL CHECKS PASSED (prettier,
eslint, tsc, vitest) — plus `uv run pytest tests/`: **6,376 passed, 23
skipped**. The permission tests were not touched and still pass.

## 🖐 This is a GUI change and HOLDS for visual review

Per the standing rule, anything with visual impact waits for the owner's eyes
however confident the build is. Do not merge on green CI.

Worktree: `.worktrees/phase-e-share-provenance` · branch
`feature/phase-e-share-provenance`.

Closes #2330.


adamflagg:
## Independent review — 6 findings, all fixed in-branch (`f6af99a5`)

I did not write this code. Reviewed at score floor 60, against the prod snapshot
(`pocketbase/pb_data/data-prod.db`, opened by name), the PocketBase migrations,
the Go ingest, and by mutation-checking the tests. **Every number in the body
reproduced exactly** — 382 rostered households, 205/104/100/28/8/0 per field,
32 CSV-only, 270 with a signal, 422 raw values with 10 containing `'; '`, the
three cm_ids against `lodging_fields.go`, the three slugs against `csvFieldMap`.
Key spaces are right: `persons.household` is a relation, so the blocks key on
the same household PB id `fetch_family_camp_registrations` does.

### 🔴 1. The two staff-authored blocks were served to any authenticated user

`BunkingNotes Notes` and `Internal Bunk Notes` are `original_bunk_requests`
rows. That table's own PocketBase listRule is `bunking.manage`
(`1500000020_original_bunk_requests.js`), and every other API route serving its
raw `content` requires an admin (`api/routers/debug.py:488`, `:528`).
`/api/lodging/roster` takes only `get_current_user`, and `/weekend/*` carries no
route guard — so this shipped internal staff commentary about named families to
**8 of the 13 users in the prod snapshot** (1 admin + 4 with `bunking.manage`;
8 without). `api/routers/lodging.py:166` already says the opposite in as many
words: *"a thing here is either behind `bunking.manage` or it is not … **So are
internal notes**, which that change left exactly where they were."*

#2330's ruling names the five fields and how they must *look*; it says nothing
about who may read them, so I took the conservative direction — nothing becomes
more visible than it was.

**Fixed:** `build_roster(..., *, include_staff_notes: bool = False)` threaded to
`_request_blocks`, resolved at the router from the caller's permission
(`_may_read_staff_notes`, admin passes). Default withholds. Family-authored
blocks and `request_text` unchanged and ungated. Gating is in the service, not
the repository, because that read is cached per year and shared across callers.

> **This one is a product call and it is reversible in one line** — if you want
> every authenticated user to see internal notes on this panel, flip the router
> argument to `True`. Worth a look during the visual review you are already
> holding for.

### 🔴 2. A collapsed block carried its fold to the next household

The panel is **never remounted** between families: all three callsites render
`<FamilyDetailsPanel party={panelParty} …>` with no `key`, and `usePanelParty`
only swaps `selectedKey`. So `RequestBlock`'s `useState(true)`, keyed on the
source field, kept its instance *and its fold* across the click. 205 of 382
households share `COVID-19 Bunking Requests` — fold it on one family and the
next family's request text arrives already hidden, which is precisely what the
"blocks start expanded, because a scanning eye must not miss request text"
ruling exists to prevent. The comment above the `useState` claimed the opposite
("Deliberately NOT remembered across households").

Reproduced before fixing:

```
× reopens a folded source field when the next household is shown
  → Unable to find an element with the text: The second family's ask
```

**Fixed:** the fold moved to `ShareRequestPanel` as a `Set<string>`, reset during
render when `share` changes identity — React's own "adjust state on a changed
prop" pattern, already used by `usePanelParty` and `WeekendRosterPage`. A second
test pins that a plain re-render of the *same* household does **not** unfold.

### 🟠 3. `needs_resolution` ignored the lane the PR exists to surface

`needs_resolution=bool(request_text)` was right while `request_text` *was* the
text. The 32 CSV-only households have a blank `request_text` and a real request
in `request_blocks`, so their ask rendered with no marker beside it. Same cause,
same 32 households: `HouseholdRosterTable.hasAnyRequest` tested only
`request_text`, so a roster made entirely of them would drop the Requests column.
**Fixed both**, plus the schema comment that asserted the old rule.

### 🟠 4. "Both lanes are one page each" is false

`PAGE_SIZE = 1000`; the bunking-CSV lane returns **1,262** rows on 2026 — two
pages. It is not narrowed to family-camp people at all, so 1,086 of those rows
belong to households with no weekend party. **Fixed** in the docstring and the
body, with a note that narrowing needs the person-id OR clause #1966 measured
returning HTTP 400, which is why the cache is the lever.

### 🟠 5. Two body claims the diff did not support

- **"51 new tests"** — the diff added **45** (9 rules + 9 repository + 9 roster
  service + 18 panel). Body now says 56 and matches after this commit.
- **`Do Not Share Bunk With` … | parent |** — the repo's authoritative 6→2
  mapping classes `staff_not_bunk_with` as **staff**
  (`bunking/sync/bunk_request_processor/core/models.py:48`,
  `frontend/src/utils/sourceFromField.ts:18`). Corrected in the table; the
  exclusion itself is right and unchanged.

### 🟡 6. Two-lane tests depended on thread scheduling

Both lanes run through `asyncio.gather` over `asyncio.to_thread` and hit the
same `pb.collection.return_value.get_full_list`, so `side_effect=[family, csv]`
handed rows to whichever worker dequeued first. It held over 15 consecutive
local runs — but nothing guarantees it, and a loaded CI runner is where it would
stop. **Fixed:** a `_lanes` helper that answers by the filter, never by order.

---

### Refuted, not filed

- **Summer bunk requests leaking onto the weekend panel.** `original_bunk_requests`
  is fed from the summer-scoped CampMinder report, so I checked whether summer
  cabin asks now render as family-camp lodging requests. Of the 176 carried rows
  reaching a rostered family household, **2** come from a requester not enrolled
  in a 2026 family session (both `main`), and 112 come from a child enrolled in
  both — one answer for the year, given in a summer context. Real, but **ruled**:
  #2330 §4 carries the coverage caveat explicitly and the owner ruled *"we can
  widen the report itself … and see what data we have after."* Not a defect.
- **`request_text_authorship` defaulting `staff` for an unregistered label is
  unreachable**, since `_request_blocks` drops unregistered fields first. It is
  defensive depth in the right direction; left alone.
- **`api/schemas` importing `api/services`** inverts the usual layering. No test
  or convention forbids it, `lodging_rules` is pure and imports nothing back, and
  the alternative is a second declaration of the family/staff split that can
  drift. The comment already justifies it. Left alone.

### Nothing filed

Every finding was in code this PR wrote or one file away from it, so all six are
fixed here rather than deferred.

### Mutation checks

| test | mutation | result |
|---|---|---|
| `test_identical_text_from_two_children_collapses_to_one_entry_naming_both` | `entries.setdefault(text.casefold(…))` → `setdefault(text, …)` | 🔴 `assert 2 == 1` |
| `renders the two staff-authored fields on a grey rail` | `isStaff ? STAFF_RAIL : FAMILY_RAIL` → `FAMILY_RAIL` | 🔴 fails |
| `test_a_caller_holding_nothing_does_not_get_them` (new) | router arg → hardcoded `True` | 🔴 `assert True is False` |
| `reopens a folded source field when the next household is shown` (new) | pre-fix implementation | 🔴 fails |

Gates: `bash scripts/pre-push-verify.sh` — ALL CHECKS PASSED (6,349 pytest,
ruff, mypy, api-types freshness, prettier, eslint, tsc, vitest). Push verified by
the remote ref (`git rev-list --count origin/…..HEAD` = 0). **Not merged, not
tagged for CodeRabbit** — the GUI hold in the body still stands, and finding 1 is
worth your eyes.

adamflagg:
**Owner ruling, 2026-08-17, after seeing the panel live** — recorded here so the change is auditable. Both tweaks are presentational; nothing about *which* fields are shown, *who* can see them, or how they are grouped moved. Shipped in 37f2ddf.

1. **One label gets a friendly name.** `BunkingNotes Notes` now *displays* as `Bunking Notes`, via a `DISPLAY_LABELS` map keyed on the source field. Every other label stays verbatim — `Internal Bunk Notes` already reads correctly, and `COVID-19 Bunking Requests` keeps its misnomer under the original "original fieldnames until staff weigh in" ruling. The owner weighed in on exactly one of them.

   The **source-field identity is untouched**: `REQUEST_TEXT_SOURCES`, the `original_bunk_requests.field` slug (`bunking_notes`), the Go ingest map and the `data-source-field` attribute all still carry `BunkingNotes Notes`. A test pins that the display and the identity diverge, and it was mutation-checked.

2. **Both staff blocks render amber, not grey.** The grey rail that distinguished staff-authored text from family-authored text is dropped; every block now gets the inherited amber blockquote.

3. **`authorship` is retained.** It is load-bearing for something the colour was never the point of: `include_staff_notes` / `_may_read_staff_notes` uses it to gate the two staff fields behind `bunking.manage`, because their source table is admin-only. Removing it would re-open the permission hole a review pass on this PR closed — 8 of 13 production users would regain staff notes. It still travels on the wire, still renders as `data-authorship`, and the rail comment in `ShareRequestPanel.tsx` says why, so the next reader does not delete an "unused" field.

Every permission test passed **untouched**: `uv run pytest tests/` → 6,376 passed, 23 skipped. `bash scripts/pre-push-verify.sh` → ALL CHECKS PASSED. PR body updated where it described the grey rail or the verbatim-label rule.

🖐 Still holds for visual review — do not merge on green CI.

PR #2424, opened 2026-08-17
feat(pb,api): move write-in occupancy to its own table and switch the read (#2382, 2 of 4)

adamflagg:
Second of four on kindred#2382. **This PR is not the one that finishes that issue** — PR 4 is, and it carries the closing reference. Nothing here does.

Stacked on `feature/phase-e-writein-table` (#2422, "1 of 4"), which created `lodging_write_ins` and `lodging_write_ins_draft` empty and dark. This one **moves the rows and switches the read**, at behavioural parity.

## The model, per the owner's 2026-08-15 clarification

`lodging_availability.family_available` answered two unrelated questions through one boolean:

| value | fact | scenario-scoped? | where it lives after this PR |
|---|---|---|---|
| `true` | staff↔family **ROLE** — a staff cabin opened to families for the weekend | **No** — an operational fact, *"we're moving staff to X for weekend Y"* | `lodging_availability`, unchanged |
| `false` | **OCCUPANCY** — somebody is in it. The write-in | **Yes** — not every write-in is non-rostered staff; some are paper registrations for families arriving with no children, which is a modelling choice | `lodging_write_ins` |
| no row | the unit's own role decides | — | — |

## Measured, and re-verified against `pocketbase/pb_data/data-prod.db`

```
SELECT family_available, count(*) FROM lodging_availability GROUP BY family_available;
0|21
```

**All 21 rows are `family_available = 0` — every one an OCCUPANCY, none a ROLE.** So `1500000162` moves all 21 and leaves `lodging_availability` **empty**. Stated plainly rather than described as a partition: the role half is live capability with no current rows, not a preserved population. There is no risky partition and no ambiguous row.

## What shipped

**Migration `1500000162_lodging_write_ins_backfill.js`** — moves the `family_available = 0` rows, carrying `unit`, `session`, `session_cm_id`, `year`, `occupant_name`, `note`, plus the row's `id` and its `created`/`updated`, so a moved row keeps the day staff actually recorded it. Two guarded statements in a deliberate order: INSERT what is not already there (keyed on the destination's unique index), then DELETE only what a copy provably exists for. Nothing is removed before it is proven saved. Down is the exact inverse.

**Read** — `build_roster` and `build_summary` each fetch `lodging_write_ins` (two separate TaskGroups, both wired; a guard test pins each). `_build_units` now takes two sources and resolves ROLE from availability, OCCUPANCY from write-ins, with occupancy outranking role for the race case no writer can produce. `write_in_covers` / `_resolve_write_in_covers` take the occupancy unit-ids as an **input** instead of re-deriving them from `family_available_override is False`, so a bare `false` on a role row — which names nobody — can no longer masquerade as somebody sleeping in a cabin.

**Write** — `PUT /api/lodging/availability` keeps its URL, its request model and everything a staff member sees, and routes behind them: `false` → `lodging_write_ins`, `true` → `lodging_availability`, `null` → deletes from both. Each write drops the other fact, because a single row used to do that for free and two tables have to keep the promise deliberately. The **new fact is written before the old one is dropped**: there is no transaction across two PocketBase tables, and write-then-drop leaves a window where both rows exist, which the read resolves as "somebody is in it" — the safe half. Drop-then-write would leave a window with neither, opening a cabin nobody meant to open.

## Behavioural parity is the acceptance criterion

`family_available_override` **still spells an occupancy as `False`**, deliberately. Every count on the stats bar (`units_family_available`, `units_reserved`, `beds_family_available`) is derived from it through `is_family_available`, and the board's forest open-tint reads `false` as a proxy for "is somebody in it". A write-in that stopped spelling itself there would silently return an occupied cabin to the open count. Disentangling the two axes **on the wire** is PR 4.

Because of that, **the frontend needs no behavioural change**: `unit.write_in` is still populated by the server, so `coveringWriteIn`, `writeInOccupant`, `writeInSource`, `reservationBadge`, `availabilityAction` and the drop-disable all keep working unmodified. The frontend diff here is table-name corrections in comments that my own change made false, nothing else.

## Deliberately NOT in this PR

- **No scenario dimension.** `lodging_write_ins_draft` stays dark — nothing reads or writes it. `fetch_draft_write_ins` is untouched and `copy_from_mirror` / `copy_scenario_to_scenario` are not opened. Two tests pin it (`test_no_scenario_dimension_is_read_yet`, and the same assertion on the lander). That is **PR 3**.
- **No re-pointing of #2093's forest open-tint proxy** and no edits to the premise comments in `writeIn.ts:1-25`, `unitBadges.ts:302`, `lodging_repository.py:18-22`. That is **PR 4**.
- **No widening of the single-cover arity** to N write-ins per container — still one row per (unit, weekend). That is kindred#2381, which does not dissolve into this.
- **`frontend/src/utils/boardLayout.ts` untouched** (in-flight PR #2404).

## ⚠️ File overlap

PR #2423 (kindred#2330, `feature/phase-e-share-provenance`) is open and also edits `api/schemas/lodging.py` and `api/services/lodging_roster_service.py`, in different regions (request-text provenance, not write-ins). This branch is not based on it and edits are scoped tightly to the write-in regions. **Whichever of the two merges second must rebase.**

## ⚠️ CI does not run on this branch — confirmed, not assumed

`.github/workflows/ci.yml` triggers on `pull_request: branches: [main]`. This PR's base is `feature/phase-e-writein-table`, so **no CI Summary runs and this branch is unverified by CI**. The only checks present are "Validate PR title", a skipped `auto-merge`, and a CodeRabbit check that is green because it says *"Review skipped: automatic reviews are disabled"* — a no-op notice, not a review.

**Its only evidence is the local gate runs below.** Re-check once #2422 merges and this PR's base moves to `main`.

## Evidence

`bash scripts/pre-push-verify.sh` — **ALL CHECKS PASSED**: ruff format, ruff check, mypy, pytest (unit), go build, golangci-lint, go test, prettier, eslint, tsc, vitest, migration headers, `options:{}` anti-pattern scan, pb-js-lint, shellcheck. Pre-push hook additionally green (`6366 passed, 41 skipped`, api-types freshness).

Targeted suites:

```
tests/unit/api/services/test_lodging_{roster,write}_service.py
tests/unit/api/services/test_lodging_repository.py
tests/unit/api/routers/test_lodging_endpoints.py     477 passed
frontend  src/components/weekend + services + types  1359 passed (54 files)
```

**The migration was verified against a real PocketBase boot**, not only against unit tests. `scripts/dev/verify-writein-table-backfill.sh` (new, following `verify-availability-writein-backfill.sh`'s pattern) boots three times — without the migration, with it, then with a byte-for-byte copy under a later filename — and asserts every column and timestamp travels, the ROLE release stays put, the draft twin is untouched, and nothing is lost between the tables. It builds the **half-applied** state (insert landed, delete did not) rather than a plain second run, because a plain second run was measured to be a weak test: the source rows are already gone, so deleting the `NOT EXISTS` guard still passes it.

```
verify-writein-table-backfill: OK (122 js migrations applied)
```

Mutation-checked, three ways: dropping the `family_available = 0` predicate → FAIL ("4 rows went in and 5 came out"); dropping the `NOT EXISTS` idempotency guard → exit 2, boot refused with `UNIQUE constraint failed: lodging_write_ins.session_cm_id, lodging_write_ins.year, lodging_write_ins.unit`; dropping the DELETE's `EXISTS` proof → FAIL (the release row deleted too). Restored → OK.

The read predicate was mutation-checked the same way: reverting `_is_written_in` to `family_available_override is False` turns 8 tests red, restoring turns 197 green.

And **against the real production snapshot** (a scratch copy of `data-prod.db`, never the original): `migrate up` → availability 0 / write-ins 21, all 21 occupant names and all 3 notes preserved; forget-and-re-apply → unchanged; `migrate down 1` → 21 rows back with `family_available = 0`, 21 names, 3 notes; `migrate up` again → moved again.

### DIFF
--- a/api/routers/lodging.py
+++ b/api/routers/lodging.py
@@ -265,16 +265,24 @@ async def set_availability(
     request: AvailabilityWriteRequest,
     user: AuthUser = Depends(require_permission(Permission.BUNKING_MANAGE)),
 ) -> LodgingWriteResponse:
-    """Reserve or release one unit for this weekend.
+    """Write somebody into one unit for this weekend, or release one to families.
 
-    Takes NO scenario, unlike every other write on this router. Availability is
-    a fact about the weekend rather than about the plan -- a burst pipe closes
-    a cabin in every scenario for that weekend -- so 1500000135 deleted the
-    dimension. Requiring one is what made this endpoint uncallable.
+    ONE ENDPOINT, TWO TABLES since kindred#2382. `family_available` answers two
+    unrelated questions and each is stored where it belongs: `false` is an
+    OCCUPANCY and goes to `lodging_write_ins`, `true` is a staff<->family ROLE
+    override for the weekend and stays in `lodging_availability`. The request
+    model, the URL and everything a staff member sees are unchanged -- the
+    split is behind them. `set_availability` carries the reasoning.
+
+    Takes NO scenario, unlike every other write on this router. The role half
+    is a fact about the weekend rather than about the plan, so 1500000135
+    deleted its dimension; the occupancy half writes the LIVE board, which is a
+    scope in its own right rather than the absence of one. Requiring a scenario
+    is what made this endpoint uncallable.
 
     `family_available: null` clears the override, which is spelled as the
-    ABSENCE of a row: there is no value meaning "normal", and writing one would
-    pin the unit against a later change to its role.
+    ABSENCE of a row -- in BOTH tables. There is no value meaning "normal", and
+    writing one would pin the unit against a later change to its role.
     """
     try:
         return await _writes().set_availability(request)
--- a/api/schemas/lodging.py
+++ b/api/schemas/lodging.py
@@ -158,9 +158,11 @@ class WriteInCover(BaseModel):
     Resolved on READ, never cascaded on write. A row per leaf would duplicate
     one fact across rows that then drift -- clear one room and the others still
     close the space -- and would strand orphans behind a re-merge. There is
-    still exactly one `lodging_availability` row; this says which units it
-    covers, and `unit_id` is the row it belongs to, so a card that inherited
-    the write-in can still clear it at the source rather than dead-ending.
+    still exactly one `lodging_write_ins` row (kindred#2382 moved occupancy off
+    `lodging_availability`, which now answers only the staff<->family role);
+    this says which units it covers, and `unit_id` is the row it belongs to, so
+    a card that inherited the write-in can still clear it at the source rather
+    than dead-ending.
     """
 
     # The unit the row actually names -- NOT necessarily the unit carrying this
@@ -249,9 +251,20 @@ class LodgingUnitSummary(BaseModel):
     # CampMinder has no sub-room concept for every building. Measured over
     # 2022-2025, resolving down instead would raise 36 false alarms.
     shareability: Shareability = "unknown"
-    # None when no lodging_availability row exists for this unit this weekend,
-    # i.e. the unit's ROLE decides. None and False are different answers and
-    # must not be flattened into one: False is "closed this weekend".
+    # None when NEITHER a `lodging_availability` row nor a `lodging_write_ins`
+    # row exists for this unit this weekend, i.e. the unit's ROLE decides. None
+    # and False are different answers and must not be flattened into one: False
+    # is "closed this weekend".
+    #
+    # TWO SOURCES since kindred#2382 split the boolean's two questions apart:
+    # True comes from the availability row (a staff cabin opened to families
+    # for the weekend), and False from a write-in row (somebody is in it),
+    # which outranks the role when a unit somehow carries both. It still spells
+    # an occupancy as False deliberately -- `is_family_available`, and through
+    # it every count on the stats bar, is derived from this one field, and the
+    # board's open-tint reads False as a proxy for "is somebody in it".
+    # Disentangling the two axes on the wire is PR 4 of kindred#2382; ask
+    # `write_in` "is somebody in this space" in the meantime.
     #
     # Stated explicitly rather than implied. The rejected encoding was a row
     # meaning "the opposite of this unit's current default", which an ordinary
@@ -260,17 +273,19 @@ class LodgingUnitSummary(BaseModel):
     # WHO is in the room. A hold IS a write-in (owner ruling, kindred#2078):
     # staff do not reserve an empty cabin, they record an occupant the system
     # does not know about -- most often non-rostered weekend staff. Read
-    # straight off the availability row's `occupant_name` column, which unlike
-    # `note`/`reason` below carries the SAME name on both sides, so there is
-    # no second translation to keep in step.
+    # straight off the write-in row's `occupant_name` column (kindred#2382;
+    # the availability row's is what a RELEASE carries, which is always
+    # blank), a column that unlike `note`/`reason` below carries the SAME name
+    # on both sides, so there is no second translation to keep in step.
     #
     # Display only, like `reason`. The rule never branches on it: what closes
     # a cabin is `family_available_override`, and an occupant with no override
     # is not a state any writer can produce.
     occupant_name: str = ""
-    # Display only. The rule never branches on it. Read from the availability
-    # row's `note` column -- see the migration header on why `note` was kept
-    # rather than renamed.
+    # Display only. The rule never branches on it. Read from the `note` column
+    # of whichever row supplied the decision -- the write-in row for an
+    # occupancy, the availability row for a release -- see 1500000118's header
+    # on why `note` was kept rather than renamed.
     #
     # OPTIONAL since kindred#2078, and empty on every historical row by
     # construction: 1500000148 moved each existing note into `occupant_name`
--- a/api/services/lodging_repository.py
+++ b/api/services/lodging_repository.py
@@ -383,9 +383,10 @@ class LodgingRepository:
         write-ins come from `fetch_draft_write_ins` and REPLACE these; there
         is no overlay.
 
-        NOTHING READS THIS YET. PR 1 of kindred#2382 lands the tables and this
-        CRUD dark; `write_in_covers` still resolves write-ins out of
-        `lodging_availability` until PR 2 switches it.
+        `build_roster` and `build_summary` both read this, and `_build_units`
+        is where a row becomes the occupancy half of a unit's answer. PR 2 of
+        kindred#2382 switched them over and 1500000162 moved the rows; the
+        DRAFT sibling below is still dark until PR 3.
         """
         return await self._page(
             LODGING_WRITE_INS,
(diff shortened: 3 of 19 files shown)


adamflagg:
## Independent review — floor 60

Reviewed `git diff origin/feature/phase-e-writein-table...HEAD` only. The parity claim held up under attack; the two real findings are both tests that this PR's own change silently made vacuous. Fixed in-branch in 2fbc33cb. **Nothing filed, nothing merged, no closing reference added.**

### Fixed in-branch

**1. Two endpoint race guards went vacuous, not red — `tests/unit/api/routers/test_lodging_endpoints.py`**

`set_availability` now performs **two** lookups before it writes (role row, then occupancy row). Three tests in `TestAvailabilityWrites` stage their PocketBase reads *positionally*, and only one of the three was re-staged for the extra lookup. In the other two, the row staged as the **recovery re-read** (`[[], [_rec(id="avail_raced")]]`) was consumed as `existing_write_in`, so `_upsert_row` took its plain update branch and the create that has to lose the unique-index race never ran.

Mutation transcript — the entire race-recovery block deleted from `_upsert_row`:

```
BEFORE the fix:  1 failed, 9 passed
  FAILED test_a_failed_re_read_after_a_lost_availability_race_keeps_its_status

AFTER the fix:   3 failed, 7 passed
  FAILED test_losing_the_availability_upsert_race_updates_instead_of_500ing
  FAILED test_a_failed_re_read_after_a_lost_availability_race_keeps_its_status
  FAILED test_a_failed_update_after_a_lost_availability_race_keeps_its_status
```

Both staged three-deep now, `create.assert_called_once()` added to each so a future re-shuffle of the lookups cannot silently skip the recovery path again, and the docstring re-pointed at `idx_lodging_write_in_unique` (the body sends `family_available: false`, so it is the write-in table's index that rejects the loser, not `idx_lodging_avail_unique`).

**2. The lander guard pinned the fetch but not the answer — `tests/unit/api/services/test_lodging_roster_service.py`**

`test_the_lander_reads_the_write_in_table_once_per_weekend` asserted only `fetch_write_ins.await_count == 2`. Mutation: replacing `write_ins_task.result()` with `[]` in `build_summary`'s `_build_units` call left **all 197 roster-service tests green** — the lander could fetch the table and throw the rows away, which is precisely the failure the comment at that fetch site names ("a weekend card that never read this table would report a written-into cabin as open beside a board that draws it closed"). The await *arguments* were unpinned too, so keying the read on the PocketBase id instead of `entry_cm_id` would also have passed.

Now: `await_args_list == [call(2026, 1000001), call(2026, 1000002)]`, plus a new `test_the_lander_counts_a_written_into_cabin_as_reserved` asserting the lander's counts against `build_roster`'s for the same weekend. Both mutations go red (197 passed → 1 failed, 197 passed each).

**3. `fetch_availability`'s docstring — `api/services/lodging_repository.py`**

Still opened "Staff reservations and releases for one session" after 1500000162 moved every reservation out. It is the one place a reader looks to learn what is left in that table, and it sat 120 lines above the new `fetch_write_ins` docstring that describes the split. Rewritten to say the table is now the role half only, and why 1500000135's reasoning is right for what remains.

### Verified, not taken on trust

- **Prod snapshot re-measured** against `pb_data/data-prod.db`: `lodging_availability` is 10 columns, all 10 carried by the migration; `SELECT family_available, count(*)` → `0|21`; 21 non-blank `occupant_name`, 3 non-blank `note`; 21 distinct `(session_cm_id, year, unit)`; zero rows with a blank `session`/`unit` or a zero key. Every number in the PR body checks out.
- **`scripts/dev/verify-writein-table-backfill.sh` run: `OK (122 js migrations applied)`.** Mutation-checked its idempotency claim myself — dropping the `NOT EXISTS` guard makes the re-run probe abort the boot with `UNIQUE constraint failed: lodging_write_ins.session_cm_id, lodging_write_ins.year, lodging_write_ins.unit`, exit 2. The guard is load-bearing and the script really catches its removal.
- **Key spaces.** Every join in 1500000162 is homogeneous — `unit` is a PocketBase id on both sides, `session_cm_id` a CampMinder id on both. No PB-id-against-CM-id join anywhere. On the read side `written_in_unit_ids` yields `_s(row, "unit")` (PB ids) and `_is_written_in` tests `unit.unit_id` (`_s(unit, "id")`, also a PB id) — same space. The verify script proves the positive case: rows actually move.
- **Grain.** `lodging_availability`, `lodging_write_ins`: `(unit, weekend)`, one row each, enforced by `idx_lodging_avail_unique` / `idx_lodging_write_in_unique`. `lodging_write_ins_draft`: `(unit, weekend, scenario)`. `fetch_write_ins(year, session_cm_id)` returns `(unit)` grain within one weekend, which is what makes `write_in_row_by_unit` safe as a plain dict. No count in the diff mixes grains.
- **`_is_written_in`** reverted to `family_available_override is False` → 8 failed, 189 passed; restored → 197 passed. Matches the PR body's claim exactly.
- **Occupancy-outranks-role** inverted → `test_a_write_in_outranks_a_release_on_the_same_unit` red. Properly pinned.
- **Scope leakage: none found.** `writeIn.ts:1-25`, `unitBadges.ts:302` and `lodging_repository.py:18-22` are untouched; `boardLayout.ts` untouched; `fetch_draft_write_ins` has no caller and both copy paths (`copy_from_mirror`, `copy_scenario_to_scenario`) are unopened — verified by grep, not by the body's say-so. `WriteInCover` arity unchanged.
- **Counts parity.** `units_family_available` / `units_reserved` / `beds_family_available` derive only from `is_family_available` ← `family_available_override`, which a write-in still spells `False`. `_resolve_write_in_covers` is still absent from `build_summary`, as it was before this PR, so the lander's inheritance behaviour has not moved either.
- `bash scripts/pre-push-verify.sh` → **ALL CHECKS PASSED**; pre-push hook on the push itself green (`6367 passed, 41 skipped`, api-types freshness).

### Refuted (not filed)

- **`record_id` can name a row that was not deleted.** In the clear branch, `record_id=write_in_id or role_id` with `deleted=write_in_deleted or role_deleted` — if the write-in delete 404s while the role delete succeeds, the response names the write-in and reports `deleted: true`. Cosmetic only: nothing in `lodgingApi.ts` or `lodgingCrud.ts` reads `record_id` off this endpoint (grepped).
- **`down` can strand a write-in.** If a role row already occupies `(session_cm_id, year, unit)`, `down`'s `NOT EXISTS` skips the insert and its `EXISTS … AND a.family_available = 0` skips the delete, so the row stays in `lodging_write_ins`. That is the non-destructive answer by construction, on a dev-only path, in a state no writer produces.
- **The frontend's `coveringWriteIn` fallback still synthesises a cover from `family_available_override === false`.** So "a bare `false` role row can no longer masquerade as an occupancy" holds at the API layer and for inheritance, but not for the row's own card. That is the documented older-server compat branch and re-pointing it is explicitly PR 4 — correct to leave alone here.
- **Migration-before-code deploy window.** If the API served the new read before PocketBase applied 1500000162, the 21 rows would draw closed with no cover. Generic to every migration in this repo, not something this PR introduced.

### Note on CI, confirming what the body already says

No CI Summary runs on this branch — base is `feature/phase-e-writein-table`, and `ci.yml` triggers on `pull_request: branches: [main]`. The only checks present are `Validate PR title` (pass), a skipped `auto-merge`, and a CodeRabbit check that is green because its body is *"Review skipped: automatic reviews are disabled"* — a no-op notice, not a review. Re-check once #2422 merges and this base moves to `main`. Local gates are the only evidence, and they are green above.

PR #2428, opened 2026-08-18
feat(api): give write-ins their scenario dimension (#2382, 3 of 4)

adamflagg:
> ## ⛔ MERGE ORDER: this PR must NOT reach `main` without PR 4
>
> **On its own this is a user-visible regression.** Reads now REPLACE, but the write path is still
> live-only (`set_availability` takes no scenario). So a staff member working inside a scenario who
> records a write-in has it land on the **live** board, and the scenario read then replaces it away —
> the write-in is not on the board they made it on. Before this PR the read fell through, so the same
> write was visible.
>
> Prior to PR 4, a scenario can only receive write-ins by copy at seed time. PR 4 closes it in three
> small steps (optional `scenario` on `AvailabilityWriteRequest`; route the OCCUPANCY half to the
> draft when it is non-blank, ROLE half always live; `LodgingBoard` passes its existing `scenario`
> prop through). Land them together, or land PR 4 first.

## What

**PR 3 of 4 for kindred issue #2382.** Stacked on #2424 (`feature/phase-e-writein-read`) — review that first.

PR 1 (#2422) created `lodging_write_ins` and `lodging_write_ins_draft` dark. PR 2 (#2424) moved the 21 rows out of `lodging_availability` and switched every reader onto the **live** table at behavioural parity, deliberately leaving the draft twin unread behind two guard tests. **This PR turns the twin on.**

Three things, exactly:

### 1. Replace-on-read

A request naming a scenario reads `fetch_draft_write_ins` and **does not read the live table at all**. This is kindred#1974's no-fall-through rule for `lodging_assignments_draft`, applied unchanged — not a new overlay semantic invented for this one table. A unit with no draft row holds **no** write-in in that scenario, whatever the live board says.

With no scenario, the live table is read exactly as before: the live board is a scope in its own right rather than the absence of one (owner, 2026-08-15 — "we need to allow write ins to happen in campminder prod, not just scenarios, for staff to properly evaluate the board").

Both `build_roster` and `build_summary` were wired — they hold parallel TaskGroups and fixing only one is the half-fix the guards in this file's tests exist to catch.

**The two PR-2 guards are replaced with the positive assertion of the new behaviour**, not merely deleted:

| PR 2 guard | Now |
|---|---|
| `test_no_scenario_dimension_is_read_yet` | `test_the_mirror_reads_the_live_table_and_never_the_draft` (the no-scenario side is still pinned) + the whole of `TestAScenariosWriteInsReplaceTheLiveOnes` |
| `fetch_draft_write_ins.await_count == 0` on the lander | kept for the **mirror** sweep, where it is now a real assertion, and mirrored by `test_the_lander_reads_each_weekends_draft_write_ins_in_a_scenario` |

### 2. Write-ins are copied in BOTH seed paths

Owner ruling, 2026-08-16. `copy_from_mirror` copies the **live board's** rows; `copy_scenario_to_scenario` copies the **source scenario's own**. Shaped after the `lodging_slot_merges` copy already sitting in `copy_scenario_to_scenario`: read the source tier, create the destination's rows, no emptiness check of its own, not counted into `LodgingCopyResponse.copied` (that number means placements).

The reason is carried into the code, because it is safety rather than convenience: without the copy a fresh scenario shows every written-into cabin as **open**, and kindred#2247's placement gate reads exactly that field — so it would let a family be dropped into a room the live board records as occupied. **The split creates that failure mode; the copy is what closes it.**

`copy_from_mirror`'s docstring said "Availability is NOT copied". That sentence had become half-false, and is rewritten: the staff↔family **ROLE** override is still not copied (it has no scenario dimension and nothing to disagree about), the **OCCUPANCY** half now is.

### 3. Scenario-delete cascade — found, not built

Checked both places the brief named:

- **`pocketbase/lodging/hooks.go`** — nothing. It registers these collections only for the year guards.
- **Python `delete_scenario`** (`api/routers/scenarios.py`) — nothing for lodging. It sweeps `bunk_assignments_draft` by hand for summer and touches no lodging table.

It is a **schema property**: `scenario` is declared `cascadeDelete: true` on `lodging_write_ins_draft` in migration `1500000161`, exactly as `lodging_assignments_draft` (1500000132) and `lodging_slot_merges` (1500000139) declare it. Adding a pre-delete loop would re-create the N+1 that `bunk_assignments_draft`'s own flip to `cascadeDelete` existed to remove.

So there is nothing to build — but PR 3 is where that guarantee becomes load-bearing (an orphan row holds a slot in `idx_lodging_write_in_draft_unique` that no board can see or clear), so it is now **pinned from both ends**:

- `TestLodgingWriteInDraftScenarioIsRequiredAndCascades` (package `main`) — the migration declares it, scoped to the draft's own field so the `unit` relation's `cascadeDelete: true` cannot satisfy it by accident.
- `TestDeletingAScenarioSweepsItsDraftWriteIns` + `TestDeletingAScenarioLeavesTheLiveWriteInsAlone` (package `lodging`) — the effect, against the running engine.

## What is deliberately NOT here

- **No migration.** `1500000161` already declares everything this needs.
- **No `frontend/src` file is touched.** No schema and no wire change, so nothing regenerates. `family_available_override` still spells an occupancy as `false` — the compat shim PR 2 kept on purpose stays.
- #2093's forest open-tint proxy, the premise comments at `writeIn.ts:1-25` / `unitBadges.ts:302` / `lodging_repository.py:18-22`, and removing the compat shim are **PR 4**.
- Widening the single-cover arity to N write-ins per container is kindred#2381, which does **not** dissolve into this.
- `frontend/src/utils/boardLayout.ts` untouched (in-flight PR #2404).

## ⚠️ Known gap PR 4 must close — the WRITE path is still live-only

`set_availability` takes no `scenario` and writes the live occupancy table. That was correct while reads fell through; it is **not** correct now that they replace. Concretely, after this PR:

> a staff member viewing scenario A records a write-in → it lands on the **live** board → scenario A's read replaces with draft rows → the write-in they just made is not on the board they made it on.

Nothing in this PR can fix that on its own, because the fix needs the frontend to send the scenario, and the brief for this PR scopes `frontend/src` out. **PR 4 owns it**, and it is small:

1. `AvailabilityWriteRequest` gains `scenario: str = Field(default="")` — blank means the live board, mirroring `SlotMergeRequest`'s optional scenario.
2. `set_availability` routes its **occupancy** half to `find/create/update/delete_draft_write_in` when the scenario is non-blank. The **role** half always stays on `lodging_availability` (owner ruling: it is not scenario-scoped), so a release written from inside a scenario is still a weekend fact.
3. `LodgingBoard` passes its existing `scenario` prop into `useUnitAvailability`, and `setUnitAvailability` puts it on the body.

Until (3) lands, `find_draft_write_in`, `update_draft_write_in` and `delete_draft_write_in` stay dark — the same one-PR lag `fetch_draft_write_ins` had. `create_draft_write_in` gains its caller here (the two seeds).

## Tests

TDD throughout: every assertion below was written first, run, and watched fail before the implementation existed. Everything that was green on first write was mutation-checked by breaking the implementation and reading the **exit code**, then restored.

**`tests/unit/api/services/test_lodging_roster_service.py`** — new `TestAScenariosWriteInsReplaceTheLiveOnes`:

- a scenario reads the draft table and not the live one (both directions asserted)
- a live write-in does **not** fall through into a scenario — the assertion a merge implementation fails
- a scenario's own write-in closes the unit, names its occupant and carries its note
- two scenarios write the same family into different cabins (the requirement, in one test)
- a scenario's write-in still covers the rooms beneath a building — catches wiring the fetch and leaving the cover walk on the live list
- the ROLE override is still read from the live table inside a scenario
- the lander reads each weekend's draft write-ins, arguments pinned not just the count
- the lander **counts** a scenario's write-in as reserved, pinned against `build_roster`'s own counts — the "the answer is used" half a call-count assertion cannot see

**`tests/unit/api/services/test_lodging_write_service.py`** — new `TestCopyFromMirrorAlsoCopiesWriteIns` and `TestCopyScenarioToScenarioAlsoCopiesWriteIns`: payload shape including `scenario`, every row copied not only the first, the right source per seed path (a mirror seed must not read the draft; a scenario seed must not read the live board), no rows means no creates, and a seed never writes the table it copied from. `test_availability_is_not_copied` is renamed and its stale docstring rewritten to say which half is still not copied and why.

**`tests/unit/api/routers/test_lodging_endpoints.py`** — the seed carries write-ins at the **endpoint**, not only in the service. `_mirror_reads` had to start keying on `expand: "units"`: one `mock_pb.collection.return_value` serves every collection, and `fetch_write_ins`' filter is byte-identical to `fetch_assignments`', so a filter-only match handed a placement row back as a write-in.

**Review pass (independent, post-authoring)** — one behavioural defect fixed in branch: `_seed_write_ins` let a `ClientResponseError` escape unconverted, so a colliding seed create reached api/main.py's catch-all as a **500** rather than the 400 its own docstring promised. Reachable, because the up-front guard counts *placements* only: a weekend whose mirror holds nothing copyable passes it on every attempt while the write-ins the first seed wrote are already there. The create now raises through `pb_error_to_http`, pinned by `test_a_failed_seed_create_keeps_its_upstream_status` on **both** seed paths (400 and 403 parametrised, so a refusal is not reported as a collision). No second emptiness guard was added — that decision stands as written.

The same pass retired the premise comments this PR falsified — `set_availability`'s "the draft twin stays dark until PR 3" (now carrying the write-path gap and the shape of PR 4's fix instead), `_build_units`' "a scenario reads the same live rows it does today", `written_in_unit_ids`' docstring, the `_repo` fixture's "every one of these has a caller", and three "stays dark until PR 3" test comments. `docs/architecture/lodging-occupancy.md` gained the same gap paragraph.

**Go** — the two cascade tests above, plus `TestLodgingWriteInLiveTableHasNoScenarioColumn` (the nullable-`scenario` sentinel this repository already turned down must not grow back onto the live table).

Mutation checks run and confirmed red: roster read reverted (6 red), summary read reverted (2 red), mirror seed copy removed (2 red + 1 endpoint red), scenario seed reading the live board (2 red), `scenario` dropped from the copy payload (3 red — the endpoint test goes red too), fixture `CascadeDelete` removed (1 red), migration `cascadeDelete: false` (1 red), migration `required: false` (1 red), `scenario` moved onto the shared `occupancyFields` (1 red).

Also corrected three stale counts PR 2 left in `lodging_roster_service.py` docstrings — `build_roster` makes thirteen fetches now, not twelve, and the lander makes five session-scoped reads per weekend, not four.

## CI

⚠️ **This PR gets no CI.** Workflows trigger on `pull_request: branches: [main]` and this targets `feature/phase-e-writein-read`. Local gates, run in this worktree:

```
$ bash scripts/pre-push-verify.sh
── Python ──      PASS ruff format   PASS ruff check   PASS mypy   PASS pytest (unit)
── Go ──          PASS go build   PASS golangci-lint   PASS go test
── Frontend ──    PASS prettier   PASS eslint   PASS tsc   PASS vitest
── Migrations ──  PASS migration headers   PASS no options:{} anti-pattern
── PocketBase JS ── PASS pb-js-lint
── Shell ──       PASS shellcheck
ALL CHECKS PASSED

$ uv run pytest tests/ -q -p no:randomly -n auto
6410 passed, 23 skipped in 141.30s

$ cd pocketbase && go test ./...
ok (all packages)

$ uv run mypy api/
Success: no issues found in 89 source files
```

The pre-push hook additionally ran `api-types-freshness` green — no generated type drift, as expected for a change with no schema or wire movement.

Part of kindred issue #2382 (3 of 4). That issue is answered by PR 4, not by this one.

### DIFF
--- a/api/services/lodging_repository.py
+++ b/api/services/lodging_repository.py
@@ -396,9 +396,10 @@ class LodgingRepository:
         is no overlay.
 
         `build_roster` and `build_summary` both read this, and `_build_units`
-        is where a row becomes the occupancy half of a unit's answer. PR 2 of
-        kindred#2382 switched them over and 1500000162 moved the rows; the
-        DRAFT sibling below is still dark until PR 3.
+        is where a row becomes the occupancy half of a unit's answer. Both pick
+        THIS read or `fetch_draft_write_ins` below on whether the request named
+        a scenario, exactly as they pick between `fetch_assignments` and
+        `fetch_draft_assignments` -- never both.
         """
         return await self._page(
             LODGING_WRITE_INS,
@@ -425,11 +426,13 @@ class LodgingRepository:
         the same kind of fact as a placement, so it follows the placement
         rule.
 
-        NOTHING READS THIS YET, unlike its live sibling above. PR 2 of
-        kindred#2382 moved the rows and switched the readers over to the LIVE
-        table at behavioural parity; giving write-ins their scenario dimension
-        -- this read, and the seed copies in both `copy_from_mirror` and
-        `copy_scenario_to_scenario` -- is PR 3.
+        `build_roster` and `build_summary` read THIS instead of
+        `fetch_write_ins` whenever the request names a scenario, and the live
+        table is then not read at all. `copy_from_mirror` and
+        `copy_scenario_to_scenario` both seed it, so a fresh scenario starts
+        with the write-ins its source had rather than blank -- without that,
+        kindred#2247's placement gate would let a family be dropped into a room
+        the live board records as occupied.
 
         `scenario_id` is client-supplied and escaped, for the reason
         `fetch_draft_assignments` spells out. The weekend is named by
--- a/api/services/lodging_roster_service.py
+++ b/api/services/lodging_roster_service.py
@@ -982,12 +982,30 @@ class LodgingRosterService:
         async with asyncio.TaskGroup() as tg:
             units_task = tg.create_task(self.repository.fetch_units(year))
             availability_task = tg.create_task(self.repository.fetch_availability(year, session_cm_id))
-            # THE OCCUPANCY HALF, split out of availability by kindred#2382.
-            # Read with NO scenario predicate, and that is not the old shape
-            # coming back: the live board is a scope in its own right, exactly
-            # as `lodging_assignments` is, and the draft twin behind this table
-            # stays dark until PR 3 gives write-ins their scenario dimension.
-            write_ins_task = tg.create_task(self.repository.fetch_write_ins(year, session_cm_id))
+            # THE OCCUPANCY HALF, split out of availability by kindred#2382,
+            # and chosen exactly the way the placement source below is chosen.
+            # A scenario reads its OWN write-ins and REPLACES the live ones --
+            # it does not read them at all -- matching kindred#1974's
+            # no-fall-through rule for `lodging_assignments_draft`. A unit with
+            # no draft row holds no write-in in that scenario, whatever the
+            # live board says.
+            #
+            # The live board is the other scope, not the absence of one: with
+            # no scenario this reads `lodging_write_ins` with no scenario
+            # predicate, exactly as `lodging_assignments` is read.
+            #
+            # A fresh scenario is seeded by an explicit COPY in both seed paths
+            # (`copy_from_mirror` and `copy_scenario_to_scenario`, owner ruling
+            # 2026-08-16), which is what stops it starting blank -- without it
+            # kindred#2247's placement gate would let a family be dropped into
+            # a room the live board records as occupied. Rendering the live
+            # rows through this read's gaps would be the overlay instead, and
+            # would make two scenarios unable to disagree.
+            write_ins_task = tg.create_task(
+                self.repository.fetch_draft_write_ins(year, session_cm_id, scenario)
+                if scenario
+                else self.repository.fetch_write_ins(year, session_cm_id)
+            )
             attendees_task = tg.create_task(self.repository.fetch_attendees_for_session(year, session_pb_id))
             households_task = tg.create_task(self.repository.fetch_households(year))
             prior_task = tg.create_task(self.repository.fetch_prior_household_cm_ids(year))
@@ -1086,7 +1104,7 @@ class LodgingRosterService:
     async def build_summary(self, year: int, scenario: str = "") -> WeekendSummaryResponse:
         """Every weekend in the year with its counts, in one pass.
 
-        `build_roster` makes TWELVE fetches (11 concurrent + `fetch_session`
+        `build_roster` makes THIRTEEN fetches (12 concurrent + `fetch_session`
         alone before them), of which SEVEN are year-scoped -- the unit
         registry, households, the prior-household set, family-camp adults,
         registrations, the unresolved-alias count and last year's cabins
@@ -1108,16 +1126,19 @@ class LodgingRosterService:
         in `_build_counts` from units already in hand rather than fetched.
 
         So the year-scoped work happens once here, and only the genuinely
-        session-scoped reads run per weekend: availability, attendees and one
-        placement read -- the synced rows, or the scenario's own. The
+        session-scoped reads run per weekend: availability, write-ins,
+        attendees and one placement read -- the synced rows, or the scenario's
+        own. The
         per-weekend numbers then come from the SAME `_build_units` /
         `_build_parties` / `_build_counts` helpers the roster uses, so the
         lander cannot drift from the page it links to, and it resolves a
         scenario the same way: replace, never fall through.
 
-        FOUR session-scoped reads per weekend, with or without a scenario:
-        availability, attendees, one placement source, and slot merges (the
-        last of these unconditional since 1500000140). A `Semaphore` below
+        FIVE session-scoped reads per weekend, with or without a scenario:
+        availability, one write-in source, attendees, one placement source, and
+        slot merges (the last of these unconditional since 1500000140). The two
+        "one source" reads are the scenario-aware pair -- the live table or the
+        scenario's draft, never both. A `Semaphore` below
         bounds how many weekends' worth of those run at once -- kindred#1920,
         which also records why a per-weekend cap was chosen over collapsing
         the placement read to one call for the whole year.
@@ -1167,14 +1188,19 @@ class LodgingRosterService:
             entry_cm_id = _i(session, "cm_id")
             async with entry_gate, asyncio.TaskGroup() as inner:
                 availability_task = inner.create_task(self.repository.fetch_availability(year, entry_cm_id))
-                # Exactly as build_roster reads it, and separately, because
-                # this is a DIFFERENT TaskGroup -- wiring one and leaving the
-                # other is the half-fix the guards in this file's tests exist
-                # to catch. The lander keeps only counts, and every count on it
-                # goes through `is_family_available`, so a weekend card that
-                # never read this table would report a written-into cabin as
-                # open beside a board that draws it closed.
-                write_ins_task = inner.create_task(self.repository.fetch_write_ins(year, entry_cm_id))
+                # Exactly as build_roster reads it, SCENARIO AND ALL, and
+                # separately, because this is a DIFFERENT TaskGroup -- wiring
+                # one and leaving the other is the half-fix the guards in this
+                # file's tests exist to catch. The lander keeps only counts,
+                # and every count on it goes through `is_family_available`, so
+                # a weekend card that read the wrong scope would report a
+                # written-into cabin as open beside a board that draws it
+                # closed.
+                write_ins_task = inner.create_task(
+                    self.repository.fetch_draft_write_ins(year, entry_cm_id, scenario)
+                    if scenario
+                    else self.repository.fetch_write_ins(year, entry_cm_id)
+                )
                 attendees_task = inner.create_task(self.repository.fetch_attendees_for_session(year, session_pb_id))
                 # One placement source, exactly as build_roster chooses it.
                 placements_task = inner.create_task(
--- a/api/services/lodging_roster_service.py
+++ b/api/services/lodging_roster_service.py
@@ -479,7 +479,12 @@ def resolve_combined(*, default: bool, override: bool | None, session_override: 
 
 
 def written_in_unit_ids(write_ins: list[Any]) -> frozenset[str]:
-    """The unit ids `lodging_write_ins` names, for `write_in_covers` to walk.
+    """The unit ids the chosen write-in source names, for `write_in_covers` to walk.
+
+    WHICHEVER source the request resolved to -- `lodging_write_ins` on the live
+    board, `lodging_write_ins_draft` inside a scenario (kindred#2382). The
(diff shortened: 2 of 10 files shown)


adamflagg:
## Independent review — PR 3 of kindred#2382

Reviewed `git diff origin/feature/phase-e-writein-read...HEAD` only. Attacked the four things the brief named: the overlay semantic, both copy paths, the cascade, and whether the new guards are vacuous. Everything below was run in this worktree.

### The overlay semantic — verified clean

Every reader of the occupancy half now picks one source and never merges:

| Consumer | Source |
|---|---|
| `build_roster` | `fetch_draft_write_ins(year, cm_id, scenario)` **or** `fetch_write_ins(year, cm_id)` |
| `build_summary` (`_entry`, its own TaskGroup) | same choice, same shape |
| `_resolve_write_in_covers` / `written_in_unit_ids` | fed the already-chosen list, not re-derived |
| `_build_units` | fed the already-chosen list |

`grep` over `api/` and `bunking/` finds no other caller of either read, so there is no path by which a live write-in can re-enter a scenario. The `if scenario else` shape is byte-for-byte the one `fetch_assignments` / `fetch_draft_assignments` already uses.

### Fixed in branch

**1. A colliding write-in seed create was a 500, not the 400 its docstring promised.** `_seed_write_ins` awaited `create_draft_write_in` with no handler. Nothing on this router catches `ClientResponseError`, so it reached `api/main.py`'s catch-all `Exception` handler — while the docstring said it "surfaces through `pb_error_to_http` the same way a colliding merge create does". It does not; neither does the merge create.

Reachable, not theoretical. The up-front guard is `count_draft_assignments`, which counts **placements only**. A weekend whose mirror holds nothing copyable — early season, before CampMinder has assigned any lodging — passes that guard on every attempt while the write-ins the first seed wrote are already sitting in the scenario, so a second `POST /api/lodging/placements/copy` collides on `idx_lodging_write_in_draft_unique`.

Probe, before the fix (temporary test, since reverted):

```
mock create -> ClientResponseError(status=400)
POST /api/lodging/placements/copy
E   pocketbase.errors.ClientResponseError: Message: duplicate      <- escaped the endpoint entirely
```

Fixed by raising through `pb_error_to_http`, which also keeps a refusal a refusal (401/403 → 403) instead of folding it into the collision. **No second emptiness guard was added** — the author's "one seed, one answer to *this scenario is already populated*" reasoning stands; only the missing conversion was the defect. Pinned by `test_a_failed_seed_create_keeps_its_upstream_status` on **both** seed paths.

**2–6. Premise comments this PR falsified, corrected in place** (the next reader implements from them):

| Where | Said | Now |
|---|---|---|
| `lodging_write_service.py` `set_availability` | "the draft twin behind it stays dark until PR 3" | carries the **write-path gap** and the shape of PR 4's fix |
| `lodging_roster_service.py` `_build_units` | "a scenario reads the same live rows it does today" | the two halves scope differently; occupancy is whichever table was chosen |
| `lodging_roster_service.py` `written_in_unit_ids` | "the unit ids `lodging_write_ins` names" | whichever source the request resolved to |
| `test_lodging_write_service.py` `_repo` | "every one of these has a caller" | `find_`/`update_`/`delete_draft_write_in` are still dark, and why |
| `test_lodging_repository.py` `TestWriteInReads` | "the DRAFT half is still dark — nothing reads or writes it" | both halves live; names the three that are not |
| `test_lodging_write_service.py:1735`, `test_lodging_endpoints.py:1309` | "stays dark until PR 3" | points at the gap and PR 4 |

The `set_availability` one mattered most: that docstring sits exactly where the hazard this PR creates lives, and it told the reader the opposite.

**7. `docs/architecture/lodging-occupancy.md`** stated the replace semantic as if complete. It now also records that the **write** path does not yet choose a scope — the durable artifact should not describe half of an invariant.

**8. PR body**: the mutation transcript said "`scenario` dropped from the copy payload (2 red)"; it is 3 — the endpoint test fails too. Corrected, since the body is the squash commit message.

### Mutation transcripts (break → confirm red → restore)

| # | Mutation | Result |
|---|---|---|
| M1 | `build_roster` reverted to always `fetch_write_ins` | **6 red** ✅ matches body |
| M2 | `build_summary` reverted to always `fetch_write_ins` | **2 red** ✅ |
| M3 | mirror seed's `_seed_write_ins` call removed | **3 red** (2 service + 1 endpoint) ✅ |
| M4 | scenario seed reads `fetch_write_ins` instead of the draft | **2 red** ✅ |
| M5 | `scenario` dropped from the copy payload | **3 red** (body said 2) |
| M6 | Go fixture: `CascadeDelete` removed from the draft's `scenario` | **red** — `the draft write-in outlived its scenario` ✅ |
| M7 | migration: draft `cascadeDelete: false` | **red** ✅ |
| M8 | migration: `scenario` pushed onto shared `occupancyFields()` | **red** — "declares 2 scenario relations, want exactly 1" ✅ |
| M10 | migration: draft `scenario` `required: false` | **red** ✅ |
| M9 | **my** new `pb_error_to_http` conversion removed | **3 red** ✅ |

The cascade test is the real thing, not a map-contains-a-key assertion: M6 proves `TestDeletingAScenarioSweepsItsDraftWriteIns` fails when the relation stops cascading, i.e. it exercises `app.Delete(scenario)` against the running engine and then looks for the row. `TestLodgingWriteInLiveTableHasNoScenarioColumn` is the weaker of the pair by construction (the live table has no such relation to lose) but is non-vacuous under M8.

### Refuted — investigated, not defects

- **Cross-weekend blindness in a scenario.** A scenario is seeded one weekend at a time, so `fetch_draft_write_ins` returns nothing for a weekend that was never seeded, and every written-into cabin there would read as open. Refuted on reachability: `WeekendRosterPage` routes through `scenarioForWeekend(currentScenario, selectedCmId)`, which resolves a scenario belonging to another weekend to the mirror, and `fetchWeekendSummary` never sends a scenario at all. Same exposure already exists for placements, where it is visible rather than silent.
- **Migration `1500000162`'s header** still says the draft "is still dark in this PR… PR 3, which starts filling it, owns its down path." Left alone: a migration header is a record of what that migration did, and PR 3 adds no migration, so there is no down path to own.
- **No second emptiness check on the write-in copy.** Considered and left as the author designed it — see above.
- **PR-4 / kindred#2381 scope leakage: none found.** No `frontend/src` file is touched (`git diff --name-only` is 9 files, all `api/`, `docs/`, `pocketbase/`, `tests/`). The `family_available_override` shim, the #2093 tint proxy, the three named premise comments and the N-write-in arity are all untouched.

### Filed

Nothing. Every finding was small enough to fix in branch.

### Gates

`bash scripts/pre-push-verify.sh` → **ALL CHECKS PASSED**. `uv run pytest tests/ -q -p no:randomly -n auto` → **6413 passed, 23 skipped** (6410 + the 3 added here). `go test ./...` → all packages ok. This PR still gets no CI, since it targets `feature/phase-e-writein-read`.

Not merged, not tagged for review. `closingIssuesReferences` verified empty.

adamflagg:
The write-path gap this PR's body flags — reads REPLACE but `set_availability` still writes the live table — is closed in **#2429** ("4 of 4"), which is stacked on this branch. #2429 also retires the `family_available_override === false` compat shim and closes #2382. Merge order: this one first, then #2429.

PR #2429, opened 2026-08-18
feat(api,frontend): write-ins follow the board they were made on, and the wire stops conflating role with occupancy (#2382, 4 of 4)

### LINKED ISSUE #2382: feat(lodging): write-ins must be scenario-scoped, not only session-scoped

adamflagg:
**Stacked on #2428** (`feature/phase-e-writein-scenario`, "3 of 4"). Review that first; this diff is only what sits on top of it.

Closes #2382.

## Why this PR exists

PR 3 made a scenario's write-ins **REPLACE** the live ones on read, and left `set_availability` taking no scenario and always writing the live occupancy table. Between the two, a staff member working inside a scenario recorded a write-in, that scenario's own read replaced it away, and the board they had just made it on did not show it. Before PR 3 the read fell through, so the same write was visible — the regression is created by PR 3 and closed here, which is why #2428's body says it must not reach `main` without this.

## 1. The write follows the board (the important half)

`AvailabilityWriteRequest` gains an **optional** `scenario`, mirroring `SlotMergeRequest`: blank is the **live board**, a scope in its own right rather than a missing value, because staff evaluate the real board and must be able to write onto it.

It steers the **OCCUPANCY** half alone:

| `family_available` | fact | table | scenario |
|---|---|---|---|
| `false` | OCCUPANCY | `lodging_write_ins` / `lodging_write_ins_draft` | **yes** |
| `true` | ROLE | `lodging_availability` | **no — ignored** |
| `null` | clear | both, in the caller's own scope | occupancy half only |

The **ROLE half stays on `lodging_availability` whatever the request says.** The owner ruled staff↔family role is not scenario-scoped ("that's more of a known 'were moving staff to X for weekend Y'"), so a release written from inside a scenario is still a weekend fact.

One consequence worth stating rather than rediscovering: the "one fact at a time" drop is now scoped to the caller's own grain. A release made inside a scenario drops **that scenario's** write-in and leaves the live one alone. The role row is shared by every scope and an occupancy row is not, so reaching across would clear a fact nobody on this board can see.

`LodgingBoard` passes its existing `scenario` prop into `useUnitAvailability` and onward to `lodgingApi.ts`. It **targets** the write; it does not gate it — `canSetAvailability` is still two conditions and not three, or staff on the CampMinder mirror (where most of them look) could not record a write-in at all.

## 2. The compat shim is out

PR 2 kept the wire conflated on purpose: a write-in still reported `family_available_override = False`, because `is_family_available` — and through it every count on the stats bar, and #2093's forest open-tint — was derived from that one field.

Every consumer now reads the occupancy source directly, so the field answers one question again:

| on the wire | means |
|---|---|
| `family_available_override` | the `lodging_availability` ROLE row, and nothing else |
| `write_in` | somebody is in this space — resolved through the unit tree |
| `is_family_available` | the DERIVED answer, which still folds both in |

`lodging_rules.is_family_available` takes a **required** third `is_occupied` argument. Required rather than defaulted: a caller that forgets it silently reports a written-into cabin as open, and this is the derivation every count and the open-tint go through. **No reported number moves** — a test pins `units_total` / `units_family_available` / `units_reserved` / `beds_family_available` across the change.

#2093's tint was already reading through `writeInOccupant` (that is what `writeIn.ts` exists for), so re-pointing it was one edit in `coveringWriteIn` rather than a hunt. The fallback there is **removed, not left unused**: it synthesised a cover from `family_available_override === false` plus the unit's own `occupant_name`, which now means a role-closed cabin and names nobody — reading it would report an occupant that exists in no row.

### One deliberate non-change, flagged for the owner

A bare `family_available_override === false` on a family cabin now gets **no badge**, where it used to badge "Write-in". That badge was a lie (the row names nobody), 1500000162 left zero such rows, and `set_availability` only ever writes `true` to `lodging_availability` — so the state is unreachable rather than rare. Naming it would mean inventing a staff-facing word for "closed by role", which is a product decision and not a side effect of a storage split. Say the word and I will add one.

## 3. Premise comments retired

`writeIn.ts`'s "a write-in is a fact about a physical space", `unitBadges.ts`'s "global, scenario-less" column, and `lodging_repository.py`'s "availability used to be the exception" all now describe the split that exists: ROLE in `lodging_availability` (weekend-scoped, one layer), OCCUPANCY in `lodging_write_ins`/`_draft` (scenario-scoped, replace-not-merge). `docs/architecture/lodging-occupancy.md` follows, including the "the READ replaces; the WRITE does not yet choose" paragraph that this PR makes false.

## Tests

Written first, run failing, then implemented. An **independent** reviewer re-ran the mutation checks on the load-bearing assertions afterwards and posted the transcripts as a review comment on this PR (six mutations, all confirmed red, all restored) — that comment, not this line, is the evidence.

- `TestAWriteInInsideAScenarioIsWrittenToTheDraftTable` — 8 tests: draft create carries the scenario, the live table is untouched, the lookup is keyed on the scenario, update-not-duplicate, the role half stays put, a scenario release drops that scenario's write-in only, clear deletes the draft + the role row, and the lost-race recovery re-reads the **draft** row.
- Endpoint-level: a scenario write-in reaches `lodging_write_ins_draft`; a scenario release still reaches `lodging_availability`.
- `TestTheWireStopsSpellingAnOccupancyAsFamilyAvailableFalse` — 4 tests, including the counts-do-not-move guard and a stale role `false` still reporting itself.
- `is_family_available` gains a parametrised "occupancy closes the unit whatever the role says" case.
- Frontend: `lodgingApi`, `useUnitAvailability`, `LodgingBoard.availability` (live board **and** scenario), `writeIn` (the proxy no longer fakes a cover), plus the `Required<AvailabilityWriteRequest>` fixture.

## ⚠️ NO CI ON THIS PR

**Update, 2026-08-17:** PRs 1-3 of this series (#2422, #2424, #2428) have all merged, this branch was rebased `--onto main`, and it now targets `main` directly — so the paragraph that used to sit here, saying CI never runs and `Closes #2382` registers nothing, no longer applies. `CI Summary` runs on this PR for real, and `closingIssuesReferences` was confirmed to flip from `[]` to `[2382]` after the body re-save the retarget requires. Local gates before the push: `pre-push-verify.sh` ALL CHECKS PASSED, plus 2,310 API tests green.

Local gates, run in the worktree on a clean tree:

```
$ bash scripts/pre-push-verify.sh        # auto-detect picked up all six areas
── Python ──      PASS ruff format · PASS ruff check · PASS mypy (815 files) · PASS pytest (6258 tests)
── Go ──          PASS go build · PASS golangci-lint · PASS go test
── Frontend ──    PASS prettier · PASS eslint · PASS tsc · PASS vitest
── Migrations ──  PASS migration headers · PASS no options:{} anti-pattern
── PocketBase JS ── PASS pb-js-lint
── Shell ──       PASS shellcheck
═══ ALL CHECKS PASSED ═══

$ git push  (pre-push hook)  6405 passed, 41 skipped
$ git fetch -q && git rev-list --count origin/feature/phase-e-writein-frontend..HEAD
0
```

## 👀 GUI impact — HOLDING for your visual review

Do not merge on my say-so. On the board:

Open a weekend, pick a **scenario**, and write somebody into a cabin — the occupant card should appear on that board immediately, and switching to the CampMinder mirror should show the cabin **open** (that is the whole fix; before this PR the write landed on the mirror instead). Then release a staff cabin from inside the same scenario and confirm the "Released" badge appears on **every** scenario and on the mirror, because a release is a weekend fact. Finally check nothing regressed on the resting board: written-into cabins keep the slate "Write-in" chip and lose the forest open-tint, released staff cabins keep the emerald "Released" chip, and the stats bar's cabin/bed counts read exactly as they did before.

## Not in this PR

- Widening single-cover arity to N write-ins per container — that is #2381, and it survives this storage change because it lives in four type signatures a new table does not touch.
- `boardLayout.ts` (owned by in-flight #2404) is untouched.
- The auto footer note and a per-write-in X — also #2381.


adamflagg:
## Independent review — floor 60, own diff only (`origin/feature/phase-e-writein-scenario...HEAD`)

I did not write this code. Everything below is fixed **in branch** (`302b8e91`) unless it says refuted. Nothing filed: no finding was pre-existing *and* multi-file *and* a design call.

**The write path is verified end to end.** I traced and re-derived each hop rather than trusting the tests:

`LodgingBoard` (`scenario` prop) → `useUnitAvailability({ year, sessionCmId, scenario })` → `setUnitAvailability` → `body.scenario` → `AvailabilityWriteRequest.scenario` → `set_availability` → `create_draft_write_in({… "scenario": S})` → `lodging_write_ins_draft` → `fetch_draft_write_ins` filters `session_cm_id = N && year = Y && scenario = "S"` → `write_in_row_by_unit` keyed on `_s(row, "unit")` → `is_family_available(…, is_occupied=True)` + `write_in` cover. The written payload and the read filter agree field for field. `useUnitAvailability` is the only consumer of the write in the whole tree (`grep onSetAvailability|setAvailability|useUnitAvailability` over `frontend/src`), so there is no second caller that could still be sending nothing.

Blank still lands live (`create_write_in`, no `scenario` key — the live table has no such column). The **ROLE** half writes `lodging_availability` with no `scenario` in the payload even when the request names one; I mutated it to leak the scenario in and two tests went red, so that is pinned rather than incidental.

**The proxy move is complete.** I grepped `family_available_override` across `frontend/src`, `api/`, `pocketbase/`, `docs/` myself. Every non-test production read is now either the ROLE (`=== true` in `unitBadges.ts` ×2 and `UnitAvailabilityControl.tsx`, `!== null` in `availabilityAction`'s residual-clear branch) or a comment. Every "is somebody in it" consumer goes through `writeInOccupant`/`writeInSource` — including kindred#2093's forest open-tint, which is `openMarkerActive = dashed && writeIn === null && …` at `LodgingUnitCard.tsx:525`, reading `writeInOccupant(unit)` at 434. `dragPlacement.resolveDrop`, the `useDroppable` `disabled` flag, `reservationBadge`/`writeInBadgeApplies` and `availabilityAction` all read the same function. No Go consumer of `family_available` exists. `boardLayout.ts` untouched; no #2381 leakage; no ARIA or `sr-only` added anywhere in the diff.

**The shim is genuinely out and the counts hold.** `is_family_available` is the only place the two facts meet (one production caller). I checked the four numbers by hand against the old derivation: `override = False if write_in_row else role` → `is_family_available(cls, override)` and the new `is_family_available(cls, role_override, is_occupied=write_in_row is not None)` agree on every input, so `units_family_available` / `units_reserved` / `beds_family_available` / `_is_planning_inventory` are unmoved.

---

### Mutation transcripts

Six mutations, each applied to the implementation alone, run, then restored. All six went red.

| # | mutation | result |
|---|---|---|
| 1 | `lodging_write_service.py`: `in_scenario = request.scenario != ""` → `in_scenario = False` | **7 failed**, 167 passed |
| 2 | role payload → `{**data, "family_available": True, "scenario": request.scenario}` | **2 failed**, 172 passed |
| 3 | `lodging_rules.py`: `if is_occupied:` → `if False:` | **13 failed**, 230 passed |
| 4 | `useUnitAvailability.ts`: pass `scenario: ''` to `setUnitAvailability` | **1 failed**, 16 passed |
| 5 | `LodgingBoard.tsx`: `useUnitAvailability({ year, sessionCmId })` | **2 failed**, 7 passed |
| 6 | `lodgingApi.ts`: drop `scenario` from the PUT body | **2 failed**, 21 passed |
| 7 | `writeIn.ts`: restore the `coveringWriteIn` fallback verbatim | **4 failed**, 1193 passed |

Selected output:

```
# 1 — the write ignores the scenario
E   AssertionError: collection('lodging_write_ins_draft') call not found
FAILED …TestAWriteInInsideAScenarioIsWrittenToTheDraftTable::test_a_write_in_naming_a_scenario_creates_a_draft_row
FAILED …::test_the_draft_lookup_is_keyed_on_the_scenario
FAILED …::test_a_release_inside_a_scenario_drops_that_scenarios_write_in
FAILED …::test_a_lost_draft_write_in_race_is_recovered_against_the_draft_table
FAILED tests/unit/api/routers/test_lodging_endpoints.py::TestAvailabilityWrites::test_writing_somebody_in_from_a_scenario_creates_a_draft_occupancy_row
======================== 7 failed, 167 passed in 3.32s =========================

# 2 — the ROLE half leaks the scenario (the defect the brief called most likely)
E   AssertionError: assert 'scenario' not in {'session': 'sess_1', …}
FAILED …::test_the_role_half_stays_on_lodging_availability_inside_a_scenario
FAILED …::TestAvailabilityWrites::test_a_release_from_a_scenario_is_still_a_weekend_fact
======================== 2 failed, 172 passed in 2.96s =========================

# 3 — occupancy stops closing the unit; the COUNTS move, which is the guard that matters
FAILED …TestIsFamilyAvailable::test_an_occupancy_closes_the_unit_whatever_the_role_says[staff_default-True]
FAILED …::test_the_counts_do_not_move_when_the_shim_comes_out - AssertionError: assert 2 == 1
FAILED …::test_a_written_into_cabin_is_still_counted_as_reserved - AssertionError: assert 2 == 1
FAILED …::test_the_lander_counts_a_written_into_cabin_as_reserved - assert 0 == 1
======================== 13 failed, 230 passed in 2.77s ========================

# 4 — the client stops sending it
- "scenario": "scn7x2k9qw3mnbv"
+ "scenario": ""
 ❯ src/hooks/useUnitAvailability.test.tsx:115
      Tests  1 failed | 16 passed (17)
```

Mutation 4 is worth one note: it left `LodgingBoard.availability.test.tsx` **green**, because that file mocks the hook. Mutation 5 is what covers the board→hook edge, and mutation 6 the client→wire edge. The three together close the chain; none of them alone does.

---

### Fixed in branch — six premises the split made false

No behaviour changed. I found these by grepping the *claims* rather than the code, which is where they were hiding.

1. **`api/schemas/lodging.py`, `occupant_name`** — still read *"what closes a cabin is `family_available_override`, and an occupant with no override is not a state any writer can produce."* Both halves are false three lines below the comment that replaced them, and this PR's own `test_a_written_into_unit_reports_no_role_override` asserts the opposite: an occupant with **no** role override is now the ordinary state.

2. **"No writer produces both rows"** (`_build_units`, `set_availability`'s docstring, and `test_a_released_cabin_that_is_also_written_into_keeps_its_role_on_the_wire`) — true when the drop was global, false now. The occupancy drop is scoped to the caller's grain while the role row is shared, so: write somebody in on the **live** board (creates the live row, drops the role row) → open a scenario and release the cabin (creates the role row, drops *that scenario's* draft write-in, which is not the live one) → both rows stand. One staff member, two clicks, no race. I corrected the claim rather than the behaviour: `is_family_available` folds both in and occupancy wins, so the cabin still reads closed and a clear on either grain still removes the role row. Narrowing the role drop instead would leave a release standing under a write-in on the *same* board, which is worse — flagging it rather than deciding it.

3. **`docs/reference/weekend-go-live-sequence.md`** — *"**Do not reintroduce a scenario dimension on availability**, in any form including a draft twin"*, sitting in the section the document calls its most durable. As written it forbids exactly what kindred#2382 shipped on an owner ruling, and it is the last thing a future agent reads before pushing back on this work. Amended in place to name the half it was actually made about, with the evidence that split the other half out; the "Live is undefined" paragraph's parenthetical was corrected the same way.

4. **`ScenarioWriteRequest`'s docstring** — gave the reason `AvailabilityWriteRequest` does not extend it as "availability is a fact about the weekend". The real reason is now the word **required**, not the word *scenario*.

5. **`get_weekend_roster`'s docstring** — *"Availability does NOT vary this way … the same rows resolve for every plan."* False since PR 3 for the occupancy half, on the docstring of the very read this PR's write path pairs with. Also `set_availability`'s *"the request model … unchanged"*, which this PR changed.

6. **`UnitAvailabilityControl`'s `canManage` and `LodgingUnitCard`'s `canSetAvailability` prop docs, and the invalidation test's comment** — all still argued from "availability carries no scenario", and `UnitAvailabilityControl`'s went further with *"a write-in is a fact about who is in the building rather than a draft"*, which is the exact proposition this PR reverses. The conclusion (don't gate on a scenario) is right; the premise is now backwards, and `useUnitAvailability.ts` and `LodgingBoard.tsx` already state the correct one.

**PR body**: it claimed *"every green-on-first-write assertion was mutation-checked (transcripts in the thread)"* — the thread held only CodeRabbit's skip notice. Since the body is the permanent squash commit message here, I edited it to point at this comment instead.

### Refuted — not filed, with the reasoning

- **Removing `coveringWriteIn`'s fallback loses the own-row occupancy for a blank-`code` unit.** Real mechanism: `write_in_covers` skips blank codes on both sides (the `""` key collides with `parent_code == ""`), so such a unit gets no `write_in`; the old fallback caught it via `family_available_override === false`, and without it `resolveDrop` would allow a drop into an occupied cabin. **Unreachable**: `lodging_units.code` is `required=1` in the collection schema, and both the prod and dev snapshots hold 118 units with **0** blank codes. The counts are unaffected regardless — `is_family_available` reads the unit's own write-in row, not the cover.
- **A `staff_default` cabin holding both a release and a write-in badges "Released" instead of showing the occupancy.** Reachable via finding 2's sequence, but not visible: `_is_planning_inventory` / `boardLayout.isPlanningInventory` exclude a `staff_default` unit whose `is_family_available` is false, so no card is drawn on the board or the map (which builds from the same slots), and `availabilityAction`'s `clear` still removes both rows.
- **The `family_available_override === false` "no badge" non-change.** Left alone deliberately — `set_availability` only ever writes `true` to `lodging_availability` (verified in the branch), `1500000162` deleted every `family_available = 0` row, and naming the state is a product decision. The body already flags it for you.

### Gates

`bash scripts/pre-push-verify.sh` → `ALL CHECKS PASSED` (6244 passed / 14 skipped pytest, ruff, mypy, Go build + lint + test, prettier, eslint, tsc, vitest, migration headers, pb-js-lint, shellcheck). Pre-push hook on the push itself: **6405 passed, 41 skipped**, plus `api-types-freshness` — `types.gen.ts` is unmoved by my edits. Push verified by remote ref: `git rev-list --count origin/feature/phase-e-writein-frontend..HEAD` = **0**.

`closingIssuesReferences` on this PR is **empty**, as the body predicts — expected for a stacked PR, not a defect. `CI Summary` does not run for the same reason; only `Documentation Linting` and `Validate PR title` fired, and the green `CodeRabbit` check is its "auto reviews are disabled" no-op. **Not merged, not tagged.** The GUI hold in the body stands.

PR #2441, opened 2026-08-18
fix(metrics): fold Family Camp into cancellation still-attending lookup

### LINKED ISSUE #2435: bug(metrics): [OVERTURNED — do not implement] cancellation_service omits Family Camp from its still-attending lookup, so weekend families read as true departures
> ## ⛔ OVERTURNED — the premise below was ruled against. DO NOT IMPLEMENT IT.
>
> **This issue was implemented and then deliberately reverted on an owner ruling. Everything under
> "Original report (overturned)" is a live-looking specification for behaviour the repo has decided
> it does not want. A fresh reader who follows it will rebuild reverted code.**
>
> | | |
> |---|---|
> | Implemented by | PR #2441 — commit `ca1f202a` |
> | **Reverted by** | **PR #2455 — commit `d8f465ff`** (merged 2026-08-18T15:24Z) |
> | Guard test added | `TestFamilyCampIsNotStillAttending` in `tests/unit/api/test_cancellation_service.py` |
>
> **The ruling:** these metrics track the **camper journey through summer**. A camper who leaves
> summer and comes only to a Family Camp weekend **is a real departure** — precisely the population
> staff want surfaced. Family Camp is a separate program, not a summer session the camper moved to.

### Which sentences below are wrong, by name

Two passages in the original text are the ones that will persuade the next reader. Both are void:

1. **"The reasoning transfers verbatim"** (under *The precedent is eight lines above the bug*) —
   it does not transfer. The SCIT/TLI fold-in is sound *because teen programs are summer*. Family
   Camp is not summer. The analogy is the whole argument, and the analogy fails.
2. **The entire section titled "Why this is safe to fix"** — its claim is that touching the
   enrollment lookup only, and not the display scope, keeps Family Camp out of summer metrics. That
   is true about *scope* and irrelevant to the objection. The objection is not "Family Camp will
   leak into display"; it is that **being enrolled in Family Camp does not make you still-attending
   summer at all**, in any scope. Narrowing the blast radius does not rescue a wrong definition.

There is a real, separate lesson here that outlived the revert: a change to *what a staff-facing
number means* is an owner decision even when it touches no frontend file and passes every check.
This one flipped a population of campers out of the cancellation metric with green CI and zero UI
diff.

### What the revert restored (verified against `main` @ `086a353e`)

| file | state on `main` |
|---|---|
| `api/services/cancellation_service.py` | `enrollment_session_ids = set(all_sessions.keys()) \| set(teen_sessions.keys())` — line 85 |
| `api/services/waitlist_service.py` | same, line 77. This file was **not named by the original report**; it was pulled in in-branch by the review pass, which extended the same premise to a second metric family |
| `api/utils/session_metrics.py` | `FAMILY_SESSION_TYPE` removed — it had no other consumer |

**Not reverted, and still correct:** PR #2444 (`7c519aad`, for #2434) scopes the historical-trends
cancellation numerator to the same `session_types` as its enrolled denominator and dedupes to
distinct persons. It excludes Family Camp from *both* sides, so it never folded Family Camp in.
`detect_session_swaps`, `true_departure_count` and every display scope were untouched throughout.

### The guard, and the hole in it

`TestFamilyCampIsNotStillAttending` pins the ruling for **cancellations**. Re-verified by mutation on
`086a353e`: re-adding a `family` fetch to `enrollment_session_ids` in `cancellation_service.py` turns
it red with `assert 0 == 1`; unmutated it is green.

⚠️ **`api/services/waitlist_service.py` has no mirror of that test.** Applying the identical fold-in
to `waitlist_service.py` leaves `tests/unit/api/test_waitlist_service.py` fully green (36/36), and the
whole `tests/unit/api/` suite raises nothing attributable to it. The word `family` does not appear
anywhere in that test file. **Half the ruling is unprotected** — a re-attempt would go red on
cancellations and green on waitlist, which is the worst shape for a partial rebuild. Filing a mirror
test is a follow-up, not part of this issue.

### On the "64 distinct persons" figure

That measurement was **exact when this was filed, at 2026-08-18T04:45Z**, and should be read as a
point-in-time figure rather than an error. A later same-day sync moved it. Re-run against the
production snapshot current to **2026-08-18T21:47Z**, the same query and the same grain now yields
**67 distinct persons**. Both numbers are correct for their moment; neither changes the ruling.

---

# Original report (overturned)

*Preserved verbatim except where an identifier drifted; the corrected identifier is marked. The
reasoning is kept for the record, not for implementation.*

`api/services/cancellation_service.py` decides whether a cancelled camper is a **true departure** or
merely moved between sessions. Family Camp is missing from that lookup, so a household that cancels a
summer session and still attends a weekend is counted as having left camp entirely.

Found by a read-only audit during the Phase E triage campaign, re-verified before filing.

## The precedent is eight lines above the bug

`api/services/cancellation_service.py:78-89` *(identifier drift: on `main` @ `086a353e` this block is
**lines 77-85**)* already solves exactly this problem for teen programs:

```python
# Fetch ALL summer sessions for display + cross-session enrollment lookup.
all_sessions = await self.repository.fetch_sessions(year, list(SUMMER_SESSION_TYPES))

# Teen programs (SCIT/TLI) also count as "still attending camp": a camper
# who cancels one teen session but keeps another reads as has_other_sessions,
# not a true departure. Fold teen session ids into the enrollment lookup
# only -- the display scope below stays summer-only / the requested filter.
teen_sessions = await self.repository.fetch_sessions(year, list(SUMMER_TEEN_TYPES))
enrollment_session_ids = set(all_sessions.keys()) | set(teen_sessions.keys())
```

The reasoning transfers verbatim: a camper who cancels a summer session but attends a Family Camp
weekend is still at camp. `family` is simply not in either set.

> ⛔ **This is the void claim.** It does not transfer — teen programs are summer, Family Camp is not.

## Why this is safe to fix

> ⛔ **This entire section is void.** It answers a scope objection nobody raised, not the definitional
> one that decided the matter. See "Which sentences below are wrong, by name" above.

The distinction the comment draws is the whole answer to the usual objection. Family sessions would
enter the **enrollment lookup only**, not the display scope — so this does not put Family Camp into
summer metrics (that is a separate, much larger question). It only stops the service asserting that
someone left when they did not.

## Measured (2026, production snapshot)

**64 distinct persons** *(as of 2026-08-18T04:45Z; 67 as of the 2026-08-18T21:47Z snapshot — see
above)* hold a cancellation transition (`cancelled` / `withdrawn` / `dismissed`) *and*
an enrolled row in a 2026 `family` session. Grain: distinct `person_id`, `attendee_status_history`
joined to enrolled `attendees`. Every one of those is currently eligible to be classified as a true
departure on the strength of the summer cancellation alone.

> ⛔ Under the ruling, being so classified is **correct**, not a defect.

⚠️ Coverage caveat: `attendee_status_history` holds 2026 rows only, so this cannot be measured for
prior years — the count above is not a rate and should not be quoted as one.

## Suggested fix

> ⛔ **Do not apply.** This is the change that was reverted by `d8f465ff`.

Add a family-session fetch beside the teen one and fold its ids into `enrollment_session_ids`, leaving
`filtered_sessions` / the display scope untouched. Mirror the existing comment so the next reader sees
one rule rather than two special cases.

- **Files:** `api/services/cancellation_service.py` (+ its tests)
- **Related:** the broader "Family Camp is invisible to the metrics stack" question is deliberately NOT part of this — that one is a product decision about grain and is not filed.

adamflagg:
`cancellation_service.py` already folds teen (SCIT/TLI) sessions into its cross-session
enrollment lookup, so a camper who cancels one summer session but keeps another summer
or teen session reads as `has_other_sessions`, not a true departure. Family Camp was
missing from that lookup entirely: 64 distinct persons in the 2026 production snapshot
hold both a summer cancellation and an enrolled Family Camp weekend row (grain: distinct
`person_id`, `attendee_status_history` joined to enrolled `attendees`), and every one of
them was eligible to be classified as having left camp entirely on the strength of the
summer cancellation alone.

## What changed

Fetch Family Camp sessions the same way teen sessions are already fetched, and fold
their ids into `enrollment_session_ids` only — the display scope (`filtered_sessions`)
is untouched, so this does not add Family Camp to summer metrics; it only stops the
service asserting a departure that did not happen.

## Deliberately NOT done

- `true_departure_count` / `session_swap_count` are a separate metric driven by
  `detect_session_swaps`' same-day cancel+enroll heuristic and never read
  `enrollment_session_ids` at all — out of scope for this fix.
- The broader "Family Camp is invisible to the metrics stack" question (adding Family
  Camp to summer display/breakdowns) is a separate product decision and is not part of
  this change, per the issue.

## Testing

Added `TestFamilyCampCrossSessionEnrollment`, mirroring the existing
`TestTeenCrossSessionEnrollment` pattern: a cancelled summer camper whose household
keeps a Family Camp weekend enrollment now counts as `has_other_sessions`, not
`no_other_sessions`. Written first, verified failing (`has_other_sessions == 0` when it
should be `1`), then implemented. Mutation-checked by reverting the
`enrollment_session_ids` union back to teen-only, confirming the new test goes red, and
restoring.

Closes #2435.

### DIFF
--- a/api/services/cancellation_service.py
+++ b/api/services/cancellation_service.py
@@ -25,6 +25,7 @@ from api.services.breakdown_calculator import calculate_percentage, compute_regi
 from api.services.extractors import extract_gender, extract_grade
 from api.utils.session_metrics import (
     DEFAULT_SUMMER_SESSION_TYPES,
+    FAMILY_SESSION_TYPE,
     SUMMER_TEEN_TYPES,
     build_ag_parent_map,
     get_session_from_expand,
@@ -35,11 +36,6 @@ from bunking.logging_config import get_logger
 
 # Back-compat alias as an explicit module export (shared default lives in session_metrics).
 SUMMER_SESSION_TYPES = DEFAULT_SUMMER_SESSION_TYPES
-
-# Family Camp session type -- not part of DISPLAY_SESSION_TYPES / DEFAULT_SUMMER_SESSION_TYPES
-# (adult-focused, separate program), but a household that cancels a summer session while
-# keeping a Family Camp weekend is still attending camp. See enrollment_session_ids below.
-FAMILY_SESSION_TYPE = "family"
 
 if TYPE_CHECKING:
     from .metrics_repository import MetricsRepository
--- a/api/services/cancellation_service.py
+++ b/api/services/cancellation_service.py
@@ -36,6 +36,11 @@ from bunking.logging_config import get_logger
 # Back-compat alias as an explicit module export (shared default lives in session_metrics).
 SUMMER_SESSION_TYPES = DEFAULT_SUMMER_SESSION_TYPES
 
+# Family Camp session type -- not part of DISPLAY_SESSION_TYPES / DEFAULT_SUMMER_SESSION_TYPES
+# (adult-focused, separate program), but a household that cancels a summer session while
+# keeping a Family Camp weekend is still attending camp. See enrollment_session_ids below.
+FAMILY_SESSION_TYPE = "family"
+
 if TYPE_CHECKING:
     from .metrics_repository import MetricsRepository
 
@@ -82,7 +87,14 @@ class CancellationService:
         # not a true departure. Fold teen session ids into the enrollment lookup
         # only -- the display scope below stays summer-only / the requested filter.
         teen_sessions = await self.repository.fetch_sessions(year, list(SUMMER_TEEN_TYPES))
-        enrollment_session_ids = set(all_sessions.keys()) | set(teen_sessions.keys())
+
+        # Family Camp also counts as "still attending camp": a household that
+        # cancels a summer session but keeps a Family Camp weekend reads as
+        # has_other_sessions, not a true departure. Fold family session ids
+        # into the enrollment lookup only -- same rule as teens above, the
+        # display scope below stays summer-only / the requested filter.
+        family_sessions = await self.repository.fetch_sessions(year, [FAMILY_SESSION_TYPE])
+        enrollment_session_ids = set(all_sessions.keys()) | set(teen_sessions.keys()) | set(family_sessions.keys())
 
         # Fetch filtered sessions for display
         effective_types = session_types or list(SUMMER_SESSION_TYPES)
--- a/api/services/waitlist_service.py
+++ b/api/services/waitlist_service.py
@@ -24,6 +24,7 @@ from api.services.breakdown_calculator import calculate_percentage, compute_regi
 from api.services.extractors import extract_gender, extract_grade
 from api.utils.session_metrics import (
     DEFAULT_SUMMER_SESSION_TYPES,
+    FAMILY_SESSION_TYPE,
     SUMMER_TEEN_TYPES,
     build_ag_parent_map,
     get_session_from_expand,
@@ -74,7 +75,14 @@ class WaitlistService:
         # not no_enrollment. Fold teen session ids into the enrollment lookup only
         # -- the display scope below stays summer-only / the requested filter.
         teen_sessions = await self.repository.fetch_sessions(year, list(SUMMER_TEEN_TYPES))
-        enrollment_session_ids = set(all_sessions.keys()) | set(teen_sessions.keys())
+
+        # Family Camp also counts as "enrolled at camp": a household waitlisted
+        # for a summer session but already enrolled in a Family Camp weekend
+        # reads as has_enrollment, not no_enrollment. Fold family session ids
+        # into the enrollment lookup only -- same rule as teens above, the
+        # display scope below stays summer-only / the requested filter.
+        family_sessions = await self.repository.fetch_sessions(year, [FAMILY_SESSION_TYPE])
+        enrollment_session_ids = set(all_sessions.keys()) | set(teen_sessions.keys()) | set(family_sessions.keys())
 
         # Fetch filtered sessions for waitlist display
         effective_types = session_types or list(SUMMER_SESSION_TYPES)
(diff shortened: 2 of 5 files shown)


adamflagg:
### Review + remediation (independent second reader)

**Verified the core fix.** `enrollment_session_ids` in `cancellation_service.py` now folds `family`-type sessions in alongside teen sessions, exactly mirroring the existing SCIT/TLI pattern; `filtered_sessions` (display scope) is untouched. Matches issue #2435's grain and reasoning. New test `TestFamilyCampCrossSessionEnrollment` was verified red before the fix (`has_other_sessions == 0`) and green after; mutation-check (revert the union, confirm red, restore) also verified.

**Found and fixed in-branch (one file, same pattern, no product judgement — folded per policy rather than filed):**

`api/services/waitlist_service.py` had the identical bug. It builds `enrollment_session_ids` with the same teen-fold-in pattern used in `cancellation_service.py`, and was likewise missing Family Camp — a household waitlisted for a summer session while already enrolled in a Family Camp weekend was misclassified `waitlisted_no_enrollment` instead of `waitlisted_has_enrollment`. Fixed with the identical fold-in, TDD (test written first, verified red, then fixed), and mutation-checked (revert union → red → restore).

Confirmed this is the full extent of the pattern: `SUMMER_TEEN_TYPES` also appears in `retention_service.py`, `retention_trends_service.py`, `forecast_service.py`, `day1_service.py`, `drilldown_service.py`, and `session_availability_service.py`, but none of them build an `enrollment_session_ids`-style cross-session "still attending" lookup — they use the teen-type list for display/scope gating only, which is a different, unaffected mechanism. No further filing needed.

Also centralized the new `FAMILY_SESSION_TYPE` constant into `api/utils/session_metrics.py` (next to `SUMMER_TEEN_TYPES`, which lives there for exactly this reason) instead of leaving it duplicated once a second service needed it.

**Refuted:** none — no false positives surfaced in this pass.

**Verification:** `tests/unit/api/` (2347 tests) and full `bash scripts/pre-push-verify.sh` (ruff format/check, mypy, pytest — 6442 passed / 41 pre-existing skips) both clean after the additional fix. Pushed as `64bc3ac6`; CI Summary green.

CodeRabbit not requested (quota reserved for the wave's highest-blast-radius PR).

PR #2442, opened 2026-08-18
fix(api): compute historical journey child age at that year's session start

### LINKED ISSUE #2420: bug(api,frontend): the family history "see members" modal shows every past year at the child’s age TODAY — 2017 rows are 8 years too old
## ⚠️ CORRECTED 2026-08-18 — READY TO BUILD. Decision-free, not gated, server-side.

Re-verified against `main` @ `1797f9bd` and the production snapshot `pocketbase/pb_data/data-prod.db`.
Every anchor in this body was re-derived; the ones that had moved are corrected in place below.

- **Decision-free.** The "no fix is proposed here on purpose" note near the bottom is retired — see it
  for the three options and why two are rejected. The fix is **server-side**, in
  `api/services/lodging_roster_service.py`: `build_household_journey` (`:1399`) already knows which
  year each journey row is, and `_party_child` (`:400-423`) does not. That gap is the whole bug.
- **Sending `birthdate` to the client is RULED OUT.**
  `GET /api/lodging/households/{household_cm_id}/journey` (`api/routers/lodging.py:155-174`) takes
  `Depends(get_current_user)` and **no** `require_permission` — it is open to any authenticated user,
  exactly like `/roster`. Deriving client-side means publishing a new personal identifier on that wire
  to save a server-side subtraction.
- **The frontend may need no edit at all.** `HouseholdYearMembersModal` renders whatever
  `PartyChild.age` holds, as `Age {n} · Grade {g}` through `displayCampMinderAge`
  (`frontend/src/utils/age.ts:23`). Correct the number on the wire and the modal is correct.
- **NOT gated on #2393.** #2393 edits `HouseholdJourneyYear` and the journey-row assembly; this edits
  `_party_child` and the arguments reaching it. Same two files, different regions — a **merge
  collision to sequence, not a dependency to wait on**.
- **Not UI-gated.** Nothing new is drawn; an existing number becomes right.

## What the owner saw

> "when looking at family historical journey members, the ages seem wrong. i can't put my finger
> on why (the kids ages)."

That is exactly the right description, because the number is *plausible* on every row — it is
simply the wrong person's age for that row's year. Confirmed against the production snapshot, with
a root cause and a measurement. **Diagnosis is settled; this issue is scoped to the fix.**

## Diagnosis

**The historical journey renders every past year's children at the age they are TODAY.**

The rendered field is `persons.age`, and `persons.age` is not "age in that season" — it is a
*live* CampMinder attribute mirrored verbatim at whatever moment the row was last synced. Every
`persons` row for 2017–2025 in the production snapshot was backfilled on 2026-02-05/06, so every
historical journey row carries the child's age as of February 2026.

The wire path, all anchors verified against `origin/main` @ `2e325fe7`:

| Where | What |
|---|---|
| `pocketbase/sync/persons.go:684-686` | `pbData["age"] = age` — mirrors CampMinder's `Age` field. The comment on `:684` claims the value "already represents age as of that year's data". **That claim is false**, and it is the origin of the bug. |
| `api/services/lodging_roster_service.py:400-423` (`_party_child`, the `age=_f(child, "age") or None` on `:421`) | publishes that stored column straight onto the wire. Called from BOTH the roster path (`_build_household_parties`, `:1927`) and the journey path (`build_household_journey`, `:1499`) — one mapping by design, so the fix must not fork it |
| `api/schemas/lodging.py:369` (`PartyChild.age`, class at `:354`) | `age: float | None` — no year attached to it |
| `api/services/lodging_roster_service.py:1399` (`build_household_journey`) | correctly builds each year's *membership* from that year's rows, then attaches an age that has no year |
| `frontend/src/components/weekend/HouseholdYearMembersModal.tsx:83-93` (`childDetail`) and `:215` | renders it as `Age {n} · Grade {g}` |

So this is **hypothesis 2, not hypothesis 1**: nothing in the render path computes `today - dob`.
The field arrives pre-computed and pre-stale from the sync, and the surface has no way to know.

### Why CampMinder's `Age` is provably live, not per-season

**93.1% of 2026 `persons` rows (2,414 / 2,593 with a birthdate and a nonzero age) were last
updated on the calendar day-of-month matching the person's own birthdate.** The nightly sync is
rewriting those rows on each person's birthday because that is the day the value it mirrors
changes. A per-season attribute would not do that.

## Measurements — production snapshot `pocketbase/pb_data/data-prod.db`, read 2026-08-17

Population is exactly what the modal can render: distinct `(household, year, child)` from
`attendees` → `persons` where `session.session_type = "family"` and `status_id = 2`.

- **5,473** child rows the modal can render, across **1,526** households and **3,104**
  `(household, year)` journey rows (**2,722** of them historical).
- **5,322 / 5,473 (97.2%)** of stored ages equal the child's age **at that row's last sync date**.
- **163 / 5,473 (3.0%)** equal the child's age **at that year's own session start** — and most of
  those are the current season, where the two dates coincide.
- **4,464 / 5,473 (81.6%)** are wrong by a full year or more. **Re-measured 2026-08-18 and it
  reproduces** (the original 4,465 was one row out). Grain and coverage, stated: one row per
  `(household PB id, year, child PB id)` reachable by the modal — `attendees` → `camp_sessions`
  → `persons` with `session_type = 'family'` and `status_id = 2`, **all years 2017-2026, no scenario
  filter, whole production snapshot**. "Wrong" compares `int(persons.age)` against whole completed
  years at that year's earliest family `camp_sessions.start_date`. 0 of the 5,473 lack a birthdate
  and 0 carry the `age == 0.0` sentinel, so nothing is excluded. On the same population **5,457
  (99.7%) match the child's whole-year age at that row's own last sync date** — which is the
  diagnosis restated as a measurement.

Error by year (`stored whole years − true whole years at that year's session start`), and it is
not a spread, it is a step:

| Year | Rows | Error distribution | Median |
|---|---|---|---|
| 2017 | 560 | +8 (284), +9 (276) | **+8** |
| 2018 | 689 | +7 (360), +8 (329) | **+7** |
| 2019 | 665 | +6 (336), +7 (329) | **+6** |
| 2020, 2021 | 0 | — (season cancelled / no family attendee rows — expected, see `EnrollmentState`) | — |
| 2022 | 610 | +3 (325), +4 (285) | **+3** |
| 2023 | 712 | +2 (315), +3 (397) | **+3** |
| 2024 | 825 | +1 (440), +2 (385) | **+1** |
| 2025 | 727 | 0 (405), +1 (322) | +0 |
| 2026 | 685 | −1 (52), 0 (603), +1 (30) | 0 |

**Every row in 2017, 2018, 2019, 2022, 2023 and 2024 — 4,061 rows — is wrong by at least one year.** 2025 is wrong for the 322 children
whose birthday fell between their 2025 weekend and the February 2026 backfill. Even 2026 is off by
one for 82 rows, in both directions — the current season is *nearly* right by luck of timing, not
by construction.

### The tell that makes it visible but unnameable

Grade **is** stored per-year and **is** correct; age is not. They print on the same line.

Across **1,249** children who appear in ≥2 journey years (2,724 adjacent year-pairs):

- **2,445 / 2,724 (89.8%)** of pairs show the age *essentially unchanged* across the gap
- **1,801 / 2,724 (66.1%)** of the same pairs show grade advancing correctly, one per year

One real household's shape, transcribed with a fictional camper (Ava Chen), values verbatim from
the snapshot:

```
year 2017: "Age 17.04 · Grade 3"    (Ava was 8 that July)
year 2018: "Age 17.04 · Grade 4"    (she was 9)
year 2019: "Age 17.04 · Grade 5"    (she was 10)
year 2022: "Age 17.04 · Grade 8"    (she was 13)
year 2023: "Age 17.06 · Grade 9"    (she was 14)
year 2025: "Age 17.04 · Grade 11"   (she was 16)
year 2026: "Age 17.11 · Grade 12"   (she was 17)  ← the only correct row
```

A staff member reading that sees a 17-year-old in 3rd grade and a family that never ages, while
the grades march correctly up the card. That is the "seems wrong, can't say why".

## What this is NOT — hypotheses checked and ruled out

- **NOT a `today − date_of_birth` computation in the render path.** Grepped `api/` and
  `frontend/src/` for `age`/`birth`/`dob`/`date_of_birth`. The only date-difference age
  computation in the API is `_completed_months` / `_consumes_a_bed`
  (`api/services/lodging_roster_service.py`, `_completed_months` at `:177` and `_consumes_a_bed`
  at `:188`), which is the infant-bed rule, correctly
  takes `session_start` as an argument, and reads `birthdate` rather than `age`. It does not feed
  this surface's display.
- **NOT date-of-birth data quality.** Of the 5,473 rows, **0 have a missing or unparseable
  birthdate and 0 carry the `age == 0.0` unknown-age sentinel**. The known family-camp DOB
  problems (#2405's normalisation, #2275) live in `family_camp_adults`, a different table — and
  the modal renders adults with name and relationship only (`PartyAdult` carries no age field at
  all, `api/schemas/lodging.py:346-352`). Family-camp DOB quality explains **0%** of this effect.
- **NOT a display-formatting bug.** `displayCampMinderAge` (`frontend/src/utils/age.ts:23`)
  faithfully renders the `yy.mm` it is handed.
- **Sibling ordering is unaffected** and does not need fixing:
  `_children_oldest_first` (`:426`) sorts on the same stale `age`, but siblings all age at the
  same rate, so the relative order is preserved.

## Blast radius

**Only this surface.** `PartyChild.age` has four other renderers — `FamilyCard.tsx:271-273`,
`HouseholdRosterRow.tsx:234-236`, `FamilyDetailsPanel.tsx:285-287`, and the sort in
`LodgingUnitCard.tsx` — and all four are **current-season board** surfaces, where "age now" and
"age this season" differ only by the ±1 seen in the 2026 row above. `HouseholdYearMembersModal` is
the only place a *historical* year's age is shown.

Summer's equivalent surface is not affected and offers no pattern to copy:
`frontend/src/components/camper/CampJourneyTimeline.tsx` renders no age at all (zero matches for
`age`). So CLAUDE.md §4's "model summer's" rule does not constrain the fix here — there is nothing
to model.

## Acceptance condition

For a household with journey rows in ≥2 years, and a child with a known birthdate:

1. The age shown on each journey year's "see members" row is that child's age **as of that year's
   family-camp session start**, so a child in a 2017 row and a 2026 row shows two ages roughly
   nine years apart, not the same one twice.
2. Age and grade on the same line are mutually consistent — no "Age 17 · Grade 3".
3. A child with no usable birthdate shows **no** age rather than a wrong one (the modal already
   omits a missing age; `childDetail` at `:88` must keep that behaviour).
4. The current-season board surfaces listed under Blast radius are unchanged, or changed
   deliberately and consistently — not left disagreeing with the modal about the same child.
5. A regression test pins a *historical* year: a fixture child whose stored `persons.age` is
   today's age must render that year's age. A test asserting on the stored value would re-pin the
   bug.

⚠️ **RETIRED 2026-08-18 — the call is made. This issue is decision-free.** The three options were:
derive the age server-side in `_party_child` (which currently receives no year or session date), send
`birthdate` on `PartyChild` and derive client-side, or have the sync stop mirroring a live field into
a per-year table.

**CHOSEN — derive server-side.** `_party_child` (`api/services/lodging_roster_service.py:400-423`) is
the one mapping shared by the roster party (`_build_household_parties`, `:1927`) and the journey party
(`build_household_journey`, `:1499`), and only the
journey caller knows a *past* year, so the year's session start is the argument that has to reach it.
`persons.birthdate` is already on the expanded rows `_consumes_a_bed` reads, so this costs no extra
query. The four current-season renderers keep "age now" unless the same session start is passed
there deliberately.

**REJECTED — `birthdate` on `PartyChild`.** `/roster` and
`GET /api/lodging/households/{household_cm_id}/journey` are open to **any authenticated user**
(`api/routers/lodging.py:155-174`, `Depends(get_current_user)`, no `require_permission`). Deriving
client-side publishes a new personal identifier to that whole audience to save a subtraction. Ruled
out on that ground before the arithmetic is argued.

**REJECTED — changing the sync.** `persons.age` is CampMinder's live attribute and the four
current-season surfaces read it correctly as "age now". Re-pointing the sync breaks four working
surfaces to fix one modal.

The measurements are here so whoever picks does not have to re-derive them.

## Note for whoever picks this up

**PR #2408 MERGED 2026-08-17**, and #2423 landed after it, so this note's forecast is spent: every
anchor above has been re-derived against `main` @ `1797f9bd` and corrected in place. `PartyChild.age`
ended at `api/schemas/lodging.py:369`, not the `:325` predicted here.

The one live overlap is **#2393** (session grain for the journey modal). It edits
`HouseholdJourneyYear` in `api/schemas/lodging.py` and the journey-row assembly inside
`build_household_journey`; this issue edits `_party_child` and what is passed to it. Same two files,
**different regions — a merge collision to sequence, not a dependency**. Neither blocks the other, and
neither needs the other's decision.

adamflagg:
The family history "see members" modal (`HouseholdYearMembersModal`) showed every past
journey year at each child's age **today**, not their age that year — 4,464 of 5,473
renderable child rows (81.6%) on the production snapshot were wrong by a full year or
more, and every 2017/2018/2019/2022/2023/2024 row (4,061 rows) was wrong by at least one
year (kindred#2420).

Root cause: `persons.age` is CampMinder's live `yy.mm` attribute, mirrored unchanged by
the nightly sync on every touch (typically the person's own birthday). `_party_child`
published that stored column straight onto the wire for every caller, and
`build_household_journey` — which already knows which year each row belongs to — had no
way to tell `_party_child` a year had passed.

Fix, server-side only, in `api/services/lodging_roster_service.py`:

- `_party_child` gains an optional `as_of: date | None` keyword. `None` (the default,
  used unchanged by the current-season roster path `_build_household_parties`) keeps
  publishing the stored `persons.age` exactly as before — that surface's blast radius is
  untouched. When `build_household_journey` passes a year's own family session start
  date, `_party_child` instead computes completed years.months from `persons.birthdate`
  at that date (`_age_at`, built on the same `_completed_months` whole-month arithmetic
  `_consumes_a_bed` already uses for the infant-bed rule), and reports no age at all when
  the child has no usable birthdate rather than falling back to the stale value.
- `build_household_journey` derives each traced year's earliest **family** session start
  from `fetch_weekend_sessions(year)` (an existing repository read, reused rather than
  adding a new one) and threads it through as `as_of`.

`birthdate` is still never sent to the client — `/api/lodging/households/{id}/journey`
has no `require_permission` gate, so publishing a new personal identifier there to save a
server-side subtraction was ruled out in the issue. The age is computed and discarded
server-side; the wire shape (`PartyChild.age: float | None`) is unchanged, so no frontend
edit was needed — `HouseholdYearMembersModal` already renders whatever `PartyChild.age`
holds.

Deliberately NOT done:
- No change to the four current-season board renderers of `PartyChild.age`
  (`FamilyCard`, `HouseholdRosterRow`, `FamilyDetailsPanel`, `LodgingUnitCard`'s sort) —
  those read "age now" correctly today and are out of this issue's blast radius.
- No change to `_children_oldest_first`'s sibling ordering, which sorts on the same
  (still-stale) `persons.age` — sibling order is unaffected by which year is stale
  because all siblings age at the same rate, per the issue's own analysis.
- Nothing in the `HouseholdJourneyYear` schema or the journey-row-assembly loop's other
  fields (cabin, housing state, enrollment) touched — that region belongs to kindred#2393,
  a merge collision to sequence rather than a dependency.

Regression tests pin the historical-year computation directly: a fixture child whose
stored `persons.age` reads as "today's age" now renders that year's computed age instead
(mutation-checked — reverting the `as_of` branch turns both new tests red), and a second
test pins that a missing birthdate renders no age rather than the stale one.

Closes #2420.

### DIFF
--- a/api/services/lodging_repository.py
+++ b/api/services/lodging_repository.py
@@ -581,7 +581,14 @@ class LodgingRepository:
                     f"person.household_id = {household_cm_id} "
                     f'&& session.session_type = "family" && {ACTIVE_ENROLLED_FILTER}'
                 ),
-                "expand": "person",
+                # `session` alongside `person` (kindred#2420): the journey
+                # needs to know WHICH family session this attendee row is
+                # enrolled in, to compute that child's age at that specific
+                # session's start rather than at the year's earliest
+                # camp-wide family session -- a season runs several, months
+                # apart, and a household does not necessarily attend the
+                # first one.
+                "expand": "person,session",
                 "sort": STABLE_SORT,
             },
         )
--- a/api/services/lodging_roster_service.py
+++ b/api/services/lodging_roster_service.py
@@ -466,6 +466,19 @@ def _children_oldest_first(children: list[Any]) -> list[Any]:
     return sorted(children, key=lambda c: -(_f(c, "age") or 0.0))
 
 
+def _child_identity(person: Any) -> Any:
+    """The same key `PartyChild.person_cm_id` is published under.
+
+    CampMinder id when there is one; the PB record id string otherwise, for
+    the rare person row with neither -- kept rather than collapsed onto
+    every other anonymous sibling. Shared by `build_household_journey`'s
+    dedup (`seen_by_year`) and its per-child session lookup
+    (`session_start_by_year`) so the two can never disagree about which
+    child is which.
+    """
+    return _i(person, "cm_id") or str(getattr(person, "id", ""))
+
+
 def _housing_state(cabin: str, year_assignments: Mapping[int, str]) -> HousingState:
     """What is known about one household's housing in one year (kindred#2073).
 
@@ -1476,6 +1489,15 @@ class LodgingRosterService:
         )
 
         children_by_year: dict[int, list[Any]] = {}
+        # kindred#2420: each traced child's age has to be computed at THAT
+        # CHILD'S OWN family session start, not at the year's earliest
+        # camp-wide family session -- a season runs several (6 to 10 a year
+        # on the production snapshot, from May through December), and a
+        # household attends whichever one it booked, not necessarily the
+        # first of the year. Keyed the same way `seen_by_year` below is, so
+        # the year's earliest of THIS CHILD'S OWN attendee rows wins, ties
+        # included, without a second pass over `attendees`.
+        session_start_by_year: dict[int, dict[Any, date]] = {}
         # A JOURNEY ROW IS A YEAR, NOT A SESSION, and this is where that stops
         # being a wording point. A family can book two of a season's weekends,
         # which gives one child TWO enrolled family attendee rows in the same
@@ -1493,11 +1515,19 @@ class LodgingRosterService:
         # than collapsed onto every other anonymous sibling.
         seen_by_year: dict[int, set[Any]] = {}
         for attendee in attendees:
-            person = (getattr(attendee, "expand", None) or {}).get("person")
+            expand = getattr(attendee, "expand", None) or {}
+            person = expand.get("person")
             if person is None:
                 continue
             year = _i(attendee, "year")
-            identity = _i(person, "cm_id") or str(getattr(person, "id", ""))
+            identity = _child_identity(person)
+            session = expand.get("session")
+            start = _as_date(_s(session, "start_date")) if session is not None else None
+            if identity and start is not None:
+                starts = session_start_by_year.setdefault(year, {})
+                existing = starts.get(identity)
+                if existing is None or start < existing:
+                    starts[identity] = start
             if identity:
                 seen = seen_by_year.setdefault(year, set())
                 if identity in seen:
@@ -1515,35 +1545,15 @@ class LodgingRosterService:
             )
             if year > 0
         ]
-        cabin_maps, weekend_sessions_by_year = await asyncio.gather(
-            asyncio.gather(*(self.repository.fetch_cabin_assignments_by_household_cm_id(year) for year in years)),
-            # kindred#2420: the age shown for a past year has to be computed
-            # at THAT year's own family session start, not read off
-            # `persons.age` (a live CampMinder attribute the sync mirrors
-            # unchanged onto every year's row). `fetch_weekend_sessions`
-            # already exists for the roster's own session picker, so this
-            # reuses it rather than adding a new repository read.
-            asyncio.gather(*(self.repository.fetch_weekend_sessions(year) for year in years)),
+        cabin_maps = await asyncio.gather(
+            *(self.repository.fetch_cabin_assignments_by_household_cm_id(year) for year in years)
         )
 
         rows: list[HouseholdJourneyYear] = []
-        for year, assignments, sessions in zip(years, cabin_maps, weekend_sessions_by_year, strict=True):
+        for year, assignments in zip(years, cabin_maps, strict=True):
             cabin = assignments.get(household_cm_id, "")
             children = _children_oldest_first(children_by_year.get(year, []))
-            # The EARLIEST family session of the year, matching how a family
-            # camp weekend book-ended by two sessions in one calendar year is
-            # already collapsed to one journey row above (`seen_by_year`).
-            # `None` when a year has no camp_sessions row at all (e.g. a
-            # registration-only year with neither a child nor an adult), in
-            # which case `_party_child` keeps the current-season fallback.
-            session_start = min(
-                (
-                    start
-                    for session in sessions
-                    if _s(session, "session_type") == "family" and (start := _as_date(_s(session, "start_date")))
-                ),
-                default=None,
-            )
+            year_starts = session_start_by_year.get(year, {})
             rows.append(
                 HouseholdJourneyYear(
                     year=year,
@@ -1555,7 +1565,17 @@ class LodgingRosterService:
                     # rendering either as an error or as a family with no kids.
                     enrollment="enrolled" if children else "none_on_file",
                     adults=[_party_adult(adult) for adult in adults_by_year.get(year, [])],
-                    children=[_party_child(child, as_of=session_start) for child in children],
+                    children=[
+                        # Each child's OWN attended session start
+                        # (kindred#2420) -- `None` when this child's
+                        # attendee row carried no readable session date
+                        # (e.g. an unexpanded `session` relation, or a
+                        # camp_sessions row missing `start_date`), in which
+                        # case `_party_child` keeps the current-season
+                        # fallback rather than guessing at a camp-wide date.
+                        _party_child(child, as_of=year_starts.get(_child_identity(child)))
+                        for child in children
+                    ],
                 )
             )
         return HouseholdJourneyResponse(household_cm_id=household_cm_id, years=rows)
--- a/api/services/lodging_roster_service.py
+++ b/api/services/lodging_roster_service.py
@@ -397,8 +397,46 @@ def _party_adult(adult: Any) -> PartyAdult:
     )
 
 
-def _party_child(child: Any) -> PartyChild:
-    """A `persons` row as the wire sees it. See `_party_adult` on sharing."""
+def _age_at(birthdate: date, as_of: date) -> float | None:
+    """Completed years.months at `as_of`, in the same yy.mm encoding
+    `persons.age` uses (kindred#2420).
+
+    Built on `_completed_months` -- the same whole-month arithmetic
+    `_consumes_a_bed` already trusts for the infant-bed rule -- so a
+    historical age and an infant-bed decision can never disagree about how
+    a month is counted.
+
+    A negative month count means `as_of` predates the birth -- bad data, not
+    a valid age -- and is reported as unknown rather than a nonsense
+    negative number.
+    """
+    total_months = _completed_months(birthdate, as_of)
+    if total_months < 0:
+        return None
+    years, months = divmod(total_months, 12)
+    return float(f"{years}.{months:02d}")
+
+
+def _party_child(child: Any, *, as_of: date | None = None) -> PartyChild:
+    """A `persons` row as the wire sees it. See `_party_adult` on sharing.
+
+    `as_of` is the ONE thing that lets this stay one mapping instead of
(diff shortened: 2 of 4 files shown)


adamflagg:
### Independent review + remediation (score floor 60)

**Fixed in-branch (this PR's own mechanism, verified against production data):** the age computation used the year's *earliest family session camp-wide* as `as_of`, not the specific session the household's child actually attended. The production snapshot shows 6-10 distinct family-camp sessions per year, spanning May through December (2023 alone ran `2023-05-26` through `2023-12-28`), and querying real attendee rows confirms most households attend a session other than the year's first. That reintroduced a smaller version of #2420's own bug — off by weeks to months instead of a full year, but still wrong for most rows.

Fix: `fetch_household_family_attendees` now expands `session` alongside `person` (`api/services/lodging_repository.py`), and `build_household_journey` tracks each child's own earliest attended session start per traced year — keyed the same way the existing `seen_by_year` dedup is, so the two can never disagree about which child is which — and passes *that* as `as_of` instead of scanning all camp-wide family sessions for the year. This also removes the now-unneeded `fetch_weekend_sessions` per-year concurrent read entirely, so the fix nets out simpler than what it replaced.

Regression test: `test_a_historical_age_uses_the_childs_own_attended_session_not_the_years_earliest_kindred_2420` — a child enrolled only in the year's *late* (September) session, with the year's early (May) session present in real proximity but never attached to her. Mutation-checked: reverting the source change (keeping only the test-fixture shape change) turns it red, asserting `9.01` where the reverted mechanism produces `17.04`/`8.09` depending on which half is reverted.

**Refuted:** none — the merged `_completed_months`/`_age_at` arithmetic, the `age == 0.0` newborn-vs-unknown handling, and the no-birthdate fallback were all checked against the fixture math and are correct.

**Filed as follow-up:** none — the finding was in code this PR itself wrote, so it was fixed here per policy rather than filed.

All 227 lodging-roster + 137 lodging-repository unit tests pass, full mockable suite (6443 tests) is green, `scripts/pre-push-verify.sh` is clean, and CI Summary is green on the pushed commit.

Issue #2443, opened 2026-08-18
feat(metrics): distinguish no-coverage years from zero cancellations in historical trends


adamflagg:
## ★ OWNER RULING — 2026-08-19: the chart's gap is a GREYED BAND

**A no-coverage year renders as a greyed band drawn behind the uncovered years.** That settles the
rendering question this issue was holding (plain break / dashed lead-in / greyed band / footnote).

### ⚠️ BUT THE COST IS NOT FREE — read this before estimating

Investigated 2026-08-19 against the tree at `691c0954` and the installed library. **The band is buildable,
and it is not a one-liner.** Two facts decide the work, and both were unknown when the ruling was given:

**1. The x-axis is CATEGORICAL, and its real scale is a `point` scale with zero bandwidth.**

- Library: **recharts `^3.10.1`** (`frontend/package.json:64`; installed `3.10.1`).
- `frontend/src/components/metrics/TrendLineChart.tsx:169` — `<XAxis hide height={0} dataKey="year" />`.
  No `type`, so it defaults to `type: 'category'`.
- `node_modules/recharts/es6/state/selectors/combiners/combineRealScaleType.js` — for `scale: 'auto'` and
  `type: 'category'` **inside a `LineChart`**, the real scale type is **`'point'`**, not `'band'`.
- `ReferenceArea` positions itself via `xAxisScale.map(x1, {position:'start'})` and
  `map(x2, {position:'end'})` (`node_modules/recharts/es6/cartesian/ReferenceArea.js`, `getRect`), and the
  `position` adjustment adds `bandwidth()` (`node_modules/recharts/es6/util/scale/RechartsScale.js`).
  **`scalePoint().bandwidth()` is `0`**, so `'start'` and `'end'` land on the same pixel as the tick.

⇒ A `ReferenceArea x1="2022" x2="2025"` draws **tick-to-tick, not slot-to-slot**: it stops dead on the 2025
data point instead of covering 2025's share of the axis, leaving a half-slot of white between the band and
the first covered year. It will not look like "a band behind the uncovered years" without extra work.
Two further traps: `ifOverflow` defaults to **`'discard'`**, so an `x1`/`x2` that is not *exactly* a value
present in `chartData` yields **no rect and no error**; and the years in `chartData` are `number`s
(`base = { year: yearData.year }`, `TrendLineChart.tsx:52`) while `xLabels` stringifies them — pass the
wrong one and the band silently vanishes.

[...]

PR #2444, opened 2026-08-18
fix(metrics): scope and dedupe the historical-trends cancellation rate

### LINKED ISSUE #2434: bug(metrics): the cancellation rate renders ~2.4x too high, and its five-year line is four fabricated zeroes
The historical-trends cancellation rate renders roughly **2.4× the true figure**, and its five-year
line is structurally incapable of showing more than one year.

Found by a read-only audit during the Phase E triage campaign and re-verified line by line before
filing. Measured against the production snapshot (`pb_data/data-prod.db`, captured ~2026-08-09/14).

## 1. The `hasattr` guard is permanently dead, so the fallback ALWAYS runs

`api/services/historical_service.py:152-161`:

```python
async def _fetch_cancellation_count(self, year: int) -> int:
    if hasattr(self.repo, "fetch_cancellation_count"):
        result: int = await self.repo.fetch_cancellation_count(year)
        return result
    try:
        transitions = await self.repo.fetch_status_transitions(year, ["cancelled", "withdrawn", "dismissed"])
        return len(transitions)
    except Exception:
        return 0
```

`fetch_cancellation_count` (no leading underscore) is **defined by no repository**. A repo-wide grep
finds it in exactly four places: this `hasattr` check, its call on the next line, and two test mocks
(`tests/unit/api/services/test_historical_service.py:69,255`,
`tests/unit/api/test_velocity_service.py:1659-1662`). So in production the guard is always False and
the fallback is the only path — while **every test installs the phantom method on a spec'd mock**,
which is why the production path has zero coverage.

## 2. Two independent inflations, both structural

`get_historical_trends` (`api/services/historical_service.py:72-95`) builds the two sides differently:

```python
cancel_futures  = [self._fetch_cancellation_count(y) for y in years]   # no session_types argument
...
filtered   = self._filter_attendees(attendees, sessions, session_types, ...)
person_ids = {pid for a in filtered if (pid := getattr(a, "person_id", None)) is not None}
```

- **Scope mismatch.** `_fetch_cancellation_count` takes no `session_types` parameter — it *cannot* be
  filtered. The numerator spans all 13 session types; the denominator is session-type scoped.
- **Grain mismatch.** The fallback returns `len(transitions)` — person-session **rows** — against a
  **distinct-person** denominator. In 2026 that is 275 rows for 263 distinct persons.

## 3. Measured impact (2026, `DISPLAY_SESSION_TYPES`, distinct persons)

| | value |
|---|---|
| Numerator as rendered (unfiltered rows) | 275 |
| Numerator correctly scoped (distinct persons in display types) | 116 |
| Denominator (distinct persons, display types) | 1,451 |
| **Rendered rate** | **18.95%** |
| **Correct rate** | **7.99%** |
| **Inflation** | **2.37×** |

An independent measurement during the audit produced 18.63% vs 8.74% (2.13×) on a slightly different
denominator construction. The two disagree on the third significant figure and agree on the shape.

## 4. The worse defect: the trend line has one real year

`attendee_status_history` holds **695 rows, all `year = 2026`**. The trends view requests five years,
so the cancellation series renders as `0, 0, 0, 0, 18.95` — four fabricated zeroes and one wrong
number. That shape misleads more than the magnitude does.

## Suggested fix

Give the numerator the same scope and grain as the denominator: pass `session_types` through to the
cancellation count, count distinct `person_id` rather than rows, and either suppress or explicitly
mark years where `attendee_status_history` has no coverage rather than plotting them as zero. Delete
the dead `hasattr` branch, or define `fetch_cancellation_count` on the repositories and make the tests
exercise the real path — the current arrangement tests a method that does not exist.

⚠️ Whichever way it goes, the tests must stop installing `fetch_cancellation_count` on the mock, or
the production path stays untested.

- **Files:** `api/services/historical_service.py`, `api/repositories/metrics_sql_repository.py`, `tests/unit/api/services/test_historical_service.py`, `tests/unit/api/test_velocity_service.py`

adamflagg:
The cancellation rate on the historical-trends view rendered ~2.4x the true figure (measured on
prod: 18.95% rendered vs 7.99% correct, 2026, display session types), from two structural defects
in `HistoricalService._fetch_cancellation_count`:

1. A dead `hasattr(self.repo, "fetch_cancellation_count")` guard. No repository implementation
   (PocketBase-backed `MetricsRepository` or the direct-SQLite `MetricsSQLRepository`) defines that
   method, so the guard was always `False` in production and the fallback branch was the only path
   ever run — while every existing test installed the phantom method directly on its mock, so the
   real fallback had zero coverage.
2. That fallback counted ALL status-transition rows for the year with no `session_types` scoping
   (the enrolled denominator IS type-scoped), and counted rows instead of distinct persons (a person
   cancelling out of two sessions in one year was counted twice).

## Fix

Fetch the raw cancellation transitions per year, then run them through the exact same
`_filter_attendees` scoping (`session_types`/`session_cm_id`/`duration`) used for the
enrolled-person denominator, and dedupe by `person_id` before counting — giving the numerator the
same scope and grain as the denominator it's divided against. The dead `hasattr` branch is deleted
rather than given a real implementation, since no repository needs a pre-aggregated count method
once the scoped/deduped path exists.

## Tests

Three new regression tests in `tests/unit/api/services/test_historical_service.py` pin the exact
defects: session-type scoping, person-grain dedup, and that a repo defining the phantom
`fetch_cancellation_count` is now ignored entirely. Each was verified failing before the fix and
mutation-checked (breaking the corresponding production line turns it red again). The two test
files that previously installed the phantom method on a mock (`test_historical_service.py`,
`test_velocity_service.py`) now mock `fetch_status_transitions` instead, exercising the real
production path for the first time.

## Not done

The five-year trend line still renders 0% for years with no `attendee_status_history` coverage
(prod currently holds rows for one year only), rather than distinguishing "no data" from "zero
cancellations". Doing that correctly needs an API response contract change
(`api/schemas/metrics.py`) and a frontend rendering change
(`frontend/src/components/metrics/TrendLineChart.tsx`), both out of scope for this file-disjoint
fix — filed as #2443.

Closes #2434.

### DIFF
--- a/api/services/historical_service.py
+++ b/api/services/historical_service.py
@@ -69,28 +69,39 @@ class HistoricalService:
         if session_cm_id is not None:
             session_name = await self._get_session_name_for_filtering(session_cm_id, years, session_types)
 
-        # Fetch attendees, persons, sessions, and cancellation counts for all years in parallel
+        # Fetch attendees, persons, sessions, and cancellation transitions for all years in parallel
         attendee_futures = [self.repo.fetch_attendees(y) for y in years]
         person_futures = [self.repo.fetch_persons(y) for y in years]
         session_futures = [self.repo.fetch_sessions(y, session_types=session_types) for y in years]
-        cancel_futures = [self._fetch_cancellation_count(y) for y in years]
+        cancel_futures = [self._fetch_cancellation_transitions(y) for y in years]
 
         all_attendees = await asyncio.gather(*attendee_futures)
         all_persons = await asyncio.gather(*person_futures)
         all_sessions = await asyncio.gather(*session_futures)
-        all_cancel_counts = await asyncio.gather(*cancel_futures)
+        all_cancel_transitions = await asyncio.gather(*cancel_futures)
 
         # Compute metrics for each year
         year_metrics_list: list[YearMetrics] = []
 
-        for year, attendees, persons, sessions, cancel_count in zip(
-            years, all_attendees, all_persons, all_sessions, all_cancel_counts, strict=True
+        for year, attendees, persons, sessions, cancel_transitions in zip(
+            years, all_attendees, all_persons, all_sessions, all_cancel_transitions, strict=True
         ):
             # Filter attendees by session type and/or session name
             filtered = self._filter_attendees(attendees, sessions, session_types, session_cm_id, session_name, duration)
 
             # Deduplicate by person_id
             person_ids = {pid for a in filtered if (pid := getattr(a, "person_id", None)) is not None}
+
+            # Cancellations use the SAME scope (session_types/session_cm_id/duration) and the
+            # same distinct-person grain as the enrolled denominator above -- see #2434, where
+            # an unscoped, row-counted numerator inflated the rendered rate ~2.4x.
+            filtered_cancellations = self._filter_attendees(
+                cancel_transitions, sessions, session_types, session_cm_id, session_name, duration
+            )
+            cancelled_person_ids = {
+                pid for t in filtered_cancellations if (pid := getattr(t, "person_id", None)) is not None
+            }
+            cancel_count = len(cancelled_person_ids)
 
             year_metric = self._compute_year_metrics(year, person_ids, persons, cancel_count)
             year_metrics_list.append(year_metric)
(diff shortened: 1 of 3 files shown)


adamflagg:
### Independent review (triage-attack Phase F Part B)

**Scan (floor 60):** Manual review at the code-review skill's checklist (CLAUDE.md compliance, bug scan, git/history context, prior-PR comments, in-file comment compliance) — no subagent tree spawned, given the diff's size (3 files, 269 lines). No findings survived at the floor-60 bar. Clean.

Verified, not just read:
- Confirmed both repository backends (`MetricsRepository` PocketBase-backed and `MetricsSQLRepository` direct-SQLite) attach `expand["session"]` to `fetch_status_transitions` rows, which `_filter_attendees`/`get_session_from_expand` requires — the new scoping path works against both backends, not just the mocked one.
- Confirmed `fetch_cancellation_count` is fully gone from production code (`grep` across the tree) and survives only as a docstring reference and the regression test proving it's ignored.
- Ran `mypy` and `ruff` on the three changed files: clean.
- Ran the full historical/velocity unit suite (177 tests): all pass.
- Mutation-checked both new regression assertions directly against the production lines they claim to pin:
  - `total_cancelled` dedup: reverting `len(cancelled_person_ids)` → `len(filtered_cancellations)` (row count) turns `test_cancellation_count_dedupes_by_person_not_rows` red (`assert 2 == 1`).
  - Session-type scoping: skipping the `_filter_attendees` call on `cancel_transitions` turns `test_cancellation_count_scoped_to_session_types` red (`assert 2 == 1`).
  - Both mutants reverted; tree is clean.

**Refuted:** none — no candidate findings reached the point of needing refutation.

**Filed as out of scope:** none new. #2443 (the trend-line "four fabricated zeroes" defect) was already correctly filed by the builder before this review — verified its body accurately describes the remaining `api/schemas/metrics.py` + `TrendLineChart.tsx` contract change and stays file-disjoint from this PR, so left as-is.

**In-branch fixes:** none needed.

CodeRabbit: not requested (quota reserved for the wave's highest-blast-radius PR per campaign policy).

Issue #2358, opened 2026-08-14
bug(sync): a widened bunk_assignments row still resolves to the FIRST session of a multi-session plan


adamflagg:
## ⚠️ CORRECTED 2026-08-16 — phase D triage. Premise HOLDS; both consumers are gone.

Verified against `main` @ `a946565a`.

**The code is still wrong and stays wrong. What changed is that nothing can reach it.** #2350 widened
`bunk_assignments` grain to `(person, session, bunk, year)`, and all three consumers now filter
`session_type` to allowlists that exclude `family` and `quest` — `social_graph_builder.py` among them.

So the resolve-to-first-session defect is **unreachable dead behaviour**, not a live bug.

**⏸ Still open — needs the owner's yes.** Two defensible outcomes and they are different diffs:

- **Close it**, on the grounds that no caller can reach the wrong path.
- **Delete the dead code in the same breath**, so the next reader does not re-derive that it is unreachable.

Was offered for closure 2026-08-16 and **not** selected, so it stays open pending that call. A PR would close
#2358 and nothing else.

---

## What is left after kindred#2259 / kindred#2264

PR #2350 widened the `bunk_assignments` write key to `(person, session, bunk, year)`, so a person holding two assignments under one multi-session bunk plan no longer loses a row. **It did not recover the session those rows belong to.** `findMatchingSession` intersects the person's enrolments with the plan's sessions and returns the **first** hit, so both surviving rows carry the *first* weekend.

Concretely: a household member booked into family weekends 1 and 2 now gets two rows both naming weekend 1, and still nothing on weekend 2. Before the widening, one row was lost; after it, the recovered row is present but attributed to the wrong session.

## Measured

Rows sitting on an ambiguous `(plan, bunk)` key, and whether the person's own enrolment resolves it:

| year | on an ambiguous key | `attendees` → exactly 1 | genuinely ambiguous |
|---|---:|---:|---:|
| 2026 | 280 | 269 | **11** |
| 2025 | 740 | 731 | **9** |
| 2024 | 790 | 777 | **13** |
| 2023 | 567 | 547 | **20** |

**The data is 98.4% sufficient already** — 3,303 of 3,358 rows resolve to exactly one session from `attendees` (`status_id = 2`), and **zero** rows lack an enrolment. The code already does this intersection; it just takes the first hit instead of noticing it received two.

All 55 residual rows are Family Camp on a single plan:

```
2026  Day Groups  family  11        2018  Session III  family  1
[...]

PR #2445, opened 2026-08-18
fix(graph): scope prior-year bunk history to the session it belonged to

### LINKED ISSUE #2425: bug(bunking): prior-year bunkmate graph has no session-type filter — 954 false BUNKMATE edges from Family Camp day groups
## Summary

`_add_historical_bunking_relationships` has **no `session_type` predicate**, so Family Camp **day
groups are scored as summer bunkmate history**. It groups by `(year, bunk PB id)`, so every
prior-year assignment counts regardless of what kind of session it belonged to.

**Re-measured against `pocketbase/pb_data/data-prod.db`** on 2026-08-17 (prod snapshot, dated
~2026-08-09..08-14) by replaying this function's own logic over the ten 2026 `main`/`embedded`/`ag`
session graphs. Grain: unique person-pairs, summed over the 10 graphs.

| | edges |
|---|---:|
| `BUNKMATE` edges the current code creates | 4,935 |
| of those, pairs a prior-year **Family Camp day group** contributes | 1,060 |
| **pairs that exist ONLY because day groups are included** (day-group pairs minus pairs also derivable from a real cabin) | **954** |

The honest numerator is **954 false edges**, not 1,060 — 106 of the 1,060 day-group pairs are also
corroborated by a genuine shared cabin and survive the fix. They touch **233 distinct 2026 campers**
of the 1,134 in these graphs and come from **23** prior-year day groups (2023, 2024, 2025).

On weight: `BUNKMATE`'s base weight is 2.0, second only to `SIBLING` (3.0) — but the weight actually
applied to a *historical* edge is decayed by recency (`bunking/sync/bunk_request_processor/social/social_graph.py:316-317`,
`2.0 / (1 + years_ago * 0.2)`), so a false edge lands at **1.67** (2025), **1.43** (2024) or
**1.25** (2023): at or below `CLASSMATE` (1.5), still well above `BUNK_REQUEST` (1.0).

This costs **summer**, not Family Camp. The direction of the defect is Family Camp data leaking
*into* summer.

## Why day groups are the wrong shape for this signal

A day group is not a cabin. Measured over every prior year this function reads (2017-2025); grain =
distinct persons per `(year, bunk)` group, which is exactly the key the code uses:

| prior-year group | groups | median members | largest |
|---|---:|---:|---:|
| Family Camp day group | 33 | 53 | **194** |
| `main`/`embedded`/`ag` bunk | 248 | 41 | 78 |

A *real* cabin — one bunk in one session — holds a median of **14** (max 16, over the 87
`(bunk, session)` pairs used in 2025). Unit names are deliberately omitted here; the sizes are the
point.

The pairwise blow-up is real, but smaller than the raw group sizes suggest, and stating it correctly
matters because it sizes the fix: the query at `bunking/sync/bunk_request_processor/social/social_graph.py:271-275` is already scoped to the
persons in *this* session's graph, so only the day-group members who are also in that graph pair up.
The largest single day group contributes **18** members to one 2026 graph and therefore **153**
pairs; 23 groups together produce the 954. (The original "a 194-member group injects ten thousand
candidate pairs before dedup" reading was wrong — it ignored that scoping.)

## Why it matters

These edges feed `social_distance`, `shared_connections` and `in_ego_network`, which score
**name-resolution confidence** for bunk-request disambiguation. A false high-weight edge does not
merely add noise: it makes the resolver *more confident about the wrong person*, because two children
who merely shared a day group two summers ago now look like former bunkmates.

## Location

`bunking/sync/bunk_request_processor/social/social_graph.py:255-332` — re-verified at `main` @
`1797f9bd`. The two lines that carry the defect:

- **`:271`** — `filter_str = f"year < {self.year} && ({person_filter})"` — no session-type predicate.
- **`:293`** — `key = (year, bunk_id)` — the grouping key, which has no session in it either. That is
  a second, larger defect; see below.

## The fix already exists in a sibling file

`bunking/graph/social_graph_builder.py:347-384` carries the session-type predicate for its own
equivalent query, added by kindred#2350 / kindred#2259 / kindred#2358, with a 30-line comment
explaining precisely this failure. **The fix was never swept sideways into this file.** Apply the
same predicate, from the same constant (see Siblings).

TDD anchor: `tests/unit/sync/bunk_request_processor/social/test_social_graph.py`.

## A SECOND defect in the same function, larger than the first — do not ship the filter alone

The grouping key at `:293` is `(year, bunk_id)` with **no session**. A bunk is a building, reused by
successive sessions: in 2025, **42 of the 57** bunks carrying an assignment were used by **more than
one session**. The key therefore merges children who occupied the same cabin in different weeks and
never met.

Replaying the function over the ten 2026 graphs (grain: unique person-pairs, summed):

| grouping | `BUNKMATE` edges |
|---|---:|
| today — `(year, bunk)`, no type filter | 4,935 |
| + session-type filter only (what this issue originally asked for) | 3,981 |
| + key `(year, bunk, session)` | **1,993** |

The type filter removes 954; the session key removes a further **1,988**. **2,942 of 4,935 (59.6%)**
of today's `BUNKMATE` edges are false. Fixing only the type filter leaves the larger half in place.

Scoping to the exact session is safe here, measured: keying on the *parent* session instead (so an
embedded session merges with its containing main session) adds only 76 pairs, and every one of those
76 comes from two embedded sessions inside the same main session **whose date ranges do not overlap**
— children who were never at camp at the same time. A date-overlap pairing rule returns exactly the
same 1,993.

## Reproduction note — the earlier passes' disagreement is resolved

A faithful replay (nodes = enrolled 2026 attendees with a truthy `gender`, plus anyone pulled in by a
sibling or classmate edge, since `add_edge` creates nodes implicitly) reproduces **4,935** total
edges exactly. The suspected cause of the earlier 4,935-vs-5,115 disagreement was wrong: **every**
enrolled 2026 attendee row in the snapshot carries a non-empty `gender` (0 empty of 2,293), so the
`add_node` gender guard changes no node set this season. Quote 954 / 4,935 = 19.3% for the type
defect, and 2,942 / 4,935 = 59.6% for both defects together.

## What is NOT wrong

- Nothing here excludes or loses Family Camp data.
- `VALID_BUNKING_SESSION_TYPES` in `session_repository.py:19` is **correct** and is not this bug:
  measured, all 1,134 distinct 2026 `original_bunk_requests` requesters hold a `main`/`embedded`/`ag`
  attendee row, so it loses zero requesters.

## Siblings — these three ship as ONE PR

kindred#2425, kindred#2426 and kindred#2427 are the same root cause at three sites and are intended
to land as a **single PR sharing one constant**. Do not split them: a fix that was never swept
sideways is exactly what produced all three, and splitting them re-creates that failure mode.

The one constant is `LAST_YEAR_HISTORY_SESSION_TYPES`
(`bunking/graph/social_graph_builder.py:54`, `("main", "taste", "embedded", "ag")`) — the fourth
site, already fixed. Import it, or move it somewhere all four call sites can import from; **do not
add a fourth copy**. Do not reuse `VALID_BUNKING_SESSION_TYPES`
(`bunking/sync/bunk_request_processor/data/repositories/session_repository.py:19`) — the comment at
`bunking/graph/social_graph_builder.py:48-53` says the two sets deliberately mean different things. There are
already three near-copies of this idea in the tree; the third is the unused `VALID_SESSION_TYPES` at
`bunking/sync/bunk_request_processor/core/constants.py:8`. Note that `"taste"` appears on **zero**
`camp_sessions` rows in the snapshot (2017-2026), so the choice between the two sets changes nothing
measurable today — it is a semantics decision, not a data one.

Relative size, all on the same snapshot: this one is the largest (954 false edges from the type
defect, plus 1,988 more from the grouping key), then kindred#2426 (147 of 782 requesters with a
prior-year row are handed a day group, and 651 of 728 summer lookups return an inflated peer set),
then kindred#2427 (33% of the rows the cache loads).

## Provenance

Found by a read-only cross-surface sync audit, 2026-08-17. Every figure in this body was
**re-measured from scratch** on 2026-08-17 against `pocketbase/pb_data/data-prod.db`, with the tree
at `main` @ `1797f9bd`, by replaying the function's own logic. Where the re-measurement disagreed
with the original pass — 954 rather than 1,060, the pairwise mechanism, the applied edge weight, and
the grouping-key defect that was missed entirely — this body carries the re-measured figure.

### LINKED ISSUE #2426: bug(bunking): find_prior_year_bunkmates has no session filter and no sort, so a Family Camp day group can become a summer prior_bunk
## Summary

`find_prior_year_bunkmates` has **no `session_type` filter and no sort**, then takes
`assignments[0]`. For a requester who holds both a prior-year summer cabin **and** a prior-year
Family Camp day group, the day group can win — after which **the entire day group is returned as
`prior_bunk`**.

**Re-measured against `pocketbase/pb_data/data-prod.db`** on 2026-08-17 (prod snapshot, dated
~2026-08-09..08-14). Grain: distinct 2026 requesters in `original_bunk_requests`; prior year = 2025.

| population | count |
|---|---:|
| distinct 2026 requesters | 1,134 |
| …holding **any** 2025 `bunk_assignments` row — the only ones this function can affect | 782 |
| …holding **both** a 2025 summer row and a 2025 day-group row → **order-dependent** | 93 |
| …holding **only** a 2025 day-group row → **deterministically wrong**, every run, no order involved | 54 |
| **combined, handed a day group as `prior_bunk`** | **147 of 782 = 18.8%** (13.0% of all 1,134) |

The 54 are the half the original body missed, and they are the worse half: for them there is no
"could go either way" — the function returns a day group *every* time.

This costs **summer**, not Family Camp.

## THREE defects, separable, all real — the third was omitted and is the biggest

1. **No session-type filter on the first query** (`:474-479`) — a Family Camp day group is eligible
   to be returned as a summer `prior_bunk` at all. Population: the 147 above.
2. **No sort, then `assignments[0]`** (`:488`) — among eligible rows the pick is arbitrary;
   PocketBase's `get_full_list` has no inherent order, so the winner can change between runs with no
   code change. Keep the sort — it is cheap and it is the right habit — but size it honestly:
   **zero** of the 1,134 requesters hold more than one 2025 row of an eligible session type, so the
   population for "non-determinism between two prior-year summer rows" is **0** this season. The
   original body's claim that fixing (1) leaves that non-determinism behind is wrong. What fixing
   (1) alone leaves behind is (3).
3. **No session predicate on the *second* query** (`:502-504`), which fetches everyone in that bunk
   in that year: `filter: bunk = "{bunk_id}" && year = {previous_year}`. A bunk is a building,
   reused by successive sessions — in 2025, **42 of the 57** bunks carrying an assignment were used
   by **more than one session** — so the returned "bunkmates" are largely children the requester
   never met.

Measured effect of (3). Grain: one row per (requester, eligible 2025 assignment), n=728, with
"returning" defined exactly as the code defines it (`bulk_get_sessions_for_persons`: any 2026
`attendees` row in a `main`/`embedded`/`ag` session, any status — a pool of 1,280 people):

| `returning_count` | median | mean | max |
|---|---:|---:|---:|
| today — bunk + year | 29 | 26.3 | 42 |
| scoped to the requester's own 2025 session | **8** | 7.6 | 11 |

**651 of those 728 lookups (89.4%) return an inflated peer set.** For the 147 requesters whose
`prior_bunk` is a day group, `total_in_bunk` runs to a median of 110 (max 178) with a median of 35
returning (max 51) — against a real 2025 cabin of 14 (median over the 87 `(bunk, session)` pairs,
max 16).

## Why it matters

Three live call sites — `bunking/sync/bunk_request_processor/services/context_builder.py:129`, `bunking/sync/bunk_request_processor/services/context_builder.py:362` and
`bunking/sync/bunk_request_processor/services/phase2_resolution_service.py:871`, all verified at `main` @ `1797f9bd` — feed **the AI
disambiguator and continuity scoring**. `prior_bunk` is used as evidence that two children have a
shared history; sourcing it from a day group makes that evidence meaningless in exactly the cases
where the resolver leans on it hardest. For a 2026 run the prior year is 2025, whose largest day
group holds **179** children; the 194 quoted in kindred#2425 is the 2024 group.

## Location

`bunking/sync/bunk_request_processor/data/repositories/attendee_repository.py:454-535` —
re-verified at `main` @ `1797f9bd`. First query `:474-479`; the arbitrary pick `:488`; second query
`:502-504`.

## The fix already exists in a sibling file

`bunking/graph/social_graph_builder.py:347-384` documents and fixes **verbatim this failure** for its
own query (kindred#2350 / kindred#2259 / kindred#2358) — its comment at `:348-379` describes the
exact "family row wins the arbitrary pick" case. It was never swept sideways.

## A caveat on one number

An earlier pass reported "family wins for 69 of the 93". That used SQLite `rowid` as a proxy for the
order PocketBase actually returns and **is not reliable** — treat 93 as the population at risk and
the split as unknown. The 54 who hold *only* a day-group row need no such caveat: for them the
outcome does not depend on order at all.

## What is NOT wrong

- No Family Camp data is excluded or lost by this.
- `VALID_BUNKING_SESSION_TYPES` (`session_repository.py:19`) is correct and is not this bug.

## Siblings — these three ship as ONE PR

kindred#2425, kindred#2426 and kindred#2427 are the same root cause at three sites and are intended
to land as a **single PR sharing one constant**. Do not split them: a fix that was never swept
sideways is exactly what produced all three, and splitting them re-creates that failure mode.

The one constant is `LAST_YEAR_HISTORY_SESSION_TYPES`
(`bunking/graph/social_graph_builder.py:54`, `("main", "taste", "embedded", "ag")`) — the fourth
site, already fixed. Import it, or move it somewhere all four call sites can import from; **do not
add a fourth copy**, and do not reuse `VALID_BUNKING_SESSION_TYPES`
(`bunking/sync/bunk_request_processor/data/repositories/session_repository.py:19`),
which the comment at `bunking/graph/social_graph_builder.py:48-53` says deliberately means something else. A third
near-copy already exists, unused, at `bunking/sync/bunk_request_processor/core/constants.py:8`.
`"taste"` appears on **zero** `camp_sessions` rows in the snapshot (2017-2026), so the two sets
select identically on today's data — the difference is semantic, not measurable.

Relative size on the same snapshot: kindred#2425 is the largest (954 false `BUNKMATE` edges from the
type defect, plus 1,988 more from its grouping key), this one is second (147 of 782, plus 651 of 728
inflated peer sets), kindred#2427 third (33% of the rows the cache loads).

## Provenance

Read-only cross-surface sync audit, 2026-08-17. Every figure in this body was **re-measured from
scratch** on 2026-08-17 against `pocketbase/pb_data/data-prod.db`, with the tree at `main` @
`1797f9bd`. Two of the original claims were inverted by that re-measurement: the 54
deterministically-wrong requesters were missing entirely, and the "non-determinism between two
summer rows" it warned about has a measured population of zero this season.

### LINKED ISSUE #2427: bug(bunking): temporal name cache treats Family Camp day groups as bunks — 31% of 2025 rows — and its "most recent" comment is false
## Summary

`_load_historical_bunking` has **no `session_type` filter**, so Family Camp day groups populate the
temporal name cache as if they were cabins. It keeps whichever row the unordered `get_full_list`
returned first, while **its own comment claims "most recent assignment per year"** — the comment is
false, and that is worth fixing alongside the filter, because it is what makes the behaviour look
intentional to the next reader.

**Re-measured against `pocketbase/pb_data/data-prod.db`** on 2026-08-17 (prod snapshot, dated
~2026-08-09..08-14). Grain: rows in `bunk_assignments`, all session types, no person filter.

| year | day-group rows | total `bunk_assignments` rows | share |
|---|---:|---:|---:|
| 2025 | **692** | 2,203 | **31.4%** |
| 2024 | **790** | **2,289** | **34.5%** |

The 2024 denominator in the original body was wrong: it is **2,289**, not 2,272, and the share is
34.5%, not 34.8%. Every 2024 row joins cleanly to a `camp_sessions` row, so no scoping produces
2,272.

The figure that actually governs a 2026 run is the two-year window this loader reads
(`year >= self.year - 2 && year < self.year`, `:210-216`): **1,482 of 4,492 rows = 33.0%** are
Family Camp day-group rows. So roughly a third of what this cache treats as prior-year bunk history
is not a cabin at all.

This costs **summer**, not Family Camp.

## Why it matters

The original body said this cache feeds the `was_bunkmate` signal. It does not: `was_bunkmate` is
set at `bunking/sync/bunk_request_processor/services/context_builder.py:387` from `find_prior_year_bunkmates`, which is kindred#2426.
`_historical_bunking` has exactly two consumers, and they pull in opposite directions:

- **`:316-324`** — presence in `_historical_bunking` is what gives a person a `"historical"` bucket
  in the name→person index read at `:451-452`. Live today. Day-group rows put people in that index
  who have no prior cabin at all: **1,200** of the window's `(person, year)` keys are day-group-only,
  covering **922** distinct people.
- **`verify_bunk_together` (`:496-545`)** — the bunk-name comparison, which is inert today for the
  reason in fix (3) below.

**So where the filter goes is a decision, not a detail.** Filtering at the load (`:211-216`) also
strips those 922 people out of the historical name index — that may well be right, but it is a
behaviour change beyond "stop treating day groups as cabins", so make it deliberately. Filtering at
the consumer, or storing the session type on the record at `:239-247` and letting each consumer
decide, keeps the two concerns apart.

Its window is two years and its consumers are weaker signals than the social graph's, which is why
this is ranked **below** kindred#2425 and kindred#2426 — but it is the same defect, and fixing the
other two while leaving this one keeps a third of the cache wrong.

## Location

`bunking/sync/bunk_request_processor/data/cache/temporal_name_cache.py:192-257` — re-verified at
`main` @ `1797f9bd`. Query `:211-216`; the false "most recent" comment `:236-237`; the keep-first at
`:238`; the stored record `:239-247`.

TDD anchor: `tests/unit/sync/bunk_request_processor/data/cache/test_temporal_name_cache.py`.

## Three things to fix, not two — and read (3) before deciding (1)

1. Add the `session_type` predicate (the same one `bunking/graph/social_graph_builder.py:347-384`
   already uses — kindred#2350 / kindred#2259 / kindred#2358), from the shared constant named under
   Siblings, at the placement chosen above.
2. Either make "most recent" true by sorting, or correct the comment to say the pick is arbitrary.
   Leaving a false comment in place is how this survived three prior passes. Measured size of what
   the pick decides: of the **3,826** distinct `(person, year)` keys in the 2024+2025 window,
   **442** carry more than one `bunk_assignments` row, so the keep-first at `:238` actually chooses
   something. **281** of those mix a day-group row with an eligible one; **131** hold more than one
   *eligible* row, so the tie-break still matters after the filter lands.
3. **`verify_bunk_together` cannot read what this loader writes.** The loader stores the bunk under
   the key `"bunk"` (`:246`); `verify_bunk_together` reads `"bunk_name"` (`:518` and `:537`), gets
   `""`, and takes the `if not requester_bunk: return False, ""` branch at `:521-522` for **every**
   person. Its only caller, `bunking/sync/bunk_request_processor/services/historical_verification_service.py:147`, therefore marks every
   historical group unverified today. The unit tests pass because they hand-build
   `_historical_bunking` with the `"bunk_name"` key the producer never writes — see
   `tests/unit/sync/bunk_request_processor/test_historical_bunking_verification.py:42`.

   State this where the next reader will see it, because it changes what (1) buys: today the
   day-group rows produce no false "verified bunked together" results, because that consumer is
   dead. Fix the key without the filter and they immediately do.

## What is NOT wrong

- No Family Camp data is excluded or lost.
- `VALID_BUNKING_SESSION_TYPES` (`session_repository.py:19`) is correct and is not this bug.

## Siblings — these three ship as ONE PR

kindred#2425, kindred#2426 and kindred#2427 are the same root cause at three sites and are intended
to land as a **single PR sharing one constant**. Do not split them: a fix that was never swept
sideways is exactly what produced all three, and splitting them re-creates that failure mode.

The one constant is `LAST_YEAR_HISTORY_SESSION_TYPES`
(`bunking/graph/social_graph_builder.py:54`, `("main", "taste", "embedded", "ag")`) — the fourth
site, already fixed. Import it, or move it somewhere all four call sites can import from; **do not
add a fourth copy**, and do not reuse `VALID_BUNKING_SESSION_TYPES`
(`bunking/sync/bunk_request_processor/data/repositories/session_repository.py:19`),
which the comment at `bunking/graph/social_graph_builder.py:48-53` says deliberately means something else. A third
near-copy already exists, unused, at `bunking/sync/bunk_request_processor/core/constants.py:8`.
`"taste"` appears on **zero** `camp_sessions` rows in the snapshot (2017-2026), so the two sets
select identically on today's data — the difference is semantic, not measurable.

Sizes on the same snapshot: kindred#2425 (`_add_historical_bunking_relationships`) is the largest — **954** false
`BUNKMATE` edges from the missing type filter, plus **1,988** more from a grouping key that has no
session in it; then kindred#2426 (`find_prior_year_bunkmates`) — 93 of 782 requesters with a
prior-year row are order-dependent and a further 54 are deterministically wrong; then this one.

## Provenance

Read-only cross-surface sync audit, 2026-08-17. Every figure in this body was **re-measured from
scratch** on 2026-08-17 against `pocketbase/pb_data/data-prod.db`, with the tree at `main` @
`1797f9bd`. The re-measurement corrected the 2024 denominator (2,289, not 2,272), replaced the
`was_bunkmate` attribution with this cache's two real consumers, and found the `"bunk"` /
`"bunk_name"` key mismatch that makes one of them inert.

adamflagg:
Three places read prior-year `bunk_assignments` to decide who a camper bunked with
last year. None of them filtered on session type, so Family Camp **day groups** —
which hold a median of 53 people and are not cabins — were scored as summer bunkmate
history. Two of them also keyed on a bunk **without its session**, and a bunk is a
building: 42 of the 57 bunks carrying a 2025 assignment served more than one session,
so children who occupied the same cabin in different weeks were treated as bunkmates.

The identical fix has been in `bunking/graph/social_graph_builder.py` since #2350 /
#2259 / #2358, with a 30-line comment describing exactly this failure. It was never
swept sideways — which is why the same defect existed at three more sites. All three
now use that file's `LAST_YEAR_HISTORY_SESSION_TYPES`; no fourth copy of the set was
added.

## What was wrong, measured on the prod snapshot

| site | defect | size |
|---|---|---:|
| `social/social_graph.py` | no type predicate; grouping key `(year, bunk)` | **2,942 of 4,935** BUNKMATE edges false (59.6%) — 954 from day groups, 1,988 from the missing session |
| `data/repositories/attendee_repository.py` | no type predicate, no sort, `assignments[0]`; second query on `bunk` + `year` only | **147 of 782** requesters handed a day group as `prior_bunk`; **651 of 728** peer lookups inflated (median 29 vs a session-scoped 8) |
| `data/cache/temporal_name_cache.py` | no type on the record; producer/consumer key mismatch | **33%** of the rows the two-year window loads are day groups |

A false BUNKMATE edge is not merely noise: these edges feed `social_distance`,
`shared_connections` and `in_ego_network`, which score name-resolution confidence for
bunk-request disambiguation. The failure mode is the resolver becoming *more* confident
about the wrong person.

## What changed

- **`social_graph.py`** — the prior-year query carries the session-type predicate, and
  the grouping key is now `(year, bunk, session)`. A row whose session cannot be
  resolved is dropped rather than merged into the building.
- **`attendee_repository.py`** — `find_prior_year_bunkmates` filters the first query by
  session type and sorts by record id (PocketBase's `get_full_list` has no inherent
  order, so the pick could otherwise change between runs with no code change), then
  scopes the bunkmate query to the requester's own session.
- **`temporal_name_cache.py`** — each record now carries its `session_type`, the load is
  sorted for a deterministic per-year pick, and a cabin-bearing row beats a day-group
  row for the same person and year. The false `# Only store most recent assignment per
  year` comment is corrected: the rows arrive unordered and nothing sorted them.
- **`temporal_name_cache.py`, behaviour revival** — the loader wrote the bunk under
  `"bunk"` while `verify_bunk_together` read `"bunk_name"`, so that consumer returned
  `(False, "")` for *everyone* and historical group verification was inert. The producer
  now writes the key the consumer reads. This is safe only because the type check lands
  in the same change: fixing the key alone would have started producing false "verified
  bunked together" results from shared day groups.

## Deliberately NOT done

- **The temporal cache load is still unfiltered, on purpose.** Presence in
  `_historical_bunking` is what gives a person a `"historical"` bucket in the name index,
  and a Family Camp attendee is a perfectly real prior-year person to resolve a name
  against. Filtering at the load would have silently dropped ~922 people from name
  resolution. The type is stored on the record and the bunk-comparison consumer filters;
  the two concerns stay apart.
- **`LAST_YEAR_HISTORY_SESSION_TYPES` was not moved to a shared leaf module.** A
  module-level import of it from inside `bunk_request_processor` is a genuine circular
  import (`bunking.graph` → `bunking.satisfaction` → `bunk_request_processor.core.models`
  → the package's own `__init__`); it fails with `cannot import name 'BucketCount' from
  partially initialized module`. Moving the constant would mean editing
  `bunking/graph/social_graph_builder.py`, which is outside this PR's file scope, so the
  three new call sites import it at call time with a `# noqa: PLC0415` and a comment.
  Consolidating it into a leaf module is worth a follow-up.
- The unused third near-copy, `VALID_SESSION_TYPES` in `core/constants.py`, is untouched.
- The three consumers (`context_builder.py`, `phase2_resolution_service.py`,
  `historical_verification_service.py`) are untouched — the returned shapes are unchanged.
- No target-side session-type check in `verify_bunk_together`: the comparison already
  requires the target to be in the *same session* as the requester, and a session has one
  type, so it would be dead code.

## Tests

Ten new tests, written and run failing first. The pre-existing fixtures that hand-built
`_historical_bunking` with keys the producer never wrote — the thing that let the inert
consumer look healthy for three passes — were corrected to the real record shape.

Closes #2425.
Closes #2426.
Closes #2427.

### DIFF
--- a/bunking/sync/bunk_request_processor/data/cache/temporal_name_cache.py
+++ b/bunking/sync/bunk_request_processor/data/cache/temporal_name_cache.py
@@ -194,8 +194,23 @@ class TemporalNameCache:
 
         Unlike monolith which queries non-existent 'historical_bunking' table,
         we use the modern schema: bunk_assignments with year < current_year.
+
+        The query is deliberately NOT filtered by session type. This dict has
+        two consumers that want opposite things (#2427): `_build_name_index`
+        uses mere PRESENCE here to give a person a "historical" bucket in the
+        name index, and a Family Camp attendee is a perfectly real prior-year
+        person to resolve a name against — filtering at the load would silently
+        drop ~922 people from name resolution. `verify_bunk_together` instead
+        wants cabins only. So each record carries its `session_type` and the
+        bunk-comparison consumer does the filtering.
         """
         logger.info("Loading historical bunking data...")
+
+        # Imported at call time, not module scope: `bunking.graph` imports
+        # `bunking.satisfaction`, which imports this package's `core.models`, so
+        # a top-level import here is a genuine cycle. Importing at call time
+        # keeps ONE definition of the eligible-session-type set.
+        from bunking.graph.social_graph_builder import LAST_YEAR_HISTORY_SESSION_TYPES  # noqa: PLC0415
 
         try:
             # Get all sessions for name lookup
@@ -208,10 +223,16 @@ class TemporalNameCache:
             # This limits memory usage while still providing useful historical context
             # For bunk request processing, 2 years is sufficient for name disambiguation
             min_year = self.year - 2
+            # `sort: "id"` makes the per-(person, year) pick below deterministic.
+            # PocketBase's get_full_list has no inherent order, so without it the
+            # kept row could change between runs with no code change — 442 of the
+            # 3,826 (person, year) keys in this two-year window carry more than
+            # one row, so the pick decides something real.
             assignments = self.pb.collection("bunk_assignments").get_full_list(
                 query_params={
                     "filter": f"year >= {min_year} && year < {self.year}",
                     "expand": "person,bunk,session",
+                    "sort": "id",
                 }
             )
 
@@ -227,15 +248,29 @@ class TemporalNameCache:
                 person_cm_id = getattr(person_data, "cm_id", None)
                 year = getattr(assignment, "year", None)
                 session_cm_id = getattr(session_data, "cm_id", None) if session_data else None
+                session_type = (getattr(session_data, "session_type", "") if session_data else "") or ""
                 bunk_name = getattr(bunk_data, "name", "") if bunk_data else ""
 
                 if person_cm_id and year:
                     if person_cm_id not in self._historical_bunking:
                         self._historical_bunking[person_cm_id] = {}
 
-                    # Only store most recent assignment per year
-                    # (in case of duplicates)
-                    if year not in self._historical_bunking[person_cm_id]:
+                    # One record per (person, year). The old comment here said
+                    # "most recent assignment per year" — that was never true;
+                    # the rows arrive unordered and this kept whichever came
+                    # first. It is now the first row in record-id order, EXCEPT
+                    # that a cabin-bearing session beats a non-bunking one: a
+                    # person who held both a summer cabin and a Family Camp day
+                    # group last year must be remembered by the cabin, or the
+                    # bunk-comparison consumer loses their real history (#2427).
+                    existing = self._historical_bunking[person_cm_id].get(year)
+                    is_eligible = session_type in LAST_YEAR_HISTORY_SESSION_TYPES
+                    replaces_existing = (
+                        existing is not None
+                        and is_eligible
+                        and existing.get("session_type", "") not in LAST_YEAR_HISTORY_SESSION_TYPES
+                    )
+                    if existing is None or replaces_existing:
                         self._historical_bunking[person_cm_id][year] = {
                             "session_cm_id": session_cm_id,
                             "session_name": session_lookup.get(session_cm_id, f"Session {session_cm_id}")
@@ -243,7 +278,13 @@ class TemporalNameCache:
                             else "",
                             "parent_session_id": session_cm_id,  # Simplified - no parent tracking for historical
                             "parent_session_name": session_lookup.get(session_cm_id, "") if session_cm_id else "",
-                            "bunk": bunk_name,
+                            # `bunk_name` is the key `verify_bunk_together`
+                            # reads. It used to be written as "bunk", so that
+                            # consumer read "" for everybody and returned
+                            # (False, "") every time — the whole historical
+                            # verification path was inert (#2427).
+                            "bunk_name": bunk_name,
+                            "session_type": session_type,
                         }
 
             total_historical = sum(len(years) for years in self._historical_bunking.values())
@@ -503,11 +544,18 @@ class TemporalNameCache:
             target_cm_ids: List of target CM IDs to check
             year: Year to check
 
+        Only sessions that actually put children in a cabin count. A shared
+        Family Camp day group is not evidence that two children bunked
+        together — day groups run to a median of 53 people (#2427).
+
         Returns:
             Tuple of (were_together, bunk_name):
             - (True, bunk_name) if all were in same bunk
             - (False, "") if not all together or data missing
         """
+        # Imported at call time — see `_load_historical_bunking` for why.
+        from bunking.graph.social_graph_builder import LAST_YEAR_HISTORY_SESSION_TYPES  # noqa: PLC0415
+
         # Get requester's historical bunking for the year
         requester_data = self._historical_bunking.get(requester_cm_id, {})
         requester_year_data = requester_data.get(year, {})
@@ -519,6 +567,9 @@ class TemporalNameCache:
         requester_session = requester_year_data.get("session_cm_id")
 
         if not requester_bunk:
+            return False, ""
+
+        if requester_year_data.get("session_type", "") not in LAST_YEAR_HISTORY_SESSION_TYPES:
             return False, ""
 
         # Empty target list is vacuously true (all zero targets were in same bunk)
@@ -537,6 +588,10 @@ class TemporalNameCache:
             target_bunk = target_year_data.get("bunk_name", "")
             target_session = target_year_data.get("session_cm_id")
 
+            # No session-type check on the target: the comparison below already
+            # requires the target to be in the SAME session as the requester,
+            # and a session has exactly one type, so the requester-side check
+            # above settles it for everyone in the group.
             # Both bunk name and session must match
             if target_bunk != requester_bunk or target_session != requester_session:
                 return False, ""
--- a/bunking/sync/bunk_request_processor/data/repositories/attendee_repository.py
+++ b/bunking/sync/bunk_request_processor/data/repositories/attendee_repository.py
@@ -529,10 +529,17 @@ class AttendeeRepository:
             # served more than one session, so filtering on bunk + year alone
             # returned children the requester never met (median peer set 29
             # against a session-scoped 8).
+            #
+            # Same `sort: "id"` STABLE_SORT convention as the query above, and
+            # for the same reason: `_try_prior_bunkmate_resolution` returns the
+            # FIRST camper in `cm_ids` whose name matches the target, so two
+            # cabinmates sharing a first name are separated by nothing but the
+            # order this query happened to return.
             bunkmates = self.pb.collection("bunk_assignments").get_full_list(
                 query_params={
                     "filter": f'bunk = "{bunk_id}" && year = {previous_year} && session = "{session_id}"',
                     "expand": "person",
+                    "sort": "id",
                 }
             )
 
--- a/bunking/sync/bunk_request_processor/data/repositories/attendee_repository.py
+++ b/bunking/sync/bunk_request_processor/data/repositories/attendee_repository.py
@@ -466,15 +466,36 @@ class AttendeeRepository:
             Dict with cm_ids, prior_bunk, prior_year, total_in_bunk, returning_count
             Empty dict if no prior year assignment found or on error
         """
+        # Imported at call time, not module scope: `bunking.graph` imports
+        # `bunking.satisfaction`, which imports this package's `core.models`, so
+        # a top-level import here is a genuine cycle. Importing at call time
+        # keeps ONE definition of the eligible-session-type set rather than
+        # adding another copy of it.
+        from bunking.graph.social_graph_builder import LAST_YEAR_HISTORY_SESSION_TYPES  # noqa: PLC0415
+
         try:
             previous_year = year - 1
 
-            # Find requester's bunk assignment from prior year
+            # Find requester's bunk assignment from prior year.
+            #
+            # The session-type predicate is pushed INTO the query so the
+            # database returns only rows from a session that actually puts
+            # children in a cabin. Without it a Family Camp DAY GROUP was
+            # eligible to be returned as a summer `prior_bunk`, and the whole
+            # day group then came back as "bunkmates" (#2426). `sort: "id"`
+            # breaks ties deterministically for the rarer case of two eligible
+            # rows -- PocketBase's get_full_list has no inherent order, so the
+            # winner could otherwise change between runs with no code change.
+            # Same predicate and same STABLE_SORT convention as
+            # `bunking/graph/social_graph_builder.py`, which already carries
+            # this fix for its own query.
+            type_clause = " || ".join(f'session.session_type = "{t}"' for t in LAST_YEAR_HISTORY_SESSION_TYPES)
             try:
                 assignments = self.pb.collection("bunk_assignments").get_full_list(
                     query_params={
-                        "filter": f"person.cm_id = {requester_cm_id} && year = {previous_year}",
-                        "expand": "person,bunk",
+                        "filter": f"person.cm_id = {requester_cm_id} && year = {previous_year} && ({type_clause})",
+                        "expand": "person,bunk,session",
+                        "sort": "id",
                     }
                 )
             except Exception:
@@ -488,19 +509,31 @@ class AttendeeRepository:
             requester_assignment = assignments[0]
             expand = getattr(requester_assignment, "expand", {}) or {}
             bunk_data = expand.get("bunk")
+            session_data = get_session_from_expand(requester_assignment)
 
             if not bunk_data:
                 return {}
 
             bunk_id = getattr(bunk_data, "id", None)
             bunk_name = getattr(bunk_data, "name", None)
+            session_id = getattr(session_data, "id", None) if session_data else None
 
-            if not bunk_id:
+            # Without the session we cannot scope the bunk to a week, and an
+            # unscoped query would return the whole building -- which is the
+            # defect this method is being fixed for. Return nothing instead.
+            if not bunk_id or not session_id:
                 return {}
 
-            # Find all campers in that bunk for prior year
+            # Find all campers in that bunk for prior year, IN THAT SESSION. A
+            # bunk is a building: 42 of the 57 bunks carrying a 2025 assignment
+            # served more than one session, so filtering on bunk + year alone
+            # returned children the requester never met (median peer set 29
+            # against a session-scoped 8).
             bunkmates = self.pb.collection("bunk_assignments").get_full_list(
-                query_params={"filter": f'bunk = "{bunk_id}" && year = {previous_year}', "expand": "person"}
+                query_params={
+                    "filter": f'bunk = "{bunk_id}" && year = {previous_year} && session = "{session_id}"',
+                    "expand": "person",
+                }
             )
 
             if not bunkmates:
--- a/bunking/sync/bunk_request_processor/social/social_graph.py
+++ b/bunking/sync/bunk_request_processor/social/social_graph.py
@@ -254,6 +254,14 @@ class SocialGraph:
 
     async def _add_historical_bunking_relationships(self, graph: nx.Graph, session_cm_id: int) -> None:
         """Add historical bunking relationships from previous years using bunk_assignments table"""
+        # Imported at call time, not module scope: `bunking.graph` imports
+        # `bunking.satisfaction`, which imports this package's `core.models`,
+        # so a top-level import here is a genuine cycle (it fails with
+        # "cannot import name 'BucketCount' from partially initialized module
+        # bunking.satisfaction"). A call-time import keeps ONE definition of
+        # the eligible-session-type set instead of adding a fourth copy of it.
+        from bunking.graph.social_graph_builder import LAST_YEAR_HISTORY_SESSION_TYPES  # noqa: PLC0415
+
         try:
             # Get all people in this session's graph
             if graph.number_of_nodes() == 0:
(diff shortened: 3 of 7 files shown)


adamflagg:
### Code review

Independent second read at score floor 60. The `code-review` skill was invoked but no
subagent tool was available in this session, so the review was done directly: full diff
plus all three production files and all four consumers read at length.

The core change holds up. The session-type predicate, the `(year, bunk, session)` key,
the session-scoped bunkmate query and the `"bunk"` → `"bunk_name"` producer/consumer fix
are all correct, and the decision to leave the temporal-cache *load* unfiltered while
filtering at `verify_bunk_together` is the right split — presence in `_historical_bunking`
is what gives a person a `"historical"` bucket in `_build_name_index`.

**Fixed in-branch (1d06172):** two order-dependence defects, both in query/loop code this
PR touched. Each test was written first and confirmed failing.

1. `find_prior_year_bunkmates`'s bunkmate query was the one query in the pair without
   `sort`. `_try_prior_bunkmate_resolution` returns the FIRST camper in `cm_ids` whose
   name matches, so two cabinmates sharing a first name were separated by nothing but the
   order PocketBase happened to return — the same hazard the sibling query one block above
   gained `sort: "id"` for.
   https://github.com/adamflagg/kindred/blob/63f9f7bbaaa4bac2e50897cd2605d00dbd611b5d/bunking/sync/bunk_request_processor/data/repositories/attendee_repository.py#L527-L536
   https://github.com/adamflagg/kindred/blob/63f9f7bbaaa4bac2e50897cd2605d00dbd611b5d/bunking/sync/bunk_request_processor/services/phase2_resolution_service.py#L899-L916
   Mutation check: `assert None == 'id'` before the fix.

2. `_add_historical_bunking_relationships` keeps exactly one edge per pair via
   `processed_pairs`, so for a pair that bunked together in more than one year the first
   grouping key visited is the one whose `recency_weight` survives. That was dict-insertion
   order off an unsorted query — nondeterministic between runs, and able to let an older
   year beat a newer one, contradicting the "more recent = stronger" rule stated three
   lines down. Now iterates most recent year first.
   https://github.com/adamflagg/kindred/blob/63f9f7bbaaa4bac2e50897cd2605d00dbd611b5d/bunking/sync/bunk_request_processor/social/social_graph.py#L332-L350
   Mutation check: `assert 1.25 == 1.6666666666666667` before the fix (2023 weight winning
   over 2025).

**Refuted, not filed:**

- *A fourth unfiltered read site exists* — `api/services/data_fetcher.py::fetch_historical_bunking`
  also reads prior-year `bunk_assignments` with no `session_type` predicate, but it filters
  `session.cm_id = {session_cm_id} && year = {prior_year}`, which pins it to one already-known
  summer session. No predicate needed.
  https://github.com/adamflagg/kindred/blob/63f9f7bbaaa4bac2e50897cd2605d00dbd611b5d/api/services/data_fetcher.py#L192-L200
- *Missing target-side `session_type` check in `verify_bunk_together`* — the PR's reasoning
  is sound: the loop already requires `target_session == requester_session`, and a session
  carries exactly one type, so a target-side check cannot change any outcome.
- *`# noqa: PLC0415` is an unused suppression* — `PLC0415` is in `ruff.toml`'s select list
  (line 45) and is only ignored under `tests/**`, so the three call-site suppressions are
  load-bearing.
- *Empty `type_clause` would emit `&& ()`* — `LAST_YEAR_HISTORY_SESSION_TYPES` is a literal
  four-element tuple, so the clause is never empty, and being a tuple rather than a set the
  emitted filter string is also stable across runs.
- *One record per (person, year) in `_historical_bunking` loses a second eligible session* —
  real, but pre-existing and structural (the dict shape allows one row per year); it needs a
  data-model change, not a fix here, and this PR strictly improves which row is kept.

**Filed as out of scope:** nothing. Both findings were small and in touched code, so both
were folded in rather than turned into issues.

Not requested from CodeRabbit — the wave spends its quota on one designated PR.

Issue #2439, opened 2026-08-18
investigate(frontend,api): a write-in carries no headcount, so a shared space's free-bed count will overstate once #2432 lands


adamflagg:
A write-in records **who** is sleeping in a space but not **how many**, so it contributes nothing to the
board's bed arithmetic. Today that is harmless. **#2432 makes it wrong.**

## Why it does not bite today

A write-in currently *closes* the space it names (`family_available = false`), so no family can be placed
alongside it and no bed sum is ever attempted against it. The unit simply reads as unavailable.

This absence is also deliberate, and the reasoning should be preserved rather than overwritten.
`WriteInCard`'s doc block:

> no party-size figure and no chip row. A write-in is a name; nothing else is known, and a `0` beside it
> would assert something.

That is correct as far as it goes: a **defaulted zero** would be a claim nobody made.

## Why #2432 changes that

**#2432's ruling (owner, 2026-08-18) is symmetric**: staff may add a family to a space that holds a
write-in, or a write-in to a space that holds a family, *regardless of which came first*. Once the two
coexist, the unit's free-bed figure is computed from the placed families alone:

- `MapUnitPopover.tsx:180` — `bedsNeeded = parties.reduce((sum, party) => sum + partyBeds(party), 0)`
- `MapUnitPopover.tsx:460` — `placed = families.reduce((total, { party }) => total + partyBeds(party), 0)`
- `boardLayout.ts:124-130` — `slotOccupancy`, `partyBeds` and `bedsNeeded` all want **beds**

A write-in occupant is in none of those sums. So a 10-bed cabin holding a placed party of 6 **and** a
written-in paper registration of 3 will read **4 free** when 1 is free — and staff can be told to place a
family into beds that are already taken. That is the exact failure the write-in mechanism exists to
prevent, reintroduced from the other side.

## The proposal

An **optional** occupant count on the write-in, entered beside the note in the same control, rendered as a
quiet detail on the card, and folded into `bedsNeeded` when present.

- **Optional is the whole point.** Absent means *"nobody said"*, exactly as today — it must NOT default to
  0 or to 1, or it re-creates the asserted-zero problem the card's doc warns about. The bed sum should
  treat an absent count as **unknown**, not as zero, and the unit should say so rather than quietly
  under-counting.
- **Storage:** `lodging_write_ins` and `lodging_write_ins_draft` currently carry `occupant_name`, `note`,
[...]

PR #2446, opened 2026-08-18
feat(frontend): one box places a family or writes a name in, and the strip loses its write-in

### LINKED ISSUE #2381: bug(frontend): write-ins do not merge/split like assignments — a merged building shows only one

### LINKED ISSUE #2430: feat(frontend): let staff edit a write-in instead of clearing and re-adding it
## What staff asked for

A write-in should be editable in place. Fixing a typo in an occupant's name, or adding a note after
the fact, currently means clearing the write-in and re-adding it from scratch.

## Why there is no edit path today

`availabilityAction` (`frontend/src/components/weekend/unitBadges.ts:306`) returns **exactly one
action per card**, and its three kinds are the whole vocabulary:

| kind | prompt | effect |
|---|---|---|
| `hold` (`:400`) | `'occupant'` | creates a write-in — required occupant name, optional note |
| `clear` | `'none'` | deletes the covering availability row |
| `release` | `'reason'` | opens a staff cabin to families (no surface reaches it today) |

There is no `edit`. Clear-and-re-add is not staff missing a hidden control — it is the only path.

The single-action shape is deliberate and its reasoning is sound as far as it goes: *"ONE action, not
a menu, because `family_available` has exactly three reachable outcomes and a unit is always in one of
them"*. But that argument is about the availability **tri-state**, not about the occupant **fields**
hanging off the row — and the fields are what staff want to change.

## Shape of the fix

`UnitAvailabilityControl` already collects exactly the two fields an edit needs — the required
occupant and the optional note (`UnitAvailabilityControl.tsx:114-119`, submitted as `occupantName` /
`reason` at `:182-183`). An edit is that same prompt, pre-filled from the covering write-in, writing
back to the existing row instead of creating one.

Worth settling while implementing: whether an edit is simply an upsert on the same row, or whether
edit and clear-then-re-add need to stay distinguishable downstream.

## Sequencing

⚠️ **#2381 first.** It widens write-in arity from one cover per drawn unit to N and adds a
per-write-in affordance (a tiny X in each card's top-right, its part 4). An edit control has to sit on
the same card and target the same specific row — built against today's single-cover shape, it gets
built twice.

- **Files:** `frontend/src/components/weekend/unitBadges.ts`, `UnitAvailabilityControl.tsx`, `WriteInCard.tsx`
- **Related:** #2381 (arity + per-write-in controls), #2252 (clear-write-in wording, merged), #2078 (a write-in is a named global occupant)

*Reported via kindred-feedback#116 — the private original, which holds the reporter, the page it was filed from, and the app version.*

### LINKED ISSUE #2431: bug(frontend): the write-in card truncates a long occupant name instead of wrapping it
## What staff asked for

> If a write-in is longer than the box, have it spill over to the next line so all the text can be seen.

## Confirmed in the code — the name line is explicitly truncated

`frontend/src/components/weekend/WriteInCard.tsx:62-63` puts `truncate` on the occupant name in both
branches:

```
named
  ? 'text-foreground min-w-0 truncate text-sm leading-tight font-semibold'
  : 'text-muted-foreground min-w-0 truncate text-sm leading-tight italic'
```

`truncate` is Tailwind's `overflow:hidden; text-overflow:ellipsis; white-space:nowrap` — one line,
ellipsis, no wrap.

The **note** line directly below it carries no `truncate` and already wraps: `:73` is the guard
(`{occupant.note !== '' && (`) and `:79` is the span itself
(`<span className="text-muted-foreground text-xs leading-tight">`). The `Written in at …` footer at `:87`
wraps too. So the card is internally inconsistent, and the name span is the **only** line on it that cannot
finish a sentence: the optional note wraps, the optional footer wraps, the required name does not.

## Why this is filed as a bug rather than the reporter's "feature request"

Same defect class as #2253 (*"the household journey card truncates the prior year's housing name"*), which
was filed as a bug and has since **shipped** (PR #2314, `fix(frontend): let the household journey card's
names wrap, not truncate`). A card that cannot finish saying a name reports the wrong occupant — two
write-ins whose names share a prefix render identically.

**Copy #2253's shipped fix rather than re-deciding it** —
`frontend/src/components/weekend/HouseholdJourneyCard.tsx:146-171`:

1. Drop `truncate`; **keep `min-w-0`.** The comment there is explicit about why: `min-w-0` is what lets the
   flex child shrink below its content width, which is what makes it wrap rather than overflow the card.
   Removing both is the easy mistake and produces an overflowing row instead of a wrapped one.
2. **Write the summer-divergence justification at the divergence**, as that fix did — root `CLAUDE.md` §4
   requires it, and summer's direct analog still truncates (`frontend/src/components/camper/CampJourneyTimeline.tsx:117-127`,
   the `record.bunkName` span). The argument #2253 made was about staff-typed unit names losing a wing or
   sub-unit suffix; here it is stronger, because an occupant's name has no alias or display-name convention
   to fall back on at all.

## Not a duplicate of #2332

#2332 is housing-**unit**-name overflow, and it resolves by shortening names through the
alias / display-name convention. This is the **occupant** name inside `WriteInCard`, where there is no
alias to shorten to: a person's name has to appear in full.

## Worth checking while fixing

- The card sits in the unit's occupant well, so wrapping changes card height and therefore board row
  height. Check a long name in a well that also holds family cards.
- `MapUnitPopover` draws no `WriteInCard` — it calls `reservationBadge` directly (header at
  `frontend/src/components/weekend/MapUnitPopover.tsx:179`, collapsed grid cell at `:598`), and that chip
  names nobody — so it is unaffected today. Verified on `main`. #2381 changes what that popover shows.

- **Files:** `frontend/src/components/weekend/WriteInCard.tsx` (+ `WriteInCard.test.tsx`)
- **Overlaps #2381, but does not collide.** Checked line by line on `main` @ 1797f9bd: this issue edits the
  name span's `className` at `WriteInCard.tsx:60-64`; #2381 part 3 deletes the `atUnitName` prop and its doc
  (`:40-49`) and the footer block (`:81-90`). Disjoint regions, and neither depends on the other. Both do
  touch `WriteInCard.test.tsx`, so land one and rebase the other rather than running them in parallel. This
  one is much the smaller and is not gated on anything.
- **Related:** #2253 (same truncation class, **shipped** in PR #2314 — copy it), #2332 (unit-name overflow — different subject), #2381 (write-in arity)

*Reported via kindred-feedback#117 — the private original, which holds the reporter, the page it was filed from, the app version, and the screenshot.*

### LINKED ISSUE #2432: feat(frontend): allow a write-in on a space that already holds a family
## What staff asked for

> I need to be able to add a write-in on a cabin that has another family as well — one paper
> registration and one CampMinder family sharing a cabin.

A paper registration has no CampMinder record, so it cannot be placed on the board at all; a write-in
is the only way to record that party. Today the write-in control disappears the moment a CampMinder
family is placed in the space — so the one case staff need it for is the one case it refuses.

## ⚠️ This reverses the #2090 ruling — do not simply delete the guard

The block is deliberate and documented. `frontend/src/components/weekend/unitBadges.ts:395-399`:

```
// An already-occupied space offers no write-in: the fix for #2090. Only
// reachable here, past both branches above, so an already-held unit keeps
// its `clear` action regardless of occupancy — clearing only ever REDUCES
// the conflict, so it is never the state that needs blocking.
if (occupied) return null
```

and `availabilityAction`'s own docstring states the ruling: *"Owner ruling on #2090: held and occupied
are mutually exclusive states, so a space that already holds a family may not also be marked held."*

That ruling was made when **hold and write-in were the same act** — the 2026-08-09 ruling that
collapsed them (*"hold IS the write-in"*, #2078). Under that collapse "occupied and held" is genuinely
contradictory. The reported case shows the collapse has a real exception: a write-in is also how a
**non-CampMinder occupant** is recorded, and such an occupant can legitimately share a cabin with a
placed family.

So the fix is *not* to reopen the hold/write-in split — #2078 settled that. It is to stop treating
*closes the space* as inseparable from *names an occupant*.

**Whoever implements this must say in the code what replaced the #2090 rule**, or the next reader
restores the guard on the strength of the comment sitting there now.

## Scope: frontend only

The API has no equivalent gate — nothing in `api/services/lodging_write_service.py` rejects a write on
occupancy grounds. The block is purely `availabilityAction`'s.

## Open questions for the implementer

- **Does a write-in on an occupied space still set `family_available = false`?** If it does, the space
  reads as closed while holding a placed family. The availability effect probably has to become
  conditional on occupancy — which is the actual substance of unpicking #2090.
- The occupant well then draws a `FamilyCard` and a `WriteInCard` together. #2091 (*"mark a space that
  holds two families, the way the master housing sheet does"*) is the sibling case and the two should
  read the same way.
- Capacity accounting: a write-in names one person and carries no party size — `WriteInCard`'s
  docstring is explicit that *"a `0` beside it would assert something"* — so a shared cabin's remaining
  beds cannot be derived from it.

- **Files:** `frontend/src/components/weekend/unitBadges.ts` (+ its tests), `UnitAvailabilityControl.tsx`, `LodgingUnitCard.tsx`
- **Related:** #2090 (the ruling being reversed — closed), #2078 (hold IS the write-in — closed), #2091 (two families in one space), #2381 (write-in arity)

*Reported via kindred-feedback#118 — the private original, which holds the reporter, the page it was filed from, the app version, and the screenshot.*

adamflagg:
feat(frontend): one box places a family or writes a name in, and the strip loses its write-in

Every tile carried two typeable rectangles asking the same question -- who is
sleeping in this cabin. The card's family picker was one; the availability
strip's "Write in" button opened the other, an occupant field plus an optional
note. Staff had to choose a control before they knew which kind of answer they
had.

Now there is one box. Type into it: if a registered family matches, place them;
if none does, the list offers `Write in "<what you typed>"`, on Enter or on a
click, and that text becomes the write-in.

## What this fixes beyond the tidy-up

#2090 gated the strip's write-in on the space being empty, and the picker was
hidden on an occupied card by a parallel rule. On a MERGED building that left no
write-in path at all -- refused on the building, and its rooms lose their own
cards to the merge:

    "if i populate 3/4 of clouds rest, i cannot add a write in directly to it"

Both gates are gone. The box is offered on every placeable card, occupied or
not, and family rows keep inheriting every refusal from
`resolvePickerPlacement` -> `resolveDrop` exactly as before.

## Kept deliberately

- **The note.** It carries real information -- one of the three notes in
  production is a third occupant's name, not a remark -- so the field stays,
  edited by the pencil on the write-in's own card. What went is the SECOND
  place to type one, mid-creation, beside a box that was already open.
- **The divergence from `canPlace`.** Recording who is in a cabin is a fact
  about the weekend, not about a plan, so the box stays live on the CampMinder
  mirror where nothing can be placed -- named "Write in an occupant for X"
  there, since offering to place a family would name an impossible action.
- **The per-card pending gate**, moved onto the box. 81 cards share one
  mutation, and a combobox invites a second write-in against an unsettled first.

## Also struck: the "Written in at <room>" line

It said which room a merged card's write-in belongs to. The room dimension is
real, but the line said it for write-ins and stayed silent about the FAMILIES in
the same well -- a half-built version of one shorthand covering every occupant,
which is filed as #2458.

## Not done

No API change and no migration: `set_availability` already routes a write-in
through `_upsert_row`, and the `note` column and its three rows are untouched.
The write-in still carries no headcount, so a shared space's free-bed figure
understates -- that is #2439, unchanged by this.

## Mix and match, either order, leaf or container

A second commit completes the reversal. Adding a write-in to a cabin holding a
family worked; the mirror did not — a cabin whose only occupant was a write-in
refused a family and offered no box. Three gates enforced the old rule and all
three are gone: `resolveDrop`'s write-in refusal (the load-bearing one, since
the picker path never touches a `useDroppable`), the card's `useDroppable`
`disabled: … || held` (the affordance half — without it the target never even
highlighted), and `canPickFamily`'s `!held`.

A write-in NAMES AN OCCUPANT; it does not CLOSE THE SPACE.

Closes #2381.
Closes #2430.
Closes #2431.
Closes #2432.

---

Supersedes the earlier shape of this PR, which kept the strip's "Write in"
button beside the card's family box and added a `Written in at …` line. Both
were struck by the owner on review, 2026-08-18.

### DIFF
--- a/api/schemas/lodging.py
+++ b/api/schemas/lodging.py
@@ -264,7 +264,7 @@ class LodgingUnitSummary(BaseModel):
     # standing `inventory_class` decides. None and False are different answers
     # and must not be flattened into one.
     #
-    # IT DOES NOT ANSWER "IS SOMEBODY IN IT". That is `write_in` below, read
+    # IT DOES NOT ANSWER "IS SOMEBODY IN IT". That is `write_ins` below, read
     # from `lodging_write_ins` / `_draft`. Until PR 4 of kindred#2382 an
     # occupancy also reported itself here as False, because
     # `is_family_available` and the board's forest open-tint were both derived
@@ -272,7 +272,7 @@ class LodgingUnitSummary(BaseModel):
     # False here means a role decision and never an occupant. A reader that
     # wants "can a family go in this space" wants `is_family_available`, which
     # folds both facts in; a reader that wants "is somebody in it" wants
-    # `write_in`.
+    # `write_ins`.
     #
     # Stated explicitly rather than implied. The rejected encoding was a row
     # meaning "the opposite of this unit's current default", which an ordinary
@@ -319,9 +319,19 @@ class LodgingUnitSummary(BaseModel):
     # Its rule is `is_family_available` in api/services/lodging_rules.py, and
     # that is the only place the two are combined.
     is_family_available: bool = False
-    # The write-in that closes this space, resolved through the unit tree --
+    # EVERY write-in that closes this space, resolved through the unit tree --
     # this unit's own row, else the nearest ancestor's, else the nearest
-    # descendant's. None means no write-in covers it.
+    # written-into descendant on each branch beneath it. Empty means no
+    # write-in covers it.
+    #
+    # A LIST since kindred#2381, and that arity is the whole point rather than
+    # future-proofing. A merged container draws in place of its rooms, so the
+    # four write-ins Clouds Rest carries in one 2026 weekend collapsed to
+    # whichever room sorted first and the other three were invisible -- while
+    # each clear silently re-populated the card with the next occupant, so
+    # destroying all four read as four failed clicks. A write-in must survive a
+    # merge and a split the way an assignment does, and an assignment does it
+    # by the drawn card carrying however many leaves it covers.
     #
     # The fields above stay STRICTLY this unit's own row: they are what the
     # write path reads back. Folding an inherited fact into any of them would
@@ -338,7 +348,7 @@ class LodgingUnitSummary(BaseModel):
     # blocks placement without moving the bed counts. That is the behaviour the
     # counts have always had, and changing it is a counts decision rather than
     # a side effect of the split.
-    write_in: WriteInCover | None = None
+    write_ins: list[WriteInCover] = Field(default_factory=list)
     map_x: float | None = None
     map_y: float | None = None
 
--- a/api/schemas/lodging.py
+++ b/api/schemas/lodging.py
@@ -326,7 +326,7 @@ class LodgingUnitSummary(BaseModel):
     #
     # A LIST since kindred#2381, and that arity is the whole point rather than
     # future-proofing. A merged container draws in place of its rooms, so the
-    # four write-ins Clouds Rest carries in one 2026 weekend collapsed to
+    # four write-ins one 2026 container carries in a single weekend collapsed to
     # whichever room sorted first and the other three were invisible -- while
     # each clear silently re-populated the card with the next occupant, so
     # destroying all four read as four failed clicks. A write-in must survive a
--- a/api/services/lodging_roster_service.py
+++ b/api/services/lodging_roster_service.py
@@ -574,8 +574,10 @@ def written_in_unit_ids(write_ins: list[Any]) -> frozenset[str]:
     return frozenset(_s(row, "unit") for row in write_ins)
 
 
-def write_in_covers(units: list[LodgingUnitSummary], write_in_unit_ids: frozenset[str]) -> dict[str, WriteInCover]:
-    """Which units each write-in closes, keyed by unit code.
+def write_in_covers(
+    units: list[LodgingUnitSummary], write_in_unit_ids: frozenset[str]
+) -> dict[str, list[WriteInCover]]:
+    """Which write-ins close each unit's space, keyed by unit code.
 
     THE UNIT A ROW NAMES IS NOT THE ONLY SPACE IT CLOSES. A write-in is a fact
     about a physical space and a building's space contains its rooms', but the
@@ -586,10 +588,27 @@ def write_in_covers(units: list[LodgingUnitSummary], write_in_unit_ids: frozense
     family could be dropped into a space somebody is already sleeping in.
 
     Order, highest first: the unit's OWN row, else the nearest ANCESTOR's, else
-    the nearest DESCENDANT's. Own beats inherited because it is the more
-    specific statement about the same space; an ancestor beats a descendant
-    because a building closed whole says something about every room in it,
-    where one room says only that the building is no longer free to let whole.
+    EVERY written-into descendant beneath it. Own beats inherited because it is
+    the more specific statement about the same space; an ancestor beats a
+    descendant because a building closed whole says something about every room
+    in it, where one room says only that the building is no longer free to let
+    whole.
+
+    THE DESCENDANT STEP RETURNS ALL OF THEM (kindred#2381), and that is the
+    arity fix. A merged container draws in place of its rooms, so returning the
+    first match dropped every other occupant off the board -- four of them on
+    the one 2026 container that carries four -- and made each clear look like a
+    failed click as the card re-resolved to the next room. The first two steps
+    stay single-answer: an own row is one row, and an ancestor chain has exactly
+    one nearest member.
+
+    Per-branch NEAREST, not every descendant. A written-into bed inside a
+    written-into wing is already inside that wing's space, so the wing's row
+    speaks for it and returning both would print one space twice. The walk
+    therefore stops descending a branch the moment it finds a row on it.
+
+    Ordered by `code` at every level, so two identical payloads never disagree
+    about the sequence a card draws its occupants in.
 
     RESOLVED FROM OWN ROWS ONLY, never transitively through a cover computed
     for somebody else -- which is what keeps a caretaker in room A off room B.
@@ -622,8 +641,11 @@ def write_in_covers(units: list[LodgingUnitSummary], write_in_unit_ids: frozense
     for unit in units:
         if unit.parent_code:
             children.setdefault(unit.parent_code, []).append(unit)
-    # Sorted so two identical payloads never disagree about WHICH row a card
-    # names when a building has several written-into rooms beneath it.
+    # Sorted so two identical payloads never disagree about the ORDER a card
+    # draws its occupants in when a building has several written-into rooms
+    # beneath it. It used to settle WHICH single row was named, because only
+    # one survived; kindred#2381 returns them all and this now fixes their
+    # sequence instead.
     for bucket in children.values():
         bucket.sort(key=lambda child: child.code)
 
@@ -640,7 +662,15 @@ def write_in_covers(units: list[LodgingUnitSummary], write_in_unit_ids: frozense
             cursor = by_code.get(cursor.parent_code) if cursor.parent_code else None
         return None
 
-    def _nearest_descendant(unit: LodgingUnitSummary) -> LodgingUnitSummary | None:
+    def _written_in_descendants(unit: LodgingUnitSummary) -> list[LodgingUnitSummary]:
+        """Every written-into descendant, nearest-first on each branch.
+
+        Breadth-first over `code`-sorted buckets, so the result is ordered and
+        two identical payloads agree. A branch is not descended past a match:
+        the matched unit's space contains whatever is below it, and its row
+        already speaks for that space.
+        """
+        found: list[LodgingUnitSummary] = []
         seen = {unit.code}
         queue = list(children.get(unit.code, []))
         while queue:
@@ -649,11 +679,12 @@ def write_in_covers(units: list[LodgingUnitSummary], write_in_unit_ids: frozense
                 continue
             seen.add(node.code)
             if _is_written_in(node):
-                return node
+                found.append(node)
+                continue
             queue.extend(children.get(node.code, []))
-        return None
+        return found
 
-    covers: dict[str, WriteInCover] = {}
+    covers: dict[str, list[WriteInCover]] = {}
     for unit in units:
         # Blank codes are excluded from BOTH sides, not just the lookup above.
         # `_build_units` reads this map back by code, so one blank-coded row
@@ -662,16 +693,23 @@ def write_in_covers(units: list[LodgingUnitSummary], write_in_unit_ids: frozense
         # a space on the strength of a row it does not hold.
         if not unit.code:
             continue
-        source = unit if _is_written_in(unit) else (_nearest_ancestor(unit) or _nearest_descendant(unit))
-        if source is None:
+        if _is_written_in(unit):
+            sources = [unit]
+        else:
+            ancestor = _nearest_ancestor(unit)
+            sources = [ancestor] if ancestor is not None else _written_in_descendants(unit)
+        if not sources:
             continue
-        covers[unit.code] = WriteInCover(
-            unit_id=source.unit_id,
-            unit_code=source.code,
-            unit_name=source.name,
-            occupant_name=source.occupant_name,
-            note=source.reason,
-        )
+        covers[unit.code] = [
+            WriteInCover(
+                unit_id=source.unit_id,
+                unit_code=source.code,
+                unit_name=source.name,
+                occupant_name=source.occupant_name,
+                note=source.reason,
+            )
+            for source in sources
+        ]
     return covers
 
 
@@ -684,7 +722,7 @@ def _resolve_write_in_covers(units: list[LodgingUnitSummary], write_in_unit_ids:
     """
     covers = write_in_covers(units, write_in_unit_ids)
     for unit in units:
-        unit.write_in = covers.get(unit.code)
+        unit.write_ins = covers.get(unit.code, [])
 
 
 def drawn_units(units: list[LodgingUnitSummary]) -> list[LodgingUnitSummary]:
@@ -1566,7 +1604,7 @@ class LodgingRosterService:
             # compat shim out. `family_available_override` is now the ROLE row
             # and nothing else -- a staff cabin opened to families for the
             # weekend, or a bare `false` closing one -- while "is somebody in
-            # this space" is answered by `write_in` alone. Until PR 4 this field
+            # this space" is answered by `write_ins` alone. Until PR 4 this field
             # reported `False` for an occupancy too, because
             # `is_family_available` and the board's open-tint were both derived
             # from it; both read the occupancy source directly now.
--- a/frontend/src/components/weekend/LodgingBoard.availability.test.tsx
+++ b/frontend/src/components/weekend/LodgingBoard.availability.test.tsx
@@ -145,7 +145,12 @@ describe('LodgingBoard — the availability gate', () => {
     // it in.
     renderBoard({ scenario: '' })
 
-    expect(screen.getByRole('button', { name: 'Write in Cedar 1' })).toBeInTheDocument()
+    // The BOX, since 2026-08-18 — the strip's "Write in" button is gone and
+    // creating a write-in is typing into the card's own control. Named for the
+    // write-in alone here, because with no scenario there is nothing to place.
+    expect(
+      screen.getByRole('combobox', { name: /write in an occupant for cedar 1/i })
+    ).toBeInTheDocument()
   })
 
   it('offers no control without bunking.manage', () => {
@@ -191,10 +196,10 @@ describe('LodgingBoard — the control becomes a write', () => {
     const user = userEvent.setup()
     renderBoard()
 
-    await user.click(screen.getByRole('button', { name: 'Write in Cedar 2' }))
-    await user.type(screen.getByRole('textbox', { name: /^occupant$/i }), 'Emma Johnson')
-    await user.type(screen.getByRole('textbox', { name: /note/i }), 'Back Monday')
-    await user.click(screen.getByRole('button', { name: /^write in$/i }))
+    await user.type(
+      screen.getByRole('combobox', { name: /place a family in cedar 2/i }),
+      'Emma Johnson{Enter}'
+    )
 
     expect(setAvailability).toHaveBeenCalledTimes(1)
     expect(setAvailability).toHaveBeenCalledWith({
@@ -202,7 +207,11 @@ describe('LodgingBoard — the control becomes a write', () => {
       unitName: 'Cedar 2',
       familyAvailable: false,
       occupantName: 'Emma Johnson',
-      reason: 'Back Monday',
+      // EMPTY, and that is the shape change. The strip collected an optional
+      // note beside the occupant; the box collects a name only, and a note is
+      // added afterwards by the pencil on the write-in's own card. One control
+      // asking one question, rather than two boxes on every tile.
+      reason: '',
     })
   })
 
@@ -212,8 +221,11 @@ describe('LodgingBoard — the control becomes a write', () => {
     pendingUnitId = 'u1'
     renderBoard()
 
-    expect(screen.getByRole('button', { name: 'Write in Cedar 1' })).toBeDisabled()
-    expect(screen.getByRole('button', { name: 'Write in Cedar 2' })).toBeEnabled()
+    // The BOX carries the per-card gate now that the strip button is gone —
+    // and it matters more on a combobox, which invites a staff member to keep
+    // typing a second write-in against an unsettled first.
+    expect(screen.getByRole('combobox', { name: /place a family in cedar 1/i })).toBeDisabled()
+    expect(screen.getByRole('combobox', { name: /place a family in cedar 2/i })).toBeEnabled()
     await Promise.resolve()
   })
 })
--- a/frontend/src/components/weekend/LodgingBoard.availability.test.tsx
+++ b/frontend/src/components/weekend/LodgingBoard.availability.test.tsx
@@ -218,12 +218,13 @@ describe('LodgingBoard — the control becomes a write', () => {
   })
 })
 
-describe('LodgingBoard — clearing a write-in this card inherited', () => {
+describe('LodgingBoard — removing a write-in this card inherited', () => {
   /*
    * The row names one unit; it closes a SPACE. Split a written-into building
    * and its ROOMS inherit the write-in, while the building — no longer drawn —
-   * has nowhere to offer the clear from. The room offers it, and both the write
-   * and the in-flight disable have to follow the row rather than the card.
+   * has nowhere to offer a removal from. The room's card carries it, as the X
+   * on the `WriteInCard` it draws, and both the write and the in-flight
+   * disable have to follow the ROW rather than the card.
    */
   const COVER = {
     unit_id: 'u-house',
@@ -232,13 +233,13 @@ describe('LodgingBoard — clearing a write-in this card inherited', () => {
     occupant_name: 'Liam Garcia',
     note: '',
   }
-  const room = unit({ unit_id: 'u-room', code: 'house-a', name: 'House A', write_in: COVER })
+  const room = unit({ unit_id: 'u-room', code: 'house-a', name: 'House A', write_ins: [COVER] })
 
   it('sends the unit that HOLDS the row, not the card it was clicked on', async () => {
     const user = userEvent.setup()
     renderBoard({ units: [room] })
 
-    await user.click(screen.getByRole('button', { name: 'Clear Write-in House A' }))
+    await user.click(screen.getByRole('button', { name: 'Remove write-in Liam Garcia' }))
 
     expect(setAvailability).toHaveBeenCalledWith({
       unitId: 'u-house',
@@ -249,14 +250,16 @@ describe('LodgingBoard — clearing a write-in this card inherited', () => {
     })
   })
 
-  it('disables the card while the row it points at is being written', async () => {
+  it('disables the card while the row it points at is being written', () => {
     // `pendingUnitId` names the unit the WRITE targets, which for an inherited
-    // clear is never this card's own id — so keying the disable on the card
-    // alone leaves the button live for the whole write and invites a second
-    // click on a row that is already going away.
+    // removal is never this card's own id — so keying the disable on the card
+    // alone leaves the X live for the whole write and invites a second click
+    // on a row that is already going away. Matched with `some` since
+    // kindred#2381: a merged card covers several rows and the pending write
+    // belongs to whichever one was clicked.
     pendingUnitId = 'u-house'
     renderBoard({ units: [room] })
 
-    expect(screen.getByRole('button', { name: 'Clear Write-in House A' })).toBeDisabled()
+    expect(screen.getByRole('button', { name: 'Remove write-in Liam Garcia' })).toBeDisabled()
   })
 })
--- a/frontend/src/components/weekend/LodgingBoard.drag.test.tsx
+++ b/frontend/src/components/weekend/LodgingBoard.drag.test.tsx
@@ -353,14 +353,15 @@ describe('LodgingBoard — placing a family from the space itself (kindred#2080)
     return screen.getByRole('combobox', { name: new RegExp(`place a family in ${name}`, 'i') })
   }
 
-  it('mounts the control on an empty space while placement is live', () => {
+  it('mounts the control on every placeable space, occupied or not', () => {
     boardWithUnplaced()
     expect(pickerFor('Cedar 2')).toBeInTheDocument()
-    // Cedar 1 already holds a family, so it offers nothing — the same rule
-    // that hides Hold on an occupied card.
-    expect(
-      screen.queryByRole('combobox', { name: /place a family in cedar 1/i })
-    ).not.toBeInTheDocument()
+    // Cedar 1 already holds a family and STILL offers the box. It did not
+    // before 2026-08-18, on the rule that hid the strip's write-in on an
+    // occupied card — a rule this change deletes, because the box is now the
+    // only way to write somebody in and a partly-filled merged building was
+    // left with no path at all.
+    expect(screen.getByRole('combobox', { name: /place a family in cedar 1/i })).toBeInTheDocument()
   })
 
   it('offers nothing without a scenario', () => {
--- a/frontend/src/components/weekend/LodgingBoard.tsx
+++ b/frontend/src/components/weekend/LodgingBoard.tsx
@@ -64,7 +64,7 @@ import { LodgingUnitCard } from './LodgingUnitCard'
 import { partyKey } from './partyKey'
 import { resolvePartyUnit } from './rosterAttention'
 import type { UnitAvailabilityWrite } from './UnitAvailabilityControl'
-import { writeInSource } from './writeIn'
+import { writeInEntries } from './writeIn'
 
 export interface LodgingBoardProps {
   parties: RosterPartyRow[]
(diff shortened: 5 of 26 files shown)


adamflagg:
### Independent review (floor 60)

No `Agent` tool in this session, so the review was done directly rather than via the multi-agent skill path: full `gh pr diff`, every touched file read at length, plus the two consumers the diff does not touch (`api/services/lodging_write_service.py`'s `family_available: null` branch, `MapUnitPopover.tsx`).

**Fixed in-branch** (4b423f69, comments only):

Three comments still named symbols the rename removed. One of them was doubly wrong — it explained `writeInBadgeApplies`'s class check by a `coveringWriteIn` fallback on `family_available_override === false` that kindred#2382 deleted two PRs ago, so it described the gate by a mechanism the code has not used since.

- `frontend/src/components/weekend/LodgingUnitCard.test.tsx` L1393 (`writeInOccupant` → `writeInEntries`) and L2341-2346 (the stale fallback paragraph → `hasWriteIn`)
- `tests/unit/api/services/test_lodging_roster_service.py` L4645, L4652 (`write_in`/`writeInOccupant` → `write_ins`/`writeInEntries`)

The one surviving mention, in [`writeIn.ts` L76](https://github.com/adamflagg/kindred/blob/4b423f697e74695e3d9f9c3d8a07d0aad30bfd9f/frontend/src/components/weekend/writeIn.ts#L70-L80), is a deliberate historical note and was left alone.

**Mutation check** — the three tests that pin the fix all fail when the fix is reverted, so none is decorative:

| mutation | result |
|---|---|
| `found.append(node); continue` → `return [node]` (back to first-only) | 3 failed, 223 passed |
| drop the `continue` (descend past a match) | 1 failed — `test_the_descendant_walk_stops_at_the_nearest_written_room_on_each_branch` |
| `writeIns.map` → `writeIns.slice(0, 1).map` in `LodgingUnitCard` | 1 failed of 120 |

**Refuted, with reasoning:**

1. *"`isRemoving={savingAvailability}` disables every X on the card, not just the one clicked."* Real but not a defect: `savingAvailability` is already `some`-matched over the covers in [`LodgingBoard.tsx`](https://github.com/adamflagg/kindred/blob/4b423f697e74695e3d9f9c3d8a07d0aad30bfd9f/frontend/src/components/weekend/LodgingBoard.tsx#L537-L545), and freezing the whole well during an in-flight delete is the conservative behaviour when the payload is about to re-resolve underneath it.

2. *"Sending `familyAvailable: null` from the X also deletes the ROLE row at the target unit."* Confirmed it does — [`lodging_write_service.py` L881-L891](https://github.com/adamflagg/kindred/blob/4b423f697e74695e3d9f9c3d8a07d0aad30bfd9f/api/services/lodging_write_service.py#L878-L893) clears both tables — but that is byte-identical to what the strip's `Clear Write-in` sent, at the same target. Not a regression.

3. *"A written-into card can no longer clear a stale role override."* The PR body's claim checks out in both directions. For an **own** row the X targets the card's own unit and the both-tables delete still clears the role row, exactly as the old clear did; for an **inherited** row neither the old clear (which targeted the source unit) nor the new X ever reached this card's role row. No behaviour moved.

**Not filed, but worth the owner's eye before merge** (design point, not a defect in this diff):

The arity fix widened the *descendant* step only, which is what the body says it does. The other two steps still return exactly one, so a drawn unit that resolves through **own** or **ancestor** still hides any written-into descendants beneath it — e.g. a merged container carrying its own row over two written-into rooms reports one occupant, or `block`(written) > `house`(merged) > `house-a`(written) surfaces `block` and drops `house-a`. That is the same shape as #2381 in a rarer position.

I did **not** change it, and I do not think it should be: the precedence is a deliberate, heavily-documented invariant ("own beats inherited because it is the more specific statement about the same space"), it is pre-existing rather than introduced here, and it is unreachable through the UI — `availabilityAction` returns `null` for any covered space, so no card offers to write a second cover into a space that already has one. It would need direct API calls or legacy data. Flagging it as a known edge of the invariant rather than as work.

**CodeRabbit:** not requested. The wave spends the rolling per-account quota on one designated PR; the green `CodeRabbit` check here reads `Review skipped: automatic reviews are disabled` and is not coverage.

**Gates:** `bash scripts/pre-push-verify.sh` (bare) → `ALL CHECKS PASSED`. CI Summary green on 4b423f69, `mergeStateStatus` CLEAN. Push confirmed by the remote ref (`git rev-list --count origin/feature/write-in-merge-split..HEAD` = 0).

**Holding for the owner.** `gh pr diff --name-only | grep -E '^frontend/src/.*\.(tsx|ts|css)$' | grep -v '\.test\.'` matches nine files, so this is a visual-impact change and parks for your eyes under the standing rule — the ~18px corner X and the strip losing its action are both things to look at rather than take on trust.

adamflagg:
### Independent review — floor 60

Reviewed by hand (no fan-out available): full diff, every touched file at length, every caller of
the moved seams (`writeInEntries`/`hasWriteIn` — `LodgingBoard`, `LodgingUnitCard`,
`dragPlacement`, `unitBadges`, and a grep confirming no fifth), plus the write path
(`set_availability`) the edit control leans on. Data claims checked against the production
snapshot read-only.

**Two fixed in-branch, one doc corrected, five refuted.**

#### Fixed

**1. Deleting the `Written in at …` footer removed the only thing saying WHICH unit a row is on —
and the plural well makes that worse, not obsolete.** The line carried two reasons; the PR retired
it citing only the navigational one ("go find the Clear the merge took away"), which is genuinely
dead. The identity reason is not:

- **Split.** Split a written-into building and every room draws a card for the *same* row —
  exactly the case `LodgingBoard.availability.test.tsx`'s "removing a write-in this card inherited"
  covers. Each room's card read as a bare name, and each card's X (all pointed at the one building
  row) emptied all the others, with nothing on screen saying so.
- **Merge.** In the 2026 snapshot the one container this PR exists for holds four write-ins on four
  *different* leaves. Four names in one well with no room beside them do not tell staff where
  anybody is sleeping. `LodgingUnitCard.test.tsx`'s "draws one card per write-in" asserted both
  names and neither room.
- Two rows named alike — or both unnamed, which `UNNAMED` exists for — otherwise render as two
  identical cards with two identical `aria-label`s. That is the same objection #2431 makes to
  truncating a name, one card over.

Restored with its prop and its old condition (shown only when the row is not the drawing card's
own), plus tests on both surfaces.
https://github.com/adamflagg/kindred/blob/c9ea26af7c8c6f72a99b6776c97f3f83c3869d6f/frontend/src/components/weekend/WriteInCard.tsx#L323-L333
https://github.com/adamflagg/kindred/blob/c9ea26af7c8c6f72a99b6776c97f3f83c3869d6f/frontend/src/components/weekend/LodgingUnitCard.tsx#L969-L978

**2. The #2430 edit form's two inputs had no `placeholder`,** unlike the prompt they were modelled
on ([`UnitAvailabilityControl`](https://github.com/adamflagg/kindred/blob/c9ea26af7c8c6f72a99b6776c97f3f83c3869d6f/frontend/src/components/weekend/UnitAvailabilityControl.tsx#L206-L236)
sets both). Migration 1500000148 cleared the note of every row it moved, so an edit opened on any
existing write-in showed a filled name box and a blank, otherwise unlabelled box under it. The
class list already carried `placeholder:text-muted-foreground/60` with no placeholder to style.
Same two strings, so one control does not teach a different vocabulary from the other.
https://github.com/adamflagg/kindred/blob/c9ea26af7c8c6f72a99b6776c97f3f83c3869d6f/frontend/src/components/weekend/WriteInCard.tsx#L214-L221

**3. `isSaving`'s doc** said "a write to THIS row"; the only caller passes a card-level flag, so a
write to one row disables every corner control in the well. The behaviour is the right
(conservative) direction — the doc was what was wrong, and now says what is passed.

The PR body was edited to match on all three, since it is the squash commit message.

#### Mutation check

New assertions re-run against reverted source (footer gated off, `atUnitName` forced undefined,
both placeholders removed): `Tests 4 failed | 145 passed`, exit 1. Restored: 41 files / 1258
passed. `scripts/pre-push-verify.sh` bare: ALL CHECKS PASSED.

#### Refuted

- **`key={entry.source.unitId}` colliding, so an open edit form binds to the wrong row after a
  sibling removal.** Refuted — each cover names a distinct unit and `unit_id` is a PocketBase
  record id, never blank. Keying by id (rather than index) is what makes the per-card form state
  survive a sibling removal correctly.
- **`_written_in_descendants` double-reporting a write-in nested under another.** Refuted — the
  walk `continue`s on a match and does not descend past it, so a written-into bed inside a
  written-into wing is not printed twice.
- **The edit's `familyAvailable: false` creating a second row rather than updating.** Refuted
  against `lodging_write_service.set_availability`, which finds-then-updates on
  `(year, session_cm_id, unit_id[, scenario])`. The `false`-not-`null` choice is right: `null`
  deletes.
- **The edit dropping a co-existing role row.** Real, but by design and pre-existing — the "one
  fact at a time" promise in that method's docstring, unchanged by this PR.
- **`availabilityAction` returning `null` stranding a stale role override.** Verified as
  documented and not a regression: an own-row write-in's X sends `familyAvailable: null`, which
  deletes *both* tables, and the inherited case was equally unreachable before.

Data claims in the body were checked, not taken on trust: the snapshot has 21 write-in rows across
four 2026 weekend sessions; exactly one container carries four in a single weekend, on four direct
leaf children, and no other container carries more than one. The body's framing is accurate.

#### Noted, not fixed

The edit form restates ~200 characters of Tailwind verbatim from `UnitAvailabilityControl`'s
prompt — two input class lists, the Save/Cancel pair, the amber refusal span. Real drift hazard,
and this file's own `WRITE_IN_FRAME` block shows the repo takes that seriously; but the honest fix
puts shared weekend form styling somewhere that does not exist yet, which is a placement decision
rather than a defect. Left alone rather than churning a diff that is already going to be looked at
on screen.

#### Not merged

`frontend/src/**` changes with visual impact, so this holds for owner review per the standing rule.
On screen: a merged container's occupant well now stacks one card per write-in — the 2026 case
shows four, each with the occupant's name wrapping over as many lines as it needs, a small muted
`Written in at <room>` line beneath, and two ~18px circular controls in the top-right (a pencil and
an ×), both always visible rather than on hover. Clicking the pencil swaps that one card for an
inline two-box form pre-filled with the name and note, with Save / Cancel underneath. The
availability strip on a written-into card no longer offers anything.

PR #2447, opened 2026-08-18
fix(sync): paginate GetSessions/GetSessionGroups past TotalCount

### LINKED ISSUE #2437: bug(sync): GetSessions/GetSessionGroups fetch one page of 100 and discard TotalCount, so session 101 would vanish silently
`GetSessions` and `GetSessionGroups` request a single page of 100, decode `TotalCount` into a struct
field, and then never read it. At 101 sessions a whole session — and therefore its entire roster —
silently ceases to exist, with no error and no log line.

Filed as a **latent** bug: there is real headroom today. Recorded now because the failure is silent and
the trend is the wrong way.

## The code

`pocketbase/campminder/client.go:376-399` (`GetSessions`) and `:689-712` (`GetSessionGroups`) are
identical in shape:

```go
params := map[string]string{
    "clientid":   c.clientID,
    "seasonid":   strconv.Itoa(c.seasonID),
    "pagenumber": "1",
    "pagesize":   "100",
}
...
var response struct {
    TotalCount int              `json:"TotalCount"`
    Results    []map[string]any `json:"Results"`
}
...
return response.Results
```

`TotalCount` is decoded and discarded. There is no loop over `pagenumber` and no comparison of
`len(Results)` against `TotalCount`.

Note the contrast with the neighbouring calls in the same file, which use `"pagesize": "500"` and
carry the comment *"Get all in one call (typically < 100)"* — the 100 here appears to be an oversight
rather than a considered limit.

## Measured headroom (production snapshot)

| year | sessions | session groups |
|---|---|---|
| 2017 | 31 | 7 |
| 2020 | 37 | 7 |
| 2022 | 48 | 9 |
| 2023 | 50 | 9 |
| **2024** | **58** | 9 |
| 2025 | 56 | 9 |
| 2026 | 45 | 9 |

Peak is 58 (2024) against a ceiling of 100. Session groups are nowhere near it. So this is **not
urgent** — but the 2020→2024 run added roughly five sessions a year, and the weekend program is the
part that grows.

## Suggested fix

Either raise `pagesize` to 500 to match the sibling calls, or — better, since it removes the class of
bug rather than moving the cliff — loop on `pagenumber` until `len(accumulated) >= TotalCount`. At
minimum, compare the two and return an error when the response is short, so a future overflow is loud
instead of silent.

- **Files:** `pocketbase/campminder/client.go` (+ tests)

adamflagg:
`GetSessions` and `GetSessionGroups` in `pocketbase/campminder/client.go` requested a single page
of 100 records, decoded `TotalCount` from the CampMinder response, and then discarded it. At 101
sessions (or session groups), the 101st record would silently disappear — no error, no log line.
The bug was latent, not yet firing: production's measured peak was 58 sessions in 2024, but the
2020-2024 trend added roughly five sessions a year.

## Fix

Both methods now loop over `pagenumber`, accumulating `Results` across pages, until
`len(accumulated) >= TotalCount` or a page comes back empty. This removes the class of bug rather
than just raising the ceiling (e.g. bumping `pagesize` to 500 would still leave a silent cliff at
501). The page size itself is unchanged (100 per page); only the "single page and done" assumption
was removed.

To make this testable, added an `apiBaseURL` field to `Client`, following the same pattern the
existing `authURL` field already uses to redirect auth calls to a test server — `apiBaseURL` does
the same for `makeRequest`'s regular API calls. It is empty in production and non-empty only in
tests.

## Tests

- `TestGetSessions_PaginatesPastFirstPage` / `TestGetSessionGroups_PaginatesPastFirstPage`: a mock
  server returns `TotalCount: 101` split across two pages; asserts all 101 records come back and
  at least two pages were requested.
- `TestGetSessions_SinglePageUnderLimit`: asserts no extra request is made when everything fits in
  page 1.

All three were run against the pre-fix single-page code first and failed (verified RED), then
against the fix (verified GREEN). Mutation-checked by re-breaking each method's loop-continuation
condition post-fix and confirming the corresponding test goes RED again.

## Verification note

`bash scripts/pre-push-verify.sh`'s combined `go test -race ./... -v` step hit Go's hardcoded
600s default test timeout on the unrelated `sync` package, purely from this worktree's host
running 9 concurrent overnight agents (each running their own heavy test suite). Isolated re-runs
prove it is a contention artifact, not a regression: `go test -race ./sync/... -timeout 25m`
passed clean in 656.751s (vs. the documented 297s baseline under normal load — consistent with
CPU starvation, not breakage), `go build ./...`, the full `go test ./...` (no `-race`), and
`golangci-lint run` (both scoped to `campminder` and repo-wide) all passed clean. The diff here
touches zero files under `sync/`.

## Deliberately NOT done

- Did not add a loud error path for the "response came back short of TotalCount" case beyond the
  loop's own retry — the loop just keeps requesting pages until it has everything or the server
  stops returning results, which is the strictly better fix the issue itself preferred over an
  error-and-bail.
- Did not touch the other single-page "get all" methods in this file (`GetDivisions`,
  `GetStaffProgramAreas`, `GetStaffOrgCategories`, `GetStaffPositions`) — they already use
  `pagesize: 500` against typical counts well under 100, and were out of scope for this issue.

Closes #2437.

### DIFF
--- a/pocketbase/campminder/client.go
+++ b/pocketbase/campminder/client.go
@@ -53,6 +53,10 @@ type Client struct {
 
 	// authURL overrides the production auth endpoint. Non-empty only in tests.
 	authURL string
+
+	// apiBaseURL overrides baseURL for regular (non-auth) API requests made via
+	// makeRequest. Non-empty only in tests.
+	apiBaseURL string
 }
 
 // Config holds CampMinder configuration
@@ -315,7 +319,11 @@ func (c *Client) makeRequest(method, endpoint string, params map[string]string) 
 	}
 
 	// Build URL with parameters
-	fullURL := fmt.Sprintf("%s/%s", baseURL, strings.TrimPrefix(endpoint, "/"))
+	base := baseURL
+	if c.apiBaseURL != "" {
+		base = c.apiBaseURL
+	}
+	fullURL := fmt.Sprintf("%s/%s", base, strings.TrimPrefix(endpoint, "/"))
 
 	if len(params) > 0 && method == "GET" {
 		values := url.Values{}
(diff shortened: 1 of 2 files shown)


adamflagg:
### Independent review (floor 60)

Scanned the diff (`pocketbase/campminder/client.go`, `client_test.go`) directly — read the full diff, git blame/history on `GetSessions`/`GetSessionGroups`, the neighboring `*Page`/`hasMore` pagination pattern used elsewhere in this file (`GetAttendeesPage`, `GetBunkPlansPage`, `GetStaffPage`, etc.), the sync-side call sites (`sync/sessions.go`, `sync/session_groups.go`, `sync/api.go`), and issue #2437's body. No findings survived verification — nothing to fix or file.

Checks performed:
- `go build ./...` clean, `golangci-lint run ./campminder/...` — 0 issues.
- Full `go test ./campminder/...` (all tests, not just the new ones) — all pass.
- Loop-continuation condition (`len(response.Results) == 0 || len(all) >= response.TotalCount`) mutated to `false` and re-run against `TestGetSessions_SinglePageUnderLimit`: the mutant hangs (infinite pagination loop against the mock server) rather than terminating, confirming the condition is load-bearing. Reverted; working tree matches `origin/feature/campminder-pagination`.
- Considered whether this should have used the codebase's existing `Get<X>Page(page, pageSize) (results, hasMore, err)` + caller-owned-loop pattern (used for `GetAttendeesPage`/`GetBunkPlansPage`/etc.) instead of looping internally and returning the full accumulated slice. Rejected as a finding: `GetSessions`/`GetSessionGroups` already had a full-slice signature pre-fix with a single call site each expecting the complete result, not page-by-page streaming — switching to the `Page` pattern would be a larger, unrelated API change to fix a silent-drop bug, not a like-for-like reuse.
- The `//nolint:dupl` duplication between the two loops is pre-existing (it covered the single-page code before this PR) and stays intentional.

CodeRabbit: not requested, per campaign policy. It ran automatically anyway and reported "Review skipped — Auto reviews are disabled on this repository" (no-op, not a real review).

CI Summary: green. `mergeStateStatus`: CLEAN.

PR #2448, opened 2026-08-18
fix(sync): admit the adult weekends' WW- prefix into financial aid applications

### LINKED ISSUE #2436: bug(sync): the HH-/FA-/CA- prefix rules miss the adult weekends' WW- prefix, dropping 24 aid requests in 2026
`household_demographics` and `financial_aid_applications` select their source fields by name prefix.
The prefix lists cover the summer/camper programs and miss the adult weekends' `WW-` prefix, so those
programs' demographic answers and financial-aid requests are dropped on the floor.

Found by a read-only audit during the Phase E triage campaign.

## Where

- `pocketbase/sync/household_demographics.go:347` — `strings.HasPrefix(name, "HH-")`, with the same
  test restated at `:363` (`IsHHField`) and the load filter at `:451`.
- `pocketbase/sync/financial_aid_applications.go` — the equivalent `FA-` / `CA-` prefix rules.

Neither admits `WW-`.

## Measured

**24 financial-aid requests in 2026 and 63 in 2025** carry the `WW-` prefix and reach no column.
Grain: person-year custom values, production snapshot.

## Why it is worth fixing

This is money and eligibility data, not a display nicety. It is also cheap: the change is extending
two prefix predicates and their tests, with no schema change and no new field.

⚠️ Check whether the `WW-` answers belong in the existing `_summer` columns or need their own — the
demographics table already splits summer from other answers, and putting an adult-weekend answer in a
`_summer` column would be a grain error, not a fix. That is the one decision in this issue.

- **Files:** `pocketbase/sync/household_demographics.go`, `pocketbase/sync/financial_aid_applications.go` (+ tests)

adamflagg:
`isFAFieldName` in `financial_aid_applications.go` admitted `FA-` fields by prefix and a fixed
`CA-` allowlist, but nothing else. The adult weekends' four `WW-` fields
(`WW-FA`, `WW-FA Amount`, `WW-Donation Amount`, `WW-Donation Amount Other`) matched neither rule,
so `loadFieldDefinitions` never mapped their `field_definition` IDs, and
`loadPersonCustomValues`'s `if !ok { continue }` threw every `WW-` answer away before
`mapFieldToApplication` ever saw it.

Measured against the production snapshot (`custom_field_defs` joined to `person_custom_values`,
non-empty values): **`WW-FA Amount` carries 63 answers in 2025 and 24 in 2026** — the exact
numbers in #2436. The other three `WW-` fields carry comparable volume per year (300-470 for
`WW-FA`/`WW-Donation Amount`, single digits for `WW-Donation Amount Other`), all dropped the same
way.

## The grain decision this issue called out

The `WW-` fields are structurally identical (interest indicator / amount awarded / donation
amount / donation-other) to the four `CA-` fields already routed by this file
(`CA-FinancialAssistanceInterest`, `CA-FinancialAssistanceAmount`, `CA-Donation amount`,
`CA-Donation other`), and land on the same four destination columns:
`interest_expressed`, `amount_awarded`, `donation_preference`, `donation_other`.

Checked directly against the table schema (`financial_aid_applications`): none of those four
columns carry a `_summer` suffix or any other program qualifier — unlike this same table's
`summer_amount_requested` / `fc_amount_requested` / `tbm_amount_requested`, which *are*
program-scoped and which this change does not touch. Admitting `WW-` values into the
program-neutral columns does not change what any existing row means: the grain here is one row
per (person, year), and a `WW-` answer only ever lands on the row for the adult who filled out
that weekend's form. This is admission into a column that already accepts a second, differently
prefixed writer (`CA-`) — not a grain change, and not a new column.

This satisfies the issue's own decision rule ("if the destination column is already
program-neutral... BUILD IT"), so this PR builds it. No PocketBase migration, no new column.

## `household_demographics.go` — investigated, not changed

The issue named `household_demographics.go` as a second file to check. Checked the production
`custom_field_defs` table for every `WW-`-prefixed field: there are exactly four, and all four are
the financial-aid/donation fields fixed above — none is a demographics-shaped question (no
`WW-Family Description`, `WW-Jewish Affiliation`, etc.). `IsHHField`/`MapHHFieldToColumn` in that
file have nothing to admit for this issue, so it is untouched here. (Grain safety for that file's
own `_summer` columns was already established by #2260, unrelated to this issue.)

## Testing

`TestFinancialAidApplicationsAdmitsWWFields` is the regression test: it pins that
`loadFieldDefinitions` admits all four `WW-` field names and that `mapFieldToApplication` routes
each to its column, using the same `newFieldDefsTestApp` fixture pattern as the existing
`kindred#1873` regression test in this file. Written first, confirmed red (all four
`loadFieldDefinitions` assertions and all four `mapFieldToApplication` assertions failing), then
made green by extending `isFAFieldName` and `mapFieldToApplication` with the `WW-` allowlist
mirroring the `CA-` arm. Mutation-checked afterward by changing the `WW-` prefix branch to a
non-matching prefix and re-running: the test failed as expected (all four fields unadmitted),
confirming it is not vacuously green — then restored and re-confirmed green.

`go build ./...`, `golangci-lint run`, and `go test ./sync/...` (scoped) all pass clean. The
unscoped `pre-push-verify.sh` run's `go test ./...` hit the go test framework's own 10-minute
per-package timeout inside the `sync` package under this run's 9-way-parallel-agent CPU
contention — the timeout fired inside unrelated pre-existing tests (`TestAliasResolverFlagsOverlappingWindows`,
lodging assignment sync tests), not this change's tests, both of which had already passed cleanly
moments earlier in the same run.

## Deliberately NOT done

- No change to `household_demographics.go` (see above — nothing there to admit).
- No PocketBase migration and no new columns — the whole point of the grain check above.
- Did not touch the pre-existing first-non-empty-wins collision behavior between `CA-` and other
  fields sharing these columns (one person+year in the whole production snapshot has both a `CA-`
  and a `WW-` value); that ordering is this file's existing design for every field, not something
  introduced or changed by this fix.

Closes #2436.

### DIFF
--- a/pocketbase/sync/financial_aid_applications.go
+++ b/pocketbase/sync/financial_aid_applications.go
@@ -22,6 +22,14 @@ const (
 	faFieldSummerQuestAmtRequested       = "Summer/Quest: Amt Requested"
 	faFieldFamilyCampAmtRequested        = "Family Camp: Amt Requested"
 	faFieldBnaiMitzvahAmtRequested       = "B'nai Mitzvah: Amt Requested"
+	// WW- fields are the adult weekends' equivalent of the CA- interest
+	// indicators above -- same four-question shape (interest / amount
+	// awarded / donation amount / donation other), routed to the same four
+	// program-neutral columns. See kindred#2436.
+	faFieldWWFinancialAssistanceInterest = "WW-FA"
+	faFieldWWFinancialAssistanceAmount   = "WW-FA Amount"
+	faFieldWWDonationAmount              = "WW-Donation Amount"
+	faFieldWWDonationOther               = "WW-Donation Amount Other"
 	// Boolean string parsing constants (shared with other sync files)
 	boolYes = "yes"
 )
@@ -306,6 +314,18 @@ func isFAFieldName(name string) bool {
 		return false
 	}
 
+	// WW- prefix: the adult weekends' equivalent of the CA- interest
+	// indicators, same fixed allowlist shape (kindred#2436).
+	if len(name) >= 3 && name[:3] == "WW-" {
+		if name == faFieldWWFinancialAssistanceInterest ||
+			name == faFieldWWFinancialAssistanceAmount ||
+			name == faFieldWWDonationAmount ||
+			name == faFieldWWDonationOther {
+			return true
+		}
+		return false
+	}
+
 	// Special amount requested fields without FA- prefix
 	if name == faFieldSummerQuestAmtRequested ||
 		name == faFieldFamilyCampAmtRequested ||
@@ -471,6 +491,27 @@ func (s *FinancialAidApplicationsSync) mapFieldToApplication(app *faApplicationD
 			app.donationPreference = value
 		}
 	case faFieldCADonationOther:
+		if app.donationOther == "" {
+			app.donationOther = value
+		}
+
+	// Interest indicators (WW- prefix, adult weekends -- kindred#2436). Same
+	// four program-neutral columns as the CA- arms above: a WW- answer only
+	// ever lands on the row for the adult who filled out that form, so this
+	// is admission into an existing column, not a grain change.
+	case faFieldWWFinancialAssistanceInterest:
+		if !app.interestExpressed {
+			app.interestExpressed = hasInterestExpressed(value)
+		}
+	case faFieldWWFinancialAssistanceAmount:
+		if app.amountAwarded == 0 {
+			app.amountAwarded = parseNumberValue(value)
+		}
+	case faFieldWWDonationAmount:
+		if app.donationPreference == "" {
+			app.donationPreference = value
+		}
+	case faFieldWWDonationOther:
 		if app.donationOther == "" {
 			app.donationOther = value
 		}
--- a/pocketbase/sync/financial_aid_applications_test.go
+++ b/pocketbase/sync/financial_aid_applications_test.go
@@ -47,6 +47,79 @@ func TestFinancialAidApplicationsLoadFieldDefinitionsTrimsNames(t *testing.T) {
 	}
 }
 
+// TestFinancialAidApplicationsAdmitsWWFields is the regression test for
+// kindred#2436. The adult weekends' four WW- financial-aid fields
+// (WW-FA, WW-FA Amount, WW-Donation Amount, WW-Donation Amount Other) were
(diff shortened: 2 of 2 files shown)


adamflagg:
### Independent review (Phase F Part B)

Reviewed the diff (`isFAFieldName`, `mapFieldToApplication`, and the new test) directly rather than
via the 5-agent code-review skill pipeline, since the change is a single 41-line addition to one
file plus its test — full context was already loaded from reading the PR.

**Verified independently against `data-prod.db`:**
- `custom_field_defs` has exactly 4 `WW-`-prefixed rows, all financial-aid/donation shaped
  (`WW-FA`, `WW-FA Amount`, `WW-Donation Amount`, `WW-Donation Amount Other`) — confirms the
  `household_demographics.go` "nothing to admit" claim; that file's `IsHHField`/`isHouseholdDemographicsField`
  logic is correctly left untouched.
- `WW-FA Amount` non-empty answers: 63 in 2025, 24 in 2026 — matches the issue and PR body exactly.
- `amount_awarded`, `interest_expressed`, `donation_preference`, `donation_other` carry no
  `_summer`/`_fc`/`_tbm` suffix in the table schema — confirms the program-neutral-column claim, so
  this is admission into an existing column, not a grain change.
- The pre-existing CA-/WW- collision (first-non-empty-wins) affects exactly 1 person+year in the
  whole snapshot (2022) — confirms it's rare, pre-existing, and untouched by this PR.
- No other `.go`/`.py`/`.ts`/`.tsx` file references the `CA-`-prefixed literals this file owns, so
  there's no other place silently dropping `WW-` data downstream of this fix.

**Mutation-checked the new test myself:** flipped the `WW-` prefix match to `XX-`, reran
`TestFinancialAidApplicationsAdmitsWWFields` — failed with all 4 fields unadmitted as expected;
reverted and reran green. Confirms the test isn't vacuous.

`go build ./...` and `golangci-lint run` both clean on the touched package.

**Findings: none survive.** The change is a straight mirror of the existing `CA-` allowlist arm in
both `isFAFieldName` and `mapFieldToApplication`, uses the same shared helpers
(`hasInterestExpressed`, `parseNumberValue`), introduces no new columns/migrations, and the test
uses fictional names. No fixes applied, nothing filed.

CodeRabbit not requested (wave budget reserved for the highest-blast-radius PR). CI Summary: SUCCESS.

Issue #2337, opened 2026-08-14
test(sync): captureSweepLogs swaps the process-global slog default inside five t.Parallel tests


adamflagg:
## What happens

`captureSweepLogs` (`pocketbase/sync/orphan_sweep_test.go:285-294`) calls the **process-global**
`slog.SetDefault` with a handler writing into a per-test `strings.Builder`, and all five of its
callers run under `t.Parallel()`:

- `TestBaseDeleteOrphansWarnsWhenNothingCanBeKeyed` (`orphan_sweep_test.go:302`)
- `TestBaseDeleteOrphansCountsOnlyCompletedDeletes` (`orphan_sweep_test.go:352`)
- `TestIsDuplicateStaffStatus` (`staff_test.go:145`)
- `TestLoadPersonCustomValues*` (`camper_transportation_test.go:791` and `:864`)

Two failures follow.

1. **Every other concurrently-running test in the package logs through `slog.Default()`**, so its
   records are appended to whichever `strings.Builder` happens to be installed — and
   `strings.Builder` is not thread-safe. That is what `-race` trips on. `base_sync.go` alone has 32
   slog sites on the sweep path, and `sync_runs.go:92-95` warns on every `recordSyncRun` with
   `Status.Year == 0` and no season configured, which is most parallel orchestrator tests.
2. **The save/restore is LIFO-unsafe.** Two overlapping captures restore in completion order, so one
   test's assertions can read another test's log, or an installed handler can outlive its buffer's
   test.

This almost certainly dates to kindred#2288, which parallelised the sync suite.

## Reproduction

1 of 4 attempts reproduced. The failing run was:

```
go test -race -count=1 ./sync/ -run 'TestBaseDeleteOrphans|TestIsDuplicateStaffStatus|TestLoadPersonCustomValues|TestSyncRun|TestBatchesAreSequential|TestConcurrentBatches'
```

It failed 8 tests with "race detected during execution of test":
`TestBaseDeleteOrphansRejectionLeavesGenuineOrphansBehind`,
`TestBaseDeleteOrphansGuardedKeepsTheRejectedRecordsRow`,
`TestBaseDeleteOrphansGuardedRefusesPartialCollapse`,
`TestBaseDeleteOrphansGuardedStillSweepsNormalChurn`,
`TestLoadPersonCustomValuesCountsStaffGateDrops`,
`TestLoadPersonCustomValuesCountsUnmappedFields`,
[...]

PR #2450, opened 2026-08-18
fix(sync): collapse a household's medical answers by total aggregation, not first-wins

adamflagg:
`processMedical` folded every person's family-camp medical answer into one household row by keeping
the first value it saw per field. `personValues` carries one entry per person per field, and
CampMinder asks the family-camp questions on a per-CAMPER form, so a household with two enrolled
children answers each question twice — and every answer after the first was discarded before any of
the per-column logic ran. That is why staff saw part of a household's medical picture and not all of
it.

Measured on the production snapshot: **105 of 464 family-camp households in 2026 (22.6%) lost at
least one real second disclosure** — narrative fields only, nullish values excluded; 121 of 464 if
you count all sixteen medical source fields — and the all-year range is 20–28% back to 2017. And because the
gate ("Yes"/"No") and its explanation are separate fields, they were selected independently, so
**836 of 8,864 stored rows read `"No; <a description of the condition it denies>"`** — 243 of the
330 allergy cases are genuine cross-attribution, one person's denial glued to another person's disclosure.

## What changed

The flatten is replaced by **total aggregation on both sides**, which is what binds a gate to its
explanation: after this change neither half has a "winner" that could come from a different person.

- The per-field map now accumulates into `answerSet` — the reviewed dedup-and-join shipped for
  `family_camp_registration_text.go` — so the common case (one parent's answer fanned onto every
  child's form) still collapses to one, while two genuine disclosures both survive, joined in a
  canonical order that does not depend on record-id load order.
- **A binary gate collapses by OR, not by join.** Joining two people's gate answers verbatim renders
  `"No; Yes; <A>; <B>"`, which is why an earlier uniform dedup-and-join across every field was
  reverted. `medicalGateFields` names the four gates `processMedical` reads and
  `medicalAnswers.parts` returns `Yes` if anyone in the household said `Yes` — the same aggregation
  `processRegistrations` already applies to the housing flags. A gate whose answers are all negative
  falls through to the join, so a vocabulary that ever widens past Yes/No is not silently narrowed.
- Each column is now capped at its **real** declared max, which is not uniform: 10,000 for
  `allergy_info` / `dietary_info` / `special_needs_info` / `additional_info`, 5,000 for `cpap_info`
  and `physician_info`, 4,000 for `bathroom_explain` and `accommodation_explain`. Joining is what
  makes overflow possible for the first time, and `record.Set()` past a PocketBase field's max fails
  the whole row's save — losing a household rather than a sentence. `joinAnswers` drops whole
  answers rather than cutting one in half and the count is logged; the values never are.
  An undeclared column falls back to keeping everything and logging an error, because a zero limit
  would otherwise blank the column silently.

- **The CPAP column is unioned across all three of its fields, then filtered.** It used to stop at
  the first of the two camper-partition names that had an answer, so a `"No"` on `Family Camp-CPAP`
  hid a real disclosure on `FAM CAMP-CPAP` outright — 27 households in 2025 and 1 in 2026, each
  carrying a `needs_power` / `needs_private_bathroom` flag that `processRegistrations` ORs from all
  three fields and a `cpap_info` that then denied it. Union-and-dedup is the same total aggregation
  `medicalAnswers` applies one level up, and dedup is what keeps the two generations of one question
  from printing twice.
- **The CPAP column keeps the gate rule by a different mechanism.** Its three fields cannot be ORed
  to one token: they are multi-option selects whose affirmative options say WHICH accommodation is
  needed (#1875), so two different affirmatives are two different needs. `cpapGateParts` therefore
  applies only the half of the OR that survives that — a pure denial is dropped whenever anyone in
  the household disclosed, both affirmatives survive, and a unanimous "No" is still kept. Without it
  the household aggregation renders `"No; Yes, outlet needed for CPAP machine"` on the one pair
  `docs/reference/family-camp-field-provenance.md` §4 names, alongside Special Needs, as doing real
  harm when the halves split: 1 household-year in 2026 and 1–2 a year back to 2022. No
  classification changes — `needs_power` and `needs_private_bathroom` come from `classifyCPAPAnswer`
  over the raw values and are untouched.

## Tests

Nine new tests against the production `processMedical` — both answerers' narratives survive, the
gate ORs so `"No; <narrative>"` becomes `"Yes; <narrative>"`, output is byte-identical when the
input slice is reversed (the order-independence probe, which does not pin which answerer wins), one
answer fanned onto three siblings collapses to one, a joined column stays inside its cap, and the
four CPAP cases: a denial beside a disclosure, two different needs, a unanimous denial, and an
answer under each of the two camper-partition field names. Two
more cover the undeclared-column fallback and pin the limits map against the migrations.

`TestMedicalColumnLimitsMatchTheSchema` READS the migrations. It first restated the same literals a
second time, which pins nothing — a migration narrowing a column would leave the map too large, the
exact direction that fails a whole row's save, and the test would still be green. It now scans every
migration mentioning `family_camp_medical` in file-number order, takes the last declared `max` per
column and compares both ways. Mutation-checked by setting `bathroom_explain` to 9999: red, then
green on restore.

`TestMedicalDeduplicationByHousehold` gained a note saying what it does not cover: it drives a
test-local reimplementation and asserts only non-emptiness, so it is blind to this defect by
construction and could not have caught it.

**It is the sync half only, and the issue stays open.** The owner ruled that the gates come out of
the narrative columns into their own nullable boolean columns, and that a YES renders a chip on the
family details panel while a NO and a null render nothing. That needs a migration,
`HouseholdMedicalResponse`, `MEDICAL_NARRATIVE_FIELD_NAMES`, regenerated `pocketbase-types.ts` and
the existing `AccessibilityFlagList` chips gaining a click — none of which is in this PR. The gate
token therefore still rides in the narrative column, exactly as before; it is simply no longer
capable of contradicting the text beside it, on any of the columns — the gates through
`medicalGateFields`, CPAP through `cpapGateParts`.

Nothing here forecloses that work: the gate values are now aggregated in one named place, which is
where the boolean will be read from.

Also untouched: `processAdults` (its merge policy was ruled unchanged, and the answers it discards
have been recorded in `attribute_conflicts` since #2421), `classifyCPAPAnswer` itself and its two
"multi-option" comments, and the CPAP gate's own boolean column — #2255 still carves that out for
its own pass.

Part of #2257.
Refs #2255.
Refs #2275.

## Summary by CodeRabbit

* **Bug Fixes**
  * Medical information now retains distinct answers from all household members.
  * Affirmative medical needs are combined accurately, including CPAP-related responses.
  * Duplicate answers are removed, and results no longer depend on response order.
  * Narrative responses are combined within field limits without splitting individual answers.
  * Missing field limits fail safely while recording an error; over-limit content reports discarded answers.

* **Tests**
  * Added coverage for multi-person medical responses, deduplication, ordering, field limits, and CPAP scenarios.

### DIFF
--- a/pocketbase/sync/family_camp_derived.go
+++ b/pocketbase/sync/family_camp_derived.go
@@ -1756,27 +1756,28 @@ func (s *FamilyCampDerivedSync) processMedical(personValues []customValueEntry) 
 			householdPBID: householdID,
 		}
 
-		// CPAP info. The two Camper-partition names are generations of the SAME
-		// question, so they collapse to one answer; Adult-CPAP is a DIFFERENT
-		// PERSON and is therefore additive.
+		// CPAP info, unioned across all three fields and deduplicated.
 		//
-		// That split has to match the flag logic in processRegistrations, which
-		// ORs across all three. A single first-wins loop over all three -- what
-		// this was -- meant a household with a camper "outlet needed" answer and
-		// an adult "bathroom ... needed" answer raised BOTH flags while the
-		// medical record kept only the camper's sentence, leaving
-		// needs_private_bathroom with nothing behind it in the one place staff
-		// can look for the reason.
+		// The union is what has to match the flag logic in processRegistrations,
+		// which ORs across all three: any answer that can raise needs_power or
+		// needs_private_bathroom has to be readable in the one place staff can
+		// look for the reason. Dedup is what keeps the union honest -- the two
+		// Camper-partition names are generations of the SAME question, and
+		// answering both must not print the sentence twice.
+		//
+		// It replaces a first-non-empty-wins loop over the two camper names,
+		// which stopped at whichever had an answer: a "No" on Family Camp-CPAP
+		// hid a disclosure on FAM CAMP-CPAP outright, in 27 households in 2025
+		// and 1 in 2026 on the production snapshot, each of them carrying a
+		// housing flag their cpap_info then denied. Taking one field's answer as
+		// the household's is the same first-wins mistake medicalAnswers exists
+		// to end, one level up.
 		cpapParts := []string{}
-		for _, key := range []string{fieldFamilyCampCPAP, fieldFamCampCPAP} {
-			if p := fields.parts(key); len(p) > 0 {
-				cpapParts = append(cpapParts, p...)
-				break
-			}
-		}
-		for _, v := range fields.parts(fieldAdultCPAP) {
-			if !slices.Contains(cpapParts, v) {
-				cpapParts = append(cpapParts, v)
+		for _, key := range []string{fieldFamilyCampCPAP, fieldFamCampCPAP, fieldAdultCPAP} {
+			for _, v := range fields.parts(key) {
+				if !slices.Contains(cpapParts, v) {
+					cpapParts = append(cpapParts, v)
+				}
 			}
 		}
 		cpapParts = cpapGateParts(cpapParts)
--- a/pocketbase/sync/family_camp_derived.go
+++ b/pocketbase/sync/family_camp_derived.go
@@ -1598,7 +1598,10 @@ var medicalColumnLimits = map[string]int{
 // "Family Camp-CPAP" is a true binary gate too and is deliberately absent. The
 // CPAP column is fed by three fields of two different shapes -- the other two
 // are multi-option selects whose text says WHICH accommodation is needed -- and
-// kindred#2255 carves it out for its own pass.
+// kindred#2255 carves it out for its own pass. Absent from this map is NOT
+// absent from the rule: cpapGateParts drops that column's denials by the half
+// of the OR that survives multi-option answers, so no household's row renders a
+// gate contradicting the need beside it.
 var medicalGateFields = map[string]struct{}{
 	"Family Medical-Allergies":     {},
 	"Family Medical-Dietary Needs": {},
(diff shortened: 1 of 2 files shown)


adamflagg:
@coderabbitai review

adamflagg:
### Review pass (independent second reader), score floor 60

Route: the project's `code-review` skill would not run its own agents here, so I did the review
directly — full diff, both files at full length, the two migrations that declare the columns, the
`answerSet` / `joinAnswers` precedent in `family_camp_registration_text.go`, `classifyCPAPAnswer`
and the `processRegistrations` flag path, `docs/reference/family-camp-field-provenance.md` §4, and
every field's real answer vocabulary measured against the production snapshot. Then CodeRabbit.

---

#### Fixed in-branch — 4 findings

**1. `cpap_info` could render a denial in front of the need it denies.** The PR body claimed the
column was "no longer capable of contradicting the text beside it". It was, for CPAP. The three
CPAP fields are deliberately not in `medicalGateFields` (their affirmatives are multi-option and
say *which* accommodation is needed, #1875), so the new household aggregation kept every distinct
answer and a household with one `No` and one `Yes` produced `"No; Yes; <the need>"` — the exact
rendering the gate OR exists to prevent, arriving through the one column the OR does not cover, on
the pair §4 of the provenance doc names as one of the two where the split does real harm.

Reproduced before the fix: `cpapInfo = "No; Yes; needs an outlet overnight"`. `cpapGateParts`
applies the half of the OR that survives multi-option answers — denials drop when anyone disclosed,
two different affirmatives both survive, a unanimous `No` is kept. No classification changes;
`needs_power` and `needs_private_bathroom` come from `classifyCPAPAnswer` over the raw values.
1 household-year in 2026, 1–2 a year back to 2022.

**2. `TestMedicalColumnLimitsMatchTheSchema` pinned nothing** (CodeRabbit agreed, minor). It said it
pinned the map "against the migrations" and in fact restated the same eight literals a second time —
it could not fail if a migration narrowed a column, which is the direction that fails a whole row's
save. It now reads every migration mentioning `family_camp_medical` and requires **every**
declaration to agree with the map (not last-wins: a regex over JS cannot tell which collection or
which direction of `migrate()` a field literal sits in, so last-wins could let a same-named field
stand in for a narrowed medical limit).

Mutation-checked twice — `bathroom_explain` set to 9999 in the map: red. Decoy `max: 123` planted in
`1500000126`: red, naming the file. Green on restore both times.

**3. The camper CPAP `break` hid real disclosures** (CodeRabbit, major — verified and confirmed).
`processMedical` stopped at whichever of `Family Camp-CPAP` / `FAM CAMP-CPAP` had an answer, so a
`No` on the first hid a disclosure on the second outright: **27 households in 2025 and 1 in 2026**
on the production snapshot. `processRegistrations` ORs all three fields into `needs_power` /
`needs_private_bathroom`, so each of those households carried a housing flag whose `cpap_info`
denied it — the reason staff can look up being empty is precisely what the flag/narrative split
exists to prevent. All three fields are now unioned and deduplicated;
`TestProcessMedicalCollapsesCamperCPAPGenerations` still pins the no-double-printing half.

**4. Three PR-body claims were false** and the body is the squash commit message, so it was
corrected: "CPAP is deliberately untouched" (the diff already changed its collapse), "no longer
capable of contradicting the text beside it", and "pin the limits map against the migrations".

Five new tests: a denial beside a disclosure, two different needs, a unanimous denial, an answer
under each camper field name, and the rewritten schema pin.

---

#### Verified, not findings

- **The four gates are correctly classified.** All of `Family Medical-Allergies`,
  `Family Medical-Dietary Needs`, `Family Camp-Special Needs` and `Family Camp-Physician` have a
  vocabulary of exactly `{Yes, No}` across every stored answer. None of the six narrative fields
  they pair with contains a bare Yes/No.
- **`Housing-Bathroom` is not a gate** and correctly stays out of `medicalGateFields` — it is free
  text; its gate is `FAM CAMP-bathroom` / `Adult-Bathroom`, which `processMedical` does not read.
  Same for `Housing Accommodation-Yes`.
- **The trailing space on `Family Camp-Physician `** does not break the new map lookups —
  `normalizeFieldName` trims at the load boundary.
- **All eight column limits match the live schema**, checked against `_collections` as well as the
  migrations.
- **Order-independence holds** through `answerSet`'s canonical sort; the case-collision rule keeps
  `Yes` over `yes` deterministically.
- `parts()`'s `slices.Clone` is genuinely needed — callers append across fields.

---

#### Filed as out of scope — #2451

`pocketbase/lodging` swaps the package globals `registryBasePath` and `registryAbsoluteRoots` from
`t.Cleanup` while parallel tests read them through `registryCandidatePaths`. Caught it under
`-race` on this branch (whose diff is confined to `pocketbase/sync`), twelve tests down, and once as
a *wrong answer* rather than a race report (`0 units after the retry; want 2`). Same class as the
already-fixed #2337. Filed rather than fixed because the three available fixes are not equivalent —
dropping `t.Parallel()` is a CI-shard budget decision, a mutex silences the detector without fixing
the cross-test leak, and threading the path through touches production signatures in a package
several concurrent branches are in. A second sighting of the same class inside `./sync/`
(PocketBase's own `TestApp.Cleanup` vs `MigrationsRunner`) is recorded as a comment on #2451 rather
than a second issue.

That flake is what reddened `Go Tests (2)` on the first run of `fd6b077`; `gh run rerun --failed`
went green on all three shards and CI Summary is green.

---

Gates: `golangci-lint run ./sync/...` clean (CI-only, so run explicitly), `go vet` clean,
`bash scripts/pre-push-verify.sh` green unpiped, full `./sync/` suite green. Push confirmed by the
remote ref. `closingIssuesReferences` is empty by design — #2255's remaining acceptance boxes are
the owner-ruled UI half, and I agree with leaving it open rather than retiring that silently.

Issue #2328, opened 2026-08-14
feat(frontend): give lodging units a short display name so compact surfaces stop truncating them


adamflagg:
## ⚠️ CORRECTED 2026-08-14 — phase C triage

Verified against `main` @ `3eb44ba4` and the production snapshot. Verdict: **SUPERSEDED** / DECISION.

> #2332 covers this entirely — its decision 1 is verbatim this issue's remedy ('curated in the registry as a `short_name` alongside `name`'), and #2332 adds the board surface, the CampMinder-side spellings and the measured widths that decide the question. Two of the three sites this issue is filed against no longer clip: #2253 shipped in 474ce260, so the journey card's row AND its header now wrap and carry no `truncate`. The registry file genuinely has no `short_name` key, so the proposal itself is not already built.

### Corrections to the body below

1. STALE: "#2253 is the *symptom*: `HouseholdJourneyCard.tsx:147` applies `truncate` to the year row … Its fix directions are all layout (wrap, widen, tooltip)." REPLACE WITH: "#2253 has SHIPPED (474ce260, PR #2314): the year row now wraps and carries no `truncate`. That remedy is already in the tree, so the choice this issue framed ('choose between the two deliberately') is settled in favour of layout on the journey card — what remains is whether a short name is ALSO wanted, which is #2332's decision 1."
2. STALE: the bullet list of clipped sites. `HouseholdJourneyCard.tsx:147` and `:222` no longer clip — both wrap as of #2314, and a test at HouseholdJourneyCard.test.tsx:304 pins the header. The surviving clip sites are `LodgingUnitCard.tsx:777` (`truncate text-lg`) and `MapUnitPopover.tsx:611`.
3. ADD at the top: "SUPERSEDED BY #2332, which asks the same question with the widths measured on both surfaces and covers the CampMinder-side spellings this one does not. Carry the ⚠️ 'do not call it an alias' warning across before closing — #2332 does not contain it, and `lodging_unit_aliases` (1500000117, seeded by 1500000121, 187 rows) really does already own that word."

<details><summary>Re-measured figures (denominator and scope stated)</summary>

- Registry holds 118 unit names; 16 exceed 20 characters (the body's figure — I did not re-measure the median/longest, but I confirmed the file has 118 units and no short_name key). Board card names still truncate at LodgingUnitCard.tsx:777.
[...]

Issue #2393, opened 2026-08-15
feat(lodging): the family history modal needs session grain — it silently collapses multi-weekend years


adamflagg:
## ⚠️ CORRECTED 2026-08-16 — phase D triage.

Verified against `main` @ `a946565a` and the **production** snapshot.

**Every figure in this body reproduced exactly** — 64 of 5,438 multi-weekend rows, and 7 of those 64 with a
child who did not attend every weekend. Re-verified 2026-08-17 against
`pocketbase/pb_data/data-prod.db`. That is unusual in this backlog and worth stating.

⚠️ **STATE THE DENOMINATOR EVERY TIME.** This issue and #2336 both count "multi-weekend
household-years" and they are **not the same population**:

| | denominator | multi-weekend |
|---|---|---|
| **#2393 (this issue)** | **5,438** — every journey household-year, i.e. the union of the three traces `build_household_journey` reads (enrolled family attendees ∪ `family_camp_adults` ∪ `family_camp_registrations`), all years, `year > 0` | **64 (1.2%)** |
| #2336 | 1,703 — household-years that carry a non-blank `cabin_assignment` AND at least one enrolled family attendee, 2022-2026 | 41 (2.4%) |

Both reproduce. Mixing them in a PR body would misstate the blast radius by a factor of three, so
name the denominator wherever the figure appears.

### One stale anchor

`AttributeSession` lives at **`pocketbase/sync/lodging_session_attribution.go:327`**, not
`lodging_assignments_sync.go:463`. That second line is the *caller's* `return // flag, do not guess` — the
quoted phrase is there, the function is not.

### It is INDEPENDENT of #2336

#2336 has been reduced to documentation only (no schema change) and folds into **#2332**, whose files it
shares. **#2393 touches neither of #2336's comment anchors**, so it neither blocks on nor closes #2336. Do
not carry `Closes #2336.` here.

### Scheduling

**#2393 and #2332 share six files and cannot run in parallel.** Sequence #2332 first — it is the larger change
and establishes the shared name-resolution helper this issue's modal should then use rather than re-deriving.

Closes #2393. Does **not** close #2336.

---

## ★ OWNER REQUEST, 2026-08-15

> relatedly, we need the actual session in the family history modal, right now it just has year and housing
> but no session info, so A) unhelpful and B) what happens if a family went to multiple sessions. it needs
> session grain.

## Confirmed — and the "what happens" answer is: it silently collapses

[...]

PR #2453, opened 2026-08-18
feat(api,frontend): render housing in today's language on the journey card and the family card

### LINKED ISSUE #2332: feat(lodging): one housing-name display convention for the journey card and the board — 32% of journey rows overflow their width
## ⚠️ CORRECTED 2026-08-16 — phase D triage.

Verified against `main` @ `a946565a` and the **production** snapshot.

### 1. ⛔ The evidence for this issue is stronger than the body knows: staff already renamed 14 units

On **2026-08-15, 03:45:01Z-03:46:47Z**, fourteen units were renamed one at a time in the admin GUI. The
registry file never got the edit:

| unit code | `config/lodging_registry.json` | production `lodging_units.name` |
|---|---|---|
| `hc-upstairs-1` … `-6`, `hc-upstairs` | `Health Center Upstairs …` | `HC Upstairs …` |
| `hc-downstairs`, `-a`, `-b` | `Health Center Downstairs …` | `HC Downstairs …` |
| `hc-doctors-house-a` / `-b` | `Doctor's House A` / `B` | `Doctor's House 1` / `2` |
| `gt-le-shack` | `Le Shack` | `L'Shack` |
| `manzanita-trailer-4` | `Trailer 4 (manzanitas)` | `Trailer 4 (Manzanitas)` |

**This is the ruling in action** — whatever staff name a unit in the GUI is what should show everywhere. It
also **closed #2328** (the "short display name" ask) as premise-false: staff solved it with the field they
already edit.

### 2. ⛔ A THIRD comment must change, not two

The ruling names two comments asserting the opposite. There is a third, and it is the most important because
it is on `cabin_name` itself — `api/schemas/lodging.py:708-712`: *"The staff-written free text out of
`family_camp_registrations`, NOT resolved to a `lodging_units` row"*. The other two are
`api/services/lodging_repository.py:873-881` (inside
`fetch_cabin_assignments_by_household_cm_id`, which starts at `:870`) and
`api/schemas/lodging.py:580-590` (on `last_year_cabin`). All three quoted texts are still exact —
**only the line numbers moved**, re-derived against `main` @ `1797f9bd`.

### 3. Read sites — both trace to one repository read, so one shared helper is achievable

Re-derived against `main` @ `1797f9bd`:

- `cabin_name`: produced `lodging_roster_service.py:1492`, typed `api/schemas/lodging.py:712` (on
  `HouseholdJourneyYear`, `:698-718`), rendered `HouseholdJourneyCard.tsx:81`.
- `last_year_cabin`: produced `lodging_roster_service.py:1945`, typed `api/schemas/lodging.py:590`,
  rendered `FamilyCard.tsx:314` and `:426-433`.

Both come from `fetch_cabin_assignments_by_household_cm_id` (`lodging_repository.py:870`).

**Helper's home: `api/services/lodging_rules.py`**, whose module docstring is verbatim this requirement —
*"Pure decision rules for the weekend lodging surface. No I/O … so the same rule is never re-implemented
differently in the repository, the service, and React."*

⚠️ **A new repository read is needed.** `LODGING_UNIT_ALIASES = 'lodging_unit_aliases'` exists at
`api/constants/collections.py:50` and is pinned by `tests/unit/api/test_lodging_constants.py:36`, but **no
Python code reads the collection** — only that constant, a docstring, and `scripts/dev/lib/diff_lodging_registry.py:136`.

### 4. ⛔ The "7 missing aliases" claim is wrong. Four already exist; three are genuinely absent.

Four of the seven exist carrying `valid_to_year = 2024` and are correct for the years they cover — resolving
at the registry's loaded year (2026) discards them. **A literal "add the 7 aliases" would collide on the
`(alias_string, valid_from_year)` unique index for those four**, since the existing rows carry
`valid_from_year = 0`.

Window the alias at the **row's own year** and translate members to 2026 — which is exactly what
`aliasRow.covers` (`pocketbase/sync/lodging_alias_resolver.go:23`) and `AliasResolver.Resolve`
(`:135`, applying the filter at `:140`) already do.

⚠️ **But do NOT sell year-windowing as the source of the 98.9%.** Re-measured 2026-08-17 against
`pocketbase/pb_data/data-prod.db`. Grain and coverage: the **1,861 `family_camp_registrations` rows
that carry a non-blank `cabin_assignment`**, one row per (household, year), spanning 2022-2026 —
and 2026 is a FLOOR at 75 rows, mid-registration. Resolution target is the union of
`lodging_units.name`, `lodging_units.code` and `lodging_unit_aliases.alias_string`, case-folded:

| resolver | rows resolving |
|---|---|
| alias windows ignored entirely | 1,841 / 1,861 (98.9%) |
| each alias windowed at the ROW's own year | **1,841 / 1,861 (98.9%)** |
| every alias windowed at the registry's loaded year 2026 | 1,792 / 1,861 |

So **1,792 is not "unwindowed" — it is "windowed at the wrong year".** Only 6 of 187 aliases carry a
window at all (4 with `valid_to_year = 2024`, 2 with `valid_from_year = 2025`), and each is used only
inside its own window, so windowing at the row's year and ignoring windows land on the identical
1,841. The 49-row gap buys protection against pinning the window to 2026, which discards the four
`valid_to_year = 2024` aliases. **Window anyway — because it is correct, not because it is what
lifts the rate.**

That leaves only **three** genuinely-missing aliases:
`Ridge 2` (2022, 4 rows), `River Side - R1` and `River Side - R2` (2022-2023, 8 rows each). Each names an
extant unit *family* but not a specific unit — mapping them needs staff knowledge. See #2392.

### 5. ⚠️ Naive resolution makes truncation WORSE on six strings

**Seven** distinct `cabin_assignment` strings resolve to 2+ units, covering 86 of the 1,861
cabin-bearing rows (the alias table holds 8 multi-member rows in all; one is never used by this
column). Joining their member names gives up to **35 characters** — longer than the 34-character
worst raw string, on a `whitespace-nowrap` span at `FamilyCard.tsx:426-433`. All seven resolve to
siblings under a single container: each has exactly 2 members sharing one non-empty
`lodging_units.parent_unit`, and all 15 distinct parents on the snapshot carry `is_container = 1`.
So **collapse to the parent's name**. That part of the recommendation stands.

#### THE COLLAPSE RULE — state it, do not re-derive it

> When a string resolves to **2+ units that share one non-empty `parent_unit`**, render **that parent
> unit's `name`**, never the joined member names. When it resolves to one unit, render that unit's
> `name`. When nothing resolves, render the raw string unchanged.

No new column is needed: `parent_unit` already reaches Python, published as `parent_code` at
`api/services/lodging_roster_service.py:1642`.

⚠️ **RE-CORRECTED 2026-08-17. This figure has now moved three times (595→1, then 595→61/50, then
595→64/50) and the data never moved — each pass used a different resolver. The difference is
DEFINITIONAL, so quote the RULE, not the number.** Re-measured against
`pocketbase/pb_data/data-prod.db`. Grain: the 1,861 `family_camp_registrations` rows carrying a
non-blank `cabin_assignment`, 2022-2026 (2026 is a FLOOR at 75 rows). Counting rows whose *rendered*
string exceeds 20 characters:

| rendering | rows > 20 ch | distinct > 20 ch | longest |
|---|---|---|---|
| raw `cabin_assignment` | 595 | 32 | 34 ch |
| resolved, member names joined ("naive") | **61** or **64** | 4 | 35 ch |
| resolved, **collapse rule above** | **1** | 1 | 24 ch |

Each prior figure is one definitional choice, none of them the data:

- **61 vs 64 is resolution ORDER only.** Trying an exact `lodging_units.name` / `code` match before
  the alias table gives 61; consulting the alias table first gives 64. Three rows differ. The join
  separator changes nothing — `+`, `,`, `/` and `&` all give the same count.
- **50 came from a third resolver** that evaluated every alias window at the registry's loaded year
  2026 rather than the row's own year (§4). That strands 49 rows at their raw length: 85 naive, 50
  collapsed. It measures the §4 mistake, not truncation.
- **1 is the collapse rule, and it is stable** — 1 of 1,861 under BOTH resolution orders and every
  join separator tested. The single survivor is a 24-character single-unit name that resolves to one
  unit with no parent, so no collapse can shorten it.

**So the original "595 → 1" was right, and right for a reason worth writing down: the COLLAPSE does
the work, not the resolution.** Naive resolution on its own leaves 61-64 rows over budget and makes
the worst case one character LONGER than the raw string, which is what §5's heading is about.
**2026 alone goes 15 of 75 → 0** under naive and collapse alike.

Display drift is still the larger and steadier reason to do this at all: **37 of 88 distinct strings
resolve to a different registry name, affecting 716 of 1,861 rows (38.5%)** — identical under naive
and collapse — and only **3** strings are genuinely unresolvable (not 7 — four of the original seven
already exist as aliases with `valid_to_year = 2024` and would collide on the
`(alias_string, valid_from_year)` unique index).

### 6. #2336 folds in HERE, not into #2393

#2336 is reduced to documentation + read-path (no schema change). Two of its three anchors **are** the
comments this issue must rewrite (`api/services/lodging_repository.py:873-881`,
`api/schemas/lodging.py:580-590` — same two comments, re-derived against `main` @ `1797f9bd`); #2393
touches neither. **Coupling re-verified 2026-08-17: #2336 is still OPEN and its own body still says
"Fold into #2332 with `Closes #2336.` on that PR", so carry `Closes #2336.` on this PR.**

### 7. ⏸ One open decision before implementation

Does `cabin_name`/`last_year_cabin` **change meaning in place** to hold the resolved registry name, or does a
**new field** carry it beside the raw string? The ruling keeps the raw string as provenance *and* displays the
registry name, which needs two fields, but does not say which keeps the published name. Redefining in place is
a zero-line frontend diff that silently changes a documented API contract with tests pinning it; adding a
field keeps both honest at the cost of touching every consumer and the generated types.

### 8. Stale anchors in the body below

Re-derived against `main` @ `1797f9bd` — this section's own replacements have since gone stale too:

- "family_camp_derived.go:886-887" → the first-wins copy
  (`if v.fieldName == fieldNameFamilyCampCabin && reg.cabinAssignment == ""`) is now
  **`pocketbase/sync/family_camp_derived.go:1304-1305`**; the two `Set("cabin_assignment", …)` writes
  are **:2259** and **:2284**.
- "LodgingUnitCard.tsx:777" → the `truncate text-lg` heading is now
  **`frontend/src/components/weekend/LodgingUnitCard.tsx:743`**.

### 9. ⚠️ #2372 must NOT be closed as no-code-needed

Its comment recommends closing because "the admin UI is the correct channel". That holds only while
`lodging_units.name` matters on one surface. **This issue makes it the display name on four.** `SeedRegistry`
runs on every boot (`pocketbase/main.go:109`) and is create-only, so an empty registry — a new deployment or
rebuilt CD seed — would render `Health Center Upstairs 1` while production renders `HC Upstairs 1`. `name` is
not even in `INVENTORY_FIELDS`, so no existing path reconciles it.

### 10. Scheduling

**#2332 and #2393 share six files and cannot run in parallel.** Neither can #2372 and #2390 (two files).

---

Verified against `main` @ `3eb44ba4` and the production snapshot. Verdict: **VALID** / DECISION.

> Every headline number reproduces exactly: 88 distinct cabin_assignment values over 1,861 rows, 32 of them longer than 20 characters, covering 595 rows = 31.97%. Stripping the area prefix leaves exactly 2 names over 20 chars covering 8 rows, as claimed. The two-source premise is correct — cabin_assignment is a verbatim first-wins copy of the CampMinder field and the registry name is curated separately. Only the code line anchors are wrong.

1. WRONG ANCHOR: "`family_camp_derived.go:1490/1515` just `Set`s the received string". REPLACE WITH: "`family_camp_derived.go:886-887` copies it verbatim and first-wins (`if … && reg.cabinAssignment == \"\"`), and `:1825`/`:1850` are the two `Set(\"cabin_assignment\", …)` writes. Lines 1490-1515 are `computeDryRunDiff` and have nothing to do with this."
2. WRONG COUNT: the prefix table lists three prefixes totalling 30 distinct units / 601 rows, but 32 distinct values / 603 rows contain ' - '. ADD a fourth row: "a fourth prefix — 2 distinct units, 2 rows, 2025 only." The '56 / 1,258 no prefix at all' row is exactly right and needs no change.
3. ADD, because it changes what a fixer must do: "#2253 has SHIPPED (474ce260, PR #2314). The journey card's housing span and its header now wrap and carry no `truncate`, so on that surface the names are no longer clipped — they take a second line. The board card (`LodgingUnitCard.tsx:777`) is the only surface here that still truncates silently."

- 1,861 rows carry a non-blank cabin_assignment out of 3,923 family_camp_registrations rows; 88 distinct values. SQL: SELECT COUNT(*),SUM(cabin_assignment<>''),COUNT(DISTINCT CASE WHEN cabin_assignment<>'' THEN cabin_assignment END) FROM family_camp_registrations
- Year scope: cabin_assignment is populated 2022-2026 only (2022:427, 2023:423, 2024:472, 2025:464, 2026:75 — 2026 is a FLOOR, mid-registration). 2017-2021 carry zero. SQL: SELECT year,COUNT(*),SUM(cabin_assignment<>'') FROM family_camp_registrations GROUP BY 1
- 32 of 88 distinct values exceed 20 characters; 595 of 1,861 rows carry one = 31.97%, which is the issue's 32%. SQL: SELECT COUNT(*) FROM (SELECT DISTINCT cabin_assignment v FROM family_camp_registrations WHERE cabin_assignment<>'') WHERE length(v)>20; and SELECT COUNT(*) FROM family_camp_registrations WHERE cabin_assignment<>'' AND length(cabin_assignment)>20
- After stripping everything up to and including the first ' - ': exactly 2 distinct values over 20 chars, covering 8 rows — matches the body's '2 names / 8 rows'. SQL: WITH s AS (SELECT CASE WHEN instr(cabin_assignment,' - ')>0 THEN substr(cabin_assignment,instr(cabin_assignment,' - ')+3) ELSE cabin_assignment END v FROM family_camp_registrations WHERE cabin_assignment<>'') SELECT (SELECT COUNT(*) FROM (SELECT DISTINCT v FROM s WHERE length(v)>20)),(SELECT COUNT(*) FROM s WHERE length(v)>20)
- Prefixed vs unprefixed: 56 distinct / 1,258 rows carry no ' - ' (matches the body exactly); 32 distinct / 603 rows do. The body's table accounts for only 30 distinct / 601 rows — there is a FOURTH prefix (2 distinct values, 2 rows, 2025 only) it omits. SQL: SELECT (cabin_assignment LIKE '% - %'),COUNT(DISTINCT cabin_assignment),COUNT(*) FROM family_camp_registrations WHERE cabin_assignment<>'' GROUP BY 1; and SELECT length(substr(cabin_assignment,1,instr(cabin_assignment,' - ')-1)),COUNT(DISTINCT cabin_assignment),COUNT(*),MIN(year),MAX(year) FROM family_camp_registrations WHERE cabin_assignment LIKE '% - %' GROUP BY 1
- The double-space anomaly is real: 2 distinct values / 2 rows contain a double space. SQL: SELECT COUNT(DISTINCT cabin_assignment),COUNT(*) FROM family_camp_registrations WHERE cabin_assignment LIKE '%  %'
- Registry side: 118 unit names, and the alias table has 187 rows. 41 of the 44 cabin strings that ever held 2+ households resolve to exactly one unit, so an ingest-side normalisation (decision 2) has a working key today.

</details>

---

Housing names are too long for the places that render them, on two surfaces, from **two different
name sources that do not agree with each other**. Filed after kindred#2253's PR made the problem
visible on the journey card by replacing truncation with wrapping.

## The two sources

| Surface | Field | Source |
|---|---|---|
| Household journey card | `family_camp_registrations.cabin_assignment` | Verbatim copy of the CampMinder custom field **"Family Camp Cabin"**. Our code composes nothing — `family_camp_derived.go:1490/1515` just `Set`s the received string |
| Lodging board / unit card | `lodging_units.name` | `config/lodging_registry.json`, curated |

They use different conventions for the same room. The registry spells a room as
`<Building> <Room>`; the CampMinder field spells the same room `<Building> - <Room>`. A staff member
reading the board and the journey card sees two spellings of one place.

## The CampMinder field's prefixing is inconsistent and has drifted

88 distinct values over 1,861 rows. **Only two areas are ever prefixed:**

| Prefix | Distinct units | Rows | Years seen |
|---|---|---|---|
| Area A + ` - ` | 19 | 402 | 2022-2026 |
| Area B + ` - ` | 9 | 183 | 2022-2026 |
| Area C + ` - ` | 2 | 16 | 2022-2023 only, then abandoned |
| **no prefix at all** | **56** | **1,258** | every other area |

Three concrete drifts, all still in the data:

- **One building appears under two different area prefixes AND unprefixed**, depending on the year —
  20 rows under one area (2022-24), 25 rows under another (2022-24), and 3 rows bare (2025-26). Same
  physical building all three times.
- One unit was recorded whole in 2025 and then split into front/back halves in 2025-26, so both
  spellings are live.
- One value carries a **double space** before its separator, which makes it a distinct string from
  its otherwise-identical sibling.

## Measured against the render width

**Journey card** (`HouseholdJourneyCard.tsx`, the housing span): the panel is `w-[26rem]` = 416px,
less `border-l` 1, less the panel's `p-4` (32) and the card's `p-5` (40), leaves **343px** per row.
The row spends 12 (dot) + 12 (gap) + 48 (year) + 12 (gap) before the span, then 14 + 4 for the home
icon, then 12 + roughly 82 for the trailing "See members" button. That leaves **≈147px ≈ 20
characters** at `text-sm`.

| Row shape | Budget | Names over | Rows affected |
|---|---|---|---|
| With "See members" (the common row) | ~147px / ~20 ch | **32 of 88** | **595 of 1,861 (32%)** |
| Without the button | ~229px / ~31 ch | 4 | 44 |
| With the button **and** the no-enrolment chip | ~62px / ~8 ch | 61 | 996 |

**Stripping the area prefix takes the common case from 32 names / 595 rows to 2 names / 8 rows.** The
only two survivors are already unprefixed and are 24 characters each. So on this surface the prefix
is essentially the entire problem.

Character counts are a rough proxy — DM Sans is proportional, so treat the 20 as ±3.

**Board** (`LodgingUnitCard.tsx:777`): `<h3 className="truncate text-lg">{unit.name}</h3>`. The
registry holds 118 unit names, median 10 characters, longest 26, with **16 over 20 characters**. This
one still truncates, so the loss is silent — and `text-lg` is a larger face than the journey card's,
so its character budget is tighter than the pixel width suggests.

## The fix already exists in this codebase, on one surface only

`MapUnitPopover.tsx` hit exactly this and solved it: it walks the common leading **words** across a
cell group, strips them, and moves the shared part to the header. Its own doc comment records the
failure mode, found in a browser rather than by a test — a four-room house rendered four cells that
were identical up to the ellipsis, because the shared building name consumed the width and truncated
away the only part that told the rooms apart. It is careful in ways any fix here should copy: it
never strips a name to nothing, it returns names untouched when they share no prefix, and it computes
over unit names only, never party names.

Nothing equivalent runs on the board card or the journey card.

## What this issue is asking for

A single housing-name **display convention**, applied to both surfaces, that fits the widths above.
Roughly three decisions, none of which should be made from this body alone:

1. **Where the canonical short name lives.** Curated in the registry as a `short_name` alongside
   `name`, or derived at render time the way `MapUnitPopover` already derives it. A curated field is
   the only one that can also fix the CampMinder-side spellings.
2. **Whether the CampMinder string is normalised on ingest.** Mapping `cabin_assignment` onto the
   registry's unit identity at sync time would collapse the three spellings of the one building, kill
   the double-space value, and make both surfaces read from one vocabulary. That is a bigger change
   than a display fix and deserves its own decision.
3. **Whether the prefix is dropped or relocated.** `MapUnitPopover`'s answer — move it to a header
   rather than delete it — is the one with precedent, and its comment argues specifically that
   deleting it outright "trades one lie for another" by leaving a room with nothing saying which
   building it is in. The journey card has no header to move it to, so this needs an answer rather
   than an inherited one.

## Scope notes

- kindred#2253's PR is **not** the cause and needs no change. It removed truncation from the journey
  card, which is what made the length visible; the names were being silently clipped before it.
- Related but distinct: kindred#2331 (shareability is derived from a guessed `sleeps`). Same registry,
  different defect.
- Per the standing rule for lodging work: **verify against the tree and the registry data before
  implementing.** Every figure above was measured against the dated production snapshot and
  `config/lodging_registry.json`, not taken from a prior artifact. Unit and area names are omitted
  here deliberately.

### LINKED ISSUE #2336: bug(sync): cabin_assignment has no session dimension, so a multi-weekend household's one assignment fans out across every weekend
## ⚠️ CORRECTED 2026-08-16 — phase D triage. NO SCHEMA CHANGE. Folds into #2332.

Owner ruling 2026-08-15/16: **staff confirmed that a multi-weekend household's housing being overwritten is
acceptable, and no snapshot or lookback is wanted.** So the schema options below are retired.

**Reduced to documentation + read-path.** Two of this issue's three anchors are the exact comments **#2332**
must rewrite (`api/services/lodging_repository.py:610-618` and `api/schemas/lodging.py:468-473`); #2393
touches neither. **Fold into #2332 with `Closes #2336.` on that PR** — a separate branch guarantees a conflict.

Classification: **MECHANICAL, XS.** No migration.

Owner, on the underlying limitation: *"i know we can't do session scoped for multi."* `cabin_assignment` is one
string per household-year, so a household attending two weekends shows the same cabin for both. That is
accepted, and the read path should say so plainly rather than implying a per-session value exists.

⚠️ Do **not** confuse this with **#2393**, which is independent and still real: the family history modal needs
session grain for its *member list*, mirroring `AttributeSession`'s single/ambiguous/none rule.

---

## What happens

`family_camp_registrations` has grain `(household, year)` and carries `cabin_assignment`. Family camp
runs **several weekends a year**. There is no session column, so a household attending more than one
weekend has its single housing assignment silently replayed into every weekend it attended.

Measured against the production snapshot, 2022-2025, households holding a `cabin_assignment` and at
least one enrolled child in a `family` session:

| family weekends attended that year | household-years |
|---|---|
| 1 | 1,662 |
| 2 | 30 |
| 3 | 7 |
| 4 | 4 |
| **more than one** | **41 of 1,703 · 2.4%** |

2.4% sounds small. It is not, because the fan-out concentrates in exactly the analyses that ask
"who was in this space during this weekend".

## This is not theoretical — it already produced a wrong answer

A shareability analysis asked which lodging units have historically held two or more households in
the same weekend. Cut at session grain and filtered to households with an enrolled child
(`persons.is_camper = 1` AND `attendees.status_id = 2`) in that exact session, it returned 17
apparent multi-household occupancies on units not designated shareable.

**Twelve of the seventeen were artefacts of this defect.** One household attended four family
weekends in one year on a single cabin string and manufactured three false shares in one unit by
itself. Another appeared four times across three units and three years for the same reason, which
read convincingly as "a family that habitually shares" and was nothing of the kind.

The same cut applied to the designated units: 73 of 76 clean. Outside them: 5 of 17. The defect
inverts the signal precisely where the sample is smallest.

## Related, and distinct

- **kindred#2305** is the *read-path* defect — the `family_camp_*` tables carry no enrollment status,
  so non-attending households render as attendees. That is a different problem and does not cause
  this one.
- The standing **fan-out vs refuse** decision for a multi-weekend household's *request* (10
  households in 2026) has the same root. Whatever is decided there should decide this too, or the
  two will diverge.

## ⚠️ The grain triple applies

Widening this table moves the **write key**, the orphan key (`TrackProcessedCompositeKey`) and the
unique index **together, in one PR**. Widening two of the three means the next run deletes the rows
the widening just created, and reports success. Index-first does not work: a coarse write key means
a widened index never sees two rows.

## Fix direction — one decision to make explicitly

Either

1. **Widen the grain** to `(household, year, session)`, which is truthful and makes every per-weekend
   read correct, at the cost of the full grain-triple migration and a decision about what a household
   with one assignment and three weekends means; or
2. **Keep the grain and forbid the fan-out at read**, so every consumer that cuts by weekend must
   either resolve the assignment to a specific weekend or refuse to answer.

Option 2 is cheaper and strictly worse for anything that wants per-weekend housing history, which is
what the weekend board is.

## Acceptance

- [ ] A stated decision between the two options, recorded in this body before any code.
- [ ] Whichever is chosen, a test that a household attending three weekends does not produce three
      identical housing rows that later read as three placements.
- [ ] If the grain widens: one PR moving write key, orphan key and unique index together, with a test
      that a second run after the widening deletes nothing.

adamflagg:
## The defect

A household's housing history rendered the string a staff member typed that season, verbatim, while the board card beside it rendered the curated registry name. **37 of 88 distinct `cabin_assignment` strings resolve to a different registry name, covering 716 of 1,861 rows (38.5%)** — and **21 of 67 live 2026 households** showed one name on the board and another in the history modal at once.

`RosterParty.last_year_cabin` was worse than "stale": it was **never alias-resolved at all**. It came straight out of `fetch_cabin_assignments_by_household_cm_id(year - 1)`, and the roster service's only alias work was `count_open_unresolved_aliases`, which counts a badge and resolves nothing.

Owner ruling 2026-08-18: *"the last year housing should use the same language via the alias year over year concept so it appears in current language."* The alias's year window says **which raw string was in use when**. It is an input to *finding* the unit, never to *naming* it. Once the unit is identified, its present-day `lodging_units.name` renders. The evidence that this matters is in the data: fourteen of the 118 units were renamed one at a time in the admin GUI on 2026-08-15, inside two minutes.

## The change

**`api/services/lodging_rules.py` — `HousingNameResolver`, a pure rule.** Deliberately a mirror of Go's `AliasResolver`, not a second opinion: same lookup key (outer whitespace and case only, inner spacing significant), same all-or-nothing member translation through `code`, same refusal when two alias rows both cover the year. What differs is the target — the Go resolver places rows in their own season, this one always *names* them in the registry's latest one.

**The rules, stated so they stop being re-derived:**

| input | rendered |
|---|---|
| a string that already names a unit today (name or code) | that unit's `name` |
| an alias covering the row's own year, one member | that member's current `name` |
| an alias covering the row's own year, 2+ members sharing one non-empty `parent_unit` | **the parent unit's `name`** |
| 2+ members sharing no parent | member names joined — no production row reaches this |
| nothing, or two overlapping alias rows | the raw string, unchanged |

**The collapse is what does the work, not the resolution.** Joining member names reaches 35 characters — one *more* than the 34-character worst raw string — on a `whitespace-nowrap` span, so naive resolution makes truncation worse. Every one of the seven in-use multi-member aliases is two siblings under one container. `lodging_units.name` averages 10.5 characters and tops out at 24, against raw strings that reach 34; the long tail lives entirely in the raw text.

**Windowed at the ROW's own year, and this is the trap.** Four aliases carry `valid_to_year = 2024` and are correct for the years they cover. Pinning every window to the registry's loaded year discards them and strands 49 rows at their raw spelling (1,792 of 1,861 instead of 1,841). Windowing is done because it is correct — **not** because it is what lifts resolution: 1,841 of 1,861 resolve with windows applied and 1,841 with them ignored, since only 6 of 187 aliases carry a window at all and each is used only inside it.

**Two new repository reads, neither year-filtered and neither cached.** `fetch_all_units()` reads every season because the season that *names* a unit has to be discovered from a table that holds 2026 only, and because an alias stores whichever season's record ids existed when it was written — `code` is the cross-year identity thread. `fetch_unit_aliases()` reads a table that has no `year` column by design. Both are admin-panel-written, so a TTL would show the old name on four surfaces at once. They are absent from `build_summary`'s TaskGroup, for the same reason last year's cabins already were: no `WeekendSummaryEntry` carries a party, so both would be paid on every weekend of the year to name a field nothing renders. A guard test pins that.

**Wire.** `HouseholdJourneyYear.cabin_name` and `RosterParty.last_year_cabin` now carry the resolved name. `cabin_name_raw` is new and carries the verbatim staff string as provenance. That answers the issue's one open decision — the value is the current registry name in both fields, and the raw string is published **once**, on the journey, rather than duplicated onto the roster party: the family card's right-anchored cabin opens the details panel, and the journey is inside it.

**Journey card.** Where the two disagree the name becomes a `ui/Tooltip` trigger reading `Recorded as "…" that season`. A real Tooltip and not a `title`, per #2177, and the same trigger shape `LodgingUnitCard` gives its occupancy figure. Offered *only* where they disagree: on the other 1,145 rows it would decorate a name with a tooltip repeating it back.

**`invalidateLodgingRegistryQueries` gained the household-journey prefix**, and this is the part the change would have been wrong without. The journey used to carry free text no registry edit could move, so `useHouseholdJourney` inherits the 30-minute app default with no writer to invalidate against. It is now registry-fed. Without the new line an admin rename would leave the history modal showing the old name for half an hour while the board behind it showed the new one — the exact disagreement this work exists to remove, re-created by the fix for it.

## #2336, folded in

Reduced by its own owner ruling to documentation plus read-path, and two of its three anchors are comments this change had to rewrite anyway. `cabin_assignment` has grain `(household, year)` because its source is one CampMinder household custom field per household-year, so a household attending two weekends carries one cabin for both — staff confirmed 2026-08-15 that the overwrite is acceptable and declined a snapshot or lookback. That is now stated at the read (`fetch_cabin_assignments_by_household_cm_id`), at both schema fields, and on `FamilyCard`, so the next reader does not re-open it as data loss. 41 of 1,703 cabin-holding household-years are multi-weekend, and treating that fan-out as per-weekend placement manufactured 12 of 17 false multi-household occupancies in one analysis.

## Deliberately NOT done

- **No headline truncation number.** The 595 → N figure has oscillated three times against unmoved data, because each pass used a different resolver. The rules table above is the durable statement; the collapse rule is stable under both resolution orders and every join separator.
- **No ingest-time normalisation.** It cannot work for history: `lodging_units` holds 2026 only, so `AliasResolver`'s `(code, year)` lookup finds nothing for a 2023 row. See #2392.
- **No new alias rows.** Three strings genuinely name a unit *family* rather than a unit and need staff knowledge; four of the originally-claimed seven already exist with `valid_to_year = 2024` and would collide on the `(alias_string, valid_from_year)` unique index.
- **No schema change and no migration.** #2336's widening option was retired by its own investigation — the upstream field never held a second weekend's answer, so a `(household, year, session)` key would replicate one string N times and bake the fan-out in.
- **No `last_year_cabin_raw`.** A second raw field on the roster party would be a wire field nothing renders.
- **No board-card change.** `LodgingUnitCard`'s `truncate text-lg` heading already renders `lodging_units.name` and is untouched.
- **No `MapUnitPopover`-style common-prefix stripping.** The registry names are short; the prefix problem was in the raw strings, and resolving removes it at the source.
- **The weekend line stays for #2393**, which should call this helper rather than re-derive it.

## Verification

`ruff`, `mypy`, `pytest` (6,329 unit), `prettier`, `eslint`, `tsc`, `vitest` — `scripts/pre-push-verify.sh --all` green. `scripts/dev/verify-no-hardcoded-lodging.sh` green. Every new test was run RED first and mutation-checked: dropping the parent collapse, windowing at the registry year instead of the row's, picking the first of an ambiguous alias pair, indexing units without the registry-year filter, resolving `last_year_cabin` at the roster's own year, dropping the raw provenance, making the lander pay for the registry reads, showing the provenance trigger unconditionally, rendering the raw instead of the resolved name, and removing the journey invalidation each turned one test red and were each restored.

⚠️ **UI-visible on two surfaces. Held for the owner's eyes.**

Closes #2332.
Closes #2336.

### DIFF
--- a/api/schemas/lodging.py
+++ b/api/schemas/lodging.py
@@ -577,12 +577,37 @@ class RosterParty(BaseModel):
     # 1,433 rows from 2017-2021. So "" is NOT "nobody assigned them", and no
     # consumer should render a placeholder or a dash for it.
     #
-    # FREE TEXT out of `cabin_assignment`, deliberately NOT resolved against
-    # `lodging_units` -- see `fetch_cabin_assignments_by_household_cm_id` for
-    # why (the registry holds only the current year, and 3 of the 88 distinct
-    # strings across 2022-2025 resolve to no alias at all). It may therefore
-    # name something no card on the board is called. Never match it against a
-    # `unit_code`.
+    # RESOLVED TO THE REGISTRY NAME (kindred#2332), not the raw string staff
+    # typed last season. Owner ruling 2026-08-18: *"the last year housing
+    # should use the same language via the alias year over year concept so it
+    # appears in current language."* This field used to carry the prior year's
+    # `cabin_assignment` free text straight through, so the family card
+    # contradicted the board card beside it -- 37 of 88 distinct raw strings
+    # resolve to a different registry name, covering 716 of 1,861 rows (38.5%).
+    #
+    # `HousingNameResolver.display_name` does the work, at the PRIOR year's
+    # alias window: the window says which raw string was in use when, and is an
+    # input to FINDING the unit, never to naming it. A string that resolves to
+    # nothing renders unchanged, which is what the three strings naming a unit
+    # FAMILY rather than a unit do (kindred#2392).
+    #
+    # THE RAW STRING IS NOT DUPLICATED HERE. It is provenance, and it lives on
+    # `HouseholdJourneyYear.cabin_name_raw` -- the same household-year fact, on
+    # the surface whose job is the record, one click away through this card.
+    #
+    # Still never match it against a `unit_code`: it is a NAME, and a name is
+    # not an identity. The board's own placement travels in `unit_code` /
+    # `unit_codes`.
+    #
+    # ⚠️ NOT PER-WEEKEND, and no consumer may render it as such (kindred#2336).
+    # `cabin_assignment` has grain (household, year) because its source is one
+    # CampMinder household custom field per household-year -- so a household
+    # attending two weekends of a season shows the SAME cabin for both, and its
+    # later weekend overwrites nothing because there was never a second value.
+    # Staff confirmed 2026-08-15 that this is acceptable and asked for no
+    # snapshot or lookback. 41 of 1,703 cabin-holding household-years are
+    # multi-weekend; treating the fan-out as per-weekend placement manufactured
+    # 12 of 17 false multi-household occupancies in one analysis.
     #
     # Only ever populated on the ROSTER, and only for household-grain parties:
     # `build_summary` keeps nothing but counts, and an adult-weekend guest is
@@ -705,11 +730,25 @@ class HouseholdJourneyYear(BaseModel):
 
     year: int = 0
     housing: HousingState = "unknown"
-    # The staff-written free text out of `family_camp_registrations`, NOT
-    # resolved to a `lodging_units` row: `lodging_units` holds only the
-    # current year, so a 2023 string can name a cabin that no longer exists
-    # under that name. See `fetch_cabin_assignments_by_household_cm_id`.
+    # The unit's name TODAY (kindred#2332), whichever year the row is.
+    #
+    # Owner ruling 2026-08-18: a prior year's housing renders in the current
+    # language, because staff recognise the cabin they use now -- a 2022 row
+    # displaying a name nobody has used since 2023 is a lookup task, not
+    # information. Renames are routine: fourteen of the 118 units were renamed
+    # one at a time in the admin GUI on 2026-08-15, inside two minutes.
+    #
+    # `HousingNameResolver.display_name` resolves it at THIS ROW'S OWN YEAR --
+    # the alias window says which raw string was in use when, which is what
+    # finds the unit; the unit's present-day `lodging_units.name` is what names
+    # it. Falls back to the raw string unchanged when nothing resolves.
     cabin_name: str = ""
+    # What staff actually typed that season, verbatim -- provenance, not a
+    # name. Equal to `cabin_name` whenever nothing resolved, and "" whenever
+    # `cabin_name` is. The client shows it only where the two DISAGREE, which
+    # is 716 of 1,861 rows on the production snapshot: rendering it beside an
+    # identical name would be noise on the other 1,145.
+    cabin_name_raw: str = ""
     enrollment: EnrollmentState = "none_on_file"
     # Every `family_camp_adults` row for the year, blanks and placeholders
     # included -- the same contract `RosterParty.adults` publishes, so the
--- a/api/services/lodging_repository.py
+++ b/api/services/lodging_repository.py
@@ -74,6 +74,7 @@ from api.constants.collections import (
     LODGING_INGEST_ISSUES,
     LODGING_SESSION_STATUS,
     LODGING_SLOT_MERGES,
+    LODGING_UNIT_ALIASES,
     LODGING_UNITS,
     LODGING_WRITE_INS,
     LODGING_WRITE_INS_DRAFT,
@@ -310,6 +311,58 @@ class LodgingRepository:
                 "sort": "area.sort_order,name",
             },
         )
+
+    async def fetch_all_units(self) -> list[Any]:
+        """EVERY lodging unit, every season -- the registry, for NAMING things.
+
+        kindred#2332. Deliberately unfiltered, and that is what separates it
+        from `fetch_units`. Naming a unit is a question about the PRESENT, so
+        the registry's latest season has to be discovered from the table
+        rather than assumed to be the year being read: `lodging_units` holds
+        2026 only (118 of 118), so resolving a 2023 cabin string against 2023's
+        units finds nothing at all. That is the trap that made display
+        resolution look impossible (kindred#2392).
+
+        Older seasons are read too, not merely tolerated. An alias stores
+        whichever season's record ids existed when it was written and is never
+        re-pointed, so translating a stored member id needs `code` for units of
+        EVERY year -- `code` is the cross-year identity thread, exactly as
+        `AliasResolver` uses it.
+
+        No `expand`: `HousingNameResolver` wants `code`, `name`, `year` and
+        `parent_unit`, and the area relation is `fetch_units`' business.
+
+        Deliberately NOT cached, for `fetch_units`' reason made sharper by this
+        issue -- `lodging_units.name` is written straight to PocketBase from
+        the admin panel, and staff rename in bursts (fourteen of the 118 units
+        on 2026-08-15, inside two minutes). A TTL here would show the old name
+        on four surfaces at once.
+        """
+        return await self._page(LODGING_UNITS, query_params={"sort": STABLE_SORT})
+
+    async def fetch_unit_aliases(self) -> list[Any]:
+        """Every `lodging_unit_aliases` row -- the historical spellings.
+
+        NO YEAR FILTER, because the table has no `year` column and never
+        should: a row's `valid_from_year` / `valid_to_year` window records what
+        a building was CALLED from a given year, which is a rename history and
+        not a per-year copy. The window is applied per raw string at the year
+        THAT STRING came from (`HousingNameResolver.display_name`), never at
+        the registry's loaded year -- pinning it to the current season
+        discards the four rows carrying `valid_to_year = 2024` and strands 49
+        of the 1,861 cabin-bearing rows at their raw spelling.
+
+        `member_units` is left as raw relation ids rather than expanded: the
+        resolver maps id -> code -> the registry year's row off
+        `fetch_all_units`, which it holds anyway, so an expand would pay for
+        rows already in hand.
+
+        Deliberately NOT cached: `mapUnresolvedAlias` in `lodgingCrud.ts`
+        writes this table straight from the admin panel, never through this
+        API -- the same argument `count_open_unresolved_aliases` makes for the
+        queue that feeds it.
+        """
+        return await self._page(LODGING_UNIT_ALIASES, query_params={"sort": STABLE_SORT})
 
     async def fetch_availability(self, year: int, session_cm_id: int) -> list[Any]:
         """Staff RELEASES for one session -- the staff<->family role override.
@@ -878,14 +931,33 @@ class LodgingRepository:
         """Where each household slept in `year`, keyed by household CampMinder id.
 
         The staff-written string out of `family_camp_registrations.cabin_assignment`
-        -- FREE TEXT, not a relation. It is deliberately NOT resolved to a
-        `lodging_units` row here: `lodging_units` holds only the current year,
-        so a 2023 string can name a cabin that no longer exists under that
-        name, and 3 of the 88 distinct strings across 2022-2025 resolve to no
-        alias at all. What staff wrote is the fact they are recalling; the
-        alias resolution (`lodging_unit_aliases` → `member_units`, honouring
-        `valid_from_year` / `valid_to_year`) is a separate job with a separate
-        failure mode.
+        -- FREE TEXT, not a relation, and RAW: this read returns exactly what
+        staff typed that season and resolves nothing.
+
+        RESOLUTION HAPPENS AT DISPLAY, ONE LAYER UP (kindred#2332). Owner
+        ruling 2026-08-18: every surface renders the unit's CURRENT registry
+        name, so `HousingNameResolver` -- fed by `fetch_all_units` and
+        `fetch_unit_aliases` -- translates this string in the service, at the
+        year the string came from. Keeping the raw value here is what lets the
+        journey publish the name and its provenance from one read, and what
+        keeps `_housing_state` deciding placed/not-placed on PRESENCE rather
+        than on whether a name happened to resolve.
+
+        ⚠ ONE STRING PER HOUSEHOLD-YEAR, WITH NO SESSION DIMENSION, and no
+        caller may present it as per-weekend (kindred#2336). The source is a
+        single CampMinder household custom field (`Family Camp Cabin`, one
+        value per household-year), so a household attending two weekends of a
+        season has one cabin for both -- there is no second weekend's answer
+        being lost, and widening the grain would replicate one string N times
+        and bake the fan-out into the schema. Staff confirmed 2026-08-15 that
+        the overwrite is acceptable and declined a snapshot or lookback. 41 of
+        1,703 cabin-holding household-years are multi-weekend; cutting this
+        column by weekend as though it were a placement manufactured 12 of 17
+        false multi-household occupancies in one analysis. The honest
+        per-weekend rule already ships in Go
+        (`LodgingAssignmentsSync.AttributeSession`), which pins the year's
+        single string to a weekend only when the household attended exactly
+        one and queues an ingest issue otherwise.
 
         ⚠️ KEYED BY cm_id, AND THAT IS THE ENTIRE POINT. `registration.household`
         is a PocketBase relation into a YEAR-SCOPED `households` table, so a
--- a/api/services/lodging_roster_service.py
+++ b/api/services/lodging_roster_service.py
@@ -1293,20 +1293,27 @@ class LodgingRosterService:
     async def build_summary(self, year: int, scenario: str = "") -> WeekendSummaryResponse:
         """Every weekend in the year with its counts, in one pass.
 
-        `build_roster` makes THIRTEEN fetches (12 concurrent + `fetch_session`
-        alone before them), of which SEVEN are year-scoped -- the unit
+        `build_roster` makes SIXTEEN reads (fourteen concurrent tasks issuing
+        fifteen reads -- `_housing_names` is one task making two -- plus
+        `fetch_session` alone before them), of which TEN are constant across
+        every weekend of the year. EIGHT of the ten are year-scoped: the unit
         registry, households, the prior-household set, family-camp adults,
-        registrations, the unresolved-alias count and last year's cabins
-        (kindred#2075) are identical for every weekend in the year. Calling it
-        once per weekend to fill the lander would repeat all seven N times,
-        which is why a weekend with zero parties still costs about three
-        seconds.
+        registrations, the unresolved-alias count, last year's cabins
+        (kindred#2075) and the raw per-field request answers (kindred#2330).
+        The other two are kindred#2332's registry-naming pair, which carry no
+        year at all and are therefore constant for the same reason. Calling
+        `build_roster` once per weekend to fill the lander would repeat all
+        ten N times, which is why a weekend with zero parties still costs
+        about three seconds.
 
-        The lander fetches only SIX of those seven. Last year's cabins feed
-        `RosterParty.last_year_cabin`, and no `WeekendSummaryEntry` carries a
-        party -- `_build_parties` runs here purely to be counted, so the
-        seventh read would buy a field nothing renders. `_build_parties` takes
-        it as a defaulted keyword for exactly that reason.
+        The lander fetches only SIX of those ten, and declines four. No
+        `WeekendSummaryEntry` carries a party -- `_build_parties` runs here
+        purely to be counted -- so last year's cabins
+        (`RosterParty.last_year_cabin`), kindred#2330's request answers, and
+        kindred#2332's two registry reads that would name the cabin all buy
+        fields nothing renders. `_build_parties` takes the first two as
+        defaulted keywords for exactly that reason, and the naming pair is
+        simply absent from the TaskGroup below.
 
         kindred#1963 measures this from eleven and eight, so that issue is
         partly pre-paid: kindred#1889 deleted `has_medical_narrative`, the only
--- a/api/services/lodging_roster_service.py
+++ b/api/services/lodging_roster_service.py
@@ -53,6 +53,9 @@ from api.schemas.lodging import (
 )
 from api.services.lodging_rules import (
     REQUEST_TEXT_SOURCES,
+    HousingNameResolver,
+    RegistryUnit,
+    UnitAlias,
     amenity_coverage,
     container_bathroom,
     effective_bathroom,
@@ -1175,6 +1178,14 @@ class LodgingRosterService:
             # deliberately absent from `build_summary`'s parallel TaskGroup,
             # which keeps nothing but counts.
             last_year_cabins_task = tg.create_task(self.repository.fetch_cabin_assignments_by_household_cm_id(year - 1))
+            # kindred#2332's registry index, for turning the raw string above
+            # into the name the unit carries TODAY. Two reads, neither
+            # year-filtered and neither cached -- see `_housing_names`.
+            # Deliberately absent from `build_summary`'s TaskGroup for exactly
+            # the reason last year's cabins are: no `WeekendSummaryEntry`
+            # carries a party, so both would be paid on every weekend of the
+            # year to name a field nothing renders.
+            housing_names_task = tg.create_task(self._housing_names())
             adults_task = tg.create_task(self.repository.fetch_family_camp_adults(year))
             registrations_task = tg.create_task(self.repository.fetch_family_camp_registrations(year))
             # kindred#2330. The RAW per-field, per-child request answers --
@@ -1242,13 +1253,24 @@ class LodgingRosterService:
         # there would be work no response can read -- once per weekend, across
         # every weekend of the year, on every lander request.
         _resolve_write_in_covers(unit_summaries, written_in_unit_ids(write_ins))
+        housing_names = housing_names_task.result()
         parties = self._build_parties(
             session_type=session_type,
             session_start=_as_date(_s(session, "start_date")),
             attendees=attendees_task.result(),
             households=households,
             prior_cm_ids=prior_task.result(),
-            last_year_cabins=last_year_cabins_task.result(),
+            # RESOLVED HERE, not in the party builder, and at `year - 1`
+            # (kindred#2332). The string came out of the PRIOR season, so the
+            # prior season is the year its alias window is tested at; testing
+            # it at the roster's own year would strand every row whose alias
+            # carries `valid_to_year = 2024` on its raw spelling. The NAME is
+            # still the present one -- the window finds the unit, it does not
+            # name it.
+            last_year_cabins={
+                household_cm_id: housing_names.display_name(raw, year - 1)
+                for household_cm_id, raw in last_year_cabins_task.result().items()
+            },
             adults_by_household=adults_task.result(),
             registrations=registrations_task.result(),
             request_values=request_values_task.result(),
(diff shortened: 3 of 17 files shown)


adamflagg:
## Independent review — floor 60, findings none-or-fixed-in-branch

The `code-review` skill wants to fan out to parallel sub-agents and this session has no Agent tool, so
the review was done by hand: the full `gh pr diff` at length, every touched file read end to end,
`Tooltip`/`housingLabel`/`_page`/`_s`/`_i`/`_journey_repo` and both `invalidateLodgingRegistryQueries`
call graphs followed to their callers, `pocketbase/sync/lodging_alias_resolver.go` read side by side
with the new Python, and **every data claim re-measured against `pb_data/data-prod.db` opened
read-only with this branch's own `HousingNameResolver` imported** rather than re-implemented.

### What the data says

Reproduced exactly, running the branch's resolver over all 1,861 cabin-bearing
`family_camp_registrations` rows:

| claim | measured |
|---|---|
| 88 distinct raw strings, 716 of 1,861 rows render differently | **716 / 1,861**, 37 distinct — matches |
| three strings genuinely unresolvable | **3 strings / 20 rows**, all 2022-2023 — matches |
| seven in-use multi-member aliases, all two siblings under one container | **7 in use of 8**, every one exactly 2 members with one shared non-empty `parent_unit` — the join branch is unreached in production, as claimed |
| `lodging_units.name` averages 10.5 ch, tops out at 24 | **118 rows, avg 10.5, max 24**, 6 over 20 ch — matches |
| the registry holds 2026 only | **118 of 118** |

Two things the body does not say that I checked and found benign:

- **122 rows get *longer* after resolution** (worst 15 ch → 18 ch). That is well inside the 34-character
  raw worst case on the `whitespace-nowrap` span, so the "collapse does the work" argument survives —
  but it is the reason the headline truncation figure was right to be dropped.
- **No ambiguous direct key exists in today's registry**, so the `None` marker in `build` is a
  latent-only branch. It now has tests (below).

### The `AliasResolver` mirror claim holds

`covers()` is line-for-line equivalent to `aliasRow.covers`; member translation is all-or-nothing on
both doors exactly as `Resolve` does it; two rows covering one year refuse in both. The only
behavioural addition beyond the stated target-year difference is the direct `name`/`code` match, which
Go has no equivalent for — **checked and refuted as a hazard**: of 187 alias strings, only two also
name a current unit directly, and on both the direct route and the alias route land on the *same*
displayed name, so precedence is unobservable on this data.

### Fixed in-branch — 868938d1

1. **`build_summary`'s docstring undercounted the roster's reads, and this branch widened the gap.**
   It said *"THIRTEEN fetches (12 concurrent + `fetch_session`), of which SEVEN are year-scoped… the
   lander fetches only SIX of those seven"*. It was already one task stale on `main` — #2330's
   `fetch_request_text_values` landed without extending it — and this branch's registry pair made it
   two. Re-derived from the TaskGroup: **sixteen reads, fourteen concurrent tasks issuing fifteen
   (`_housing_names` is one task making two), ten constant across every weekend, six batched and four
   declined.** That paragraph is the canonical justification for the lander's shape, so a wrong count
   there is the kind of thing the next reader plans against. `TestBuildSummary`'s docstring carried the
   same two numbers and is corrected with it.

2. **`HousingNameResolver.build` described a refusal that the read path does not perform.** The comment
   said a key two units both answer to is refused *"the same call `Resolve` makes on two overlapping
   alias rows"*. It is a **deferral**: `display_name` does `self._direct_by_key.get(key, "")`, so a
   `None` marker is indistinguishable from an absent key and an alias row covering the year still
   resolves. That is the better behaviour — an alias is a deliberate staff mapping, a name collision is
   an accident — so the comment was corrected to say so rather than the code changed, and **two tests
   now pin it**, mutation-checked against both a refuse-outright variant and a pick-the-last-writer
   variant (each turned one or both red; restored).

### Refuted, with reasoning — not filed, not fixed

- **A two-member alias appearing to render unresolved.** It carries `valid_from_year = 2025`; probing it
  at 2024 correctly declines. Correct behaviour, not a gap.
- **Client-side trim asymmetry** — `rawHousing` is trimmed, `housing` is not, so a padded unresolvable
  string would show a tooltip repeating itself. **Zero** `cabin_assignment` values in the snapshot carry
  outer whitespace, and any padded string that *does* resolve produces a different name anyway. Not
  reachable.
- **`registry_year = max(unit.year)` degrading if a future season is partially seeded.** Real in
  principle: one row for year N+1 would strand every other unit at its raw spelling. But there is no
  such row, the season roll-forward is a bulk operation, and the alternative (latest season *per code*)
  is a behaviour change with no failing case to justify it. Documented design, left alone.
- **The board half of #2332.** `LodgingUnitCard`'s `truncate text-lg` is untouched, but the issue's ask
  was *one convention* and the board already rendered `lodging_units.name` — this PR is what brings the
  other two surfaces onto it. The residual silent truncation is now **6 of 118 names over 20 characters
  with a 24-character maximum**, down from the 16 the issue measured, because staff shortened the names
  themselves in the GUI (the same evidence that closed #2328 as premise-false). Not worth an issue.
- **A second full `lodging_units` read per roster load** (`fetch_units(year)` plus `fetch_all_units()`).
  Deliberate and required: `code` is the cross-year identity thread and the alias members can point at
  any season's ids. 118 rows, uncached by design on both sides.
- **#2392's body.** Checked for staleness under the "a wrong issue body gets fixed" rule — its §2 and §3
  already retract the keystone framing and its residue table matches this branch's 3-strings/20-rows
  exactly. Nothing to correct.

### Verification

`ruff format`, `ruff check`, `mypy` (817 files), `pytest` 6,331 unit — `scripts/pre-push-verify.sh` run
bare, `ALL CHECKS PASSED`. Push confirmed by the remote ref
(`git rev-list --count origin/feature/journey-name-convention..HEAD` = 0). CI Summary green.

CodeRabbit deliberately **not** requested — the wave spent the per-account quota on #2450.

⚠️ **Holds for the owner's eyes.** Frontend render paths changed on two surfaces.

PR #2454, opened 2026-08-18
feat(frontend): let a write-in and a family share one space, in either order

### LINKED ISSUE #2432: feat(frontend): allow a write-in on a space that already holds a family

adamflagg:
Staff asked to write a paper registration into a cabin that a CampMinder family is already placed in — one paper registration and one CampMinder family sharing a cabin. A paper registration has no CampMinder record, so it cannot be placed on the board at all; a write-in is the only way to record that party. The write-in control vanished the moment a family was placed in the space, so **the one case the control exists for was the one case it refused**.

Owner ruling 2026-08-18 reverses #2090 symmetrically:

> *"we should be able to add families to any write in space, or add a write in to a family space — regardless of which came first."*

`names an occupant` and `closes the space` are now separate facts.

## The measured number: four refusals, not one

The issue body said *"Scope: frontend only"* and named **one** gate. Re-derived against the tree, **four** sites refused, across both directions of the ruling:

| Site | What it was | Direction it refused |
|---|---|---|
| `unitBadges.availabilityAction` | `if (occupied) return null` — plus the `occupied` parameter itself | write-in onto a family |
| `dragPlacement.resolveDrop` | `if (hasWriteIn(unit)) return null` — its own comment called it *"the load-bearing one"* | family onto a write-in |
| `LodgingUnitCard`'s `useDroppable` | `disabled: … \|\| held` | family onto a write-in |
| `LodgingUnitCard`'s `canPickFamily` | `… && !held && …` | family onto a write-in (#2080's picker) |

The last two are why a resolver-only fix would not have worked: `disabled` keeps dnd-kit from ever reporting `isOver`, so a drop the resolver accepts could never be aimed at the card, and the picker would stay absent for a placement drag performs happily. `resolvePickerPlacement` needed no edit at all — it is a thin adapter over `resolveDrop`, which is the whole point of it having no rules of its own.

The `occupied` argument is **deleted rather than defaulted or ignored**, in both `availabilityAction` and `UnitAvailabilityControl`'s props: a caller still holding an occupancy fact has nowhere to spell it, which is what stops the gate growing back by inches. `tsc` is the enforcer.

Every site carries the ruling that replaced the old one, in the place the old one sat. Without that, the next reader restores the guard on the strength of the comment already there.

## What deliberately did NOT change

- **The `hasWriteIn` gate in `availabilityAction` stays.** It is write-in **arity** (#2381 — one button cannot name which of four rows a click destroys), not occupancy. The two refuse the same card for unrelated reasons and are now named apart in the code.
- **`canPickFamily`'s `parties.length === 0` stays.** A second family reaches a shareable space by drag, which remains the deliberate path for a share. #2091 is unbuilt and nothing here invents its treatment.
- **The open-space forest tint still stands down on a written-into cabin**, on the surviving half of its reason: EMPTY and OPEN are different predicates, and a room somebody is sleeping in is not empty even though it is now placeable. The comment says which half died.
- **"Drop families here" still stands down** for the same reason — the well is not empty, a `WriteInCard` is sitting in it. The card remains a live drop target regardless.
- **No API behaviour change.** `is_family_available` (`api/services/lodging_rules.py`) still folds occupancy in, so the stats bar's "available" count stays conservative about a written-into space and a *released staff cabin* that is then written into still drops off the board via `isPlanningInventory`. Out of scope here, flagged rather than silently widened.
- `held` is renamed `writtenInto` throughout `LodgingUnitCard` — the surviving readers say **who is in it**, never that the space is shut.

## #2091's visual precedent, set on purpose

This draws a `FamilyCard` and a `WriteInCard` in one occupant well for the first time, and #2091 (*mark a space that holds two families*) is unbuilt — so whatever this ships becomes its precedent by default. The chosen form is the most conservative one available and is written down as such in the well's own comment: **one well, occupants stacked in the order they already stack, no divider, no new chip, no new colour**, the existing `CARD_FRAME`/`card-lodge` frames untouched. A shared space is not a new *kind* of card; it is a card with two occupants in it. If #2091 later wants a shared-space treatment it should extend this rather than contradict it.

## Known interaction with #2439 — read before merging

A write-in carries **no headcount** (`lodging_write_ins` has `occupant_name` and `note` and no count field), so once both occupants are in one space the free-bed figure **overstates**: a 10-bed cabin holding a party of 6 plus a written-in party of 3 reads **4 free** when 1 is. That is filed as **#2439** and ruled an *optional investigation, explicitly not a blocker* — an understated count on a space staff can share beats a space they could not share at all. The card's occupancy comment names #2439 and says not to "fix" it by guessing a headcount.

## ⚠️ Stacked on #2446 — this PR gets NO CI

Based on `feature/write-in-merge-split` (PR #2446, the Lane B card work for #2381 + #2431 + #2430), which it shares `unitBadges.ts` and `dragPlacement.ts` with. This repo's CI is `pull_request: branches: [main]`, so **a PR based on a feature branch runs nothing — an absence, not a failure.** Its only evidence is the local gates below.

The `Closes` line will also register nothing until #2446 merges and this body is re-saved.

Observed on this PR: `Validate PR title` passed, `auto-merge` skipped, and **CodeRabbit's green check is the "Review skipped: automatic reviews are disabled" no-op** — not a review. CodeRabbit was deliberately not requested: the wave's rolling quota is better spent on a PR that gets CI, and this one parks for the owner's eyes regardless.

## Independent review (round 2) — fixed in-branch

A second reader re-derived the change against the tree and the production
snapshot. No behaviour defect was found; the four gates are the right four and
the reversal is complete in both directions. What it did find is that the rule
#2432 reverses is still **asserted in eight places outside the gates**, and
each is what the next reader would restore it from. All eight are corrected
here (`ce540ab1`), comments and docs only:

| Where | What it still said |
|---|---|
| `docs/architecture/lodging-occupancy.md` | The whole *"What a hold represents"* section — write-in and placement are *"mutually exclusive states: a unit cannot be both held and occupied"* — naming the exact three code sites this PR struck as its proof. Rewritten around the 2026-08-18 ruling. |
| 5 API comments (`lodging_write_service.py` ×3, `lodging_repository.py`, `lodging_roster_service.py`) | Each gave *"kindred#2247's placement gate"* as the REASON write-ins are seeded into a scenario. That gate is gone; a reader who checked would find nothing there and could delete the seed, dropping every write-in occupant from seeded scenarios. The seed's real argument was always **disclosure**, and now says so. |
| `api/schemas/lodging.py` | *"an inherited write-in badges and **blocks placement**"* — it badges and discloses; it blocks nothing. |
| `needsFit.ts`, `needsFit.test.ts`, `LodgingUnitCard.test.tsx` | The board's signal vocabulary — `opacity-40` is owned by *"the invalid merge target and by a held space"*. This PR corrected the two copies inside `LodgingUnitCard.tsx` and left the canonical statement in `needsFit.ts` plus two test restatements. |
| `LodgingBoard.tsx` | `placeParty`'s copy of `resolvePickerPlacement`'s refusal list still led with *"a held space"*, disagreeing with the list in `dragPlacement.ts` this PR rewrote. |
| `unitBadges.ts` | The Write-in chip justified its slate tone by *"the underlying fact (this cabin is not available to a family this weekend)"*. The fact is **somebody is in it**. |
| `LodgingUnitCard.tsx` | The new ⚠️ said the picker has *"NO rules of its own by design"* — which reads as licence to delete the `parties.length === 0` gate listed three lines above. Qualified: that gate removes an affordance without removing an outcome, which is a scoping choice, not the drift. |

Checked against the production snapshot rather than by reading: of **16**
written-into units in the 2026 weekend, **8 already carry a draft placement**,
so the shared-occupant well is a state real data is in today, not a legacy-row
curiosity — and two write-ins name a paper registration explicitly, which is
the reported case. All 16 have a recorded capacity, so `countUnmeasuredSpaces`
reports the same number before and after.

Also verified, and the reason nothing here holds the PR: `place_party`
(`lodging_write_service.py`) performs an unconditional upsert with **no**
occupancy check, so the client-side reversal succeeds end to end rather than
optimistically moving and rolling back. And the released-staff-cabin edge is
**not widened** — `isPlanningInventory` keeps any unit that has a party on it,
and every case #2432 newly enables has one.

## Local evidence

- `bash scripts/pre-push-verify.sh` → **ALL CHECKS PASSED**, exit 0 (run bare, never piped)
- `vitest run` (full frontend): **450 files, 6956 tests passed**
- `tsc --noEmit`: clean. `eslint src/components/weekend`: **0 errors** (19 pre-existing warnings, byte-identical set before and after)
- `./scripts/dev/verify-no-hardcoded-lodging.sh` → OK
- Strict TDD: 9 tests written failing first — 4 RED in `dragPlacement.test.ts`, 4 RED in `LodgingUnitCard.test.tsx`, then green. The 3 tests that passed on first write were mutation-checked (well made either/or → RED; `wholesaleWriteIn` forced true → RED; `mergeDragActive` dropped from `disabled` → RED) and restored.

Fictional data throughout (Emma Johnson, Liam Garcia, Ava Martinez, Noah Johnson).

Closes #2432.

### DIFF
--- a/api/schemas/lodging.py
+++ b/api/schemas/lodging.py
@@ -344,10 +344,14 @@ class LodgingUnitSummary(BaseModel):
     # open every room beneath a released building.
     #
     # ⚠️ `is_family_available` deliberately does NOT read this field: it folds
-    # in the unit's OWN occupancy row only, so an inherited write-in badges and
-    # blocks placement without moving the bed counts. That is the behaviour the
-    # counts have always had, and changing it is a counts decision rather than
-    # a side effect of the split.
+    # in the unit's OWN occupancy row only, so an inherited write-in badges
+    # without moving the bed counts. That is the behaviour the counts have
+    # always had, and changing it is a counts decision rather than a side
+    # effect of the split.
+    #
+    # It said "badges and BLOCKS PLACEMENT" until kindred#2432, which struck
+    # every client-side refusal of a placement onto a written-into space. An
+    # inherited write-in now badges and discloses; it blocks nothing.
     write_ins: list[WriteInCover] = Field(default_factory=list)
     map_x: float | None = None
     map_y: float | None = None
--- a/api/schemas/lodging.py
+++ b/api/schemas/lodging.py
@@ -264,7 +264,7 @@ class LodgingUnitSummary(BaseModel):
     # standing `inventory_class` decides. None and False are different answers
     # and must not be flattened into one.
     #
-    # IT DOES NOT ANSWER "IS SOMEBODY IN IT". That is `write_in` below, read
+    # IT DOES NOT ANSWER "IS SOMEBODY IN IT". That is `write_ins` below, read
     # from `lodging_write_ins` / `_draft`. Until PR 4 of kindred#2382 an
     # occupancy also reported itself here as False, because
     # `is_family_available` and the board's forest open-tint were both derived
@@ -272,7 +272,7 @@ class LodgingUnitSummary(BaseModel):
     # False here means a role decision and never an occupant. A reader that
     # wants "can a family go in this space" wants `is_family_available`, which
     # folds both facts in; a reader that wants "is somebody in it" wants
-    # `write_in`.
+    # `write_ins`.
     #
     # Stated explicitly rather than implied. The rejected encoding was a row
     # meaning "the opposite of this unit's current default", which an ordinary
@@ -319,9 +319,19 @@ class LodgingUnitSummary(BaseModel):
     # Its rule is `is_family_available` in api/services/lodging_rules.py, and
     # that is the only place the two are combined.
     is_family_available: bool = False
-    # The write-in that closes this space, resolved through the unit tree --
+    # EVERY write-in that closes this space, resolved through the unit tree --
     # this unit's own row, else the nearest ancestor's, else the nearest
-    # descendant's. None means no write-in covers it.
+    # written-into descendant on each branch beneath it. Empty means no
+    # write-in covers it.
+    #
+    # A LIST since kindred#2381, and that arity is the whole point rather than
+    # future-proofing. A merged container draws in place of its rooms, so the
+    # four write-ins Clouds Rest carries in one 2026 weekend collapsed to
+    # whichever room sorted first and the other three were invisible -- while
+    # each clear silently re-populated the card with the next occupant, so
+    # destroying all four read as four failed clicks. A write-in must survive a
+    # merge and a split the way an assignment does, and an assignment does it
+    # by the drawn card carrying however many leaves it covers.
     #
     # The fields above stay STRICTLY this unit's own row: they are what the
     # write path reads back. Folding an inherited fact into any of them would
@@ -338,7 +348,7 @@ class LodgingUnitSummary(BaseModel):
     # blocks placement without moving the bed counts. That is the behaviour the
     # counts have always had, and changing it is a counts decision rather than
     # a side effect of the split.
-    write_in: WriteInCover | None = None
+    write_ins: list[WriteInCover] = Field(default_factory=list)
     map_x: float | None = None
     map_y: float | None = None
 
--- a/api/schemas/lodging.py
+++ b/api/schemas/lodging.py
@@ -326,7 +326,7 @@ class LodgingUnitSummary(BaseModel):
     #
     # A LIST since kindred#2381, and that arity is the whole point rather than
     # future-proofing. A merged container draws in place of its rooms, so the
-    # four write-ins Clouds Rest carries in one 2026 weekend collapsed to
+    # four write-ins one 2026 container carries in a single weekend collapsed to
     # whichever room sorted first and the other three were invisible -- while
     # each clear silently re-populated the card with the next occupant, so
     # destroying all four read as four failed clicks. A write-in must survive a
--- a/api/services/lodging_repository.py
+++ b/api/services/lodging_repository.py
@@ -502,9 +502,11 @@ class LodgingRepository:
         `fetch_write_ins` whenever the request names a scenario, and the live
         table is then not read at all. `copy_from_mirror` and
         `copy_scenario_to_scenario` both seed it, so a fresh scenario starts
-        with the write-ins its source had rather than blank -- without that,
-        kindred#2247's placement gate would let a family be dropped into a room
-        the live board records as occupied.
+        with the write-ins its source had rather than blank -- without that, a
+        room the live board records as occupied shows up empty on the scenario
+        and its occupant is not on the board at all. (The older reason here was
+        kindred#2247's placement gate refusing such a drop; kindred#2432 struck
+        the gate, and disclosure is what the seed protects.)
 
         `scenario_id` is client-supplied and escaped, for the reason
         `fetch_draft_assignments` spells out. The weekend is named by
--- a/api/services/lodging_roster_service.py
+++ b/api/services/lodging_roster_service.py
@@ -1138,8 +1138,10 @@ class LodgingRosterService:
             # A fresh scenario is seeded by an explicit COPY in both seed paths
             # (`copy_from_mirror` and `copy_scenario_to_scenario`, owner ruling
             # 2026-08-16), which is what stops it starting blank -- without it
-            # kindred#2247's placement gate would let a family be dropped into
-            # a room the live board records as occupied. Rendering the live
+            # a room the live board records as occupied renders empty on the
+            # scenario, occupant and all. (Was "kindred#2247's placement gate
+            # would let a family be dropped into" one; kindred#2432 struck that
+            # gate -- the disclosure is the part that matters.) Rendering the live
             # rows through this read's gaps would be the overlay instead, and
             # would make two scenarios unable to disagree.
             write_ins_task = tg.create_task(
--- a/api/services/lodging_roster_service.py
+++ b/api/services/lodging_roster_service.py
@@ -574,8 +574,10 @@ def written_in_unit_ids(write_ins: list[Any]) -> frozenset[str]:
     return frozenset(_s(row, "unit") for row in write_ins)
 
 
-def write_in_covers(units: list[LodgingUnitSummary], write_in_unit_ids: frozenset[str]) -> dict[str, WriteInCover]:
-    """Which units each write-in closes, keyed by unit code.
+def write_in_covers(
+    units: list[LodgingUnitSummary], write_in_unit_ids: frozenset[str]
+) -> dict[str, list[WriteInCover]]:
+    """Which write-ins close each unit's space, keyed by unit code.
 
     THE UNIT A ROW NAMES IS NOT THE ONLY SPACE IT CLOSES. A write-in is a fact
     about a physical space and a building's space contains its rooms', but the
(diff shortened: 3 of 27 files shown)


adamflagg:
### Independent review — score floor 60

No `Agent`/`Task` tool was available, so the parallel-subagent fan-out the review skill wants did not run; the review was done single-threaded instead — full `gh pr diff`, every touched file read at length, every caller of a moved signature read, and each data claim checked against the production snapshot read-only.

**No behaviour defect found.** The four gates are the right four, the `occupied` argument is genuinely deleted rather than defaulted, and the reversal is complete in both directions. What survived verification is a single class: the rule this PR reverses is still **asserted in eight places outside the gates**, and each is what the next reader restores it from. All eight fixed in-branch, comments and docs only — `ce540ab138282e3be06bcf0185055a8ef6328ae1`.

#### Fixed in-branch

1. **The architecture doc still ruled the opposite.** `docs/architecture/lodging-occupancy.md` § *"What a hold represents"* held write-in and placement to be *"mutually exclusive states: a unit cannot be both held and occupied"*, and named the exact three code sites this PR struck as its proof. Rewritten around the 2026-08-18 ruling, with the four struck refusals listed, what deliberately did not change stated beside them, and #2439's cost recorded.
https://github.com/adamflagg/kindred/blob/ce540ab138282e3be06bcf0185055a8ef6328ae1/docs/architecture/lodging-occupancy.md#L186-L200

2. **Five API comments named a struck gate as the reason write-ins are seeded into a scenario.** `lodging_write_service.py` (×3), `lodging_repository.py`, `lodging_roster_service.py` each said the seed exists because *"kindred#2247's placement gate would let a family be dropped into a room the live board records as occupied"*. That gate is gone — a reader who checked would find nothing there and could delete the seed, which drops every write-in occupant from seeded scenarios. The seed's real argument was always **disclosure**, and each now says so.
https://github.com/adamflagg/kindred/blob/ce540ab138282e3be06bcf0185055a8ef6328ae1/api/services/lodging_write_service.py#L320-L336

3. **`api/schemas/lodging.py`** — *"an inherited write-in badges and **blocks placement** without moving the bed counts"*. It badges and discloses; it blocks nothing.
https://github.com/adamflagg/kindred/blob/ce540ab138282e3be06bcf0185055a8ef6328ae1/api/schemas/lodging.py#L346-L356

4. **The signal vocabulary was corrected in two of its four copies.** *"`opacity-40` = REFUSAL. Owned by the invalid merge target and by a held space (#2087)"* — this PR fixed both copies inside `LodgingUnitCard.tsx` and left the **canonical** statement in `needsFit.ts`, plus restatements in `needsFit.test.ts` and `LodgingUnitCard.test.tsx`.
https://github.com/adamflagg/kindred/blob/ce540ab138282e3be06bcf0185055a8ef6328ae1/frontend/src/components/weekend/needsFit.ts#L14-L18

5. **`LodgingBoard.tsx`** — `placeParty`'s copy of `resolvePickerPlacement`'s refusal list still led with *"a held space"*, disagreeing with the list in `dragPlacement.ts` this PR rewrote.
https://github.com/adamflagg/kindred/blob/ce540ab138282e3be06bcf0185055a8ef6328ae1/frontend/src/components/weekend/LodgingBoard.tsx#L322-L334

6. **`unitBadges.ts`** — the Write-in chip justified its slate tone by *"the underlying fact (this cabin is not available to a family this weekend)"*. Post-#2432 the fact is **somebody is in it**, which is a different sentence and the one the chip must not grow back out of.
https://github.com/adamflagg/kindred/blob/ce540ab138282e3be06bcf0185055a8ef6328ae1/frontend/src/components/weekend/unitBadges.ts#L110-L120

7. **`LodgingUnitCard.tsx`** — the new ⚠️ claims the picker has *"NO rules of its own by design, so a gate here that the resolver does not share is the exact drift that design exists to prevent"*, three lines below a bullet keeping `parties.length === 0`, which the resolver does not share. As written it is the next reader's licence to delete that gate. Qualified: a gate that removes an **affordance** without removing an **outcome** is a scoping choice; only one that changes what the board accepts is the drift.
https://github.com/adamflagg/kindred/blob/ce540ab138282e3be06bcf0185055a8ef6328ae1/frontend/src/components/weekend/LodgingUnitCard.tsx#L659-L676

#### Refuted, with reasoning

- **"The client now accepts a placement the server will reject."** No. `place_party` performs an unconditional upsert with no occupancy check anywhere on the path, so the reversal succeeds end to end rather than optimistically moving and rolling back.
- **"The released-staff-cabin disappearance is widened."** No. `isPlanningInventory` keeps any unit carrying a party, and every case #2432 newly enables has one. The edge needs a release *and* a write-in on the same staff unit *and* no party — reachable before this PR, unchanged by it.
- **`countUnmeasuredSpaces` drops written-into cabins**, and its docstring says it asks about *"spaces a family could be put in RIGHT NOW"* — now a wider set than it filters. Checked against the snapshot: **all 16** written-into units carry a recorded capacity, so the printed number is identical before and after. Loosening the filter would desync it from `beds_family_available` on the same stats-bar line — a counts decision, not a defect, and correctly left with the `is_family_available` note.
- **Stats-bar tooltip "Written in for this weekend, excluded from family spaces"** — still literally true of the count it labels (`units_reserved` / `units_family_available` are unchanged).
- **`writeInBadgeApplies`'s `staff_default` exclusion** and the "Released" branch winning ahead of the write-in branch — pre-existing, untouched, and the newly-reachable combination still draws its occupant as a `WriteInCard` in the well.

#### Grounded in data, not reading

Production snapshot, read-only. Of the **16** written-into units in the 2026 weekend, **8 already carry a draft placement** — so the shared occupant well is a state real data is in *today*, not the legacy-row curiosity the card comment called it, which strengthens the #2091-precedent argument rather than weakening it. Two write-ins name a paper registration explicitly, which is the reported case. All 16 have a recorded capacity.

#### Filed out of scope

None. Every finding was M-or-smaller and went into the branch.

#### Evidence

- `bash scripts/pre-push-verify.sh` run **bare** → `ALL CHECKS PASSED`, exit `0`
- `vitest run src/components/weekend` → 41 files, 1259 tests passed; `tsc --noEmit` clean; `ruff check` / `ruff format --check` clean
- `Documentation Check` on the pushed commit: **success**
- Push confirmed by the remote ref: `git rev-list --count origin/feature/write-in-coexistence..HEAD` → `0`

#### CI and closing refs

No `CI Summary` run exists and none will — the base is `feature/write-in-merge-split` and this repo's CI is `pull_request: branches: [main]`. That is an **absence, not a pass**; the local gates above are the only evidence. `closingIssuesReferences` is **empty**, re-checked after the body was re-saved — #2432 registers nothing until #2446 merges.

**Holds for the owner's eyes** — six non-test files under `frontend/src/` are in the diff, so this parks for a look regardless of confidence. CodeRabbit deliberately not requested; the wave's rolling quota was spent on #2450.

adamflagg:
⚠️ **This PR is stale against its own base and must not be merged as it stands.**

PR #2446, which this is stacked on, was reworked on 2026-08-18 after the owner reviewed it. The rework
deletes code this PR edits:

- the availability strip's `hold`/"Write in" action is **gone** — creating a write-in is now typing into
  the card's family box — so `availabilityAction`'s `occupied` gate, the `occupant` prompt, and the
  `AvailabilityAction`/`AvailabilityPrompt` members this PR modified no longer exist;
- `LodgingUnitCard`'s `canPickFamily` **already drops** the `parties.length === 0` gate on the reworked
  base, which overlaps this PR's own change;
- the `Written in at …` line this PR's tests reference has been struck (see #2458).

A rebase conflicts in six files, most of them because the target was removed rather than moved.

**What still stands, and is the real substance here:** the `dragPlacement.ts` change — `resolveDrop` no
longer refuses a written-into space, so a family can be dragged onto a cabin that already holds a
write-in. #2446 does not touch that file, so that direction is still unimplemented on main.

**Plan:** once #2446 merges, this branch rebases onto `main` — at which point it also stops being a
stacked PR and gets real CI for the first time (this repo's CI is `pull_request: branches: [main]`, so a
PR based on a feature branch runs nothing). The parts that are now moot get dropped in that rebase.

No work is lost and nothing here needs re-deciding — the owner's 2026-08-18 ruling on symmetric
coexistence is unchanged.

PR #2461, opened 2026-08-18
feat(api): give the family journey session grain — weekends on the panel, weekend tabs in the modal

### LINKED ISSUE #2393: feat(lodging): the family history modal needs session grain — it silently collapses multi-weekend years

adamflagg:
## The defect

A journey row is a household-**year**, and nothing on the family history panel said *which* of a season's six-to-ten family weekends the year was. When a household books two weekends the row does not duplicate — `lodging_roster_service` dedupes children by CampMinder person id and emits exactly one row per year — so N weekends collapsed into one row with a merged member list.

**Measured on the production snapshot: 64 of 5,438 journey household-years (1.2%) are multi-weekend.** The denominator is *every traced household-year* — the union of the three reads `build_household_journey` issues (enrolled family attendees ∪ `family_camp_adults` ∪ `family_camp_registrations`), all years with `year > 0`. Distribution: 2 weekends → 48, 3 → 11, 4 → 5.

**In 7 of those 64 the merged member list overstates at least one weekend's party** — a child who did not attend every weekend appeared as though they did. That is the half the session grain actually had to fix, and it becomes *visibly* wrong the moment the weekends appear on the panel.

## The change

Both halves fall out of the attendee rows the members already come from, so this costs **zero extra round trips** — the journey's attendee read has expanded `session` since #2420.

**Wire** (`api/schemas/lodging.py`, `api/services/lodging_roster_service.py`)

- `HouseholdJourneyYear.sessions` — the weekends the household attended that year, earliest first, each carrying `session_cm_id`, the verbatim CampMinder `name`, and `start_date`. Deduped per year: the attendee table is child-grain, so the naive read repeats every weekend once per sibling.
- `HouseholdJourneyYear.housing_session_cm_id` — nullable, populated **only** when the year has exactly one attended weekend *and* a cabin string to attribute. Deliberately the same refusal `AttributeSession` makes in the Go ingest (`pocketbase/sync/lodging_session_attribution.go:327`): CampMinder's one per-year value cannot say which weekend it describes, and a read surface that guessed where the ingest declines would put the two into disagreement about the same fact.
- `PartyChild.session_cm_ids` — which weekends *this child* attended. Journey-only; the roster is already one weekend, so it publishes an empty list there. Collected **before** the per-child dedup, because the second attendee row for a child *is* the second weekend.

**Panel** (`HouseholdJourneyCard.tsx`)

`FC1 · FC4` as plain text on the line beneath the housing name. The cabin is still rendered **once**, and there is no "recorded for the year, weekend unresolved" note — the owner struck it on 2026-08-18.

**Modal** (`HouseholdYearMembersModal.tsx`)

A weekend tab strip, `All · FC1 · FC4`, using the same segmented-control grammar as `HouseholdRosterTable`'s need filters. The children filter to the selected weekend and the header headcount follows the tab. A stale selection resets itself rather than through an effect: the card keeps the modal mounted across opens, so a weekend picked on 2025 would otherwise still be selected on 2024.

**Label** (`weekendNames.ts`)

New `weekendLabel(name)` — the slug, uppercased, falling back to `shortWeekendName` and **never** to a CampMinder id (it takes a name, not a session, so it structurally cannot emit one). `weekendSlug`'s doc block previously read *"THIS IS NOT A DISPLAY NAME"*; that block and the file's top note are rewritten in this PR rather than left contradicting the screen, and the three forms are now stated as a rule about which surface may use which.

- **No housing-per-weekend.** `family_camp_registrations` holds a single cabin string per household-year and no source holds a second. Repeating it against each weekend is the fan-out that manufactured 12 of 17 false multi-family occupancies in the phase-C shareability analysis. #2336 is where the widening question lives; nothing here touches its anchors and it stays open.
- **The adult list is not tabbed, and cannot be.** `family_camp_adults` is household-year grain with no session dimension. The adults render unchanged on every tab; inventing a per-weekend attendance claim for them is #1943, blocked on a 2027 form change.
- **No tab strip for a single-weekend year.** `All · FC1` is a control that cannot change anything, and the card's own weekend line already names it.
- **Tab state is not in the URL.** A deliberate divergence from the weekend-surface rule, justified at the divergence: the modal itself is not addressable — the card holds the open row in `useState` — so a search param for its inner filter would outlive the modal that owns it.
- **Sequencing.** #2393's ruling asked for #2332 first, on the grounds that this grows each row a weekend line underneath a housing name that still wraps to two lines. #2332 is not merged, so the panel is at its tightest until it is.

## One correction to the ruling

The ruling names `rs` and `w` as the two 2026 slugs outside the day-group plan. **Both are off**, and the tree wins: the slug rule keeps every initial of the identity, so `Ready, Set, Camp` is `rsc` (the example `weekendSlug`'s own doc block already gives) and `JFAM Winter Family Camp` is `wfc` once the shared JFAM token is dropped. Pinned as computed rather than as written down — the alternative was special-casing two weekends to match a note. Labels for the ten 2026 family sessions are therefore `RSC`, `FC1`–`FC8`, `WFC`, and a test asserts all ten are distinct.

## Verification

- 12 new service tests, written failing first; five mutations (default session ids, publishing weekend 0, dropping the ordering, pinning a cabin that does not exist, collecting sessions after the dedup) each confirmed RED and restored.
- 6 `weekendLabel` tests, 7 weekend-line tests, 13 weekend-tab tests. Every test that went green on first run was mutation-checked — including one that did **not** die and was strengthened: a mutation opening the modal on the first weekend survived a names-only assertion, because every child in that fixture attended the first weekend.
- `bash scripts/pre-push-verify.sh` → ALL CHECKS PASSED (ruff format/check, mypy, 6,344 pytest, prettier, eslint, tsc, vitest).
- `./scripts/dev/verify-no-hardcoded-lodging.sh` → OK.
- `types.gen.ts` regenerated; `git diff -w` confirms the change is additive only.

Closes #2393.

### DIFF
--- a/api/schemas/lodging.py
+++ b/api/schemas/lodging.py
@@ -378,6 +378,23 @@ class PartyChild(BaseModel):
     last_name: str = ""
     age: float | None = None
     grade: int | None = None
+    # WHICH FAMILY WEEKENDS THIS CHILD ATTENDED that year, earliest first
+    # (kindred#2393). Populated by the household journey ONLY, and empty on
+    # every other surface: the roster is already one weekend, so a per-child
+    # weekend list there would restate the page's own title once per camper.
+    #
+    # The journey spans years and its rows are household-YEAR grain, which is
+    # where a family that booked two of a season's weekends collapses into one
+    # merged member list. Measured on the production snapshot, 64 of 5,438
+    # journey household-years are multi-weekend and 7 of those 64 carry a
+    # child who did not attend every weekend -- so the merged list overstates
+    # at least one weekend's party. This is what lets the client undo that.
+    #
+    # Empty is "not knowable", NOT "attended nothing": an attendee row whose
+    # `session` relation did not expand yields no weekend at all, and the
+    # client keeps such a child visible rather than hiding them from every
+    # weekend.
+    session_cm_ids: list[int] = Field(default_factory=list)
 
 
 class RequestTextEntry(BaseModel):
@@ -730,6 +747,27 @@ HousingState = Literal["placed", "not_placed", "unknown"]
 EnrollmentState = Literal["enrolled", "none_on_file"]
 
 
+class HouseholdJourneySession(BaseModel):
+    """One family weekend a household attended in a given year (kindred#2393).
+
+    THE WEEKEND, NOT THE HOUSING. A journey row is a year and a year holds
+    exactly ONE cabin string -- `family_camp_registrations` has no second
+    field for a second weekend -- so this list says which weekends the
+    household was at and says nothing about where it slept in each. Repeating
+    the year's cabin against every entry is the fan-out that manufactured 12
+    of 17 false multi-family occupancies in the phase-C shareability analysis.
+
+    `start_date` is the raw PocketBase string, exactly as
+    `WeekendSessionSummary` publishes it -- the client already reads that
+    shape, and the server's only use for it here is the ordering it has
+    already applied.
+    """
+
+    session_cm_id: int = 0
+    name: str = ""
+    start_date: str = ""
+
+
 class HouseholdJourneyYear(BaseModel):
     """One year of a household's family-camp record.
 
@@ -759,6 +797,32 @@ class HouseholdJourneyYear(BaseModel):
     # is 716 of 1,861 rows on the production snapshot: rendering it beside an
     # identical name would be noise on the other 1,145.
     cabin_name_raw: str = ""
+    # WHICH FAMILY WEEKENDS the household attended that year, earliest first
+    # (kindred#2393). Derived from the attendee rows this row's members
+    # already come from, so it costs no extra round trip -- the journey's
+    # attendee read has expanded `session` since kindred#2420.
+    #
+    # Empty for a year with no enrolled child (2020's cancelled season, 2021's
+    # adults-only rows) and for the pre-kindred#2420 payload shape where the
+    # relation did not expand. Empty is "no weekend is knowable", never "the
+    # household attended none".
+    sessions: list[HouseholdJourneySession] = Field(default_factory=list)
+    # Which weekend `cabin_name` belongs to -- and `None` whenever that cannot
+    # be said, which is the whole point of the field.
+    #
+    # ⚠️ DELIBERATELY THE SAME REFUSAL the Go ingest makes. `AttributeSession`
+    # (`pocketbase/sync/lodging_session_attribution.go:327`) pins the year's
+    # single cabin string to a weekend only when the household attended
+    # exactly one and otherwise declines, because CampMinder's one per-year
+    # value cannot say which weekend it describes. A read surface that guessed
+    # where the ingest refuses would put the two into disagreement about the
+    # same fact.
+    #
+    # `None` therefore covers three different situations and the client words
+    # none of them: several weekends (41 of the 1,861 cabin-bearing
+    # household-years on the production snapshot), no weekend on file (158),
+    # and no cabin to attribute in the first place.
+    housing_session_cm_id: int | None = None
     enrollment: EnrollmentState = "none_on_file"
     # Every `family_camp_adults` row for the year, blanks and placeholders
     # included -- the same contract `RosterParty.adults` publishes, so the
--- a/api/services/lodging_roster_service.py
+++ b/api/services/lodging_roster_service.py
@@ -27,6 +27,7 @@ from api.schemas.lodging import (
     AmenityCoverage,
     EffectiveBathroom,
     HouseholdJourneyResponse,
+    HouseholdJourneySession,
     HouseholdJourneyYear,
     HouseholdMedicalResponse,
     HousingState,
@@ -420,8 +421,14 @@ def _age_at(birthdate: date, as_of: date) -> float | None:
     return float(f"{years}.{months:02d}")
 
 
-def _party_child(child: Any, *, as_of: date | None = None) -> PartyChild:
+def _party_child(child: Any, *, as_of: date | None = None, session_cm_ids: Sequence[int] = ()) -> PartyChild:
     """A `persons` row as the wire sees it. See `_party_adult` on sharing.
+
+    `session_cm_ids` is the same kind of journey-only argument `as_of` is
+    (kindred#2393): the weekends THIS child attended that year, earliest
+    first. `_build_household_parties` omits it and publishes an empty list,
+    because the roster is already one weekend and a per-child weekend list
+    there would restate the page's own title once per camper.
 
     `as_of` is the ONE thing that lets this stay one mapping instead of
     forking (kindred#2420). `_build_household_parties` (the current-season
@@ -461,6 +468,10 @@ def _party_child(child: Any, *, as_of: date | None = None) -> PartyChild:
         # reads `birthdate` instead.
         age=age,
         grade=_i(child, "grade") or None,
+        # A LIST COPY, not the caller's own: the journey builds one list per
+        # (year, child) and a shared reference would let a later mutation
+        # reach a wire object already handed out.
+        session_cm_ids=list(session_cm_ids),
     )
 
 
@@ -480,6 +491,25 @@ def _child_identity(person: Any) -> Any:
     child is which.
     """
     return _i(person, "cm_id") or str(getattr(person, "id", ""))
+
+
+def _session_order(entry: HouseholdJourneySession) -> tuple[int, date, int]:
+    """A season read left to right (kindred#2393).
+
+    Start date first, because that is the order staff say the weekends in;
+    the CampMinder id breaks a tie, so two weekends opening on the same day
+    print in a stable order rather than in whatever order the attendee rows
+    happened to arrive.
+
+    A weekend with no readable `start_date` sorts LAST rather than first. An
+    unparseable date is an unknown, and `date.min` would file the unknown at
+    the head of the season -- claiming a position the data does not support,
+    in the one spot a reader trusts most.
+    """
+    start = _as_date(entry.start_date)
+    if start is None:
+        return (1, date.min, entry.session_cm_id)
+    return (0, start, entry.session_cm_id)
 
 
 def _housing_state(cabin: str, year_assignments: Mapping[int, str]) -> HousingState:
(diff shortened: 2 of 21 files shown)


adamflagg:
## Independent review — round 3, floor 60

The `code-review:code-review` skill fans out to sub-agents and no Agent/Task tool is available in this
session, so the review was done by hand: the full diff, every touched file at length, `weekendLabel`'s
callers, and every claim about data checked against the production snapshot read-only.

### Fixed in-branch — 1 finding (CONFIRMED, in code this PR wrote)

**`weekendLabel` abbreviated weekends CampMinder never numbered, and the initials collide.** `63eeb7b`.

The distinctness test covered the **ten 2026 weekends**, where the collision does not occur. The journey
panel renders **every** year a household has a trace, and the pre-2020 catalogue is named in prose. Swept
all 79 family sessions in `camp_sessions` and labelled each season:

| Season | Collision |
|---|---|
| 2017–2019 | `Spring Family Camp` → **SFC**, `Summer Family Camp` → **SFC** |
| 2017–2019 | `Fall Family Camp I` / `II` / `III` → **FFCI**, **FFCI**, **FFCI** |
| 2020–2026 | none |

`FFCI` for *Fall Family Camp II* is not merely terse — it **names a different weekend**.

Blast radius, measured on the snapshot (`attendees` ⋈ `persons` ⋈ `camp_sessions`, `status_id = 2`,
`session_type = 'family'` — the same population `fetch_household_family_attendees` reads):

* **891 of 3,040** single-weekend household-years print a label that does not identify its weekend within
  its own season.
* **5 of the 64** multi-weekend household-years printed the same label twice on one weekend line —
  `SFC · SFC`, and one row `SFC · SFC · FFCI · FFCI` — and offered the members modal **two or four tabs
  with identical text that filter differently**.

Fix: `weekendLabel` abbreviates only when the identity's last word is a number — the same final word
`weekendSlug`'s trailing-number rule already keeps whole — and returns `shortWeekendName` otherwise.
Every season 2017–2026 now labels its weekends distinctly. `FC1 · FC4`, which is what the ruling was
actually about, is untouched.

**The trade-off, stated plainly so it can be overruled in one line:** the two modern non-numbered
weekends now print in full — `Ready, Set, Camp` and `Winter Family Camp` instead of `RSC` and `WFC`.
That is longer on **19** multi-weekend rows from 2020 on (e.g. `FC1 · FC5 · FC6 · Winter Family Camp`).
They were *not* ambiguous in their own seasons; they lost the abbreviation because no per-name rule can
tell a safe prose abbreviation from an unsafe one, and the unsafe ones are 891 rows. If you would rather
have `WFC` back, that is a one-line change to `weekendLabel`.

TDD: three tests written and verified failing first — the legacy-season distinctness sweep, the
"never labels one weekend with another weekend's name" pin, and the changed non-numbered expectation.
Two pinned expectations changed (`RSC`, `WFC`) with the measured reason recorded at both sites; the
spec was wrong on real data, not the implementation.

### Refuted — checked, not filed, not fixed

* **Blank weekend label / blank tab button.** `weekendLabel('')` returns `''`, so a session with no name
  would render an empty entry. Not reachable: all 79 family sessions have a non-blank `name`, a non-zero
  `cm_id` and a non-blank `start_date` on the snapshot.
* **`headcount === 0` → "No members on file for this year" while a weekend tab is selected** would be a
  year-scoped sentence on a weekend-scoped count. Not reachable: a weekend tab exists only because some
  attendee row carried it, and that row's child either has the id in `session_cm_ids` or has an empty
  list (kept visible on every tab). Every weekend tab therefore has ≥1 child.
* **Duplicate React keys from `session.session_cm_id ?? 0`** in the tab strip. The server drops a session
  with no `cm_id` before publishing (pinned by
  `test_a_session_with_no_campminder_id_is_not_published_as_weekend_zero`), so the `?? 0` is unreachable
  defence, not a live collision.
* **`camp_sessions.cm_id` repeats across years** (`1236371` is Spring Family Camp in 2017, 2018 *and*
  2019), which would break a cm_id-keyed map — but every map here is keyed `year → cm_id`, and no year
  contains a duplicate `cm_id`. Verified.
* **`housing_session_cm_id` has no client consumer.** True — the note that would have read it was struck
  by the 2026-08-18 ruling. Left in deliberately rather than deleted: it is the wire-side mirror of
  `AttributeSession`'s refusal, and deleting it would make the next reader re-derive that parity
  argument from scratch. Flagging it here so you can rule; it is one field and three server tests.

### Data claims in the PR body — independently reproduced

Both headline numbers check out exactly against the snapshot: **64** multi-weekend journey
household-years, and **7** of those carrying a child who did not attend every weekend.

### Test quality

Mutation-checked the three claims that carry the feature, reading the exit codes rather than trusting
the assertions:

| Mutation | Result |
|---|---|
| drop the `session_cm_ids` filter entirely | 2 failed — *drops a child who did not attend*, *headcount follows the tab* |
| drop the `.length === 0` clause | 1 failed — *keeps a child whose weekends are unknown visible* |
| pre-fix `weekendLabel` vs the new season sweep | 3 failed |

The modal tabs are plain `onClick` buttons, so `fireEvent.click` is the whole interaction — no focus
step is being skipped here.

### Gates

`bash scripts/pre-push-verify.sh` run bare: prettier, eslint, tsc, vitest all PASS. `vitest run
src/components/weekend/` 1285 passed. `pytest tests/unit/api/services/test_lodging_roster_service.py`
250 passed. Push confirmed by the remote ref (`git rev-list --count origin/…..HEAD` = 0).
CI Summary **pass** on `63eeb7b`. CodeRabbit deliberately not requested — the quota is spent this campaign.

### Holds for your eyes

`visual: true` by the mechanical grep. See below for what changed on screen.

PR #2463, opened 2026-08-18
fix(sync,pb): two transform-phase failures found in a production log

adamflagg:
Two independent defects surfaced by reading the Docker logs of a full
transform-phase run in production. Neither had a test, and neither was
reachable from CI.

## 1. `geo_overrides` has never loaded

`loadGeoOverrides` asked PocketBase to sort by `"year ASC"`. PocketBase sort
expressions are `field`, `+field` or `-field`, so that string was parsed as a
FIELD NAME, `FindRecordsByFilter` refused the query, and the caller swallowed
the error and continued with empty maps:

```
WARN Could not load geo_overrides, continuing without overrides
     error=loading geo_overrides: invalid sort field "year ASC"
```

Every alias, merge and rejection override has been ignored by this job for as
long as the line has existed, on every run, with nothing going red. A bare
field name is already ascending, which is the direction the loop needs: it
overwrites on each hit, so ascending year is what makes the newest year's value
the surviving one.

**Why it survived.** The three `TestOverrideMapDedup_*` tests look like
coverage of this feature, but each assigns into a hand-built Go map and asserts
on the map — none issues the query, so the sort argument was never executed by
a test. The two tests added here call `loadGeoOverrides` against a real app:
one asserts all three override kinds load, the other asserts newest-year-wins
*through* the query rather than around it. Both fail with the exact production
error before the fix, confirmed by reverting the line.

`override_type` had to be added to the shared geo fixture, which never carried
the column the loader actually switches on — `type` is a different column and
fed nothing.

**Blast radius.** Production holds one non-canonical override row, a city alias
correcting a misspelling, which starts applying on the next normalization run.
The remaining rows are `override_type = "canonical"`, which this loader does
not consume — `api/services/geo_service.py` does, and it already sorts
correctly and is unaffected.

## 2. One long free-text answer fails the whole `household_demographics` job

A parent answered a date question with a 130-character sentence saying their
plans were not settled yet. The column is capped at 100, PocketBase refused the
save, and the job reported the whole run failed:

```
ERROR Error saving household_demographics record ... year=2026
      error=away_from_date: Must be no more than 100 character(s).; ...
ERROR Phase job failed phase=transform job=household_demographics
      error=1 database operations failed
```

Every other record wrote; the job still exited `failed`, and would again on
every run until someone edited that answer in CampMinder.

Migration `1500000163` widens eleven columns to 1000. **All eleven, not the two
that failed:** each mirrors a free-text CampMinder custom field a parent types
into, and CampMinder enforces no length on any of them. The two date fields are
simply the ones a long answer reached first; `form_filler` at 200, the two
`congregation_*` and two `jcc_*` at 300, `away_phone` at 400, and
`away_location` and the two `jewish_affiliation*` at 500 are the same red job
waiting for a longer sentence, each to be found from a production log the same
way.

These caps were never a data-quality control — nothing downstream reads a
length and no validation rests on them. 1000 is the table's own existing answer
for "a paragraph of free text": `family_description_other` and
`parent_immigrant_origin` already carry it. Columns at or above 1000 are
deliberately untouched; this only raises floors. Nothing costs storage —
SQLite TEXT is variable-length and PocketBase's `max` is a record-save
validation, not a column width.

## Verified against production data, not just asserted

The migration was applied to a copy of the production snapshot: all eleven land
at 1000, no other column moves, and the real 129-character answer that broke
the job then saves (HTTP 200) where it previously could not.

The down path restores every original cap and touches no rows. Verified with a
129-character value already stored: `max` applies at record-save time, so
narrowing the field succeeds and the answer survives rather than being
truncated. Truncating to make the revert "clean" would destroy the answer,
which the no-destructive-migrations rule forbids.

`scripts/pre-push-verify.sh` reports ALL CHECKS PASSED, golangci-lint included.

## 3. The exact caps are now asserted in CI

Raised by CodeRabbit on this PR, and it names a gap the Go tests genuinely have:
they read the migration FILE and grep it, so they prove the source says 1000 and
prove nothing about the schema PocketBase ends up with. That gap is structural —
the migration widens fields by mutating them in place, so no collection literal
in `pb_migrations/` states the resulting cap.

`validate_migrations.py` already consumes `/api/collections?perPage=1000` from a
booted PocketBase after every migration has applied, which is the only place the
real number is observable — and therefore the only place a later migration
silently narrowing one of these would be caught.

`TEXT_FIELDS_REQUIRE_CUSTOM_LIMIT` could not carry it: it asks only whether a
field is sitting on PocketBase's 5000 default, which says nothing about one
capped far too LOW. So `TEXT_FIELD_EXACT_LIMITS` names the cap and
`validate_text_field_exact_limits` asserts it. A missing COLLECTION is skipped
(`validate_required_collections` already reports that); a missing FIELD is an
error, since a rename that leaves the spec behind must not read as "nothing to
check".

Round-tripped against a copy of the production snapshot: with 1500000163
reverted the check reports all eleven columns at their old caps; re-applied, it
reports none. The unit tests narrow each of the eleven in turn, so a spec
listing only the two columns that actually failed would fail nine of them.

## Deliberately not in scope

- **The job still fails wholesale when one record refuses to save.** Worth
  revisiting — one family's free text should not redden a phase — but it is a
  Go change with its own blast radius and does not belong in a schema
  migration.
- **The lodging work queue renders only `kind = "unresolved_alias"`.** The
  other kinds are recorded and have no surface, so a queue holding several
  `ambiguous_session` items reads as empty. Separate issue.

## Summary by CodeRabbit

* **New Features**
  * Household demographics text fields now support entries up to 1,000 characters without altering existing records.

* **Bug Fixes**
  * Geographic overrides are now processed in year order, ensuring the most recent applicable alias takes precedence.
  * Improved loading of geographic override records from PocketBase.

* **Tests**
  * Added coverage validating demographic field limits, reversible migrations, and geographic override precedence.

### DIFF
--- a/pocketbase/main_household_demographics_text_caps_test.go
+++ b/pocketbase/main_household_demographics_text_caps_test.go
@@ -1,0 +1,101 @@
+package main
+
+import (
+	"os"
+	"strings"
+	"testing"
+)
+
+const householdDemographicsTextCapsMigration = "pb_migrations/1500000163_household_demographics_text_caps.js"
+
+// householdDemographicsWidenedFields is every text column on
+// household_demographics that 1500000041 capped below 1000 characters.
+//
+// The two that forced the change are the date pair. A parent answered
+// `HH-Away From (mm/dd/yy)` with a 130-character sentence about plans not
+// being settled yet, PocketBase refused the save at max=100, and
+// household_demographics reported the WHOLE JOB failed -- one household's
+// free text turning a transform-phase job red:
+//
+//	ERROR Error saving household_demographics record ... error=away_from_date:
+//	      Must be no more than 100 character(s).; away_return_date: ...
+//	ERROR Phase job failed phase=transform job=household_demographics
+//
+// The rest are in the list because they are the same KIND of column -- free
+// text a parent types into a CampMinder custom field, with no length limit at
+// the source -- so each one is the same red job waiting for a longer answer.
+// A cap on these was never a data-quality control: nothing downstream reads a
+// length, and CampMinder does not enforce one.
+var householdDemographicsWidenedFields = []string{
+	"jewish_affiliation",
+	"jewish_affiliation_other",
+	"congregation_summer",
+	"congregation_family",
+	"jcc_summer",
+	"jcc_family",
+	"away_location",
+	"away_phone",
+	"away_from_date",
+	"away_return_date",
+	"form_filler",
+}
+
+func readHouseholdDemographicsTextCapsMigration(t *testing.T) string {
+	t.Helper()
+	content, err := os.ReadFile(householdDemographicsTextCapsMigration)
+	if err != nil {
+		t.Fatalf("read migration %s: %v", householdDemographicsTextCapsMigration, err)
+	}
+	return string(content)
+}
+
+// TestHouseholdDemographicsTextCapsNamesEveryShortField guards the list above
+// against a migration that widens only the two columns that happened to fail
+// in production. Every one of them is fed from the same place.
+func TestHouseholdDemographicsTextCapsNamesEveryShortField(t *testing.T) {
+	body := readHouseholdDemographicsTextCapsMigration(t)
+
+	for _, field := range householdDemographicsWidenedFields {
+		if !strings.Contains(body, `'`+field+`'`) {
+			t.Errorf("migration %s must widen %q", householdDemographicsTextCapsMigration, field)
+		}
+	}
+}
+
+// TestHouseholdDemographicsTextCapsWidensToOneThousand pins the target. 1000 is
+// the cap `family_description_other` and `parent_immigrant_origin` already
+// carry on this same table, so it is the table's own existing answer for "a
+// paragraph of free text" rather than a new number.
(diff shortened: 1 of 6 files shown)


adamflagg:
@coderabbitai review

adamflagg:
Folded in — thanks. Confirmed the point applies: `validate_migrations.py` runs against `/api/collections` on a booted PocketBase after every migration applies, so it *can* observe a cap the migration sets by mutating a field in place, which no static read of `pb_migrations/` can.

Added `TEXT_FIELD_EXACT_LIMITS` + `validate_text_field_exact_limits`. `TEXT_FIELDS_REQUIRE_CUSTOM_LIMIT` could not carry it: it only asks whether a field is on PocketBase's 5000 default, which says nothing about one capped far too low — the failure this PR exists for.

Round-tripped against a copy of the production snapshot rather than fixtures alone: with the migration reverted the check reports all eleven columns at their old caps, and re-applied it reports none. The unit tests narrow each of the eleven in turn, so a spec listing only the two columns that actually failed would fail nine of them.

Issue #2465, opened 2026-08-18
bunk_assignments: every active bunk staff assignment is created by a full sync and deleted by the next hourly sync


adamflagg:
Observed end to end in one production log, inside a single 20-minute window.

## What happens

| line | event |
|---|---|
| 289 | `bunk_assignments` **run 1** starts (inside a full sync) |
| 301 | `No orphaned records found entity=bunk assignment` |
| 302 | run 1 completes — **created=262** |
| 415 | scheduled hourly sync fires |
| 416 | `bunk_assignments` **run 2** starts |
| 426–687 | **262** × `Ambiguous (bunkPlan, bunk) staff session lookup, skipping assignment … candidateCount=2` |
| 690–951 | **262** × `Deleting orphaned record entity=bunk assignment` |
| 954 | run 2 completes |

The same 262 rows. All 70 distinct people involved are `status = 'active'`,
`bunk_staff = true`.

## Root cause: the lookup maps are never reset between runs

`bunkPlanSessionsList` and `bunkPlanBunkToSession` are allocated once in
`NewBunkAssignmentsSync` and only ever **appended** to. The orchestrator
registers a single long-lived instance, and the scheduler reuses it every hour,
so the same `bunk_plans` rows are appended again on every run:

- run 1 → `bunkPlanBunkToSession["14896:56959"] == [1356533]`, length 1 → resolves
- run 2 → `[1356533, 1356533]`, length 2 → **"ambiguous"**
- run N → length N

`resolveStaffSession` switches on `len(candidates)`, not on the number of
DISTINCT sessions, so two copies of the *same* session read as a conflict. The
data is unambiguous: plan 14896 has nine rows in 2026, one per bunk, all naming
session 1356533. `candidateCount=2` on the second run is the giveaway — it
equals the run count, not anything in the data.

Only the staff path is affected. Campers resolve earlier in the ladder, via
enrollment; staff are not in `attendees`, so `resolveStaffSession` is the only
step that can answer for them.

## Why the sweep then deletes them

`processAssignment` is what calls `TrackProcessedCompositeKey`, and the skip
`continue`s before reaching it. So a skipped assignment is absent from the
processed set, and `deleteOrphans` treats absence as "no longer in CampMinder".

`protectNonActiveStaffAssignments` does not help: it protects staff with
`status != 'active'`, and all 70 here are active. That protection exists for the
opposite population — staff CampMinder has stopped returning.

[...]

PR #2469, opened 2026-08-19
fix(sync): reset Stats at the top of reconcile_request_lifecycle's Sync

### LINKED ISSUE #2468: reconcile_request_lifecycle never resets Stats, so the first error makes every later run report failed until restart
Found while auditing the accumulation class behind #2465. Same class — mutable
state on a long-lived service instance that is never reset between runs — but a
different file and a different symptom, so it wants its own change.

**Latent today, not currently firing.** All 5 recorded runs have
`errors_count = 0`, so nothing has tripped it yet. It arms the moment anything
does.

## The defect

`ReconcileLifecycleSync` (in `pocketbase/sync/reconcile_request_lifecycle.go` —
note the type name drops "Request", unlike the file and the service name) does
this:

```go
func (s *ReconcileLifecycleSync) WasSuccessful() bool {
    return s.Stats.Errors == 0        // :56
}

func (s *ReconcileLifecycleSync) Sync(_ context.Context) error {
    // ... no `s.Stats = Stats{}` anywhere
    s.Stats.Errors++                  // :69  (year resolution failure)
    s.Stats.Errors += markErrors      // :74
    s.Stats.Errors++                  // :76
}
```

`Sync` only ever increments `Stats.Errors` and never resets it. The orchestrator
registers **one** long-lived instance (`orchestrator.go:2328`, and `:1731` for the
year-scoped variant), so the counter is per-process, not per-run.

`applyCompletionStatus` derives a run's status from `stats.Errors > 0`. So after
the first error of a process's life:

- every subsequent run records as **failed**, however clean it was
- `errors_count` climbs monotonically, so the number is a since-boot total rather
  than a run total
- it only clears on a container restart

## The fix

`s.Stats = Stats{}` at the top of `Sync`, as five siblings already do:

| file | line |
|---|---|
| `bunks.go` | 58 |
| `bunk_plans.go` | 71 |
| `enrollment_snapshots.go` | 106 |
| `financial_lookups.go` | 38 |
| `multi_workbook_export.go` | 80 |

## Test

Run `Sync` twice on **one** instance with a failure injected into the first run
only, and assert the second run reports `WasSuccessful() == true` and
`Stats.Errors == 0`. A test that constructs a fresh service per run cannot catch
this — that is exactly why #2465 survived, and the same trap applies here.

## Scope

Deliberately its own change rather than riding along with #2465: different file,
different service, different failure mode, and it should be revertable on its own.

## Related siblings found in the same audit, NOT fixed here

Recorded so they are not re-discovered from scratch. All are log-only or bounded,
none deletes data:

- `persons.go:66` (`missingDataStats`) and `:489` (`skippedStaff`) — never reset,
  while `Sync` (`:125-130`) resets four other fields. The data-quality log at
  `:1152-1158` therefore reports counts multiplied by runs since boot.
- `base_sync.go:104` `FieldDiffStats` — accumulates for every embedding service
  (~16 of them). `ClearFieldDiffStats` (`:1709-1712`) has zero production callers.
  Two effects: `LogFieldDiffSummary` reports since-boot totals, and the `<= 3`
  debug gate at `:753` silences per-field logging permanently after run 1.
- Stale-but-bounded (these overwrite rather than append, so they do not grow — but
  a key admitted once is never pruned): `attendees.go:36/:169`,
  `bunk_plans.go:48-52/:103`, and `bunk_assignments.go`'s own five bool maps
  (`:278, :306, :307, :318, :385`). Concretely: a camper who leaves
  `status_id = 2` after run 1 stays permanently "valid" for that process.

## Verified safe, so nobody needs to re-audit them

`lodging_assignments_sync.go` (the closest structural twin — re-makes at
`:813/:825/:830`, assigns wholesale at `:269/:336`), `family_camp_derived.go:368-371`,
`normalize_geographic.go:131`, `staff_skills.go:116`, `bunk_requests.go:56/:196`,
`sessions.go:88`, and the `ratelimit` package (cross-run state self-heals on
`Success()`).

adamflagg:
The orchestrator registers **one** long-lived `ReconcileLifecycleSync` for the life of
the process (`orchestrator.go:2328`, plus `:1731` for the year-scoped path), and its
`Sync` only ever incremented `Stats.Errors` — it never reset the counter.

`WasSuccessful()` is `s.Stats.Errors == 0`, and `applyCompletionStatus` derives a run's
status from `stats.Errors > 0`. So the first error of a process's life made every later
run record as **failed**, however clean it was, with a monotonically climbing
`errors_count` that read as a since-boot total rather than a run total, and cleared only
on container restart.

Latent today rather than firing: all 5 recorded runs have `errors_count = 0`. It arms the
moment anything errors.

## The change

`s.Stats = Stats{}` at the top of `Sync`, matching the five siblings that already do it —
`bunks.go`, `bunk_plans.go`, `enrollment_snapshots.go`, `financial_lookups.go`,
`multi_workbook_export.go`. One line plus the comment explaining why it is load-bearing.

## The test

`TestReconcileLifecycleSync_ResetsStatsBetweenRunsOnSameInstance` runs `Sync` twice on
**one** instance, with a failure injected into the first run only (the collections the
reconciliation queries do not exist yet, so the attendees lookup errors out), then asserts
the second, clean run reports `WasSuccessful() == true` and `Stats.Errors == 0`.

Reusing the instance is the whole point. A test that constructs a fresh service per run
cannot observe carry-over at all, which is how this class of bug survives its own suite —
the same trap that let #2465 through. Verified by mutation: with the reset removed the new
test fails on run 2 with `Stats.Errors = 1, want 0`; with it restored the test passes.

## Verification

- `bash scripts/pre-push-verify.sh` — ALL CHECKS PASSED (go build, golangci-lint 0 issues, go test)
- `go test ./sync/` — ok, 282s
- `golangci-lint run ./...` — 0 issues

## Scope

Deliberately its own change so it is revertable on its own. `bunk_assignments.go` is
untouched — it belongs to the sibling change for #2465. The other accumulation sites the
same audit recorded (`persons.go`, `base_sync.go`'s `FieldDiffStats`, and the
stale-but-bounded maps) are left alone here; they stay written down in the issue body.

Closes #2468.

### DIFF
--- a/pocketbase/sync/reconcile_request_lifecycle.go
+++ b/pocketbase/sync/reconcile_request_lifecycle.go
@@ -61,6 +61,12 @@ func (s *ReconcileLifecycleSync) WasSuccessful() bool {
 // to CAMPMINDER_SEASON_ID via ParseSeasonYear, matching the convention used
 // by every other yearless service in this package.
 func (s *ReconcileLifecycleSync) Sync(_ context.Context) error {
+	// Stats describe THIS run, not the process. The orchestrator registers one
+	// long-lived instance, so without this reset the first error of a process's
+	// life makes WasSuccessful() false — and applyCompletionStatus records every
+	// later run as failed — until the container restarts.
+	s.Stats = Stats{}
+
 	year := s.Year
 	if year == 0 {
 		var err error
--- a/pocketbase/sync/reconcile_request_lifecycle_test.go
+++ b/pocketbase/sync/reconcile_request_lifecycle_test.go
@@ -507,3 +507,54 @@ func TestReconcileRequestLifecycle_SuccessPathReturnsZeroMarkErrors(t *testing.T
 		t.Errorf("markErrors = %d, want 0", markErrors)
 	}
 }
+
+// TestReconcileLifecycleSync_ResetsStatsBetweenRunsOnSameInstance pins the
+// per-run contract for Stats on the ONE long-lived instance the orchestrator
+// registers (orchestrator.go registers a single ReconcileLifecycleSync for the
+// life of the process, plus one for the year-scoped path). A run's Stats.Errors
+// must describe that run alone: applyCompletionStatus derives a run's status
+// from stats.Errors > 0, so a counter that carries over makes every later run
+// record as failed, with a monotonically climbing errors_count, until the
+// container restarts.
+//
+// The instance is reused deliberately. A test that constructs a fresh service
+// per run cannot observe carry-over at all, which is how this class of bug
+// survives its own suite.
+func TestReconcileLifecycleSync_ResetsStatsBetweenRunsOnSameInstance(t *testing.T) {
+	t.Parallel()
+	app, err := pbtests.NewTestApp()
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer app.Cleanup()
+
+	s := NewReconcileLifecycleSync(app)
+	s.SetYear(2025)
+
+	// Run 1 -- failure injected here and only here: the collections the
+	// reconciliation queries do not exist yet, so the attendees lookup errors
+	// out and the run records one error.
+	if syncErr := s.Sync(context.Background()); syncErr == nil {
+		t.Fatal("run 1: want an error from the missing collections to inject the failure, got nil")
+	}
+	if s.Stats.Errors != 1 {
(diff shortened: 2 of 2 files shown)


adamflagg:
### Code review

PR #2471, opened 2026-08-19
fix(api): the solver refuses a scenario-less run instead of writing production

### LINKED ISSUE #2467: Applying a solver run with no scenario attempts a production write instead of refusing
> **Corrected 2026-08-18.** The original body recommended making the production
> branch write correctly. That was wrong, and the owner's ruling is the reason:
> **the solver cannot write to production — it applies to scenarios only, and
> production is read-only for both the solver and drag/drop.** Making that branch
> functional would build a write path the product deliberately forbids. The defect
> is real; the fix is the opposite of what this issue first proposed. The original
> analysis of the payload is preserved below because it is still accurate about
> what the code does.

## What happens

`POST /api/solver/apply/{run_id}` branches on whether the run carries a scenario
(`solver.py:590`). With one, it writes `bunk_assignments_draft` correctly. Without
one, it takes an `else` branch that tries to write **`bunk_assignments`** — the
production table — with this payload (`:740-748`):

```python
assignment_data = {
    "person_cm_id": person_cm_id,
    "session_cm_id": actual_session_cm_id,
    "bunk_cm_id": bunk_cm_id,
    "year": run_year,
    "assigned_at": ...,
    "assigned_by": "solver",
}
```

`bunk_assignments` has six writable columns — `cm_id`, `person`, `session`,
`bunk`, `bunk_plan`, `year`. Only `year` is real here; `person`/`session`/`bunk`
are required **relations** and are never set. The write fails validation and the
`except Exception` at `:758` logs it and continues to the next camper, so the
endpoint returns success having written nothing.

## Why this is a refusal, not a repair

Production is read-only for the solver. The branch is therefore attempting
something that must not succeed — and it is only failing to succeed by accident,
because the payload happens to be malformed. A correct payload here would be the
actual bug.

The lodging side already encodes exactly this rule and is the precedent to match:
`PlacementWriteBase.scenario` is documented as *"REQUIRED and non-empty. A blank
scenario is a 422, never a write to the live plan."*

## It is reachable

`scenario_id` is required only when `session_cm_id` is unset (`solver.py:154-155`).
A run created with a `session_cm_id` and no scenario therefore reaches the
production branch on apply.

## Suggested fix

1. **Refuse explicitly.** A run with no scenario should get a 4xx from
   `apply_solver_results` — "solver output applies to a scenario; production is
   read-only" — instead of a per-camper error log and a success response. Decide
   whether to refuse at apply time, at run creation, or both.
2. **Delete the production write branch** rather than leaving dead code that reads
   as an intended capability.
3. **Guard the class, not the instance.** A test asserting the solver never writes
   to `bunk_assignments` is worth more than one asserting this payload is
   well-formed — it pins the architectural rule instead of the symptom.
4. Narrow the `except Exception` at `:758` regardless. A swallowed write failure
   that still returns success is the reason this went unnoticed.

## Unverified

The payload construction, the column list, the reachability condition and the
swallowing `except` are all read from the code. The endpoint was not run end to
end, so "returns success while writing nothing" is inferred rather than observed.

adamflagg:
Production bunk assignments are read-only for the solver. Applying a solver run
with no scenario used to take an `else` branch that tried to create a
`bunk_assignments` row from a payload whose only real column was `year` —
`person`, `session` and `bunk` are required relations and were never set. The
create failed validation, a bare `except Exception` logged it once per camper,
and the endpoint returned 200 having written nothing.

The fix is to refuse, not to repair the payload. A *working* production write
there would be the actual defect: the solver applies to a scenario's draft
assignments and nowhere else. That is the rule the lodging side already
encodes on `PlacementWriteBase.scenario` — *"REQUIRED and non-empty. A blank
scenario is a 422, never a write to the live plan."* — and the reasoning is
quoted at the refusal site so it travels.

## What changed

**Refusal at both ends, which is a decision worth a second look.** The issue
left it open whether to refuse at apply time, at run creation, or both. This
does **both**, so half can be reverted cheaply if that is not wanted:

- **Apply time** — `POST /api/solver/apply/{run_id}` returns 422 when the run
  carries no scenario, before it touches PocketBase. This is the half that
  covers runs already stored.
- **Creation time** — `POST /api/solver/run` and
  `POST /api/solver/run-multi-session` return 422 when `scenario` is missing or
  blank. This is the stronger guarantee: a run that can only ever be applied to
  production is never created in the first place.

Creation-time refusal is a visible behaviour change: the board can currently
run the solver with no scenario selected, and the run appears to succeed. What
that flow actually produces today is an apply that writes nothing and reports
success, so nothing that works is being taken away — but staff will now see an
error where they previously saw a silent no-op.

**The production write branch is deleted** rather than left as dead code that
reads like an intended capability. With the refusal in place `scenario` is
always truthy inside the loop, so the branch, the `table_name` ternary and the
`invalidate_session` fallback all collapse to the draft path.

**The swallowing `except Exception` is narrowed** to
`(ClientResponseError, ValueError)`, and failures are counted rather than
discarded: per-camper resilience is kept, but if any write failed the endpoint
now returns 500 with the counts instead of "Applied N assignments". Reporting
success over a failed write is why this went unnoticed.

## Tests

The guard pins the **class**, not the instance — a payload-shape test would
start passing the moment someone "fixed" the production branch into a working
write, which is the outcome being prevented. `test_solver_never_writes_production.py`:

- a recording PocketBase stand-in drives the real apply path with a camper that
  resolves all the way to the write, and asserts **no mutation ever reaches
  `bunk_assignments`** — with a scenario or without one;
- an AST tripwire asserts `apply_solver_results` does not so much as name the
  production collection, in either spelling (the literal or the
  `BUNK_ASSIGNMENTS` constant);
- refusal tests for both apply paths (in-memory run and run loaded from
  PocketBase), for both creation endpoints, and for blank/whitespace scenarios;
- a test that a failed draft write is not reported as success.

Both guards were mutation-checked: pointing the write at `bunk_assignments`
fails the behavioural guard and the AST guard, and deleting the apply-time
refusal fails the three refusal tests.

Three existing test files moved with the specification:

- `test_draft_writes_invalidate_cache.py` — `test_prod_apply_invalidates_session_slot`
  asserted the old production behaviour (200 + session-slot invalidation). It now
  asserts the refusal and that nothing is invalidated, since there is no write to
  go stale. This is a spec change, not a test bent to fit the code.
- `test_solver_singleflight.py` — request bodies now carry a scenario. These
  tests are about the single-flight guard, which the scenario refusal now
  precedes.
- `test_http_exception_500_str_e.py` — the two multi-session no-leak probes need
  a scenario to reach the PocketBase call they raise on.

## Deliberately not touched

- **`POST /api/solver/run-sweep` with `session_cm_id`** still creates
  scenario-less runs. A sweep is a benchmark that never applies, and refusing it
  would delete the session-based sweep feature — a bigger product call than this
  issue settles. Its children are covered by the apply-time refusal.
- **`POST /api/sessions/{cm_id}/clear-assignments`** still *deletes* from
  production `bunk_assignments` when no scenario is given. Same file, same
  read-only question, but a different decision and out of scope here.
- **The hourly orphan sweep** (`pocketbase/sync/bunk_assignments.go`) is
  untouched. It is the reason a repaired payload would have silently reverted
  within the hour, and it is being worked in other branches.
- **`docs/api/solver-api.md`** still says the apply endpoint applies "to
  production bunk assignments" and documents `apply_results` as auto-applying to
  production. That is now wrong and wants a follow-up; it is outside this PR's
  scope lock.

## Verification

`bash scripts/pre-push-verify.sh` — ALL CHECKS PASSED (ruff format, ruff check,
mypy, 6361 passed / 14 skipped). No Go or frontend files changed, so
golangci-lint and eslint are not implicated.

Closes #2467.

### DIFF
--- a/api/routers/solver.py
+++ b/api/routers/solver.py
@@ -33,6 +33,7 @@ from bunking.solver.impossibility import filter_immaterial_requests, validate_im
 
 from ..constants.collections import (
     ATTENDEES,
+    BUNK_ASSIGNMENTS_DRAFT,
     BUNKS,
     CAMP_SESSIONS,
     SOLVER_RUNS,
@@ -74,6 +75,31 @@ def _resolve_time_limit(value: int | None, default: int = 60) -> int:
     return value if value is not None else default
 
 
+_SCENARIO_REQUIRED_DETAIL = (
+    "A scenario is required: solver output applies to a scenario's draft assignments. "
+    "Production bunk assignments are read-only for the solver."
+)
+
+
+def _require_scenario(scenario: str | None) -> str:
+    """Return *scenario*, or refuse with 422 when it is missing or blank.
+
+    The solver cannot write production. It applies to a scenario's
+    ``bunk_assignments_draft`` rows and nowhere else — production bunk
+    assignments are read-only for both the solver and drag/drop. A scenario-less
+    run therefore has nowhere legal to land, so it is refused rather than
+    repaired (kindred#2467: the old production branch built a payload the
+    collection rejected, logged the failure per camper and still returned 200).
+
+    Same rule the lodging side already encodes on ``PlacementWriteBase.scenario``
+    (``frontend/src/services/lodgingApi.ts``): *"REQUIRED and non-empty. A blank
+    scenario is a 422, never a write to the live plan."*
+    """
+    if scenario is None or not scenario.strip():
+        raise HTTPException(status_code=422, detail=_SCENARIO_REQUIRED_DETAIL)
+    return scenario
+
+
 # ========================================
 # Solver Run Endpoints
 # ========================================
@@ -86,6 +112,10 @@ async def run_solver(
     user: AuthUser = Depends(require_permission(Permission.BUNKING_MANAGE)),
 ) -> SolverResponse:
     """Run the bunking solver for a session."""
+    # Refuse before anything else: a run with no scenario can only ever be
+    # applied to production, which is read-only for the solver.
+    _require_scenario(request.scenario)
+
     # Single-flight guard: reject duplicate in-progress runs for the same session.
     # FastAPI serializes coroutines on the event loop and solver_runs is in-process,
     # so no locking is needed beyond a plain dict scan.
@@ -571,9 +601,9 @@ async def analyze_solver_run(
 async def apply_solver_results(
     run_id: str, user: AuthUser = Depends(require_permission(Permission.BUNKING_MANAGE))
 ) -> dict[str, str]:
-    """Apply the results of a solver run to the database."""
+    """Apply the results of a solver run to its scenario's draft assignments."""
     session_cm_id = None
-    scenario = None
+    scenario: str | None = None
 
     results: dict[str, Any] = {}
     if run_id not in solver_runs:
@@ -598,6 +628,10 @@ async def apply_solver_results(
         session_cm_id = run_data.get("session_cm_id")
         scenario = run_data.get("scenario")
 
+    # Refuse before touching PocketBase. A run with no scenario has nowhere
+    # legal to land — the solver does not write production.
+    scenario = _require_scenario(scenario)
+
     assignments = results["assignments"]
 
     # Get year from run config, fall back to results or current year
@@ -621,6 +655,8 @@ async def apply_solver_results(
     ctx = await build_session_context(int(session_cm_id), run_year, pb)
     session_filter = ctx.session_relation_filter
 
+    write_failures: list[str] = []
+
     for person_cm_id_str, bunk_name in assignments.items():
         try:
             person_cm_id = int(person_cm_id_str)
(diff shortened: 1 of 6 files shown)


adamflagg:
## Triage-attack Part B follow-up

Pushed `118efc60` with the three M-sized review findings from this pass, rolled into the branch per the campaign's fix-not-file rule:

- **`docs/api/solver-api.md`** — the "Basic Solving" walkthrough still showed a scenario-less `POST /solver/run`, which now 422s under this PR's refusal. Corrected end to end (create a scenario, pass it through the run, apply message matches the real `"...to bunk_assignments_draft"` text).
- **`api/routers/solver.py:736`** (`apply_solver_results`) — `existing_record.id` on a malformed existing-draft-record shape raised `AttributeError`, which the narrowed `except (ClientResponseError, ValueError)` no longer catches. That would have aborted the whole remaining batch as an unhandled 500 instead of degrading the one camper. Fixed with `getattr(..., "id", None)` and routed through the existing `write_failures` counting — no new mechanism, no re-widened `except`. Added a regression test that's confirmed red against the pre-fix code (Starlette's bare 500 vs. the endpoint's own JSON failure response) and green after.
- **`test_apply_function_never_names_the_production_collection`** — docstring now says explicitly that it's a secondary, foolable tripwire (a module-level alias for the production collection walks past the AST scan) and that the `RecordingPB`-based behavioral tests are the load-bearing coverage. Nobody should simplify the suite down to just the AST check.

### The L+ UI-gated finding — investigated, not filed

The fourth finding from this pass proposed filing an issue for "no frontend guard" on the creation-time refusal: staff viewing the live board with no scenario selected would hit this PR's new 422 where they previously got a silent 200.

I verified the actual call path before filing anything (`useSolverOperations.handleRunSolver` → `solverService.runSolver` → `POST /solver/run` has exactly one caller in the app: `OptimizeBunksButton`, rendered from `SessionHeader.tsx`). That render is gated:

```
frontend/src/components/session/SessionHeader.tsx:196
{canManage && !isProductionMode && (
  <OptimizeBunksButton ... onRunSolver={onRunSolver} ... />
)}
```

`isProductionMode` is `currentScenario === null` (`ScenarioContext.tsx:24`). So the button that would send a scenario-less run is not rendered at all while viewing the live board — there's no click path to the 422 through the UI. This is deliberate, tested behavior, not an oversight: `SessionHeader.test.tsx:145` asserts `does NOT render the Optimize Bunks (run-solver) button` when `isProductionMode: true`, and the "Clear" (delete) button is gated identically at `SessionHeader.tsx:213`. Both trace back to #1413 ("show precheck button in prod (keep solve/clear hidden)"), closed well before this PR.

Filing an issue on the stated premise would have been a wrong-premise issue on day one — the kind this repo's triage culture explicitly guards against. No issue filed. If a *different* path to a scenario-less request turns out to exist (a stale cached bundle, a direct API call, a future caller that doesn't reuse `OptimizeBunksButton`), this PR's 422 is exactly the backend-side defense that covers it — which is the point of doing this at the API layer rather than only in the UI.

PR #2472, opened 2026-08-19
fix(sync): bunk staff assignments no longer created by one sync and deleted by the next

### LINKED ISSUE #2465: bunk_assignments: every active bunk staff assignment is created by a full sync and deleted by the next hourly sync

adamflagg:
The orchestrator constructs one BunkAssignmentsSync at boot and dispatches
every scheduled run to it, but loadMappings only ever APPENDED to the maps it
builds. From run 2 onward each hourly sync read the previous run's candidates
still sitting in the map alongside its own fresh copy, so a (bunkPlan, bunk)
pair with exactly ONE bunk_plans row behind it presented as two candidate
sessions. resolveStaffSession switched on len(candidates), called that
ambiguous, and skipped -- and the skip continued before
TrackProcessedCompositeKey, so the row was absent from the processed set and
deleteOrphans read absence as "CampMinder no longer returns this" and deleted
it. 262 rows for 70 active bunk staff, deleted every hour, restored only by a
container restart.

Four interlocking changes; shipping any one alone leaves the others armed.

Re-make all eight maps at the top of loadMappings. Not in Sync()'s reset block:
loadMappings is their only writer, and bunkPBIDToCMID must be rebuilt before
the bunk_plans pass reads it, which the existing load order guarantees. The
constructor's make() calls stay -- several tests build the struct literally and
never call loadMappings. stranded_assignment_cleanup.go carries the same reset
for Stats and names the same hazard.

Decide on DISTINCT sessions at both read sites. resolveStaffSession switched on
the raw candidate count; resolveViaBunk counted matches per candidate
OCCURRENCE. The decision belongs at the read sites, not the build site: a list
legitimately holds one entry per bunk_plans row, and deduping on insert would
erase kindred#2264's regression guard.

Give an unresolved assignment its own counter, Stats.UnresolvedSession. Folded
into Stats.Skipped it was invisible -- base_sync increments the same field for
every unchanged row -- which is why 119 destructive runs all reported
status='success' with a flat skipped_count while the rows merely moved from the
no-change bucket into the ambiguous one. Skipped still increments too, so the
persisted skipped_count keeps its meaning; the new field is a named subset.
No migration: like SkippedValues it is not persisted to sync_runs.

Track the keys before both skips, from a person:bunk index off
preloadExistingAssignments. The run SAW this person in this bunk and only
failed to name the session; absence from ProcessedKeys claims the opposite.
This is persons.go's kindred#2394 patch applied to the same symptom.

Also restored by the map reset, and worth stating because it was silent: the
same accumulation disabled kindred#2259/#2264's bunk-specific camper narrowing
from run 2 onward, reverting it to the plan-wide fallback. That is a latent
regression, not a second live bug -- no 2026 row shows the collapse signature.

Verifying this in production needs care: deploying restarts the container,
which clears the maps, so the FIRST run after deploy looks healthy whether or
not the fix is real. The acceptance test is the run after that -- staff
assignments must still be present on the following hour. Watch
unresolved_session in the bunk_assignments completion line, not skipped_count.

Tests, written failing first: a two-run test on ONE instance driving the real
loadMappings (the existing two-run test builds a fresh instance per run AND
hand-assigns the maps, which is why it was green through all of this), plus
read-site tests for both distinct-session decisions and one for the counter and
the unresolved-key tracking. All five mutations checked red.

Closes #2465.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved bunk-assignment synchronization across multiple sessions and repeated sync runs.
  * Prevented valid assignments from being duplicated, incorrectly removed, or merged across bunks.
  * Improved handling of ambiguous, unresolved, or session-less assignments.
  * Ensured departed assignments are removed while valid unresolved records are preserved.
  * Fixed stale matching data from affecting subsequent synchronization runs.

* **New Features**
  * Sync results now report the number of bunk assignments that could not be matched to a session.

---

Added in review remediation (see the review-remediation comment for the mutation
transcripts):

A FIFTH interlocking skip, pre-existing and one line above the two above. The
assignment loop dropped an entire (bunkPlan, bunk) group on
`len(bunkPlanSessions) == 0` without tracking OR counting -- the same
untracked-skip-feeds-the-sweep mechanism, a whole group at a time. It now warns
once for the group and falls through, so every assignment routes through
resolveSessionOrTrackUnresolved. That cannot write a row to a wrong session:
loadMappings gates bunkPlanSessionsList and bunkPlanBunkToSession on the same
`sessionCMID > 0`, so an empty plan-session list guarantees an empty candidate
list and every step of the ladder fails. The trigger is narrow -- bunk_plans.session
is required, so it needs a dangling relation or a null cm_id -- and the blast
radius is a group rather than a row.

Sync()'s per-result body moved into processResultGroup to make that branch
reachable from a test at all: Client is a concrete *campminder.Client, so Sync()
cannot be driven without an HTTP fake, which is why nothing pinned these
branches. The extraction is mechanical -- three `continue`s became `return`s.
The three shape checks above it (missing BunkID, missing BunkPlanID, no
Assignments array) deliberately stay untracked and now say why: they name no
person and no bunk, so there is no key to keep from the sweep.

Both new DB-backed tests were also re-shaped. Each seeds enough resolving rows
that OrphanSweepGuard's empty-computed-set arm cannot fire, and an unrelated
stored row the sweep MUST delete -- so removing the tracking now fails on the
row that was destroyed rather than on the guard refusing, and a sweep that did
nothing at all can no longer satisfy them.

### DIFF
--- a/pocketbase/sync/bunk_assignments.go
+++ b/pocketbase/sync/bunk_assignments.go
@@ -81,7 +81,7 @@ func (s *BunkAssignmentsSync) Sync(ctx context.Context) error {
 	// We need to build keys based on person/session CampMinder IDs stored during creation
 	year := s.Client.GetSeasonID()
 
-	existingAssignments, err := s.preloadExistingAssignments(year)
+	existingAssignments, existingByPersonBunk, err := s.preloadExistingAssignments(year)
 	if err != nil {
 		return err
 	}
@@ -196,17 +196,13 @@ func (s *BunkAssignmentsSync) Sync(ctx context.Context) error {
 				// full ladder; ambiguous is only ever true on the staff path,
 				// where a (bunkPlan, bunk) key has more than one candidate
 				// session and guessing would be worse than skipping (kindred#2264).
-				sessionID, ambiguous := s.resolveAssignmentSession(personCMID, bunkPlanID, bunkID, bunkPlanSessions)
-				if ambiguous {
-					candidateCount := len(s.bunkPlanBunkToSession[fmt.Sprintf("%d:%d", bunkPlanID, bunkID)])
-					slog.Warn("Ambiguous (bunkPlan, bunk) staff session lookup, skipping assignment",
-						"bunkPlanCMID", bunkPlanID, "bunkCMID", bunkID, "personCMID", personCMID,
-						"candidateCount", candidateCount)
-					s.Stats.Skipped++
-					continue
-				}
-				if sessionID == 0 {
-					s.Stats.Skipped++
+				// Both no-session outcomes go through one wrapper because both
+				// owe the same bookkeeping before they skip -- counting the
+				// failure where it can be seen, and keeping the assignment's
+				// stored rows out of the orphan sweep (kindred#2465).
+				sessionID, unresolved := s.resolveSessionOrTrackUnresolved(
+					personCMID, bunkPlanID, bunkID, bunkPlanSessions, existingByPersonBunk, year)
+				if unresolved {
 					continue
 				}
 
@@ -254,7 +250,11 @@ func (s *BunkAssignmentsSync) Sync(ctx context.Context) error {
 	}
 
 	// Use extra stats to show fetched count
-	s.LogSyncComplete("Bunk assignments", fmt.Sprintf("fetched=%d assignments", totalAssignments))
+	// unresolved_session is named in the completion line, not left to the JSON
+	// alone: it is the counter an operator watches to know kindred#2465 has not
+	// come back, and LogSyncComplete's own stats string cannot show it.
+	s.LogSyncComplete("Bunk assignments", fmt.Sprintf("fetched=%d assignments, unresolved_session=%d",
+		totalAssignments, s.Stats.UnresolvedSession))
 
 	return nil
 }
@@ -264,6 +264,34 @@ func (s *BunkAssignmentsSync) loadMappings() error {
 	slog.Info("Loading valid CampMinder IDs")
 
 	year := s.Client.GetSeasonID()
+
+	// Every map this function builds describes THIS run and is rebuilt from
+	// scratch here (kindred#2465). The orchestrator constructs one
+	// BunkAssignmentsSync at boot and dispatches every scheduled run to it, so
+	// without this the append-shaped maps -- personEnrollments,
+	// bunkPlanSessionsList, bunkPlanBunkToSession -- gained a full extra copy of
+	// their contents every hour: a (bunkPlan, bunk) pair with ONE bunk_plans row
+	// behind it read as two candidates on run 2, resolveStaffSession called that
+	// ambiguous, the assignment was skipped without being tracked, and
+	// deleteOrphans deleted the row for being untracked. 262 rows for 70 active
+	// bunk staff, every hour, restored only by a container restart. The same
+	// accumulation silently disabled kindred#2259/#2264's bunk-specific
+	// narrowing on the camper path from run 2 onward.
+	//
+	// It belongs HERE, not in Sync()'s reset block: loadMappings is the only
+	// writer of all eight, and bunkPBIDToCMID must be rebuilt before the
+	// bunk_plans pass below reads it -- which the existing load order already
+	// guarantees. The constructor's make() calls stay: several tests build
+	// BunkAssignmentsSync as a struct literal and never call loadMappings.
+	// stranded_assignment_cleanup.go's Sync() names the same hazard for Stats.
+	s.validPersonCMIDs = make(map[int]bool)
+	s.validBunkCMIDs = make(map[int]bool)
+	s.validSessionCMIDs = make(map[int]bool)
+	s.personEnrollments = make(map[int][]int)
+	s.bunkPlanSessionsList = make(map[int][]int)
+	s.staffPersonCMIDs = make(map[int]bool)
+	s.bunkPlanBunkToSession = make(map[string][]int)
+	s.bunkPBIDToCMID = make(map[string]int)
 
 	// Load person enrollments: personCMID -> list of sessionCMIDs they're enrolled in
 	// This is the source of truth for which session a person belongs to
@@ -408,7 +436,16 @@ func (s *BunkAssignmentsSync) loadMappings() error {
 // two is silently lost. See the write key in processAssignment and the
 // orphan key in deleteOrphans -- all three (plus this preload key, the
 // fourth of the grain) must move together.
-func (s *BunkAssignmentsSync) preloadExistingAssignments(year int) (map[string]*core.Record, error) {
+//
+// The second return is an index of the same rows by "personCMID:bunkCMID" ->
+// the tracking keys ("personCMID:sessionCMID:bunkCMID") they were stored
+// under. It exists for the branch that cannot name a session
+// (resolveSessionOrTrackUnresolved): person and bunk are the two thirds of the
+// grain an unresolved assignment still knows, and this is how it finds the
+// stored rows to keep from the sweep (kindred#2465, the kindred#2394 pattern).
+func (s *BunkAssignmentsSync) preloadExistingAssignments(
+	year int,
+) (existing map[string]*core.Record, byPersonBunk map[string][]string, err error) {
 	filter := fmt.Sprintf("year = %d", year)
 
 	assignmentMappings, err := s.BuildRecordCMIDMappings("bunk_assignments", filter, map[string]string{
@@ -417,10 +454,11 @@ func (s *BunkAssignmentsSync) preloadExistingAssignments(year int) (map[string]*
 		"bunk":    "bunks",
 	})
 	if err != nil {
-		return nil, fmt.Errorf("loading assignment mappings: %w", err)
+		return nil, nil, fmt.Errorf("loading assignment mappings: %w", err)
 	}
 
-	return s.PreloadCompositeRecords(
+	byPersonBunk = make(map[string][]string)
+	existing, err = s.PreloadCompositeRecords(
 		"bunk_assignments", filter, func(record *core.Record) (string, bool) {
 			mapping := assignmentMappings[record.Id]
 			personCMID := mapping["personCMID"]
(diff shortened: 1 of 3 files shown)


adamflagg:
@coderabbitai review

Focus, if it helps: this rewrites the `bunk_assignments` orphan-sweep path. The failure being fixed is that `loadMappings` accumulated onto constructor-allocated maps across runs on a long-lived service instance, so from run 2 every staff assignment resolved "ambiguous", skipped before `TrackProcessedCompositeKey`, and was then deleted by `deleteOrphans` as an untracked orphan — 262 rows for 70 active bunk staff, every hour.

The highest-value things to check are the ones that would look correct and still delete data:

1. Whether the tracked key format matches exactly what `deleteOrphans` looks for. A near-miss format looks fixed and still sweeps.
2. Whether `distinctSessions` and the `resolveViaBunk` accumulator handle the empty and single-candidate cases identically to before.
3. Whether the new tests can distinguish "tracking kept the row" from "OrphanSweepGuard refused the sweep" — the guard refuses when the computed set is empty, which can mask a mutation.
4. Any accumulating field on the service that the reset misses.

adamflagg:
@coderabbitai review

adamflagg:
## Review remediation — all seven findings fixed in-branch (39d5467)

No new issues filed; every finding was M-or-smaller and in code this PR wrote.

---

### 1. The "kept from the sweep" assertion was never reached — confirmed, and fixed

The finding was exactly right. `TestBunkAssignment_UnresolvedAssignmentIsCountedAndKeptFromTheSweep` seeded one row and tracked one key, so removing the tracking loop emptied `ProcessedKeys`, `OrphanSweepGuard.Check` fired its `Computed == 0` arm and **refused the sweep**, and the test died at the `t.Fatalf` above the row-count assertion. Reproduced against the pre-remediation fixture, with the tracking loop mutated out:

```
--- FAIL: TestBunkAssignment_UnresolvedAssignmentIsCountedAndKeptFromTheSweep (0.33s)
    bunk_assignments_grain_test.go:898: protectThenSweepOrphans: orphan sweep refused: refusing to
    sweep 1 existing bunk_assignments rows for year 2026 against a computed set of 0 (0.0% of what
    is on disk, floor 50%): check that the CampMinder bunk-assignment feed returned this season
```

Red, but it proved the guard fires — not that tracking kept the row.

**Fix:** three campers who resolve normally (same reason `TestBunkAssignmentGrain_SecondSyncRunNeitherLosesNorDuplicates` widens its fixture, and the same shape production had — the staff rows sat in a table whose camper rows still resolved and still tracked), plus a departed camper's stored row the sweep must delete.

**Mutation check on the final code** — tracking loop removed from `resolveSessionOrTrackUnresolved`:

```
--- FAIL: TestBunkAssignment_UnresolvedAssignmentIsCountedAndKeptFromTheSweep (0.37s)
    bunk_assignments_grain_test.go:1022: after the sweep: the unresolved staffer's row (cm_id
    740001) is gone -- an assignment the run SAW but could not resolve is not an orphan (kindred#2394)
    bunk_assignments_grain_test.go:1030: after the sweep: bunk_assignments rows = 3, want 4
    (3 campers + the unresolved staffer); survivors = map[9740010:true 9740011:true 9740012:true]
MUTATED_EXIT=1
```

No guard error; `protectThenSweepOrphans` succeeded and the sweep genuinely deleted the row. Restored:

```
ok  	github.com/camp/kindred/pocketbase/sync	0.361s
RESTORED_EXIT=0
```

---

### 2. Neither test had a row the sweep must still delete — fixed

Both now carry one. The two-run test gets one orphan per run (seeded before run 1, another before run 2), so each run's sweep has to delete something; the unresolved test gets one alongside the staffer's. Assertions moved from a bare row count to a cm_id set, so they say *which* rows survived.

Proven load-bearing by short-circuiting `deleteOrphans` to a no-op — both tests catch it, which neither did before:

```
    bunk_assignments_grain_test.go:1026: after the sweep: the departed camper's row (cm_id 740002)
    survived -- if the sweep deletes nothing, the assertion above proves nothing either
    bunk_assignments_grain_test.go:1030: after the sweep: bunk_assignments rows = 5, want 4 ...
    bunk_assignments_grain_test.go:717: run 1: bunk_assignments rows = 5, want 4 (3 campers +
    1 counselor); survivors = map[740003:true 9730005:true 9730006:true 9730007:true 9730008:true]
```

**One thing worth recording from the two-run mutation.** Reverting *only* the `loadMappings` reset now fails on the map assertion alone — the staff row survives, because `distinctSessions` absorbs the duplicate. Reproducing the actual data loss takes reverting **both** halves:

```
    bunk_assignments_grain_test.go:735: after two loadMappings() calls on one instance,
    bunkPlanBunkToSession["8005:6501"] = [5501 5501] (2 candidates), want exactly 1
    bunk_assignments_grain_test.go:740: run 2: staff resolution reported ambiguous for a
    (plan, bunk) pair with ONE bunk_plans row
    bunk_assignments_grain_test.go:756: after run 2: the counselor's row (cm_id 9730005) is gone
    bunk_assignments_grain_test.go:764: after run 2: bunk_assignments rows = 3, want 4;
    survivors = map[9730006:true 9730007:true 9730008:true]
```

That is the layered defense behaving as designed, not a gap — but it means the map-reset half is pinned by the map assertion, and the row assertions pin the pair.

---

### 3. Comments asserted an invariant the schema forbids — corrected, behavior untouched

Verified: the map key is `bunkPlanCMID:bunkCMID` and `bunk_plans` is unique on `(year, bunk, session, cm_id)` (`1500000017_bunk_plans.js:106`), so within one key a repeated session is a bug, not data. `distinctSessions` is a **no-op on a clean run**; its value is that a recurrence of the accumulation degrades into a correct answer instead of a mass delete. Both comments now say that.

Attribution fixed too: `TestLoadMappings_KeepsEveryCandidateSessionForASharedBunk` pins `[A, A, B]` in **`bunkPlanSessionsList`** and pins `bunkPlanBunkToSession` **duplicate-free** (`[A, B]`) — the comments had it backwards, which could have invited someone to "fix" the build site into emitting duplicates it does not emit.

`distinctSessions` and the `map[int]bool` accumulator are unchanged.

I did **not** touch `resolveStaffSession`'s comment: it says the raw list holds one entry per `bunk_plans` row and attributes the duplication to the accumulation, which is accurate as written.

### 4. `runOnce` said "today" for pre-PR behavior — fixed

Now "what Sync()'s loop DID, before this fix", plus a note that hand-rolling the old branch is deliberate: it isolates the map reset, and with the tracking wrapper standing in, the test would otherwise stop pinning the bug it is named for.

### 5. "All three must move together" is four — fixed in both places

`processAssignment` and `deleteOrphans` now name `preloadExistingAssignments`' person:bunk index as the fourth producer of the tracking-key format; the preload's own doc says so from its side.

### 6. The Sync tab is not a surface — comment corrected, no badge added

Confirmed: `SyncTab.tsx` renders a hardcoded badge list (created / updated / skipped / skipped_values / errors) in two places and `unresolved_session` is in neither — zero hits across `frontend/src`. The comment now says it rides along in the JSON but nothing displays it, and that the completion log line is the acceptance surface #2465 names. No frontend change.

### 7. The pre-existing untracked skip — fixed, and it needed one structural change

`if len(bunkPlanSessions) == 0 { continue }` dropped a whole (plan, bunk) group untracked and uncounted. It now warns once for the group and falls through, so each assignment goes through `resolveSessionOrTrackUnresolved` — counted in `unresolved_session`, stored rows kept from the sweep.

Falling through cannot write a row to a wrong session: `loadMappings` gates `bunkPlanSessionsList` and `bunkPlanBunkToSession` on the same `bpCMID > 0 && sessionCMID > 0`, so an empty plan-session list guarantees an empty candidate list and every step of the ladder fails.

**Where I went slightly wider than the finding, deliberately:** `Sync()`'s per-result body moved into `processResultGroup`. `Client` is a concrete `*campminder.Client`, so `Sync()` cannot be driven without an HTTP fake — which is precisely why nothing pinned these branches. The extraction is mechanical (three `continue`s became `return`s) and buys a real test:

```
--- FAIL: TestProcessResultGroup_PlanWithNoSessionsStillTracksItsAssignments (0.00s)   [mutation: continue restored]
    bunk_assignments_grain_test.go:961: Stats.UnresolvedSession = 0, want 1
    bunk_assignments_grain_test.go:966: Stats.Skipped = 0, want 1
    bunk_assignments_grain_test.go:972: ProcessedKeys = map[], want it to contain "9000011:5901:6901|2026"
```

**What I deliberately did not route:** the three shape checks above it (missing `BunkID`, missing `BunkPlanID`, no `Assignments` array). Those name no person and no bunk, so there is no key to keep from the sweep and nothing to count against a grain — they stay untracked, and the code now says why so nobody has to re-derive it.

---

### Verification

| Check | Result |
|---|---|
| `bash scripts/pre-push-verify.sh` | `ALL CHECKS PASSED`, exit 0 (auto-detect: Go only — no other surface changed) |
| `golangci-lint run ./...` from `pocketbase/` | `0 issues.`, exit 0 |
| `go test ./... -count=1` | all packages ok, exit 0 |
| Push | verified by remote ref: `git rev-list --count origin/feature/bunk-assignment-map-reset..HEAD` = `0` |

`golangci-lint` (CI-only here) caught three `misspell` hits in the new prose — `favour`, `defence`, `serialises` — which pre-push did not. Fixed to US spellings.

adamflagg:
@coderabbitai review

PR #2485, opened 2026-08-19
fix(api): the Returning badge requires an enrolled prior weekend, not any household row

adamflagg:
## What was wrong

`fetch_prior_household_cm_ids` (`api/services/lodging_repository.py`) computed the weekend
"Returning" badge's prior-year signal by paging `households` on a bare `year < {year}` — no
enrollment predicate at all. A `households` row exists for a year because a *person* record
existed that year: registration, waitlist, cancellation and inquiry alike, since `households`
carries no status column of its own. That badged a family "Returning" off a year they only ever
cancelled or waitlisted — the exact bug reported from the weekend housing board (kindred-feedback#120).

The file already documented the fix it wasn't applying to itself:
`status_id = 2` is called "the single source of truth for enrollment" at `:580-581`, and the
2020-cancelled-season trap is spelled out at `:615-618`. Verified independently against the
production snapshot: 2020 has 1,264 family/adult attendee rows and **0** enrolled, while 1,981
`households` rows for that year fed the badge anyway.

## The fix

`fetch_prior_household_cm_ids` now reads `attendees` instead of `households`: filtered to
`status_id = 2` (`ACTIVE_ENROLLED_FILTER`) on a family/adult session (`_attendee_weekend_session_filter`,
already used elsewhere in this file for the same weekend/summer split), for years before the
requested one, bridged to the household through `person.household_id` — the CampMinder id, not
the PocketBase relation, because this is a cross-year read exactly like the existing
`fetch_household_family_attendees`.

Deliberately **not** `persons.years_at_camp`: it counts only summer attendance and is blind to
weekends entirely — 225 of 360 genuine 2026 returners have `max(years_at_camp) = 0`, so that
signal would misclassify most real returners as first-time.

## What moved — CORRECTED after independent verification (see PR comment)

The originally-drafted number below undercounted the population this PR actually flips. The
`_attendee_weekend_session_filter()` predicate this fix adds does not only exclude
cancelled/waitlisted-only households — it also excludes a household whose only prior ENROLLED
attendance was a **summer** session, because that predicate requires a family/adult session
type. That second bucket is the "summer-only" half of kindred#2475, which the issue explicitly
marks as **owner-ruled but not cleared to build without asking first**.

Independently re-measured against the read-only production snapshot
(`pocketbase/pb_data/data-prod.db`), replicating the exact predicates this function now uses:

| bucket | households (2026) | this PR's actual effect |
|---|---:|---|
| no prior ENROLLED row of any kind (cancelled/waitlisted-only) | 21 | flips to first-time — the bug as reported |
| prior ENROLLED row, but **summer session only**, never a weekend session | 15 | **also flips to first-time** — this is kindred#2475's "summer-only" half, which the issue says is not ruled for building yet |
| genuine returners (enrolled prior weekend attendee) | 360 | stays Returning |
| **total flip** | **36 of 396** | |

Confirmed by direct SQL against the same predicates the code uses
(`ACTIVE_ENROLLED_FILTER` = `status_id = 2`, `WEEKEND_SESSION_TYPES` = `family`/`adult`): if the
fix used `status_id = 2` alone, with **no** session-type restriction, the flip count is exactly
**21** — matching kindred#2475's own table. Adding `_attendee_weekend_session_filter()` (the
predicate actually shipped in this PR) is what pulls in the other 15.

The two households from the original report have prior status sets of exactly `{waitlisted}` and
`{cancelled, waitlisted}` — both in the 21-bucket, so the reported bug is fixed either way. The
open question is only whether the additional 15-household effect is acceptable to ship now or
should wait, per kindred#2475's "ask before building" note.

2026 is only ~16% placed on the snapshot, so any 2026 count here is a floor, not a final tally —
the settled-year figures in kindred#2475 (34/443 in 2025, 55/502 in 2024, 53/471 in 2023) are the
more reliable measurements of blast radius, and the same 21-vs-36-style split applies to each of
those years (see the issue's own "summer-only" column).

## Scope — this PR's actual reach vs. its intent

kindred#2475 also names a second half: a family whose *only* prior visit was a **summer** session
should read as first-time to Family Camp. That half is **owner-ruled but mechanism-blocked** per
the issue, and the issue says explicitly that separating the two is "not ruled, so ask before
building":

> "we can't accurately calculate the returning as first time to family camp until we fix the
> prior year enrolled but not housing and no kids / cancelled actually fix i think"

This PR does not touch `pocketbase/sync/family_camp_derived.go` or `api/schemas/lodging.py`, and
does not build any new machinery toward the summer-only half. **However**, as verified above, the
query shape it does ship functionally resolves that half's badge outcome too, as a side effect of
adding the weekend-session-type predicate needed to fix the reported bug. That was not the stated
intent and needs an explicit owner call: keep the broader (36-household) fix as shipped, or narrow
`fetch_prior_household_cm_ids` back to `status_id = 2` with no session-type restriction so only the
21-household cancelled/waitlisted-only case flips, leaving the summer-only case as Returning until
kindred#2305 lands.

## Files touched

- `api/services/lodging_repository.py` — `fetch_prior_household_cm_ids`
- `tests/unit/api/services/test_lodging_repository.py` — new coverage for the enrolled-only
  filter, the weekend-session-type restriction, and the cancelled/waitlisted-only case

## Holds for the owner's eyes

This is a **condition 8** hold (a staff-facing number's meaning changes) with zero frontend files
touched. It does not merge itself — see the campaign report and the PR comment for the corrected
number, the population that moves, and the scope decision this now needs.

Refs kindred#2475 (partial — cancelled/waitlisted-only half by intent; summer-only half affected
in practice, pending an owner call on whether that's acceptable).

### DIFF
--- a/api/services/lodging_repository.py
+++ b/api/services/lodging_repository.py
@@ -761,26 +761,58 @@ class LodgingRepository:
 
     @cached_by_year(lodging_cache)
     async def fetch_prior_household_cm_ids(self, year: int) -> set[int]:
-        """CampMinder ids of every household seen in an EARLIER year.
+        """CampMinder ids of every household with an ENROLLED weekend attendee
+        in an EARLIER year -- the returning-family signal (kindred#2475).
 
-        This is the returning-family signal: households rows are per-year, so
-        a cm_id present before `year` means the family has been here before.
+        This used to page `households` on a bare `year < {year}`, with no
+        enrollment predicate at all. A `households` row exists for a year
+        because a *person* record existed that year -- registration,
+        waitlist, cancellation and inquiry alike, since `households` carries
+        no status column of its own. That badged a family Returning off a
+        year they only ever cancelled or waitlisted (21 of 396 badged
+        2026 households on the production snapshot, 8-11% in settled years).
 
-        The expensive one -- 20,256 rows on 2026 prod data for one boolean
-        per party (kindred#1966). #1966/#1976 already bought back the
-        round-trip cost (batch=PAGE_SIZE, 21 requests instead of 203) and
-        explicitly rejected narrowing the filter to the current year's
-        household ids: a 250-term OR clause returns HTTP 400 (undocumented,
-        fails closed), so there is no query-shape lever left to pull. Caching
-        is what is left, and it is safe here for the same reason as
-        fetch_households: year-scoped, sync-written only. See
-        api/services/lodging_cache.py.
+        `attendees` is where enrollment actually lives, so this reads THAT
+        table instead: `status_id = 2` (`ACTIVE_ENROLLED_FILTER`) is the
+        single source of truth for enrollment everywhere else in this file,
+        and `_attendee_weekend_session_filter` keeps summer sessions
+        (main/embedded/ag/quest/...) from counting as a prior weekend visit.
+
+        Bridged through `person.household_id` (the CampMinder id) rather than
+        `person.household` (the PocketBase relation): this is a cross-YEAR
+        read, exactly like `fetch_household_family_attendees`, and
+        `household` is a relation into the year-scoped `households` table --
+        it can only ever match one season. `household_id` is the identity
+        thread across seasons (CLAUDE.md section 1).
+
+        Deliberately NOT `persons.years_at_camp`: it counts only SUMMER
+        attendance and is blind to weekends (225 of 360 genuine 2026
+        returners have `max(years_at_camp) = 0`) -- it would misclassify most
+        real returners as first-time.
+
+        `attendees` is the pricier table to page here (each row expands
+        `person`), but this is still the once-per-year cross-season read
+        #1966/#1976 already bought the round-trip cost down for
+        (batch=PAGE_SIZE); caching is what is left, and it is safe here for
+        the same reason as fetch_households: year-scoped, sync-written only.
+        See api/services/lodging_cache.py.
         """
         rows = await self._page(
-            HOUSEHOLDS,
-            query_params={"filter": f"year < {year}", "fields": "cm_id", "sort": STABLE_SORT},
+            ATTENDEES,
+            query_params={
+                "filter": (f"year < {year} && ({_attendee_weekend_session_filter()}) && {ACTIVE_ENROLLED_FILTER}"),
+                "expand": "person",
+                "sort": STABLE_SORT,
+            },
         )
-        return {int(getattr(row, "cm_id", 0)) for row in rows if getattr(row, "cm_id", 0)}
+        ids: set[int] = set()
+        for row in rows:
+            expand = getattr(row, "expand", None) or {}
+            person = expand.get("person") if isinstance(expand, dict) else None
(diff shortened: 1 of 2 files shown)


adamflagg:
## Independent scan (triage-attack Part B) — verified, held, PR body corrected

**Code review at floor 60:** no correctness defects survive. `_page`/`expand` handling matches
the established pattern used elsewhere in this file (lines 931, 959, 1078 —
`getattr(row, "expand", None) or {}` then `.get(...)`, checked `isinstance(expand, dict)`
before use). `person.household_id` (not `.household`) is the right bridge for a cross-year read,
consistent with `fetch_household_family_attendees`'s own reasoning. Tests are TDD, mutation-checked
(reverting `ACTIVE_ENROLLED_FILTER` turns the enrolled-only test red), and green: 145/145 in this
file, CI Summary pass.

**Premise verification — found a real discrepancy, corrected in the PR body just now.**

I re-derived the "21 of 396" claim independently against the production snapshot
(`pocketbase/pb_data/data-prod.db`), replicating the exact predicates the shipped code uses:

```sql
-- roster household population: enrolled 2026 attendees on family/adult sessions
-- old_prior: bare `households` row, year < 2026 (what the code used to do)
-- new_prior: attendees, status_id = 2, session_type IN ('family','adult'), year < 2026 (what it does now)
```

Result: **36 households flip from Returning to first-time, not 21.** The extra 15 are households
whose only prior ENROLLED attendance was a **summer** session — never a weekend one. Dropping the
`_attendee_weekend_session_filter()` predicate and keeping only `status_id = 2` (any session type)
reproduces exactly 21 — matching kindred#2475's own table precisely. So the weekend-session-type
restriction this PR adds (needed to correctly define "prior weekend attendee") also resolves
kindred#2475's *other* half — the summer-only case the issue explicitly flags as "owner-ruled but
not cleared to build without asking first," pending kindred#2305.

The PR body's "Scope" section claimed this half was "not touched." That's not accurate — the code
doesn't touch `family_camp_derived.go`/`lodging.py`, but the query shape it ships functionally
resolves that half's badge outcome as an unavoidable side effect of the fix. I've corrected the PR
body's "What moved" and "Scope" sections to state the true 36/396 split and name the decision this
now needs, rather than leave an inaccurate scope claim as the permanent squash-commit message.

**I did not restrict the query back to the claimed 21-only scope.** That would itself be making
the "not ruled, ask before building" call unilaterally. I left the code as shipped and surfaced the
choice instead.

### What the owner needs to decide before merging

1. **Ship the broader fix as-is** (36 households flip: 21 clear-cut bug fix + 15 that also settle
   kindred#2475's summer-only half, ahead of kindred#2305 landing), or
2. **Narrow `fetch_prior_household_cm_ids`** to `status_id = 2` with no session-type restriction,
   so only the 21 cancelled/waitlisted-only households flip and the 15 summer-only households keep
   reading Returning until kindred#2305 lands and the summer-only half can be built and reviewed on
   its own.

Both are defensible; neither is mine to pick. CI is green, tests are solid, no other findings.
Holding per condition 8 either way — flagging this so the number the owner reviews is the real one.

PR #2486, opened 2026-08-19
fix(sync): dedupe family_camp_adults rows that coalesce to the same adult

### LINKED ISSUE #2483: fix(sync): dedupe family_camp_adults rows that coalesce to the same adult, at write time
## ★ OWNER RULING 2026-08-19 — fix at WRITE time, in the Go sync

**Dedupe in `pocketbase/sync/family_camp_derived.go` on ingest, not in the API read path.**

Consequences to build against:

1. **It must reuse #2275's ruled merge policy** — first-wins with a **CampMinder-id tiebreak**, via the
   existing `mergeFirstNonEmpty` helper. Do **not** write a second merge implementation; six of the
   seven adult attributes already route through that one function, and two policies on one table will
   diverge.
2. **Use the corrected key** (see the correction section below): same casefolded coalesced name **AND
   non-conflicting DOB**. The name-only key merges 27 groups of different people.
3. **Handle n > 2** — one 2024 group holds three rows.
4. **Field survival:** when a pair collapses, non-blank fields from the losing row (`email`,
   `date_of_birth`) must attach to the survivor rather than being dropped. That *is* a merge, which is
   why it must be the same helper.
5. ⚠️ **This trades away evidence, knowingly.** The raw adult slots record what a family typed into the
   CampMinder form, and #2275's finding is that those slots are evidence. Collapsing at write time
   destroys the duplicate. The owner has accepted that trade in exchange for fixing every consumer at
   once. If the raw form submission ever needs to be recoverable, that is a separate capture concern —
   do not solve it by leaving the duplicate in place.

---

## ⛔ CORRECTION 2026-08-19 — the proposed dedupe key would MERGE 27 GROUPS OF DIFFERENT PEOPLE

**The reproduction is exact and the defect is real — I reproduced both 2026 pairs against a fresh
production snapshot. But the suggested key, "coalesced, casefolded name" alone, is unsafe and must
not be implemented as written.**

### The falsification

Run over **all years** of `family_camp_adults`, grouping by household + casefolded coalesced name:

| | groups |
|---|---:|
| same household, same casefolded name, >1 row | **42** |
| …of which the rows carry **CONFLICTING dates of birth** | **27** |

Twenty-seven of the forty-two are two *different humans who share a name*. Sample DOB pairs from
distinct groups: `05/03/2013` vs `8/7/1960` (a child and an adult — almost certainly a junior/senior
pair), `01/20/1976` vs `10/06/1971`, `1/22/1983` vs `7/11/1983`.

**A name-only key merges all 27.** That is the same class of error as the "delete the blank rows"
proposal that would have erased 196 real adults — a per-row rule applied to a cross-row question.

### The key that works

> **Same casefolded coalesced name AND non-conflicting DOB** — where non-conflicting means the DOBs are
> equal, or **at most one of the rows carries a DOB at all**.

Measured over all years:

| | groups |
|---:|---:|
| **MERGE** | **15** (2026: **2** · 2025: 1 · 2024: 4 · 2023: 1 · 2019: 4 · 2018: 1 · 2017: 2) |
| **REFUSE** | **27** |

**2026 merges = exactly 2 — the two pairs this issue reports.** So the safe key loses nothing that was
actually reported, and refuses every collision.

⚠️ **Why a plain name+DOB equality key is NOT enough:** it catches only **1 of the 2** 2026 pairs,
because one pair has a blank DOB on one row. The "at most one populated" clause is load-bearing, not a
nicety.

⚠️ **One 2024 group holds THREE rows**, so the merge must handle n>2 rather than assuming a pair.

### Two things this issue does not settle, and they are the real design questions

**1. Dedupe at WRITE time (Go sync) or READ time (Python API)?** The body proposes changing
`is_attending_adult_name` and the bed-count sum — both **read-side**, which means every future consumer
must remember to dedupe. Fixing in `family_camp_derived.go` fixes it once. **But** the raw adult slots
are the record of what a family typed into the CampMinder form, and #2275's whole finding is that
those slots are evidence — collapsing them at write time destroys it. This is an owner call.

**2. It must share #2275's merge policy, not invent a second one.** #2275 is ruled: **first-wins with a
CampMinder-id tiebreak**, and six of seven attributes already route through the single helper
`mergeFirstNonEmpty`. This issue's *"merge the non-blank fields upward onto a single logical adult"* is
the same operation on the same table at a different grain (#2275 merges across source *copies* of one
slot; this merges across *slots*). **If the two ship separate merge implementations they will
diverge.** Reuse the helper.

---

## Bug

`family_camp_adults` can hold the same human twice in one household — two valid,
non-blank, non-placeholder adult slots that coalesce to the identical display
name — and nothing dedupes it. The existing guards only cover blank names
(`ADULT_NAME_PLACEHOLDERS` in `api/constants/lodging.py`) and malformed/absent
emails. Neither guard fires here because both rows pass validity individually;
the defect only shows up when you compare **across** the two rows for the same
household.

## Reproduction (representative example, names fictionalized)

A 2026 household has two `family_camp_adults` rows for `year=2026`:

| adult_number | name | first_name | last_name | email | date_of_birth |
|---|---|---|---|---|---|
| 1 | `Emma Johnson` | *(blank)* | *(blank)* | *(blank)* | *(blank)* |
| 2 | *(blank)* | `Emma Johnson` | *(blank)* | `test@example.com` | `1980-06-15` |

`_adult_display_name` (`api/services/lodging_roster_service.py`, lines 220-237)
coalesces `name or f"{first_name} {last_name}".strip()`:

```python
return _s(adult, "name") or f"{_s(adult, 'first_name')} {_s(adult, 'last_name')}".strip()
```

Both rows resolve to the identical string `"Emma Johnson"` — one person, counted
as two adults.

## Season-wide scope (2026)

A `GROUP BY household, casefold(coalesced_name) HAVING COUNT(*) > 1` query
against `family_camp_adults` for `year=2026` finds **exactly 2 households**:

1. The household above — exact-case duplicate.
2. A second household with a case-only duplicate: adult_number=1 has
   `first_name='Liam Garcia'` (email `test@example.com`, dob `1985-03-22`),
   adult_number=2 has `first_name='liam garcia'` (no email, same dob
   `1985-03-22`). A byte-exact string comparison misses this pair; only a
   casefolded comparison catches both, which is why any fix should casefold
   before comparing (matching the pattern `is_attending_adult_name` already
   uses for its placeholder check).

## Impact

- `is_attending_adult_name` (`api/constants/lodging.py`, lines 39-54) returns
  `True` for both rows in each pair — it only rejects blank/placeholder
  tokens, so a valid duplicate sails through:

  ```python
  def is_attending_adult_name(name: str | None) -> bool:
      if name is None:
          return False
      token = name.strip()
      return bool(token) and token.casefold() not in ADULT_NAME_PLACEHOLDERS
  ```

- The bed-count line in `lodging_roster_service.py` (~line 2267) sums over
  exactly this predicate:

  ```python
  beds = sum(1 for adult in adults if is_attending_adult_name(_adult_display_name(adult)))
  ```

  For both affected households this overcounts `party_size` / beds-needed by
  one adult.
- Any person-grained rendering of the roster (e.g. `PartyAdult` built in the
  same file) shows the same human as two separate adults.

## Suggested fix direction

Dedupe on the **coalesced, casefolded** name within a household, not on
emptiness alone — both existing guards (`ADULT_NAME_PLACEHOLDERS`, blank
check) are per-row validity checks and structurally cannot catch two valid
rows naming the same person. When a duplicate pair is found, merge the
non-blank fields upward onto a single logical adult (e.g. one row's
`email`/`date_of_birth` should attach to the one adult that survives) rather
than dropping data from whichever row is discarded.

---

*Reported via kindred-feedback#127 — the private original, which holds the
real household identifiers used to verify this against production.*

adamflagg:
## Bug

`family_camp_adults` could hold the same human twice in one household: two non-blank adult
slots that coalesce (`name`, or `first_name`+`last_name`) to the identical casefolded display
name. Nothing deduped it, so `is_attending_adult_name` counted both, overcounting
`party_size`/beds-needed downstream and double-rendering the person in any person-grained
roster view.

## The corrected key

A name-only key is unsafe. Measured against the production snapshot, grouping
`family_camp_adults` by household + casefolded coalesced name across **all years** finds
**42 groups**, but **27 of them are two different humans who happen to share a name** (sample
DOB pairs: `05/03/2013` vs `8/7/1960` — a child and an adult).

**The key that works:** same casefolded coalesced name **and non-conflicting DOB**, where
non-conflicting means the group's non-empty dates of birth are all equal, or **at most one
slot in the group carries a DOB at all**. A plain equality key is not enough — it misses one
of the two reported 2026 pairs, which has a blank DOB on one side.

Measured (SQL replay against `family_camp_adults`, all years): **MERGE 15 / REFUSE 27**.
**2026 merges = exactly 2** — the two pairs this issue originally reported, and the exact
staff-facing movement this PR produces: 2 households' adult headcount (and therefore
beds-needed) drops by one.

## The fix

Dedupe happens in `pocketbase/sync/family_camp_derived.go`'s `processAdults`, at write time,
before the existing nameless-row admission filter:

- `dedupeAdultSlots` groups a household's adult slots by casefolded coalesced name and refuses
  the whole group if two or more DISTINCT non-empty DOBs are present (`adultGroupDOBConflicts`).
- Handles groups larger than a pair — one 2024 production group holds three rows for the same
  person.
- The survivor is the lowest `adult_number` in the group (deterministic; Go map iteration order
  is not).
- `mergeAdultSlot` folds the losing slot's non-blank fields onto the survivor using the SAME
  per-field policy `processAdults` already applies for sibling-vs-sibling merges
  (`mergeFirstNonEmpty` for name/first/last/pronouns/gender/DOB/relationship, `preferEmail` for
  email) — reusing kindred#2275's merge helper rather than writing a second policy on the same
  table.
- A dropped slot's key is simply not written for this run, so the existing orphan sweep in
  `deleteOrphanedAdults` deletes its stored row on the next sync — the same path kindred#2335
  already uses for rows the computed set no longer accounts for.

This knowingly destroys the duplicate raw slot as evidence (the owner ruling on #2483 accepts
that trade explicitly, in exchange for fixing every consumer of `family_camp_adults` at once).

## Scope guard — what this PR deliberately does NOT do

- No change to `api/services/lodging_roster_service.py` or `api/schemas/lodging.py` — those
  belong to other issues.
- No PocketBase migration — this is a data-shape fix inside the existing schema.
- No re-grain of `family_camp_adults` — the (household, year, adult_number) grain from
  kindred#2275/#2335 is unchanged; this only decides which *rows* get written to it.

Closes #2483.

### DIFF
--- a/pocketbase/sync/family_camp_derived.go
+++ b/pocketbase/sync/family_camp_derived.go
@@ -1033,6 +1033,13 @@ func (s *FamilyCampDerivedSync) processAdults(
 		}
 	}
 
+	// Dedupe rows that coalesce to the same adult (kindred#2483) BEFORE the
+	// admission filter below, so a merged slot's survivor is judged for
+	// admission exactly like any other adult.
+	for _, adults := range adultMap {
+		dedupeAdultSlots(adults)
+	}
+
 	// Convert map to slice, only include adults with a real name somewhere.
 	// kindred#1946: the email/gender arms this filter used to carry let 194
 	// wholly nameless rows into production -- an adult with only a gender
@@ -1049,6 +1056,128 @@ func (s *FamilyCampDerivedSync) processAdults(
 	}
 
 	return result
+}
+
+// coalescedAdultName reproduces the API's `_adult_display_name` coalesce
+// (`api/services/lodging_roster_service.py`) so the dedupe key groups the
+// same way the read path does: `name` (the household column of record) if
+// set, otherwise the split first/last columns joined and trimmed.
+func coalescedAdultName(a *adultData) string {
+	if a.name != "" {
+		return a.name
+	}
+	return strings.TrimSpace(strings.TrimSpace(a.firstName) + " " + strings.TrimSpace(a.lastName))
+}
+
+// dedupeAdultSlots collapses adult slots within one household that coalesce
+// to the same casefolded display name with a non-conflicting date of birth
+// (kindred#2483 -- see the correction on the issue for why name alone is
+// unsafe: it merges 27 groups of genuinely different people who share a
+// name, sample DOB pairs include a child and an adult).
+//
+// "Non-conflicting" means the group's non-empty date_of_birth values are all
+// equal, or at most one slot in the group carries a DOB at all. A plain
+// equality key is not enough -- one of the two reported 2026 pairs has a
+// blank DOB on one side, so it looks like a single non-empty DOB, not an
+// equal pair. Any group with two or more DISTINCT non-empty DOBs is refused
+// entirely, matching the issue's measurement (MERGE 15 / REFUSE 27 over all
+// years; 2026 merges = exactly 2).
+//
+// Deleting a losing slot's map entry removes it from processAdults' result,
+// which is also what marks its stored row (if any) unprocessed for this run
+// -- the existing orphan sweep in deleteOrphanedAdults then deletes it, the
+// same path kindred#2335 already uses for rows the computed set no longer
+// accounts for. That is the intended outcome here: the duplicate slot is
+// destroyed, not archived (owner ruling 2026-08-19 -- this trades away the
+// raw-slot evidence deliberately).
+//
+// The survivor is the LOWEST adult_number in the group. adult_number is the
+// one deterministic ordering available here (Go map iteration is not), and
+// it mirrors "first-wins" -- Adult 1 is filled first on the form.
+func dedupeAdultSlots(adults map[int]*adultData) {
+	groups := make(map[string][]*adultData, len(adults))
+	for _, adult := range adults {
+		key := strings.ToLower(coalescedAdultName(adult))
+		if key == "" {
+			continue // nameless slots are not a duplicate of anything; left to the admission filter
+		}
+		groups[key] = append(groups[key], adult)
+	}
+
+	for _, group := range groups {
+		if len(group) < 2 {
+			continue
+		}
+		if adultGroupDOBConflicts(group) {
+			continue // 27-groups-of-different-people case: leave every slot as its own adult
+		}
+
+		slices.SortFunc(group, func(a, b *adultData) int {
+			return a.adultNumber - b.adultNumber
+		})
+		survivor := group[0]
+		for _, loser := range group[1:] {
+			mergeAdultSlot(survivor, loser)
+			delete(adults, loser.adultNumber)
+		}
+	}
+}
+
+// adultGroupDOBConflicts reports whether a same-name group holds two or more
+// DISTINCT non-empty date_of_birth values -- the signal that the group is
+// actually different people who share a name, not one person duplicated
+// across slots. date_of_birth is normalised (normalizeDateOfBirth) before it
+// ever reaches processAdults' map, so a straight string comparison is a
+// straight date comparison.
+func adultGroupDOBConflicts(group []*adultData) bool {
+	seen := ""
+	for _, adult := range group {
+		if adult.dateOfBirth == "" {
+			continue
+		}
+		if seen == "" {
+			seen = adult.dateOfBirth
+			continue
+		}
+		if adult.dateOfBirth != seen {
+			return true
+		}
+	}
+	return false
+}
+
+// mergeAdultSlot folds loser's non-blank attributes onto survivor using the
+// SAME per-field merge policy processAdults already applies when two
+// siblings answer for one slot (kindred#2483 ruling: reuse mergeFirstNonEmpty
+// rather than invent a second merge implementation). email keeps its
+// existing preferEmail validity tie-break rather than plain first-wins, for
+// the same reason the sibling-merge loop above does.
+func mergeAdultSlot(survivor, loser *adultData) {
+	survivor.name = survivor.mergeFirstNonEmpty("name", survivor.name, loser.name)
+	survivor.firstName = survivor.mergeFirstNonEmpty("first_name", survivor.firstName, loser.firstName)
+	survivor.lastName = survivor.mergeFirstNonEmpty("last_name", survivor.lastName, loser.lastName)
+	survivor.pronouns = survivor.mergeFirstNonEmpty("pronouns", survivor.pronouns, loser.pronouns)
+	survivor.gender = survivor.mergeFirstNonEmpty("gender", survivor.gender, loser.gender)
+	survivor.dateOfBirth = survivor.mergeFirstNonEmpty("date_of_birth", survivor.dateOfBirth, loser.dateOfBirth)
+	survivor.relationship = survivor.mergeFirstNonEmpty(
+		"relationship_to_camper", survivor.relationship, loser.relationship)
+
+	if loser.email != "" {
+		if preferEmail(survivor.email, loser.email) {
+			if !sameAnswer(survivor.email, loser.email) {
+				survivor.noteConflict("email", survivor.email)
+			}
+			survivor.email = loser.email
+		} else if !sameAnswer(survivor.email, loser.email) {
+			survivor.noteConflict("email", loser.email)
+		}
+	}
+
+	for column, others := range loser.conflicts {
+		for _, other := range others {
+			survivor.noteConflict(column, other)
+		}
+	}
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


adamflagg:
## Verification summary

**CI**: CI Summary green (all checks pass, including 3-shard Go Tests with -race).

**Floor-60 scan (code-review skill, medium effort)**: one finding — dedupeAdultSlots merges a same-name group when BOTH slots have a blank date_of_birth, which could in theory collapse two different people who share a name and never answered DOB. Investigated and closed as not-a-bug: the issue's own CORRECTION text defines "non-conflicting DOB" as equal, **or at most one slot carries a DOB at all** — zero populated satisfies "at most one" by construction. This boundary is not hypothetical: it is already inside the issue's own ratified MERGE=15 measurement. Household `<household-id-redacted>` (2024, "emma johnson") has adult_number 1 with `name="Emma Johnson"`/`first_name="Olivia Martinez"` and adult_number 2 with `first_name="Emma Johnson"`, both DOBs blank — clearly one person's row split across two mislabeled-field slots. Re-derived the full measurement against `pb_data/data-prod.db`: MERGE 15 / REFUSE 27 across all years, 2026 merges = exactly 2, matching the issue's numbers exactly. No code change made for this finding.

**Staff-facing number movement (merge condition 8)**: exactly 2 households' `family_camp_adults` row count for 2026 drops from 2 to 1 (both households' adult headcount / beds-needed decreases by 1):
- `0taz56t3vuwxkva` ("morgan shapiro") — same name, same DOB `1987-04-15` on both rows.
- `bz02mn6khl87x5f` ("jaime jenett") — same name, one row's DOB blank, the other `1976-01-05`.

This is fixing a double-count against a definition nobody disputes (the same-year measurement standard applied elsewhere in the codebase), not deciding a new population — no other year or household moves.

Merging per the triage-attack Part B merge criteria.


---

_Edited to replace illustrative real data with the fictional test set, per the repository's no-real-personal-information rule. The measurement and the conclusion are unchanged._

adamflagg:
Independent triage-attack scan (floor 60, high effort) — PR was already merged when this scan started.

**Verified independently against `pocketbase/pb_data/data-prod.db` (main repo, not the worktree copy which is 0 bytes) by replaying the same grouping the code uses (household+year, casefolded coalesced name):**
- 42 same-name groups, MERGE 15 / REFUSE 27 — exact match.
- 2026 merges = 2 distinct households — exact match, this is the staff-facing movement (adult headcount/bed-count −1 for those 2 households only).
- Both-blank-DOB edge case: exactly 1 real household matches the shape (household `p2dcbxpx0aidegb`, 2024, "ariel rubin") and is inside the measured 15 — confirms the pinning test isn't guarding a false positive.
- 3-row group: exactly 1 exists (household `uc470qdswijo95w`, 2024, "roya yasharpour") — confirms `TestProcessAdultsDedupesHandlesThreeRowGroup` reflects real production shape.

**Verified in code:**
- `dedupeAdultSlots` groups within one household's `adultMap` entry (called per-household, not globally) — no cross-household leakage.
- Runs after both the household-value and person-value loops finish populating `adultMap`, so `dedupeAdultSlots` sees fully-merged slots, not partial ones.
- `date_of_birth` is normalised (`normalizeDateOfBirth`) before it's ever stored on `adultData`, so `adultGroupDOBConflicts`'s raw string comparison is a correct date comparison.
- Deleted losers are removed from `adultMap` before the write set is built, so their `(household, year, adult_number)` key is simply not written this run and the existing orphan sweep (`deleteOrphanedAdults`, kindred#2335 path) deletes the stored row — no new deletion path introduced.
- `mergeAdultSlot` reuses `mergeFirstNonEmpty`/`preferEmail` byte-for-byte rather than a new policy; loser's `conflicts` map is folded into survivor's via the existing `noteConflict` (dedupes by design).

**Mutation-checked independently:** disabled `adultGroupDOBConflicts`'s effect (`if false && adultGroupDOBConflicts(group)`) and reran `TestProcessAdultsRefusesConflictingDOBSameName` — it went red as expected (`expected 2 adults, got 1`), confirming the guard is load-bearing. Reverted, `git status` clean.

**Scope guard confirmed:** diff is exactly `pocketbase/sync/family_camp_derived.go` + its `_test.go`, no migration, no `api/services/lodging_roster_service.py` or `api/schemas/lodging.py` touched, no frontend files (`gh pr diff --name-only` — 2 files total).

**CI:** all 24 checks pass, CI Summary confirmed actually ran (not absent). `go build ./...` and `go vet ./sync/...` clean. `closingIssuesReferences` = [2483], and #2483 is closed by this merge.

**Findings:** none. Scan result: clean (0 discarded, 0 found) — this is a genuinely clean, well-measured fix.

PR #2487, opened 2026-08-19
fix(api): clear-assignments refuses a scenario-less call instead of deleting production

### LINKED ISSUE #2473: bug(api): clear-assignments deletes from PRODUCTION bunk_assignments when no scenario is given
## ★ RULED 2026-08-19 — refuse it. This is no longer an open question.

> **Owner:** *"technically this is irrelevant because it would flow down again from the bunk
> assignment sync, but in reality, we should in fact fix it properly. it's not allowed even in admin —
> it's faster to just resync. staff aren't going to ask for it."*

So both readings offered below are settled at once:

- **It is not a staff admin operation.** The capability is not wanted, in admin or anywhere else.
  Resyncing is faster than clearing, so there is no workflow this endpoint serves.
- **It must refuse without a scenario**, exactly as `apply` now does after #2467 / PR #2471.

**Do not** merely add a comment at the call site saying the behaviour is intentional — that branch of
the question is closed.

### ⚠️ One correction to this issue's own severity argument, which the ruling exposed

The body below says recovery after an accidental clear is *"a container restart plus a full sync, not
the next hourly run"*, citing #2465. **That is true only while #2465 is unfixed.** The reason a full
sync could not recreate the rows is precisely the accumulation bug — the long-lived
`BunkAssignmentsSync` instance kept growing its candidate maps, so from run 2 onward every staff
assignment resolved as ambiguous and was skipped. **Once PR #2472 merges, a resync does recreate them**,
which is what the owner's "it would flow down again" refers to.

⇒ **The data-loss argument for this issue is weaker than the body claims**, and it should not be the
reason this gets fixed. The reason is the product rule: production is read-only for the solver, and an
endpoint that deletes from it contradicts a ruling the codebase now enforces one function away.

### Implementation, now decision-free

1. Refuse a scenario-less `clear_session_assignments` with a 4xx, reusing **PR #2471's** refusal shape
   and message style rather than inventing a second one.
2. Extend #2471's **class-level** guard so it pins *the solver router never writes **or deletes**
   `bunk_assignments`* — one rule, not two endpoint-specific tests. A test that only pins this endpoint
   would let the next production path through.
3. **Sequence after PR #2471 merges**, and build on `main` rather than stacking — a stacked PR gets no
   CI of its own and its closing reference does not register until the parent lands.

---

Found while implementing #2467's refusal (PR #2471). Same file, same product rule, **opposite
direction** — and unlike #2467's write, this path **works**.

## The question this issue is really asking

The owner's ruling of 2026-08-18, recorded on #2467, is:

> *"you cannot solve against production, only scenarios — prod is designed as read only for solver,
> and drag/drop."*

#2467 applies that to the solver's **write**. This issue asks whether it also applies to a **delete**:

**Is clearing a session's assignments a solver operation — in which case it must refuse without a
scenario, exactly as apply now does — or is it a staff admin operation that is legitimately allowed to
touch the production table?**

Both readings are defensible and they produce materially different diffs, so this is not filed as a
defect with an obvious fix.

## What the code does

`clear_session_assignments` in `api/routers/solver.py` branches on the scenario:

```python
if request.scenario:
    collection_name = "bunk_assignments_draft"        # :905
    base_filter = f'scenario = "{request.scenario}"'  # :906
else:
    collection_name = "bunk_assignments"              # :908
...
await asyncio.to_thread(pb.collection(collection_name).delete, assignment.id)   # :922
```

With no scenario it resolves to the **production** `bunk_assignments` collection and deletes every
matching row. There is no guard, no confirmation flag, and no refusal.

## Why it is not the same severity as #2467

#2467's production branch builds a payload with no valid columns, so it fails validation and writes
nothing — it is only harmless **by accident**. This path constructs no payload; it deletes by record id,
so nothing stops it. Whatever the ruling, this is the one that can actually change production data.

## Scale, for context

`bunk_assignments` holds **14,407** rows on the production snapshot, **1,727** of them for 2026. The
endpoint is session-scoped, so a single call deletes that session's share rather than all of it — but the
rows are a CampMinder mirror, and #2465 establishes that **the mirror does not self-heal within a
process**: services are constructed once at boot and a later full sync does not recreate swept rows.
So the recovery story after an accidental clear is a container restart plus a full sync, not the next
hourly run.

## If the answer is "it must refuse"

The shape is already established by PR #2471 and should be copied rather than reinvented, including the
precedent it cites — `PlacementWriteBase.scenario`: *"REQUIRED and non-empty. A blank scenario is a 422,
never a write to the live plan."* The guard test should likewise pin the **class** (the solver router
never deletes from `bunk_assignments`) rather than this one endpoint.

## If the answer is "it is a legitimate admin operation"

Then it wants the opposite treatment: say so at the call site, so the next reader does not fix it into a
refusal on the strength of #2467's ruling. It would also be worth asking whether it belongs in
`api/routers/solver.py` at all, since being in the solver router is what makes it read as solver-owned.

## Not in scope

- `POST /api/solver/run-sweep` still creates scenario-less runs with a `session_cm_id`. Deliberate — a
  sweep is a benchmark that never applies, and its children are covered by #2467's apply-time refusal.
  Refusing sweeps would remove the session-based sweep feature, which is a larger product question.
- The `bunk_assignments` orphan sweep (#2465) is a different mechanism and is fixed separately.

- **Files:** `api/routers/solver.py` (+ tests)
- **Predecessor:** none, but it should be decided *after* #2471 merges so it can reuse that refusal's
  shape rather than diverge from it.

adamflagg:
Production bunk assignments are read-only for the solver. `clear_session_assignments` took a scenario-less request down an `else` branch that deleted straight from `bunk_assignments` by record id — unlike PR #2471's write-side branch, this one had nothing to fail validation on, so it actually worked and could delete production rows.

The owner's ruling settles both readings the issue left open: it is not a staff admin operation ("not allowed even in admin — it's faster to just resync"), and it must refuse without a scenario, exactly as apply now does after #2471.

Reuses #2471's `_require_scenario` helper and its 422 detail message verbatim, rather than inventing a second refusal shape. The production branch is deleted outright, same treatment as apply's.

The issue's own severity argument (recovery needs a container restart, per #2465) no longer holds — PR #2472 fixed the underlying accumulation bug, so a resync now recreates swept rows. The reason to fix this is the product rule that production is read-only for the solver, not data loss, and this PR does not restate the weaker argument.

## Tests

In the same two files #2471 already used for this rule:

- `test_solver_never_writes_production.py`'s class-level guard is extended to also pin deletes, not just writes — one rule covering both `apply_solver_results` and `clear_session_assignments`, so the next production-touching path added to either function is still caught. A behavioral test drives clear-assignments through a `RecordingPB` and asserts no delete reaches `bunk_assignments`; the AST tripwire now walks both functions for the production collection name.
- New refusal tests for clear-assignments (missing/blank/whitespace scenario, a scenario-having call still deletes the draft), and one that asserts the clear refusal's message is byte-identical to run-creation's.
- `test_draft_writes_invalidate_cache.py`'s `test_prod_clear_invalidates_session_slot` asserted the old production behaviour (200 + session-slot invalidation). It now asserts the refusal and that nothing is invalidated — a spec change, not a test bent to fit the code, mirroring what #2471 did to this file's apply-side counterpart.

Both guards were mutation-checked: reverting the collection/filter back to `bunk_assignments` with no scenario filter fails the class guard's behavioral test and its AST tripwire.

## Deliberately not touched

- `POST /api/solver/run-sweep` still creates scenario-less runs with a `session_cm_id`. A sweep is a benchmark that never applies; refusing it would remove the session-based sweep feature, a bigger product call than this issue settles.
- The hourly orphan sweep (`pocketbase/sync/bunk_assignments.go`, #2465/#2472) is untouched — a different mechanism, already fixed separately.

## Verification

`bash scripts/pre-push-verify.sh` — ALL CHECKS PASSED (ruff format, ruff check, mypy, 6419 passed / 14 skipped). Pre-push (pushed run): 6580 passed / 41 skipped.

Closes #2473.

### DIFF
--- a/api/routers/solver.py
+++ b/api/routers/solver.py
@@ -907,7 +907,11 @@ async def clear_session_assignments(
     request: ClearAssignmentsRequest,
     user: AuthUser = Depends(require_permission(Permission.BUNKING_MANAGE)),
 ) -> dict[str, Any]:
-    """Clear all assignments for a session and its related sessions."""
+    """Clear all assignments for a session and its related sessions' scenario draft."""
+    # Refuse before touching PocketBase: clearing is a delete, and production
+    # bunk assignments are read-only for the solver, same as writes (kindred#2473).
+    scenario = _require_scenario(request.scenario)
+
     try:
         # Build session context from request (validates session exists for year)
         ctx = await build_session_context(session_cm_id, request.year, pb)
@@ -916,18 +920,11 @@ async def clear_session_assignments(
         deletions_by_session: defaultdict[int, int] = defaultdict(int)
         total_deleted = 0
 
-        if request.scenario:
-            collection_name = "bunk_assignments_draft"
-            base_filter = f'scenario = "{request.scenario}"'
-        else:
-            collection_name = "bunk_assignments"
-            base_filter = ""
+        collection_name = BUNK_ASSIGNMENTS_DRAFT
+        base_filter = f'scenario = "{scenario}"'
 
         for sid in ctx.related_session_ids:
-            if base_filter:
-                filter_str = f"{base_filter} && session.cm_id = {sid} && year = {ctx.year}"
-            else:
-                filter_str = f"session.cm_id = {sid} && year = {ctx.year}"
+            filter_str = f"{base_filter} && session.cm_id = {sid} && year = {ctx.year}"
 
             assignments = await asyncio.to_thread(
                 pb.collection(collection_name).get_full_list, query_params={"filter": filter_str}
@@ -960,15 +957,12 @@ async def clear_session_assignments(
         # assignments materially changes the social graph, so any cached
         # rendering would be stale.
         for sid in ctx.related_session_ids:
-            if request.scenario:
-                graph_cache.invalidate_scenario(int(sid), int(ctx.year), request.scenario)
-            else:
-                graph_cache.invalidate_session(int(sid), int(ctx.year))
+            graph_cache.invalidate_scenario(int(sid), int(ctx.year), scenario)
 
         return {
             "message": f"Cleared {total_deleted} assignments across {len(ctx.related_session_ids)} related sessions",
             "total_deleted": total_deleted,
-            "scenario": request.scenario,
+            "scenario": scenario,
             "session_breakdown": breakdown,
         }
 
--- a/tests/unit/api/routers/test_draft_writes_invalidate_cache.py
+++ b/tests/unit/api/routers/test_draft_writes_invalidate_cache.py
@@ -148,7 +148,12 @@ class TestApplySolverResultsInvalidatesCache:
 
 
 class TestClearSessionAssignmentsInvalidatesCache:
-    """Bulk-clearing draft (or prod) assignments must drop the matching cache."""
+    """Clearing a scenario's draft assignments must drop the matching cache slot.
+
+    Production is out of reach here — kindred#2473: a scenario-less clear is
+    refused with a 422 before any PocketBase call, so there is no production
+    cache slot to invalidate any more.
+    """
 
     def _run(self, *, scenario: str | None) -> MagicMock:
         from api.routers.solver import router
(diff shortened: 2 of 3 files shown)


adamflagg:
## Independent scan (post-merge verification)

PR was already merged (`0686ea7d`) when this scan started. Reviewed the merged diff at floor 60 against the current tree rather than blocking on it.

**Findings: clean, 0 discarded.** Nothing to fix or file.

**Verified independently (not just re-reading the builder's claims):**
- `_require_scenario(request.scenario)` runs before any PocketBase access in `clear_session_assignments`; production `else` branch is gone, only `BUNK_ASSIGNMENTS_DRAFT` remains.
- Ran the full 28-test slice (`test_solver_never_writes_production.py` + `test_draft_writes_invalidate_cache.py`) clean.
- **Mutation check, full revert** (not just the collection-name swap): restored the pre-PR `clear_session_assignments` body verbatim from `4618f86c` and reran. 6 tests failed as expected — the class guard's behavioral test, its AST tripwire, both `TestClearAssignmentsRequiresScenario` refusal tests, and the shared-message test. Reapplied the fix; file diff is clean (`git diff --stat` empty) and the 28-test slice is green again.
- Grepped `frontend/src` for `clear-assignments` / `clearSessionAssignments` — only the generated `.openapi-schema.json` / `types.gen.ts` reference it, confirming the builder's "no live consumer" claim.
- Premise check (§5.5b): the issue body carries an explicit owner ruling (`★ RULED 2026-08-19 — refuse it`) predating this PR, so the "admin op vs. solver refusal" question was pre-settled, not argued for by the PR itself. Not a product decision in a bug's clothes.
- `gh pr checks 2487`: CI Summary pass, Python/Go Lint pass, Python Tests pass, CodeQL x3 pass, Secret Scanning pass. CodeRabbit shows "Review skipped: automatic reviews are disabled" (not requested, matches the mechanical/single-surface skip criteria) — no false-green risk since it wasn't relied on.
- `gh issue view 2473`: state CLOSED, closed by #2487 (not hand-closed).
- `verify-pr-state.sh push`: origin/main is at HEAD.

No GUI impact (0 frontend/src files in diff) and no staff-facing number/label redefinition — this is a refusal gate on a delete path, not a metric change. Both merge gates the builder self-certified check out under independent verification.

PR #2488, opened 2026-08-19
feat(metrics): distinguish no-coverage years from zero cancellations in historical trends

### LINKED ISSUE #2443: feat(metrics): distinguish no-coverage years from zero cancellations in historical trends

adamflagg:
## Summary

`attendee_status_history` only holds rows for one year in production (coverage began 2026), but the historical-trends endpoint requests a five-year window regardless, so the cancellation-rate trend rendered `0, 0, 0, 0, X` — four fabricated zeroes indistinguishable from a measured zero-cancellation year.

Adds `YearMetrics.has_cancellation_data`, set from the existing `fetch_status_history(year)` probe (both status filters left at their defaults, so it answers the year-level coverage question) — **no new repository method**, per the issue's warning about repeating the `hasattr`-guard trap PR #2444 fixed.

## Both consumers

- **Year-by-Year Summary table**: `Cancelled`/`Cancel %` render an em-dash (`—`) when `has_cancellation_data` is false, instead of `0` / `0.0%`. This is the house convention for "no data" (`SessionBunkHeatmap`, `DrillDownModal`, `RegistrationOverview`) — **adopted here as a default, explicitly vetoable by the owner**, per the issue.
- **Cancellation Rate trend chart**: a greyed band behind uncovered years.

### ★ Before/after — this changes what a staff-facing number means

```
Year-by-Year Summary, cancellation columns
  before: 2022 -> 0 / 0.0%   2023 -> 0 / 0.0%   2024 -> 0 / 0.0%   2025 -> 0 / 0.0%   2026 -> 276 / X%
  after:  2022 -> — / —      2023 -> — / —      2024 -> — / —      2025 -> — / —      2026 -> 276 / X%
Cancellation-rate trend chart
  before: a line through four zeroes rising to 2026
  after:  a greyed band across 2022-2025, the line drawn for 2026 only
```

No population, arithmetic, or denominator changes — only the distinction between *measured zero* and *never measured*.

## Chart rendering — the named build risk, and which route shipped

The issue flagged a build risk before estimating: `TrendLineChart`'s x-axis is a categorical `point` scale with zero bandwidth (recharts `^3.10.1`, `combineRealScaleType` for `type:'category'` inside `LineChart`), so a `ReferenceArea` would land tick-to-tick rather than covering each year's share of the axis — and the visible axis is drawn in the DOM by `ChartCard`, not by recharts. That analysis was already verified against the installed library and cited exact file/line evidence, so **this PR ships the issue's named DOM-layer fallback directly rather than building the broken SVG path first and reverting it**:

- `computeCoverageBands` (new, in `cssChartUtils.ts`) computes band extents using the **same edge-aligned percentage math** `VerticalXAxis` already uses to place the year labels (`left = i/(n-1) * 100%`, reduced by the same right padding) — so the band's alignment with the labels is correct by construction, not asserted from the SVG scale.
- Verified with: unit tests on the band-extent math (leading/trailing/middle uncovered runs, multiple runs) and component tests asserting the band renders/doesn't render per `has_cancellation_data` input and per metric.
- **Not yet eyeballed in a live browser** — this is a chart-rendering change and holds for the owner's visual check.

## Type coupling

`api/schemas/metrics.py` (source of truth) → regenerated `frontend/src/types/api-generated/types.gen.ts` (`npm run generate:api-types`, diff is whitespace-clean via `git diff -w`) → hand-maintained `frontend/src/types/metrics.ts` (the one the UI actually imports).

## Deliberately NOT done

No change to the existing chart-visibility gate (`TrendsOverview.tsx`, `data.years.some(y => (y.total_cancelled ?? 0) > 0)`) — a separate, pre-existing asymmetry the issue only notes in passing, out of scope here.

## Test plan

- [x] `uv run pytest tests/unit/api/services/test_historical_service.py` — 27 passed, including 4 new coverage tests (TDD: written first, verified red, then implemented)
- [x] `uv run pytest tests/` (full, `SKIP_POCKETBASE_TESTS=true -n auto`) — 6578 passed, 41 skipped
- [x] `cd frontend && npx vitest run` — 449/450 files green; the one failure (`PostCheckPopout.test.tsx`) is an unrelated pre-existing flake, confirmed by re-running that file in isolation (3/3 pass)
- [x] `bash scripts/pre-push-verify.sh` — ALL CHECKS PASSED (pytest, prettier, eslint, tsc, vitest)
- [x] Mutation-checked both new frontend guards (table em-dash condition, chart band condition) by breaking each and confirming the pinning test goes red, then restored

## Merge gate

This is GUI-affecting (`TrendLineChart.tsx`, `TrendsOverview.tsx`) **and** changes what a staff-facing number means (measured-zero vs never-measured). Per this repo's standing rule, it holds for the owner's eyes before merge regardless of CI color.

Closes #2443.

## Summary by CodeRabbit

- **New Features**
  - Historical cancellation metrics now distinguish unavailable data from measured zero cancellations.
  - Trend charts display shaded coverage bands for years without cancellation history.
- **Improvements**
  - Cancellation totals and rates show an em dash when historical data is unavailable, preventing misleading values.
  - Other trend metrics continue to display normally without coverage shading.
  - Existing data remains backward compatible when coverage information is not provided.

### DIFF
--- a/api/schemas/metrics.py
+++ b/api/schemas/metrics.py
@@ -420,6 +420,15 @@ class YearMetrics(BaseModel):
     new_vs_returning: NewVsReturning = Field(description="New vs returning breakdown")
     total_cancelled: int = Field(default=0, description="Total cancelled campers for this year")
     cancellation_rate: float = Field(default=0.0, description="Cancelled / (enrolled + cancelled)")
+    has_cancellation_data: bool = Field(
+        default=True,
+        description=(
+            "Whether attendee_status_history holds any rows for this year at all. "
+            "attendee_status_history only began being written in 2026, so prior years "
+            "have no coverage and total_cancelled/cancellation_rate are unmeasured, not "
+            "zero -- consumers should render 'no data' rather than the fabricated 0/0.0%."
+        ),
+    )
 
 
 class HistoricalTrendsResponse(BaseModel):
--- a/api/services/historical_service.py
+++ b/api/services/historical_service.py
@@ -97,15 +97,25 @@ class HistoricalService:
             # Cancellations use the SAME scope (session_types/session_cm_id/duration) and the
             # same distinct-person grain as the enrolled denominator above -- see #2434, where
             # an unscoped, row-counted numerator inflated the rendered rate ~2.4x.
+            #
+            # cancel_transitions is None when the underlying fetch failed (see
+            # _fetch_cancellation_transitions) -- that failure must not be reported as a
+            # confidently-measured zero: fold it into has_cancellation_data so the count and
+            # the coverage flag can never disagree about whether this year was actually
+            # measured this call.
+            cancellation_fetch_failed = cancel_transitions is None
             filtered_cancellations = self._filter_attendees(
-                cancel_transitions, sessions, session_types, session_cm_id, session_name, duration
+                cancel_transitions or [], sessions, session_types, session_cm_id, session_name, duration
             )
             cancelled_person_ids = {
                 pid for t in filtered_cancellations if (pid := getattr(t, "person_id", None)) is not None
             }
             cancel_count = len(cancelled_person_ids)
+            year_has_cancellation_data = has_cancellation_data and not cancellation_fetch_failed
 
-            year_metric = self._compute_year_metrics(year, person_ids, persons, cancel_count, has_cancellation_data)
+            year_metric = self._compute_year_metrics(
+                year, person_ids, persons, cancel_count, year_has_cancellation_data
+            )
             year_metrics_list.append(year_metric)
 
         return HistoricalTrendsResponse(years=year_metrics_list)
@@ -162,7 +172,7 @@ class HistoricalService:
 
         return None
 
-    async def _fetch_cancellation_transitions(self, year: int) -> list[Any]:
+    async def _fetch_cancellation_transitions(self, year: int) -> list[Any] | None:
         """Fetch raw cancellation status-transition rows for a year.
 
         Returns the unfiltered rows so the caller can scope them by
@@ -170,12 +180,17 @@ class HistoricalService:
         denominator (see calculate_historical_trends), and dedupe by
         person_id for the correct grain. No repository defines a
         pre-aggregated ``fetch_cancellation_count`` -- see #2434.
+
+        Returns ``None`` (not ``[]``) on a repository failure, so the caller
+        can distinguish "queried, zero rows" from "never queried" -- a fetch
+        failure must not be reported as a confidently-measured zero
+        cancellation count (see #2443's has_cancellation_data).
         """
         try:
             result: list[Any] = await self.repo.fetch_status_transitions(year, ["cancelled", "withdrawn", "dismissed"])
             return result
         except Exception:
-            return []
+            return None
 
     async def _fetch_cancellation_data_coverage(self, year: int) -> bool:
         """Whether attendee_status_history has ANY rows for this year.
--- a/api/services/historical_service.py
+++ b/api/services/historical_service.py
@@ -74,17 +74,19 @@ class HistoricalService:
         person_futures = [self.repo.fetch_persons(y) for y in years]
         session_futures = [self.repo.fetch_sessions(y, session_types=session_types) for y in years]
         cancel_futures = [self._fetch_cancellation_transitions(y) for y in years]
+        coverage_futures = [self._fetch_cancellation_data_coverage(y) for y in years]
 
         all_attendees = await asyncio.gather(*attendee_futures)
         all_persons = await asyncio.gather(*person_futures)
         all_sessions = await asyncio.gather(*session_futures)
         all_cancel_transitions = await asyncio.gather(*cancel_futures)
+        all_coverage = await asyncio.gather(*coverage_futures)
 
         # Compute metrics for each year
         year_metrics_list: list[YearMetrics] = []
 
-        for year, attendees, persons, sessions, cancel_transitions in zip(
-            years, all_attendees, all_persons, all_sessions, all_cancel_transitions, strict=True
+        for year, attendees, persons, sessions, cancel_transitions, has_cancellation_data in zip(
+            years, all_attendees, all_persons, all_sessions, all_cancel_transitions, all_coverage, strict=True
         ):
             # Filter attendees by session type and/or session name
             filtered = self._filter_attendees(attendees, sessions, session_types, session_cm_id, session_name, duration)
@@ -103,7 +105,7 @@ class HistoricalService:
             }
             cancel_count = len(cancelled_person_ids)
 
-            year_metric = self._compute_year_metrics(year, person_ids, persons, cancel_count)
+            year_metric = self._compute_year_metrics(year, person_ids, persons, cancel_count, has_cancellation_data)
             year_metrics_list.append(year_metric)
 
         return HistoricalTrendsResponse(years=year_metrics_list)
@@ -175,8 +177,29 @@ class HistoricalService:
         except Exception:
             return []
 
+    async def _fetch_cancellation_data_coverage(self, year: int) -> bool:
+        """Whether attendee_status_history has ANY rows for this year.
+
+        Not filtered by session_types, session_cm_id, duration, or status --
+        coverage is a year-level fact, independent of the scoping applied to
+        the enrolled/cancelled counts. Reuses the existing, already-defined
+        ``fetch_status_history(year, old_status=None, new_statuses=None)``
+        probe (both filters default) rather than adding a new repository
+        method -- see #2443.
+        """
+        try:
+            rows: list[Any] = await self.repo.fetch_status_history(year)
+            return len(rows) > 0
+        except Exception:
+            return False
+
     def _compute_year_metrics(
-        self, year: int, person_ids: set[int], persons: dict[int, Any], total_cancelled: int = 0
+        self,
+        year: int,
+        person_ids: set[int],
+        persons: dict[int, Any],
+        total_cancelled: int = 0,
+        has_cancellation_data: bool = True,
     ) -> YearMetrics:
         """Compute metrics for a single year from deduplicated person IDs.
 
@@ -220,4 +243,5 @@ class HistoricalService:
             new_vs_returning=new_vs_returning,
             total_cancelled=total_cancelled,
             cancellation_rate=cancellation_rate,
+            has_cancellation_data=has_cancellation_data,
         )
--- a/frontend/src/components/metrics/TrendLineChart.test.tsx
+++ b/frontend/src/components/metrics/TrendLineChart.test.tsx
@@ -1,6 +1,7 @@
 import { describe, it, expect, vi, beforeEach } from 'vitest'
 import { render, screen } from '@testing-library/react'
 import { TrendLineChart } from './TrendLineChart'
+import { computeCoverageBands } from './cssChartUtils'
 import type { YearMetrics } from '../../types/metrics'
 
 let capturedStrokeLinecaps: string[] = []
@@ -77,5 +78,114 @@ describe('TrendLineChart', () => {
         expect(capturedStrokeLinecaps.every((v) => v === 'round')).toBe(true)
       }
     )
+  })
+})
+
+// ============================================================================
+// #2443 — no-coverage years render a greyed band, not a fabricated zero.
+//
+// The x-axis is a categorical `point` scale with zero bandwidth (recharts
+// ^3.10.1, combineRealScaleType for type:'category' inside LineChart), so a
+// ReferenceArea drawn against it lands tick-to-tick, not slot-to-slot, and
+// this chart's visible axis is drawn in the DOM by ChartCard, not by
+// recharts -- see the issue body's build-risk analysis. This ships the
+// named DOM-layer fallback instead: a band positioned by the SAME
+// edge-aligned percentage math VerticalXAxis uses for the year labels
+// (cssChartUtils.ts, edgeAligned branch: left = i/(n-1)*100%, with the
+// same rightPadding reduction), so alignment is correct by construction
+// rather than asserted from the SVG scale.
+// ============================================================================
+
+describe('computeCoverageBands', () => {
+  it('returns no bands when every year has coverage', () => {
+    expect(computeCoverageBands([true, true, true])).toEqual([])
+  })
+
+  it('returns no bands for fewer than 2 points', () => {
+    expect(computeCoverageBands([false])).toEqual([])
+    expect(computeCoverageBands([])).toEqual([])
+  })
+
+  it('covers a leading run of uncovered years, stopping halfway to the first covered year', () => {
+    // 5 years, 2022-2025 uncovered (indices 0-3), 2026 covered (index 4).
+    const bands = computeCoverageBands([false, false, false, false, true])
+    expect(bands).toHaveLength(1)
+    expect(bands[0]!.leftPct).toBeCloseTo(0, 5)
+    // rightIndex = 3 + 0.5 = 3.5 of 4 -> 87.5%
+    expect(bands[0]!.widthPct).toBeCloseTo(87.5, 5)
+  })
+
+  it('covers a trailing run of uncovered years', () => {
+    const bands = computeCoverageBands([true, true, false, false])
+    expect(bands).toHaveLength(1)
+    // leftIndex = 2 - 0.5 = 1.5 of 3 -> 50%
+    expect(bands[0]!.leftPct).toBeCloseTo(50, 5)
+    // rightIndex = min(3, 3.5) = 3 -> 100%; width = 100-50 = 50
+    expect(bands[0]!.widthPct).toBeCloseTo(50, 5)
+  })
+
+  it('covers a middle run bounded by coverage on both sides', () => {
+    const bands = computeCoverageBands([true, false, false, true, true])
+    expect(bands).toHaveLength(1)
+    // leftIndex = 1-0.5=0.5 of 4 -> 12.5%; rightIndex = 2+0.5=2.5 of 4 -> 62.5%
+    expect(bands[0]!.leftPct).toBeCloseTo(12.5, 5)
+    expect(bands[0]!.widthPct).toBeCloseTo(50, 5)
+  })
+
+  it('returns one band per contiguous uncovered run', () => {
+    const bands = computeCoverageBands([false, true, false, true])
+    expect(bands).toHaveLength(2)
+  })
+})
+
+describe('TrendLineChart — coverage band rendering (#2443)', () => {
+  const coverageData: YearMetrics[] = [
+    {
+      year: 2022,
+      total_enrolled: 0,
+      by_gender: [],
+      new_vs_returning: {
+        new_count: 0,
+        returning_count: 0,
+        new_percentage: 0,
+        returning_percentage: 0,
+      },
+      total_cancelled: 0,
+      cancellation_rate: 0,
+      has_cancellation_data: false,
+    },
+    {
+      year: 2026,
+      total_enrolled: 100,
+      by_gender: [],
+      new_vs_returning: {
+        new_count: 40,
+        returning_count: 60,
+        new_percentage: 40,
+        returning_percentage: 60,
+      },
+      total_cancelled: 276,
+      cancellation_rate: 43.9,
+      has_cancellation_data: true,
+    },
+  ]
+
+  it('renders a coverage band for the cancellation_rate metric when a year lacks data', () => {
+    render(
+      <TrendLineChart data={coverageData} title="Cancellation Rate" metric="cancellation_rate" />
+    )
+    expect(screen.getAllByTestId('coverage-band').length).toBeGreaterThan(0)
+  })
+
+  it('renders no coverage band when every year has data', () => {
+    const covered = coverageData.map((y) => ({ ...y, has_cancellation_data: true }))
+    render(<TrendLineChart data={covered} title="Cancellation Rate" metric="cancellation_rate" />)
+    expect(screen.queryByTestId('coverage-band')).not.toBeInTheDocument()
+  })
+
+  it('renders no coverage band for metrics other than cancellation_rate', () => {
+    // total_enrolled has no coverage concept -- the band is scoped to cancellation data only.
+    render(<TrendLineChart data={coverageData} title="Total Enrolled" metric="total" />)
+    expect(screen.queryByTestId('coverage-band')).not.toBeInTheDocument()
   })
 })
--- a/frontend/src/components/metrics/TrendLineChart.tsx
+++ b/frontend/src/components/metrics/TrendLineChart.tsx
@@ -17,7 +17,7 @@ import {
 import { useMemo } from 'react'
 import type { YearMetrics } from '../../types/metrics'
 import { ChartCard } from './ChartCard'
-import { getNiceTicks, calculateVerticalLayout } from './cssChartUtils'
+import { getNiceTicks, calculateVerticalLayout, computeCoverageBands } from './cssChartUtils'
 import { formatLabelListValue, formatLabelListPercent } from '../../utils/chartFormatters'
 
 const COLORS = {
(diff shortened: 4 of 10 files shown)


adamflagg:
@coderabbitai review

adamflagg:
Addressed the CodeRabbit finding about the coverage flag disagreeing with a failed cancellation-count fetch (deb659a1): a failed \`fetch_status_transitions\` now returns \`None\` instead of \`[]\`, and \`has_cancellation_data\` folds in that failure so the count and the flag can never disagree about whether a year was actually measured this call. TDD: added a failing test first (mocked \`fetch_status_transitions\` to raise while \`fetch_status_history\` succeeds), verified red, implemented, mutation-checked.

On the other actionable comment (IDs at test_historical_service.py:854/892): those are \`9001\`/\`5001\`, following this file's own established synthetic-ID convention throughout (\`1001\`, \`5001\`, \`5002\`, \`2001+i\`, \`3001\`, etc. — see \`make_mock_person\`/\`make_mock_attendee\` calls above them in the same file). Not real CampMinder IDs; dismissing as a false positive.

adamflagg:
Independent scan (triage-attack Part B, floor 60, high effort) on top of the builder's own scan-it pass. Reviewed the full diff plus premise; verified every claim in the builder's report against current code rather than trusting it.

**Verified, not just re-read:**
- `has_cancellation_data` / `_fetch_cancellation_transitions` fix (deb659a1): confirmed a failed `fetch_status_transitions` now returns `None`, and `year_has_cancellation_data = has_cancellation_data and not cancellation_fetch_failed` folds that failure into the coverage flag. New regression test (`test_coverage_false_when_cancellation_fetch_fails_even_if_history_probe_succeeds`) reproduces the CodeRabbit-flagged scenario and asserts `has_cancellation_data is False, total_cancelled == 0`. Ran the suite: 28/28 pass.
- CodeRabbit's second finding (test IDs `9001`/`5001` at lines 854/892 flagged as "real CampMinder IDs"): opened the file and cross-checked against the file's own established synthetic-ID convention (`1001`, `1002`, `2001+i`, `3001`, `5001`, `5002` used throughout `make_mock_person`/`make_mock_attendee` calls above). Confirmed false positive — dismissal is correct.
- `computeCoverageBands` alignment claim: opened `VerticalXAxis`'s edge-aligned branch (`cssChartUtils.ts`) — it places labels at `left: (i/(n-1))*100%` inside a box with `marginRight: rightPadding` (20px). `TrendLineChart`'s band container uses the identical `leftPct = (leftIndex/(n-1))*100` formula with `right: '20px'`, and the LineChart's own margin is `{ right: 20, left: 0 }`. The math is mirrored correctly by construction, as claimed — still an unrendered claim until eyeballed live, which is why this holds regardless.
- Em-dash convention: confirmed real in `SessionBunkHeatmap.tsx`, `DrillDownModal.tsx`, `RegistrationOverview.tsx` — not invented for this PR.
- Premise: issue #2443 carries an explicit 2026-08-19 OWNER RULING for the greyed band, including the exact recharts `point`-scale/zero-bandwidth analysis and the DOM-layer fallback this PR ships. Not a decision wearing a bug's clothes — the decision was already made in the issue body before this PR existed.
- `closingIssuesReferences` → `[2443]` (GraphQL), confirmed non-empty. `origin/feature/metrics-coverage-band` is at HEAD (verify-pr-state.sh push: 0). CI Summary: pass.

**No new findings.** Nothing to fix or file beyond what's already in deb659a1.

**Holds for the owner's eyes per the standing GUI-impact rule** — `TrendLineChart.tsx`, `cssChartUtils.ts`, `TrendsOverview.tsx` are touched, and the coverage band has not been eyeballed in a live browser. Dev server: `frontend` on VITE_PORT 3117 in worktree `.worktrees/metrics-coverage-band` (branch `feature/metrics-coverage-band`). What to look at: the Historical Trends page → Cancellation Rate chart, confirm the grey band sits behind 2022-2025 aligned with those year labels (not drifted), and the Year-by-Year Summary table's Cancelled/Cancel % columns read fine as em-dashes for 2022-2025 vs 276/X% for 2026.

Issue #2482, opened 2026-08-19
feat(sync): capture cabin-value change history so each weekend's cabin is derived, not guessed


adamflagg:
## ★ Status: FULLY RULED, 2026-08-19. Nothing here is blocked on an owner decision.

Split out of #2464 by owner ruling on 2026-08-19. #2464 keeps its own scope (the
lodging work queue has no surface for six of its seven kinds); **this** issue
carries the cabin-value change-capture design that was ruled in #2464's comments
and does not belong in a work-queue issue. **Not a duplicate of #2464** — #2464
is the *consumer* that benefits, this is the mechanism.

| | |
|---|---|
| **Ruled** | The rule (value-as-of-weekend), the amendment (interval is `(prev end, own end]`), the new table's shape, the write-hook location, the retention scope (cabin fields only) |
| **✅ RULED 2026-08-19** | **Cadence.** Retain the weekly unrestricted sweep; ADD a bounded daily pass (family-camp attendees, ANY status) inserted between the source and transform phases. Measured cost +11.5 min. See "CADENCE — RULED" below. **Nothing in this issue is now blocked on an owner decision.** |

---

## The problem this solves

`Family Camp Cabin` (cm_id 218072, household grain) holds **one value per
household per year**. `family_camp_registrations` is unique on
`(household, year)` — `CREATE UNIQUE INDEX idx_fc_reg_unique ON family_camp_registrations (household, year)`,
confirmed on the production DB. A household attending two weekends has one slot,
so the sync cannot tell which weekend a cabin string belongs to and files
`ambiguous_session` instead of guessing (`pocketbase/sync/lodging_assignments_sync.go:463`,
`// flag, do not guess (spec 3.6)`).

Production, 2026-08-18 21:47 snapshot: ~~**7 open `ambiguous_session` rows**, 7
distinct households~~. **RE-MEASURED 2026-08-19 at build time against the same
snapshot file, whose newest `lodging_ingest_issues` row is now 2026-08-18 22:11:
9 open `ambiguous_session` rows across 8 distinct households, `occurrences = 1`
each, all 9 on `Family Camp Cabin` (household grain).** The count moved because
the queue is live and kept filing after 21:47; the direction is more ambiguity,
not less, so nothing in the argument below changes. Exactly 10 households are
enrolled in 2+ family weekends in 2026 (9 in two, 1 in three — re-verified);
8 of them have now written a cabin value and are those 9 rows.

## The rule, as ruled

[...]

PR #2489, opened 2026-08-19
feat(sync): insert a bounded daily family-camp custom-values pass between source and transform

adamflagg:
## What

Cadence half of #2482 (owner-ruled 2026-08-19). Adds a **bounded daily custom-values pass** for
family-camp attendees, inserted into the daily sync between the source jobs and the transform
jobs. Does **not** implement the capture half (the `lodging_value_history` table / write hook) —
that is separate scope. **This PR closes nothing on its own; it is the prerequisite for the
capture half.**

## The defect this fixes

`getDailySyncJobs()` in `pocketbase/sync/orchestrator.go` jumps straight from the source jobs
(ending at `financial_transactions`) to the transform jobs (starting at `family_camp_derived`),
even though `GetAllPhases()` already declares the intended order
`[PhaseSource, PhaseExpensive, PhaseTransform, ...]` and `person_custom_values` /
`household_custom_values` are already registered as `PhaseExpensive`. Because the daily cron
never ran a `PhaseExpensive` job, every daily run transformed fresh source data against custom
values up to 7 days stale (the weekly sweep runs Sunday 04:00) — and `family_camp_derived`,
which writes `needs_fridge` and resolves the cabin/accommodation fields, is in that transform
list.

## The fix

- **`session_resolver.go`**: `GetPersonIDsForSession` / `GetHouseholdIDsForSession` hardcoded
  `status = 'enrolled'`. Per the ruling, the bounded pass must observe a household moving *in or
  out of* enrolled (a cancellation or waitlist entry), so enrolled-only would defeat the point.
  Added status-agnostic **siblings** — `GetPersonIDsForSessionAnyStatus` /
  `GetHouseholdIDsForSessionAnyStatus` — rather than relaxing the existing functions, because
  those back manual `?session=` runs where enrolled-only stays correct. Also added
  `GetFamilyCampSessionCMIDs` / `GetFamilyCampPersonIDsAnyStatus` /
  `GetFamilyCampHouseholdIDsAnyStatus`, which union the any-status cohort across every
  family-camp weekend session in the year (the cohort is selected via `attendees`, not custom
  values — reading custom values to decide who to sync custom values for would be circular).
- **`person_custom_field_values.go`** / **`household_custom_field_values.go`**: added a
  `FamilyCampBounded` field. When set, `getPersonIDsToSync` / `getHouseholdIDsToSync` use the new
  any-status family-camp cohort instead of `Session` or the year-wide fallback. The plain
  `Session=""` (unbounded) path used by the weekly sweep and manual runs is untouched.
- **`orchestrator.go`**: registers two new dedicated service instances,
  `person_custom_values_family_camp` and `household_custom_values_family_camp`
  (`FamilyCampBounded = true`), distinct from the existing `person_custom_values` /
  `household_custom_values` singletons. Inserted into `getDailySyncJobs()` between
  `financial_transactions` and `family_camp_derived`, and updated that function's stale comment
  about "custom values from the most recent weekly sync." Added both new jobs to `syncJobMeta`
  as `PhaseExpensive`. **Left `GetDefaultUnifiedSyncJobs` (the manual `service=all` API trigger)
  untouched** — the ruling is specifically about the daily cron flow.
- **`scheduler.go`**: **unchanged.** The weekly unrestricted sweep (`0 4 * * 0`,
  `runCustomValuesSync`) is retained exactly as-is — `person_custom_values` feeds
  `household_demographics`, `camper_dietary`, `camper_transportation`, and
  `financial_aid_applications`, none of which this bounded pass touches.

## Measured cost (from `sync_runs`, one recorded full batch, 2026-08-16)

| | value |
|---|---|
| unrestricted pair (weekly), wall clock | 2,584 s = 43.1 min (2,596 persons / 2,070 households) |
| bounded family-camp cohort, any status | 781 persons / 447 households ⇒ ~11.5 min |
| daily run today | ~5 min |
| **projected daily run with this pass** | **~16.5 min**, against the daily cron's unbounded context (no fixed timeout on `RunDailySync`) |

n=1 for the per-entity rates (`sync_runs` only begins 2026-08-14) — re-measure once a few daily
runs exist under this change.

- No `lodging_value_history` migration and no write-hook changes — that's the capture half,
  tracked separately on #2482.
- No changes to `family_camp_derived.go`.
- No changes to `GetDefaultUnifiedSyncJobs` / the manual `service=all` API trigger — only the
  daily cron's `getDailySyncJobs()`.
- No relaxation of `GetPersonIDsForSession` / `GetHouseholdIDsForSession`'s enrolled-only
  behavior — manual `?session=` runs are unaffected.

## Testing

New test file `pocketbase/sync/family_camp_daily_cadence_test.go`:
- `TestSessionResolver_AnyStatus_IncludesNonEnrolled` — pins enrolled-only vs. any-status on both
  the person and household resolvers.
- `TestGetFamilyCampIDsAnyStatus_SpansWeekendsExcludesOtherSessions` — cohort spans every
  family-camp weekend, excludes a non-family session's attendees.
- `TestPersonCustomFieldValuesSync_FamilyCampBounded` /
  `TestHouseholdCustomFieldValuesSync_FamilyCampBounded` — bounded-mode wiring on the sync
  services themselves.
- `TestGetDailySyncJobs_FamilyCampBoundedPassBetweenSourceAndTransform` — ordering: strictly
  after `financial_transactions`, strictly before `family_camp_derived`; asserts the unrestricted
  job names stay out of the daily list.
- `TestSyncJobMeta_FamilyCampBoundedJobsAreExpensivePhase` — new jobs registered as
  `PhaseExpensive`.
- `TestWeeklyCustomValuesCronRetained` — pins `scheduler.go`'s `0 4 * * 0` / `runCustomValuesSync`
  survive untouched.

All new tests were written first, verified failing (compile error against the not-yet-added
symbols), then made to pass. Two were mutation-checked by deliberately breaking the
implementation and confirming red before restoring.

```
cd pocketbase && go build ./...      # PASS
cd pocketbase && go vet ./...        # PASS
cd pocketbase && go test ./...       # PASS
bash scripts/pre-push-verify.sh      # PASS (go build, golangci-lint, go test -race)
```

The parent issue (kindred#2482) stays open -- this PR is the cadence half only; the capture half (the `lodging_value_history` table and write hook) is separate, future work.

### DIFF
--- a/pocketbase/sync/family_camp_daily_cadence_test.go
+++ b/pocketbase/sync/family_camp_daily_cadence_test.go
@@ -1,0 +1,389 @@
+package sync
+
+import (
+	"os"
+	"sort"
+	"strings"
+	"testing"
+
+	"github.com/pocketbase/pocketbase/core"
+	pbtests "github.com/pocketbase/pocketbase/tests"
+)
+
+// cadenceTestApp builds a throwaway app carrying only the collections the
+// bounded daily family-camp cadence pass touches (kindred#2482). It is
+// deliberately separate from newSyncTestApp: that shared fixture's attendees
+// collection has no "status" text field (only status_id), and the resolver's
+// existing enrolled-only path filters on the text column
+// (`status = 'enrolled'`), so a test exercising both the enrolled-only and
+// any-status paths needs both columns present.
+func cadenceTestApp(t *testing.T) core.App {
+	t.Helper()
+	app, err := pbtests.NewTestApp()
+	if err != nil {
+		t.Fatalf("NewTestApp: %v", err)
+	}
+	t.Cleanup(app.Cleanup)
+
+	sessions := core.NewBaseCollection("camp_sessions")
+	sessions.Fields.Add(&core.NumberField{Name: "cm_id"})
+	sessions.Fields.Add(&core.TextField{Name: "name"})
+	sessions.Fields.Add(&core.TextField{Name: "session_type"})
+	sessions.Fields.Add(&core.NumberField{Name: "parent_id"})
+	sessions.Fields.Add(&core.NumberField{Name: "year"})
+	if err := app.Save(sessions); err != nil {
+		t.Fatalf("save camp_sessions: %v", err)
+	}
+
+	households := core.NewBaseCollection("households")
+	households.Fields.Add(&core.NumberField{Name: "cm_id"})
+	households.Fields.Add(&core.NumberField{Name: "year"})
+	if err := app.Save(households); err != nil {
+		t.Fatalf("save households: %v", err)
+	}
+
+	persons := core.NewBaseCollection("persons")
+	persons.Fields.Add(&core.NumberField{Name: "cm_id"})
+	persons.Fields.Add(&core.NumberField{Name: "household_id"})
+	persons.Fields.Add(&core.RelationField{Name: "household", CollectionId: households.Id, MaxSelect: 1})
+	persons.Fields.Add(&core.NumberField{Name: "year"})
+	if err := app.Save(persons); err != nil {
+		t.Fatalf("save persons: %v", err)
+	}
+
+	attendees := core.NewBaseCollection("attendees")
+	attendees.Fields.Add(&core.RelationField{Name: "person", CollectionId: persons.Id, MaxSelect: 1})
+	attendees.Fields.Add(&core.NumberField{Name: "person_id"})
+	attendees.Fields.Add(&core.RelationField{Name: "session", CollectionId: sessions.Id, MaxSelect: 1})
+	attendees.Fields.Add(&core.TextField{Name: "status"})
+	attendees.Fields.Add(&core.NumberField{Name: "status_id"})
+	attendees.Fields.Add(&core.NumberField{Name: "year"})
+	if err := app.Save(attendees); err != nil {
+		t.Fatalf("save attendees: %v", err)
+	}
+
+	return app
+}
+
+func cadenceAddSession(t *testing.T, app core.App, cmID int, sessionType string, year int) string {
+	t.Helper()
+	return saveRecord(t, app, "camp_sessions", map[string]any{
+		"cm_id": cmID, "name": sessionType, "session_type": sessionType, "year": year,
+	})
+}
+
+func cadenceAddHousehold(t *testing.T, app core.App, cmID, year int) string {
+	t.Helper()
+	return saveRecord(t, app, "households", map[string]any{"cm_id": cmID, "year": year})
+}
+
+func cadenceAddPerson(t *testing.T, app core.App, cmID, householdCMID, year int, householdPBID string) string {
+	t.Helper()
+	return saveRecord(t, app, "persons", map[string]any{
+		"cm_id": cmID, "household_id": householdCMID, "household": householdPBID, "year": year,
+	})
+}
+
+func cadenceAddAttendee(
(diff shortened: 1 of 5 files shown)


adamflagg:
## Independent scan (post-merge verification, floor 60)

This PR was already merged when the independent scan pass started. Verified retrospectively against the merge commit (`1b9b281`) rather than holding.

**Verified against current code, not just the builder's report:**
- Diff scope: exactly `session_resolver.go`, `person_custom_field_values.go`, `household_custom_field_values.go`, `orchestrator.go`, `family_camp_daily_cadence_test.go` — matches the file allowlist claimed.
- `getDailySyncJobs()` ordering: pinned by `TestGetDailySyncJobs_FamilyCampBoundedPassBetweenSourceAndTransform`, confirmed to place the two bounded jobs strictly between `financial_transactions` and `family_camp_derived`, and confirmed the weekly-only job names (`person_custom_values`, `household_custom_values`) are absent from the daily list.
- Weekly cron (`"0 4 * * 0"`) unchanged in `scheduler.go` — confirmed by grep, not just by the new test.
- Doc-comment fix (person/household `CustomFieldValuesSync` type comments) landed correctly — reads "A second, `FamilyCampBounded` instance ... IS part of the daily cron" on both types.
- Ran the new/touched tests locally (`go test ./sync/... -run 'FamilyCamp|AnyStatus' -v`): all pass, including `TestSessionResolver_AnyStatus_IncludesNonEnrolled`, `TestGetFamilyCampIDsAnyStatus_SpansWeekendsExcludesOtherSessions`, `TestPersonCustomFieldValuesSync_FamilyCampBounded`, `TestHouseholdCustomFieldValuesSync_FamilyCampBounded`, `TestSyncJobMeta_FamilyCampBoundedJobsAreExpensivePhase`, `TestWeeklyCustomValuesCronRetained`.
- Premise check (§5.5b): the cadence change is not persuasion-shaped — the issue's ordering claim (`GetAllPhases()` already declares `PhaseExpensive` between source and transform, a real production full run already placed these jobs at 10/26–11/26 in that slot) is a contradiction between the daily job list and the orchestrator's own declared phase order, not an argued-for new behavior. No product decision hiding here.
- No GUI/visual-impact files (`gh pr diff --name-only | grep frontend/src` — no match). No staff-facing number or label redefinition — pure sync-cadence plumbing.
- CI Summary and all 3 Go Tests shards: confirmed green on the actual run (not just an absent-check false pass).
- Push verified: `verify-pr-state.sh push` → remote ref at HEAD.

**Filed issue #2491 (api.go run-guard gap) re-verified, not just re-read:**
- Confirmed `handleCustomValuesSync` at `api.go:1474` guards on the literal strings `"person_custom_values"` / `"household_custom_values"` exactly as quoted.
- Confirmed `handlePersonCustomFieldValuesSync` / `handleHouseholdCustomFieldValuesSync` route through `RunSingleSyncWithService`, which reserves `o.runningJobs[syncType]` keyed by the **literal syncType string** — for these two handlers that's the unrestricted names, never the `_family_camp` job names the daily cron uses. So the two reservation maps genuinely don't overlap: a manual trigger and the daily bounded pass can run concurrently against the same underlying collections. The finding is real, correctly sized L (needs a name-agnostic "is anything writing to this collection" check, `api.go` outside this PR's file scope), and correctly filed rather than fixed in-branch.

**Process gap noted, not blocking:** no PR-comment record of the original scan (§6 "post one comment on the PR") was found on #2489 or on #2482 — this comment is that record, filed after the fact.

**Nothing discarded, nothing further found at floor 60.** No additional findings from this independent pass.

Issue #2475, opened 2026-08-19
bug(api,frontend): the Returning badge counts any prior-year household row, so a family that only ever cancelled or waitlisted reads as Returning


adamflagg:
## ★★ RULED AND SHIPPED 2026-08-19 — BOTH halves. This issue is no longer blocked on anything.

**Owner, on being shown the corrected blast radius:** *"numbers make sense. proceed."*

Shipped in **PR #2485** (merged `87859f53`). `fetch_prior_household_cm_ids` now reads `attendees`
with `year < {year} && (_attendee_weekend_session_filter()) && ACTIVE_ENROLLED_FILTER` — a prior
**enrolled weekend** visit. That single predicate discharges both halves at once:

| half | what it fixes | status |
|---|---|---|
| **A/B** — cancelled/waitlisted-only reads as Returning | `status_id = 2` is now required | ✅ shipped |
| **summer-only** — a summer-only family is first-time to Family Camp | `session_type IN ('family','adult')` is now required | ✅ shipped, **ruled 2026-08-19** |

**Measured on the production snapshot, 2026:** of **357** households currently badged Returning,
**~35** flip to first-time. Restricting to `status_id = 2` alone (the A/B half by itself) flips
**~23**; the extra **~12** are households whose only prior *enrolled* attendance was a summer
session. Two independent replications agreed on the structure and differed by 1–2 households on how
the household join is walked — the split, not the exact integer, is the load-bearing fact.

⚠️ **The PR was originally described as flipping 21 and as "not touching the summer-only half."
Both claims were wrong** and were corrected on the PR before merge. The session-type predicate is
precisely what resolves the summer-only half, so it was never separable in the shipped shape.

⛔ **`persons.years_at_camp` remains the wrong tool** and must not be reached for later: it counts
SUMMER attendance and is blind to weekends — **225 of 360** genuine 2026 returners have `max = 0`.

### ⛔ SUPERSEDED 2026-08-19 — the passage below marked the summer-only half as blocked

The body previously recorded the owner's concern that the classification *"cannot be computed
accurately until we fix the prior year enrolled but not housing and no kids / cancelled"*, and named
**#2305** as a hard predecessor. **That is no longer operative.** The shipped mechanism derives
enrollment from `attendees.status_id` joined to `camp_sessions.session_type` — a path that does not
read the `family_camp_*` tables at all, so #2305's missing enrollment column never gated it. #2305
[...]

Issue #2438, opened 2026-08-18
feat(sync,api,frontend): resolve step-free access into needs_step_free and publish ramp coverage


adamflagg:
## ⚙️ BUILT IN PR #2494 (2026-08-19). Body corrected against the tree — read this first.

The build is done and open as PR #2494, **held for the owner's eyes, not merged**. Four figures
below were re-measured on the same snapshot at the same grain and came out different; the tree
wins, and the corrections are marked ✅ inline. **Nothing about the ruling changes** — the mobility
signal is still comfortably larger than the fridge signal, and the gate conclusion (3 ungated
households) reproduces exactly.

| body said | re-measured | where |
|---|---|---|
| 84 households with any narrative | **86** | the demand table below |
| mobility = 10 | **11** on the body's own surface; **14** on the surface actually shipped | the demand table below |
| bathroom-narrative-only = 4 | **5** | the routing-trap table below |
| 7 of 10 gated | **8 of 11** on the body's surface, **11 of 14** on the shipped one — 3 ungated either way | the gate-coverage section |
| migration `1500000165` | **`1500000167`** — 165 and 166 went to #2482 and #2305 | the method table and files list |

The shipped surface adds `knee` and `hip`, which this body listed as a widening option. They are
admitted on evidence: **every knee/hip mention in the whole corpus, in every year, is a mobility
limitation** (zero false positives), and three are households nothing else in the surface catches.
Bare `close to` stays out — it is proximity, which the registry answers with map coordinates and
`near_bathhouse`, not with `has_ramp`.

---

## ★ OWNER RULING, 2026-08-19 — BUILD IT. Step-free access becomes a third graded need dimension.

**Decided: `has_ramp` is surfaced, as a need dimension mirroring `needs_fridge`, on the pattern PR
#2462 shipped.** Not as an amenity checkbox, and not dropped.

Two new fields, one on each side of the match:

| side | new field | where it is derived | shape |
|---|---|---|---|
| **demand** | `needs_step_free` on `family_camp_registrations` | narrative resolver in `pocketbase/sync/family_camp_derived.go`, beside `mentionsFridge` | `bool` |
| **supply** | `ramp_coverage` on the lodging unit payload | resolved server-side in `api/services/lodging_roster_service.py`, recursive to leaves | graded, **with a `partial` case a boolean amenity does not have** |

Then one further entry in `NEEDS_DIMENSIONS` (`frontend/src/components/weekend/needsFit.ts:96`), which
[...]

Issue #2305, opened 2026-08-13
bug(sync): family camp tables carry no enrollment status, so non-attending households render as attendees


adamflagg:
## ⚠️ RE-ANCHORED 2026-08-27 — every line number below drifted; locate by NAME

> ⚠️ **Re-checked 2026-08-27, later the same day: TWO of the ten rows below moved AGAIN.** PR #2572 (`b184fb56`) added nine lines to `api/schemas/lodging.py` at `:800`, shifting everything under it. Against `main` @ `2cd6950b`: **`EnrollmentState` is `:1117`** (not `:1108`) and **`HouseholdJourneyYear.enrollment` is `:1196`** (not `:1187`). The other eight rows held exactly — including `"enrolled" if children else "none_on_file"` at `lodging_roster_service.py:2460`, whose file changed but only in same-size hunks. This is the fourth move in four passes; **locate by name.**

**Nothing about the argument, the ruling or the three named edits changes.** But #2555 (`c27466d6`) and #2558 (`56832b41`) merged on 2026-08-24 into the same three files this issue edits, and **every `file:line` written into this body is now wrong.** Re-derived against `main` @ `1849f1a1`:

| symbol | this body says | actual on `1849f1a1` |
|---|---|---|
| `EnrollmentState` (`api/schemas/lodging.py`) | `:776`, then `:841` | **`:1108`** — still `Literal["enrolled", "none_on_file"]` |
| `HouseholdJourneyYear.enrollment` | `:920` | **`:1187`** |
| `fetch_household_registration_years` (`api/services/lodging_repository.py`) | `:688`, `:624` | **`:691`** |
| `fetch_household_family_attendees` | `:539`, `:410` | **`:599`** |
| `fetch_household_adults_by_year` | `:589` | **`:656`** |
| `WEEKEND_SESSION_TYPES` | `:100` / `:101` | **`:104`** |
| `_weekend_type_filter` | `:139` / `:143` | **`:146`** |
| `_attendee_weekend_session_filter` | `:149` / `:147` | **`:150`** |
| `"enrolled" if children else "none_on_file"` (`api/services/lodging_roster_service.py`) | `:1305-1307`, then `:1497` | **`:2460`** |
| `row.enrollment === 'none_on_file'` (`HouseholdJourneyCard.tsx`) | `:186-193`, then `:298-303` | **`:298`** — unchanged, and the card still needs **no** change |

[...]

PR #2492, opened 2026-08-19
feat(sync): family_camp_* rows carry the household's weekend enrollment status

adamflagg:
The three `family_camp_*` tables are derived from CUSTOM VALUES -- a form a family
filled in -- and filling the form is not attending. Nothing on the row said which
it was, so a household that registered and never came was indistinguishable from
one that did. Staff reported it from the other end: a family waitlisted for a
weekend, whose child went to summer camp instead, still shows that year in the
household journey.

Adds `enrollment_status` to `family_camp_adults`, `family_camp_registrations` and
`family_camp_medical` (migration `1500000166`) and derives it in
`family_camp_derived`.

## The derivation

Two stages, from `attendees` x `camp_sessions`:

1. ANY actively enrolled member (`status_id = 2`) on a **family or adult** weekend
   makes the whole household `enrolled`. Mixed households are real -- 43 of 594 at
   (household, year) grain in 2026 -- and a family with one cancelled child and one
   who came did attend.
2. Otherwise the single best non-enrolled status, by the ordering
   `frontend/src/utils/enrollmentFilter.ts` already applies when the board picks the
   one status to show a family that did not enroll.
3. No family/adult attendee row at all becomes `none_on_file`.

**`session_type IN ("family", "adult")`, not `"family"` alone.** An adult weekend is
a family-camp weekend; it differs only in enrolling the parent directly rather than
their children. The predicate matters and this is where it was going to be got
wrong: measured on the production snapshot for 2026, 89 of 480 household-journey
rows are badged "No enrollment" today and **33 of those households did enroll** --
on a session typed `adult`. Copying `fetch_household_family_attendees`'s family-only
filter would have carried that error straight into the stored column.

**`none_on_file`, not `none`.** CampMinder has an occupied status spelled `none`
(409 weekend attendee rows carry it), so reusing it would collapse "we hold a row
and it says none" into "we hold no row".

**The status ordering is duplicated, and pinned.** The derivation runs in Go and
`getStatusPriority` is TypeScript, so there is no shared source. The Go pairing this
would otherwise have reused (`familySessionTypes` in `camper_history.go`) went away
with the `camper_history` table in migration `1500000157`, so there is nothing left
in Go to import. `TestFamilyCampStatusPriorityMatchesTheFrontend` reads the real
`.ts` file and fails if the two ever disagree.

## Projected values, 2026 production snapshot

| table | rows | enrolled | registered, none enrolled | no attendee row |
|---|---|---|---|---|
| `family_camp_registrations` | 479 | 423 | 55 | 1 |
| `family_camp_adults` | 860 | 768 | 90 | 2 |
| `family_camp_medical` | 900 | 519 | 77 | 304 |

## Forward-only, and no replay

D12 stands: no `family_camp_*` table is replayed here. The column is an ATTRIBUTE,
not a key -- the write key, the three `Processed*Keys` sets and the three unique
indexes are untouched, so the "grain is a triple" trap in
`docs/reference/family-camp-grain-collapse.md` does not apply. The column populates
from the next derive of a year and the eventual post-#2255/#2275 replay picks up
history for free.

The additive-column trap is handled rather than dodged: `enrollment_status` is part
of all three change comparisons, so a row written before the column existed is
UPDATED rather than skipped as unchanged. A row left holding `""` would be exactly
the "could not determine" case the column exists to prevent.
`TestFamilyCampEnrollmentStatusRewritesAPreExistingRow` pins it.

The field is `required: false` deliberately: rows for 2017-2025 keep `""` until that
replay, and a required field would make every one of them unsaveable in the meantime.

## Refusing what it cannot know

`none_on_file` is a POSITIVE claim, so the derivation refuses a year whose
`camp_sessions` were never synced rather than writing it. Absence from the status
map is what produces `none_on_file`; with the sessions table empty for the year
every household is absent for a reason that has nothing to do with the household,
and the column being part of all three change comparisons means the wrong answer
lands rather than sitting inert. `POST /api/custom/sync/run?year=NNNN&service=family_camp_derived`
takes an arbitrary year, so this was reachable.

The discriminator is the year's `camp_sessions` rows, not its weekends: a season
that ran sessions and no family or adult weekend is a real answer (`none_on_file`
for everyone), a season with no sessions row at all has not been synced. Every year
2017-2026 on the snapshot carries between 6 and 18 weekends, so the second case is
never a real season.

⚠️ **The stored vocabulary is wider than the wire's.** `EnrollmentState` in
`api/schemas/lodging.py` is `Literal["enrolled", "none_on_file"]` and types
`HouseholdJourneyYear.enrollment`. The follow-on below must WIDEN that Literal
before it publishes this column -- 55 of 2026's 479 registration rows store a status
slug the two-value Literal would reject with a Pydantic `ValidationError`.

**The household journey still badges from `if children`, so no staff-visible number
moves in this PR.** Switching `build_household_journey` to read the stored status is
the other half of #2305 and needs `api/services/lodging_repository.py`, which #2475
owns and PR #2485 currently has open. Measured on the 2026 snapshot: all 33 of the
wrongly-badged households hold a `family_camp_registrations` row and only 1 holds a
`family_camp_adults` row, so the status cannot be read off the rows
`build_household_journey` already has in hand -- `fetch_household_registration_years`
requests `fields: "year"` and strips it. Doing it here would mean editing a file
another open PR is changing. #2305 therefore stays open; the follow-on is one
repository read, the `EnrollmentState` widening noted above, and the one-line
switch.

`api/schemas/lodging.py` and `needsFit.ts` (#2438) are untouched for the same reason,
and no frontend component changed -- `frontend/src/types/pocketbase-types.ts` is the
regenerated artifact the CI freshness check requires, three added lines.

## Verification

- `bash scripts/pre-push-verify.sh` -- ALL CHECKS PASSED
- `cd pocketbase && go test ./...` -- green
- `golangci-lint run --config ../.golangci.yml` -- 0 issues
- Eight new tests, each mutation-checked: reverting the derivation to
  `session_type = "family"` fails the adult-weekend test; swapping a priority rank
  fails both the parity test and the fallback test; dropping the column write fails
  the create and update tests; removing the no-sessions guard fails the refusal
  test while leaving the no-weekends test green.

Refs #2305.


adamflagg:
## Independent scan (triage-attack Part B) — floor 60, high effort

**Result: clean (0 new findings; builder's own 2 pre-fix findings independently re-verified as real and correctly fixed).**

**What I checked, beyond re-reading the diff:**
- Reproduced the builder's headline numbers directly against `pb_data/data-prod.db` for year 2026, replicating `build_household_journey`'s year-union and `fetch_household_family_attendees`'s family-only filter independently in Python: **480 journey rows, 89 carry the "No enrollment" chip, 33 of those 89 have an enrolled attendee on a family-or-adult weekend, 56 are genuine.** Matches the PR/issue body exactly. (Note: the campaign handoff's own figures — 465/84/32/52 — are the stale pre-drift numbers the builder already flagged in `partAGaps`; the current-tree numbers are the ones above.)
- Reproduced the registration-status distribution the same way: 479 rows, 424 enrolled / 39 cancelled / 10 incomplete / 4 waitlisted / 1 withdrawn / 1 applied (my count landed on 424 enrolled vs. the claimed 423 — a 1-row difference not worth chasing, doesn't change the structural claim: **55 of 479 rows carry a non-enrolled/non-none_on_file status slug**, confirmed).
- Verified `api/schemas/lodging.py:776` — `EnrollmentState = Literal["enrolled", "none_on_file"]` — exactly as cited; a follow-on publishing the column verbatim would indeed raise on the first slug-carrying row.
- Verified migration number 1500000166 is free on `main` (163/164 both taken, as the PR notes).
- Confirmed all three tables' create AND update branches set `enrollment_status` (adults/registrations/medical) — no silent-no-op gap.
- Ran the full `TestFamilyCamp*` suite locally: all pass. Mutation-checked `TestFamilyCampEnrollmentStatusRefusesAYearWithNoSessionsAtAll` by deleting the refusal branch — test goes red as expected, confirms it pins real behavior rather than trivially passing.
- Confirmed push: `origin/feature/fc-enrollment-status` == local HEAD (0/0 ahead-behind). CI Summary green, all 30 checks. `closingIssuesReferences` empty, confirmed deliberate (#2305 stays open for the journey-read half).
- CodeRabbit: opt-in, not requested — appropriately skipped per the campaign's guidance for a Go-sync/schema PR where a second independent reader (this scan) already covers it; the auto-check on this PR is the no-op `Review skipped: auto reviews disabled` notice, not a real review, and should not be read as coverage.

**Premise review (§5.5b):** this is a real bug, not a decision wearing a bug's clothes — `fetch_household_family_attendees` names a contradiction (adult-weekend attendees are person-grain and get filtered out by `session_type = "family"`), and the fix (a new column + derivation) doesn't itself change any staff-facing read. The number-meaning change is correctly identified as belonging to the *follow-on* PR that switches `build_household_journey` onto this column, not this one — confirmed by grep: no changes to `lodging_repository.py`, `lodging_roster_service.py`, `needsFit.ts`, or `HouseholdJourneyCard.tsx` in this diff.

**Holding per campaign condition 8**, not for anything found in this scan — see the closing report for the number-meaning detail.

PR #2493, opened 2026-08-19
feat(sync): capture cabin-value changes so each weekend's cabin becomes derivable

### LINKED ISSUE #2482: feat(sync): capture cabin-value change history so each weekend's cabin is derived, not guessed

adamflagg:
`Family Camp Cabin` (cm_id 218072) is a household-grain custom field holding exactly one value
per household per year — `family_camp_registrations` is unique on `(household, year)`. A
household attending two family weekends therefore has one slot for two answers: when staff type
the second weekend's cabin, the first weekend's is gone. The assignment ingest cannot tell which
weekend a string belongs to, so it files `ambiguous_session` rather than guess
(`pocketbase/sync/lodging_assignments_sync.go`, `// flag, do not guess (spec 3.6)`).

**Measured, production snapshot at 2026-08-18 22:11: 9 open `ambiguous_session` rows across 8
distinct households for 2026, `occurrences = 1` each, all 9 on `Family Camp Cabin`.** Ten
households are enrolled in 2+ family weekends this year; 8 have written a cabin value and are
those 9 rows. Nothing today keeps the overwritten string: the only surviving copy of one is
`lodging_ingest_issues.raw_value`, and only by accident, because `RawValue` happens to be part
of `dedupKey()`.

This adds the capture, and only the capture.

## What lands

**`pocketbase/pb_migrations/1500000165_lodging_value_history.js`** — a new append-only table,
modelled on `1500000057_attendee_status_history.js` with four deliberate departures documented
in its header: no `session` column (session is the unknown, derived on read, not stored);
`old_value` / `new_value` are `text` rather than `select`, because there are 88 distinct
hand-typed cabin strings and a select would reject the first unanticipated name and lose the
change at the moment it matters; a dual clock, `source_changed_at` (CampMinder's own
`last_updated`) beside `observed_at` (when the sync saw it), because the retroactive-entry case
needs both — 21 of 2025's household cabin values were last edited in December, after every 2025
weekend; and the create branch writes too, as `is_genesis`, because the first observed cabin is
a fact worth keeping.

`source_changed_at` is `text`, **not** the `date` the issue's shape sketch named. The column it
copies — `household_custom_values.last_updated` and its person twin — is declared `text` in
`1500000029`. A `date` field would have to parse a CampMinder string, and a parse failure
rejects the row and loses the change, which is the same failure mode the text-not-select call
avoids for the value itself. The issue body has been corrected in place. `observed_at` stays a
real `date`: it is our own clock.

The collection's reads are **admin-only**, deliberately tighter than
`attendee_status_history`. It stores `household_custom_values` /
`person_custom_values` data verbatim and both of those source collections are admin-only, so
publishing a copy at authed-read would hand every non-admin account cabin strings it cannot read
from the source. The forward-looking half matters more: the retention scope is a Go map, so a
later widening needs no migration and nobody would re-read the access rule.

**`pocketbase/sync/lodging_value_history.go`** — the writer, plus
`lodgingRetainedHistoryFields`, the retention scope. Cabin fields only:
`Family Camp Cabin` (household) and `Reportable Family Camp Cabin` (person). The
medical-adjacent lodging fields — bathroom, CPAP, infant, opt-out — are held out **not on the
merits** but because `api/routers/lodging.py` records a deliberate ruling that that surface has
no access log and that one was removed on purpose; reversing that as a side effect of a
cabin-attribution change would be the wrong way to reverse it. Widening later is an edit to that
Go map and needs no migration.

**The write hook**, at the `existing.GetString("value") != newValue` comparison in
`processHouseholdCustomFieldValue` and `processPersonCustomFieldValue` — the only point where
old and new values are simultaneously in scope, mirroring where `logStatusChange` sits in
`attendees.go`. Three details that are load-bearing:

- It fires on a **value** change only. The enclosing branch condition is `value changed OR
  last_updated changed`, and a bare `last_updated` bump is not a change to where anyone slept.
- `oldValue` is read *before* the `Set` loop overwrites it.
- It sits **after** the custom-values `App.Save`, which is how `DryRun` is honored
  structurally — a dry run returns before it and cannot reach it — and also means history never
  claims a change that failed to persist. Failures are logged and swallowed; capture is
  non-critical and must not fail a sync.

**`frontend/src/types/pocketbase-types.ts`** — regenerated. It is `@generated` from live schema
introspection and CI diffs it against a scratch database built from `pb_migrations/` alone; a
new collection makes it stale. No hand edits, no component or style touched.

**No published value moves.** Nothing reads the new table — verified by grep across Go, Python
and TypeScript; the only non-test references are the writer, the migration and the three
generated type-map lines. No cabin is assigned where the sync previously flagged
`ambiguous_session`, the attribution path is untouched, and all 9 open queue rows stay exactly
as they are. Deriving a weekend from these rows is a separate change and is not attempted here.

**One honest limit on what "captures every change" means.** The hook can only observe a change
the custom-values sync actually processes, and the existing fast path skips a record whose
`last_updated` is unchanged — so a value edited in CampMinder without its `last_updated` moving
is invisible here, exactly as it is invisible to `household_custom_values` itself today. That
is pre-existing behaviour this PR neither introduces nor fixes.

**The backfill is a floor, not a history — and there is no backfill in this PR at all.** Only
the current value and its `last_updated` survive in the source tables, so 2026's first
observations will all be stamped mid-May (64 of 75 values were last edited in 2026-05, 11 in
2026-04). A household with an April weekend and a September weekend reads "unknown for April".
The table earns from the next observed change forward. Do not read it as answering historical
questions on day one.

## Scan remediation

A floor-60 review found six issues; all six are fixed in this branch, none filed:

| # | Finding | Fix |
|---|---|---|
| 1 | The table published verbatim custom-values data at authed-read while both source collections are admin-only | reads are now admin-only, which also makes widening the Go retention map safe by default |
| 2 | `is_genesis` read as "no earlier value existed", which an orphan sweep followed by re-entry makes false | documented as "the first observation this table holds", in both the migration header and the Go type |
| 3 | With an empty `source_changed_at` the dedupe key degenerated and a cabin going A → B → A had its third observation silently dropped | `old_value` joins the dedupe predicate and the unique index; regression test added |
| 4 | The test fixture had no indexes, so the "idempotency contract" was never exercised | the fixture now carries the migration's three indexes verbatim |
| 5 | Assertions indexed `rows[1]` after sorting on a millisecond autodate that can tie | rows are selected by `is_genesis`, never by position |
| 6 | The migration header's row count disagreed with the evidence it cited | corrected to nine rows across eight households |

Separately, found before the scan and fixed the same way: the dedupe pre-check bound
`source_changed_at` and `new_value` as parameters, and a bound `""` does not match a stored
empty string in PocketBase — so a value with no `lastUpdated`, or a cleared cabin, re-attempted
its insert on every run and the unique index turned that into an error logged forever.

## Verification

Local, all exit 0: `go build ./...`; `go test ./...` in `pocketbase/` (full suite);
`golangci-lint run --config ../.golangci.yml` — **0 issues** (CI-only, not in pre-push);
`eslint` on the new migration; `scripts/dev/verify-migration-history.sh`;
`bash scripts/pre-push-verify.sh` — ALL CHECKS PASSED, including the frontend prettier, eslint,
tsc and vitest lanes.

The migration was applied to a scratch database built from `pb_migrations/` alone and the
produced schema inspected directly: the 500/200/100-char caps landed as declared rather than
falling back to PocketBase's 5000 default, and all three indexes exist. `1500000165` is the next
free number — the highest on `origin/main` is `1500000164`.

Nine tests, written failing first. Every one that was green on first write was
mutation-checked: dropping the retention-scope guard, the empty-genesis guard, the dedupe
pre-check, the value-changed guard, moving the genesis write ahead of the `DryRun` return, the
literal-`''` empty comparison, and removing `old_value` from both the predicate and the index
each turned its own test red, and each was restored.

The migration was also reverted and re-applied on a scratch database (`migrate down 1` then
`migrate up`) to confirm the down arm drops the collection cleanly and the up arm rebuilds it.

Closes #2482.

### DIFF
--- a/frontend/src/types/pocketbase-types.ts
+++ b/frontend/src/types/pocketbase-types.ts
@@ -56,6 +56,7 @@ export const Collections = {
   LodgingSlotMerges: 'lodging_slot_merges',
   LodgingUnitAliases: 'lodging_unit_aliases',
   LodgingUnits: 'lodging_units',
+  LodgingValueHistory: 'lodging_value_history',
   LodgingWriteIns: 'lodging_write_ins',
   LodgingWriteInsDraft: 'lodging_write_ins_draft',
   NormalizedMappings: 'normalized_mappings',
@@ -1212,6 +1213,21 @@ export type LodgingUnitsRecord<Tbeds = unknown> = {
   year: number
 }
 
+export type LodgingValueHistoryRecord = {
+  created: IsoAutoDateString
+  field_cm_id: number
+  household_cm_id?: number
+  id: string
+  is_genesis?: boolean
+  new_value?: string
+  observed_at?: IsoDateString
+  old_value?: string
+  person_cm_id?: number
+  source_changed_at?: string
+  source_field?: string
+  year: number
+}
+
 export type LodgingWriteInsRecord = {
   created: IsoAutoDateString
   id: string
@@ -1873,6 +1889,8 @@ export type LodgingUnitsResponse<Tbeds = unknown, Texpand = unknown> = Required<
   LodgingUnitsRecord<Tbeds>
 > &
   BaseSystemFields<Texpand>
+export type LodgingValueHistoryResponse<Texpand = unknown> = Required<LodgingValueHistoryRecord> &
+  BaseSystemFields<Texpand>
 export type LodgingWriteInsResponse<Texpand = unknown> = Required<LodgingWriteInsRecord> &
   BaseSystemFields<Texpand>
 export type LodgingWriteInsDraftResponse<Texpand = unknown> = Required<LodgingWriteInsDraftRecord> &
@@ -1993,6 +2011,7 @@ export type CollectionRecords = {
   lodging_slot_merges: LodgingSlotMergesRecord
   lodging_unit_aliases: LodgingUnitAliasesRecord
   lodging_units: LodgingUnitsRecord
+  lodging_value_history: LodgingValueHistoryRecord
   lodging_write_ins: LodgingWriteInsRecord
   lodging_write_ins_draft: LodgingWriteInsDraftRecord
   normalized_mappings: NormalizedMappingsRecord
@@ -2070,6 +2089,7 @@ export type CollectionResponses = {
   lodging_slot_merges: LodgingSlotMergesResponse
   lodging_unit_aliases: LodgingUnitAliasesResponse
   lodging_units: LodgingUnitsResponse
+  lodging_value_history: LodgingValueHistoryResponse
   lodging_write_ins: LodgingWriteInsResponse
   lodging_write_ins_draft: LodgingWriteInsDraftResponse
   normalized_mappings: NormalizedMappingsResponse
--- a/pocketbase/pb_migrations/1500000165_lodging_value_history.js
+++ b/pocketbase/pb_migrations/1500000165_lodging_value_history.js
@@ -1,0 +1,189 @@
+/// <reference path="../pb_data/types.d.ts" />
+/**
+ * lodging_value_history — append-only capture of cabin-value changes.
+ *
+ * Dark on arrival for every reader. This migration creates one empty table and
+ * moves nothing; no published value changes because of it, and the seven open
+ * `ambiguous_session` rows in the lodging work queue stay exactly as they are.
+ * The write hook in pocketbase/sync/lodging_value_history.go fills it going
+ * forward. Deriving a weekend from these rows is a separate change.
+ *
+ * ── WHY: ONE SLOT, TWO WEEKENDS ─────────────────────────────────────────────
+ *
+ * `Family Camp Cabin` (cm_id 218072) is a HOUSEHOLD-grain custom field holding
+ * exactly one value per household per YEAR — `family_camp_registrations` is
+ * unique on (household, year). A household attending two family weekends
+ * therefore has one slot for two answers: when staff type the second weekend's
+ * cabin, the first weekend's is gone. The assignment ingest cannot tell which
+ * weekend a string belongs to, so it files `ambiguous_session` rather than
+ * guess (`sync/lodging_assignments_sync.go`, "flag, do not guess (spec 3.6)").
+ *
+ * Keeping what the source overwrites is what makes "which cabin was in effect
+ * for THIS weekend" answerable from recorded observations instead of from a
+ * timestamp heuristic that picks one weekend per observation and leaves the
+ * other blank.
+ *
+ * ── WHY NOT lodging_assignment_history ──────────────────────────────────────
+ *
+ * It is already a per-change table and it cannot serve this. `ingestValue`
+ * returns on `attr.Reason != attrSingleSession` BEFORE every `writeHistory`
+ * call, so it is blind in exactly the ambiguous case that needs it, and its
+ * organising key is `session` — the unknown. This table sits UPSTREAM of
+ * attribution, which is why it has no session column at all.
+ *
+ * ── SHAPE: FOUR DELIBERATE DEPARTURES FROM attendee_status_history ──────────
+ *
+ * 1500000057_attendee_status_history.js is the existing lookback-table shape in
+ * this codebase and this is modelled on it. It differs in four places, each on
+ * purpose:
+ *
+ *   1. NO `session` COLUMN. It is `required: true` there. Here the session is
+ *      the unknown the table exists to make derivable, so requiring one would
+ *      beg the question. Session is derived on read.
+ *
+ *   2. old_value / new_value are `text`, NOT `select`. `attendee_status_history`
+ *      can enumerate its statuses; there are 88 distinct hand-typed cabin
+ *      strings for 218072 across all years. A select rejects the first
+ *      unanticipated name and loses the change at the moment it matters.
+ *      `lodging_assignment_history` already learned this and says so at
+ *      `sync/lodging_assignments_sync.go`: "History records the OBSERVED label
+ *      whether or not it resolved -- old_unit / new_unit are TEXT for exactly
+ *      this reason."
+ *
+ *   3. DUAL CLOCK. `source_changed_at` is CampMinder's own `last_updated` for
+ *      the value; `observed_at` is when this sync saw it. They are different
+ *      facts and the retroactive-entry case needs both: 21 of 2025's household
+ *      cabin values were last edited in December, after every 2025 weekend.
+ *
+ *   4. THE CREATE BRANCH WRITES TOO, as `is_genesis`. The first observed cabin
+ *      is a fact worth keeping; `attendee_status_history` logs only
+ *      transitions.
+ *
+ * ── source_changed_at IS text, NOT date ─────────────────────────────────────
+ *
+ * The issue's shape sketch said `date`. The source column it copies is
+ * `household_custom_values.last_updated` / `person_custom_values.last_updated`,
+ * and BOTH are declared `type: "text"` (1500000029, and the person twin). It is
+ * stored here verbatim, in the same type, for two reasons: a `date` field would
+ * have to parse a CampMinder string, and a parse failure would reject the row
+ * and LOSE the change — the same failure mode the text-not-select call above
+ * avoids; and a value that does not round-trip identically to the source column
+ * cannot be joined back to it. `observed_at` stays a real `date` because it is
+ * our own clock and is always well-formed.
+ *
+ * ── RETENTION SCOPE: CABIN FIELDS ONLY ──────────────────────────────────────
+ *
+ * The write hook retains only `Family Camp Cabin` (218072, household) and
+ * `Reportable Family Camp Cabin` (223823, person). The medical-adjacent lodging
+ * fields — bathroom, CPAP, infant, opt-out — are held out NOT on the merits but
+ * because `api/routers/lodging.py` records a deliberate ruling that that
+ * surface has no access log and that one was removed on purpose; reversing that
+ * as a side effect of a cabin-attribution change would be the wrong way to
+ * reverse it. The scope lives in `lodgingRetainedHistoryFields` in
+ * `sync/lodging_value_history.go`, so widening it later is a Go edit and needs
+ * no migration.
+ *
+ * ── THE UNIQUE INDEX IS THE IDEMPOTENCY CONTRACT ────────────────────────────
+ *
+ * (year, field_cm_id, household_cm_id, person_cm_id, source_changed_at,
+ * new_value) — re-running a sync re-observes the same change, and the table is
+ * append-only, so without this a weekly sweep would stack duplicates. The Go
+ * side checks for the row before inserting so a re-run is quiet rather than a
+ * swallowed constraint error; the index is the guarantee behind that check.
+ *
+ * Exactly one of household_cm_id / person_cm_id is set per row, the other stays
+ * 0 — which is what keeps two grains in one table without a nullable relation,
+ * and why both are in the index rather than one nullable column.
+ *
+ * ── PocketBase v0.23 ────────────────────────────────────────────────────────
+ *
+ * Field properties are DIRECT. The wrapper object some older migrations used is
+ * silently ignored in v0.23 and the field falls back to PocketBase's default
+ * cap, so `old_value` / `new_value` would quietly become 5000-char fields
+ * instead of the 500 declared below.
+ */
+
+// VERBATIM from the sibling history table (1500000057) and the lodging
+// migrations. Do not paraphrase — the permission is read off
+// `cached_permissions` as a dotted string, and the plausible alternative
+// matches nothing and denies every write SILENTLY.
+const AUTHED_READ = '@request.auth.id != ""';
+const ADMIN_ONLY = '@request.auth.is_admin = true';
+
+// Used for the skip guard and the down arm. The `name:` inside the
+// `new Collection({...})` below is a LITERAL rather than this constant, and
+// that is load-bearing: scripts/dev/verify-migration-history.sh's CHECK 2 finds
+// the collections a migration CREATEs by regexing `name: "..."` out of the up
+// arm, and a constant reads as no collection at all.
+const HISTORY = 'lodging_value_history';
+
+migrate(
+  (app) => {
+    // Idempotent create. A re-run is a no-op rather than a boot failure on
+    // "Collection name must be unique"; this is not licence to accept whatever
+    // shape is already present, only to make the second run quiet.
+    let exists = false;
+    try {
+      app.findCollectionByNameOrId(HISTORY);
+      exists = true;
+    } catch {
+      // Not present — create it below.
+    }
+    if (exists) {
+      return;
+    }
+
+    app.save(
+      new Collection({
+        type: 'base',
+        name: 'lodging_value_history',
+        listRule: AUTHED_READ,
+        viewRule: AUTHED_READ,
+        // Written by the sync, never by the UI — same posture as
+        // attendee_status_history.
+        createRule: ADMIN_ONLY,
+        updateRule: ADMIN_ONLY,
+        deleteRule: ADMIN_ONLY,
+        fields: [
+          // CampMinder reuses ids across years — every data table carries the
+          // year that disambiguates them.
+          { type: 'number', name: 'year', required: true, presentable: false, min: 2010, max: 2100, onlyInt: true },
+          // The custom field this observation is about (218072 / 223823).
+          // Matching is on cm_id, never the user-editable display name.
+          { type: 'number', name: 'field_cm_id', required: true, presentable: false, min: 1, max: null, onlyInt: true },
+          // Exactly one of the two is set; the other stays 0. Neither is a
+          // relation: cross-table relationships in this repo key on CampMinder
+          // ids so they survive a resync that recreates a PocketBase row.
+          { type: 'number', name: 'household_cm_id', required: false, presentable: false, min: 0, max: null, onlyInt: true },
+          { type: 'number', name: 'person_cm_id', required: false, presentable: false, min: 0, max: null, onlyInt: true },
+          // The display name, for humans reading a row. Documentation, not the
+          // matching key — field_cm_id above is the contract.
+          { type: 'text', name: 'source_field', required: false, presentable: false, min: 0, max: 200, pattern: '' },
+          // Empty on a genesis row, and empty is also a legitimate NEW value
+          // (staff clearing a cabin), which is why neither is required.
+          { type: 'text', name: 'old_value', required: false, presentable: false, min: 0, max: 500, pattern: '' },
+          { type: 'text', name: 'new_value', required: false, presentable: false, min: 0, max: 500, pattern: '' },
+          // CampMinder's clock, verbatim and untyped — see the header.
+          { type: 'text', name: 'source_changed_at', required: false, presentable: false, min: 0, max: 100, pattern: '' },
+          // Our clock.
+          { type: 'date', name: 'observed_at', required: false, presentable: false, min: '', max: '' },
+          { type: 'bool', name: 'is_genesis', required: false, presentable: false },
+          { type: 'autodate', name: 'created', required: false, presentable: false, onCreate: true, onUpdate: false },
+        ],
+        indexes: [
+          'CREATE UNIQUE INDEX idx_lvh_observation ON lodging_value_history ' +
+            '(year, field_cm_id, household_cm_id, person_cm_id, source_changed_at, new_value)',
+          'CREATE INDEX idx_lvh_household_year ON lodging_value_history (household_cm_id, year)',
+          'CREATE INDEX idx_lvh_person_year ON lodging_value_history (person_cm_id, year)',
+        ],
+      })
+    );
+  },
+  (app) => {
+    try {
+      app.delete(app.findCollectionByNameOrId(HISTORY));
+    } catch {
+      // Already gone — nothing to undo.
+    }
+  }
+);
--- a/pocketbase/pb_migrations/1500000165_lodging_value_history.js
+++ b/pocketbase/pb_migrations/1500000165_lodging_value_history.js
@@ -3,8 +3,9 @@
  * lodging_value_history — append-only capture of cabin-value changes.
  *
  * Dark on arrival for every reader. This migration creates one empty table and
- * moves nothing; no published value changes because of it, and the seven open
- * `ambiguous_session` rows in the lodging work queue stay exactly as they are.
+ * moves nothing; no published value changes because of it, and the nine open
+ * `ambiguous_session` rows in the lodging work queue (8 distinct households,
+ * production snapshot 2026-08-18 22:11) stay exactly as they are.
  * The write hook in pocketbase/sync/lodging_value_history.go fills it going
  * forward. Deriving a weekend from these rows is a separate change.
  *
@@ -59,6 +60,16 @@
  *      is a fact worth keeping; `attendee_status_history` logs only
  *      transitions.
  *
+ *      ⚠️ `is_genesis` MEANS "the first observation THIS TABLE HOLDS for this
+ *      key", NOT "no earlier value existed". Do not read it as the latter. If
+ *      CampMinder stops returning the field entry for a household, the orphan
+ *      sweep deletes the `household_custom_values` row and no history is
+ *      written; a later re-entry then arrives at the CREATE branch and is
+ *      stamped `is_genesis` with an empty `old_value`, even though a different
+ *      cabin really did precede it. A reader deriving a weekend's cabin must
+ *      treat a genesis row as a floor on what is known, exactly as the backfill
+ *      is a floor rather than a history.
+ *
  * ── source_changed_at IS text, NOT date ─────────────────────────────────────
  *
  * The issue's shape sketch said `date`. The source column it copies is
@@ -86,10 +97,22 @@
  * ── THE UNIQUE INDEX IS THE IDEMPOTENCY CONTRACT ────────────────────────────
  *
  * (year, field_cm_id, household_cm_id, person_cm_id, source_changed_at,
- * new_value) — re-running a sync re-observes the same change, and the table is
- * append-only, so without this a weekly sweep would stack duplicates. The Go
- * side checks for the row before inserting so a re-run is quiet rather than a
- * swallowed constraint error; the index is the guarantee behind that check.
+ * old_value, new_value) — re-running a sync re-observes the same change, and
+ * the table is append-only, so without this a weekly sweep would stack
+ * duplicates. The Go side checks for the row before inserting so a re-run is
+ * quiet rather than a swallowed constraint error; the index is the guarantee
+ * behind that check.
+ *
+ * `old_value` IS IN THE KEY, and dropping it would be a silent data-loss bug
+ * rather than a tidier index. `source_changed_at` is empty whenever CampMinder
+ * returns no `lastUpdated` (the transform only sets it when present and
+ * non-empty), and without `old_value` the key then degenerates to
+ * (year, field, household, person, '', new_value): a household whose cabin goes
+ * A -> B -> A has its third, genuine observation matched against the first and
+ * SILENTLY DROPPED, leaving B as the last recorded state. With `old_value` the
+ * two are ('', A) and (B, A) and both survive. Latent today — every cabin row
+ * in the production snapshot carries a `last_updated` — but the transform
+ * explicitly anticipates the empty case, so the key does too.
  *
  * Exactly one of household_cm_id / person_cm_id is set per row, the other stays
  * 0 — which is what keeps two grains in one table without a nullable relation,
@@ -103,11 +126,12 @@
  * instead of the 500 declared below.
  */
 
-// VERBATIM from the sibling history table (1500000057) and the lodging
-// migrations. Do not paraphrase — the permission is read off
-// `cached_permissions` as a dotted string, and the plausible alternative
-// matches nothing and denies every write SILENTLY.
-const AUTHED_READ = '@request.auth.id != ""';
+// VERBATIM from the sibling history table (1500000057). Do not paraphrase —
+// these rule strings are matched literally, and a plausible-looking alternative
+// denies every request SILENTLY.
+//
+// Every rule on this collection is ADMIN_ONLY, reads included; see the create
+// site below for why that is tighter than attendee_status_history on purpose.
 const ADMIN_ONLY = '@request.auth.is_admin = true';
 
 // Used for the skip guard and the down arm. The `name:` inside the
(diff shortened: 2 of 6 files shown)


adamflagg:
## ⛔ HELD for owner review — condition 7 (GUI grep) matched

Every other merge gate is satisfied, and the hold is mechanical rather than a doubt about the
change.

| Gate | Result |
|---|---|
| CI Summary actually ran and is green | ✅ pass, and **zero** non-pass / non-skipping checks |
| Floor-60 scan completed | ✅ six findings, **all fixed in-branch**, none filed |
| Push verified by the remote ref | ✅ `verify-pr-state.sh push` exit 0, head `cd6b9f25` |
| `closingIssuesReferences` | ✅ non-empty — `[2482]` |
| Diff is the item the campaign described | ✅ migration + write hook + table, capture only |
| **Condition 7 — no GUI/visual impact** | ❌ **MATCHED** |
| Condition 8 — a staff-facing number or label moves | ✅ no |

**Why it held.** The mechanical test matches one file:

```
gh pr diff 2493 --name-only | grep -E '^frontend/src/.*\.(tsx|ts|css)$' | grep -v '\.test\.'
frontend/src/types/pocketbase-types.ts
```

That file is `@generated` from live schema introspection — a new collection makes it stale and
CI diffs it against a scratch database built from `pb_migrations/` alone, so regenerating it is
mandatory, not optional. Its whole diff is 20 added lines: one `Collections` entry, one
`Record` type, one `Response` type and two map entries. No component, style, hook or route is
touched, and nothing imports the new type. The rule says to test it and not to judge it, so it
holds regardless of that.

**Condition 8, decided explicitly and stated with the counts.** The extra test for this item was
whether a cabin now gets ASSIGNED where the sync previously flagged `ambiguous_session`. It does
not. Nothing reads the new table — verified by grep across Go, Python and TypeScript, whose only
non-test hits are the writer, the migration and three generated type-map lines. The attribution
path in `lodging_assignments_sync.go` is untouched. Re-measured against the production snapshot
(newest queue row 2026-08-18 22:11): **9 open `ambiguous_session` rows across 8 distinct
households for 2026, all on `Family Camp Cabin`** — and all 9 stay exactly as they are. This
captures history without altering any published value.

**One call I made that the campaign did not rule on**, flagged rather than buried: the
collection's read rules are **admin-only**, not the authed-read that `attendee_status_history`
uses. The scan raised it, and the reasoning is that this table stores `household_custom_values` /
`person_custom_values` data verbatim while both source collections are admin-only — so
authed-read would publish cabin strings that 12 non-admin accounts cannot read from the source.
It is strictly tighter than the alternative and nothing reads the table yet, so it costs nothing
to revisit later.

Worktree `/home/adam/kindred/.worktrees/fc-cabin-capture`, branch `feature/fc-cabin-capture`,
Vite port 3160 / Caddy 8360.

adamflagg:
## Independent scan (triage-attack Part B)

Verified independently, not re-derived from the builder's report:

- **Build/tests/lint**: `go build ./...`, `go test ./sync/...` (full package, all `Lodging*`/`Cabin` tests including the 8 new ones), `golangci-lint run --config ../.golangci.yml` — all clean, 0 issues.
- **Row counts**: queried `pocketbase/pb_data/data-prod.db` directly (a snapshot newer than the one cited in the PR body — `last_seen` now `2026-08-19 06:19:48`) — still **9 open `ambiguous_session` rows / 8 distinct households**, `Family Camp Cabin`, year 2026. Matches the body exactly; the queue moved but the count didn't.
- **No reads of the new table**: grepped Go/Python/TS myself — only hits outside tests are the writer (`lodging_value_history.go`), the migration, and the three generated type-map lines in `pocketbase-types.ts`. Nothing derives a session or publishes a value from it.
- **Placement claims checked against the actual diff context** (not just the PR body's excerpt): `oldValue` is read before the `Set` loop; the history call sits after `App.Save` and after the `DryRun` early return in both files; the value-changed guard (`oldValue != newValue`) is separate from the persisted-changed condition (`value OR last_updated`), so a bare `last_updated` bump doesn't write history.
- **ADMIN_ONLY rule**: literal string in the new migration (`@request.auth.is_admin = true`) matches `1500000057_attendee_status_history.js`'s `createRule` verbatim, as claimed.
- **Migration numbering**: `1500000165` is confirmed the next free number on `origin/main` (highest is `1500000164`).
- **`closingIssuesReferences`**: `[2482]`, confirmed via GraphQL. `verify-pr-state.sh push` exits 0.
- **CI Summary**: pass, confirmed by name (run `32232783995`), along with every other lane including Go Lint, Frontend Lint, Migration Smoke Test, Lodging Name Guard.
- **Scan-remediation commit** (`cd6b9f25`): read the diff directly — all six floor-60 findings plus the pre-scan dedupe-literal-`''` bug are genuinely fixed (admin-only reads, `old_value` in both the dedupe predicate and the unique index with a regression test for A→B→A, real indexes in the test fixture, `is_genesis`-based row selection instead of position, corrected row count in the header). No re-introduction, no partial fix.
- **Frontend types diff**: read all 20 added lines directly — one `Collections` entry, one `Record` type, one `Response` type, two `CollectionRecords`/`CollectionResponses` map entries. Nothing else touched; no component or hook imports the new type.

**No new findings from this pass.** Nothing discarded either — the builder's own scan already covered the surface area I'd have flagged.

**Condition 8 (does a published number/label move?)**: No. Nothing reads `lodging_value_history`; the `ambiguous_session` queue count is unchanged (9/8, verified above) and the attribution path in `lodging_assignments_sync.go` is untouched. This is capture-only.

**Merge decision: HELD**, on condition 7 alone — `gh pr diff --name-only` matches `frontend/src/types/pocketbase-types.ts`, and the standing rule is to detect that mechanically and hold regardless of confidence, not to judge the diff and wave it through. I read the diff (above) and agree with the builder that it's a clean, mandatory `@generated`-type regeneration with no component/style/route impact — but that's exactly the kind of confidence the rule says not to act on. Holding for your eyes; merge with `gh pr merge 2493 --squash` once you've glanced at it.

PR #2462, opened 2026-08-18
feat(sync): resolve the accommodation narrative into needs_fridge and publish fridge coverage

adamflagg:
## The defect

`needs_accommodation` is a **gate question, not a need**. CampMinder asks a plain Yes/No — `Housing Accommodation` on the camper form, `Housing Accomodation` (one `m`) on the adult twin — and the substance of the request lands in a separate free-text field the product never read. So the board could match a family to a cabin on exactly **two axes** (private bathroom, power) while the registry carried columns that answer several more, and the asks that need them sat in prose nobody parsed.

Measured against the production snapshot, 2026:

| | |
|---|---|
| `family_camp_registrations` rows | 464 |
| ...carrying `needs_accommodation` | **42** |
| ...whose narrative names a refrigerator | **6** |
| `lodging_units` carrying `has_fridge` | **12 of 118** |
| ...also carrying `has_shared_fridge` | 4 |
| carrying `has_shared_fridge` WITHOUT `has_fridge` | **0** |

2026 is only 16% placed, so **6 is the shape of the demand, not a rate**.

Before this, `needs_fridge`, `fridge_coverage` and `needsFridge` were zero hits repo-wide, `has_shared_fridge` was absent from `api/` entirely, and `NEEDS_DIMENSIONS` held exactly one entry.

## The change

**Sync (`family_camp_derived.go`)** — a new `needs_fridge` boolean on `family_camp_registrations` (migration `1500000163`), OR'd across the **raw per-person** narrative values, exactly as the gate arm already ORs. Not from the collapsed `family_camp_medical.accommodation_explain`: that column's first-non-empty flatten has already discarded 13 of 43 sibling narratives before anything can read them.

Routing now goes through one shared list, `accommodationExplainFieldNames`, which `processMedical` also uses — two copies of a name-keyed route drift the moment a generation is added, which is how the Adult twin came to be read in one of the two and not the other. The list carries the two live names plus two **defensive successors**: admission is by `cm_id` and survives a CampMinder rename, but routing is by display name and does not, so a rename silently stops population with no error. CampMinder's one-`m` misspelling of "Accomodation" is real — it is how the live adult gate is spelled — so it is the likeliest shape a successor takes.

[...]

Issue #2224, opened 2026-08-10
feat(sync,frontend): resolve the accommodation free text into needs the registry can actually answer


adamflagg:
## ⚠️ CORRECTED 2026-08-16 — phase D triage. RULED, and the defect is in the CODE, not the data.

Verified against `main` @ `a946565a` and the **production** snapshot.

### 1. The ruling (owner, 2026-08-15)

**A shared fridge IS a fridge** — `has_shared_fridge` reads **`fits`**, not `partial`. This generalises: **a
child column may never downgrade its parent's verdict** (`has_tub` ⊂ bathroom, `has_kitchenette` ⊂
`has_kitchen`).

### 2. ⛔ The registry data does NOT violate the rule — so this is a verdict-logic fix, not a data fix

Measured against production:

- `has_shared_fridge = 1 AND has_fridge = 0` → **0 rows**
- `has_tub = 1 AND bathroom = ''` → **0 rows**
- `has_kitchenette = 1 AND has_kitchen = 0` → **0 rows** (only 4 units carry kitchenette at all, all of which
  already carry kitchen — see #2390)

**Every violation lives in the matching code's fits/partial verdict logic.** That changes the shape of the fix
and of its test: the test must exercise the verdict function, not assert something about registry contents.
Enumerate every child ⊂ parent pair the verdict logic touches — that set is the PR's scope.

### 3. Sequencing — ⚠️ **REVERSED 2026-08-17. THIS ISSUE'S BACKEND NOW GOES FIRST.**

**The 2026-08-16 ruling that chained this behind #2072 is superseded.** #2072's staff conversation —
the thing it was HELD on — happened on **2026-08-17** and ruled the glyph set to
**bathroom · power · fridge**, dropping generic `needs_accommodation`. **Fridge is this issue's
output**, so #2072 now depends on *this*, not the other way round.

Verified against `main` @ `1797f9bd`:

- `needs_fridge` / `needsFridge` / `fridge_coverage` — **zero hits repo-wide** (`.py/.go/.ts/.tsx`)
- `has_shared_fridge` — **absent from `api/` entirely**
- `NEEDS_DIMENSIONS` (`frontend/src/components/weekend/needsFit.ts:91-102`) — **exactly one entry**,
  `needs_power`

So #2072 cannot ship its ruled three-glyph set until this issue lands.

**Break the cycle here.** The backend half — derive `needs_fridge` in Go from the raw 274058 values,
[...]

PR #2494, opened 2026-08-19
feat(sync,api,frontend): resolve step-free access into needs_step_free and publish ramp coverage

### LINKED ISSUE #2438: feat(sync,api,frontend): resolve step-free access into needs_step_free and publish ramp coverage

adamflagg:
## The defect

The registry has recorded step-free access since migration `1500000131` and the
product never read it. `lodging_units.has_ramp` is a **three-value select** —
`yes` / `no` / `partial`, blank = NOT ASSESSED — and the only reference to it
anywhere under `frontend/src/` was the generated type. It was absent from
`api/` and `bunking/` entirely.

It also *looked* empty, and that is why it stayed unread. A **boolean** read of
a three-value select reports `has_ramp` true on **0 of 118 units**, which is
the figure a prior pass acted on. The real distribution, measured over all 118
rows of `pocketbase/pb_data/data-prod.db`:

| `has_ramp` | `is_accessible` false | `is_accessible` true | total |
|---|---:|---:|---:|
| *(blank — not assessed)* | 104 | 0 | **104** |
| `no` | 4 | 0 | **4** |
| `partial` | 5 | 0 | **5** |
| `yes` | 3 | 2 | **5** |

**14 staff assessments**, which the boolean read erased. Neither column
determines the other: `has_ramp = 'yes'` splits 2 `is_accessible` true / 3
false, and the 4 explicit `no`s plus 5 `partial`s are invisible to
`is_accessible`, which reads them identically to the 104 nobody has looked at.
No fold in either direction is information-preserving.

Nobody had measured the **demand** side at all. Measured at the exact grain
`fridgeKeywords` uses — both narrative fields on `family_camp_medical`, 2026,
household grain, case-insensitive substring:

| signal | 2026 households |
|---|---:|
| carrying **any** narrative text | 86 |
| **fridge** (the shipped `fridgeKeywords` verbatim) | **6** |
| **mobility / step-free** (this PR's surface) | **14** |
| both | 1 |

**The mobility signal is more than twice the fridge signal that justified
shipping a whole need dimension in #2462.** 2026 is only ~16% placed, so 14 is
the SHAPE of the demand, not a rate.

## The change

`needs_step_free` (demand) + `ramp_coverage` (supply), mirroring
`needs_fridge` + `fridge_coverage` one dimension over.

**Sync (`family_camp_derived.go`)** — a new `needs_step_free` boolean on
`family_camp_registrations` (migration `1500000167`), OR'd across the **raw
per-person** narrative values, never the collapsed `family_camp_medical`
columns whose first-non-empty flatten has already discarded sibling narratives.
`bathroomExplainFieldNames` is lifted out of `processMedical`'s local slice to
package level, so the sentence-storing reader and the boolean-deriving reader
share ONE routing list — the discipline `accommodationExplainFieldNames`
already follows.

**⚠️ It routes BOTH narratives, and copying the fridge routing verbatim would
have lost more than a third of the signal.** `needs_fridge` reads the
accommodation narrative alone, which sufficed because 0 of the 6 fridge
households narrate elsewhere. Mobility does not split that way:

| narrative field | fridge households | mobility households |
|---|---:|---:|
| `accommodation_explain` only | 6 | 6 |
| `bathroom_explain` only | 0 | 5 |
| both | 0 | 3 |
| **union** | **6** | **14** |

A family explaining why it needs a private bathroom is often explaining that
someone cannot walk to the shared one. The two routes stay separate for the
same measured reason, and a test pins that the bathroom narrative does not set
`needs_fridge`.

**⚠️ It is NOT accommodation-gated.** All 6 fridge households carry
`needs_accommodation`; only **11 of the 14** mobility households do. So
`needs_step_free` joins the has-some-data guard that decides whether a
registration row is written at all — without that, those 3 rows are dropped
before they are stored, and the flag is missing for exactly the households
nothing else records.

The keyword surface prefers **recall over precision**, for the reason
`fridgeKeywords` states: the flag is advisory, it hatches a card and never
refuses a drop. Contributions, so the surface can be argued rather than
inherited: `walk` finds 10 (a substring, so it also catches *walking* /
*walkway*, and every one of the ten reads as a genuine step-free ask on
inspection); `mobilit` 2, one of them new; `knee` and `hip` 3 each, **three of
them new**; `crutch` 1, none new. The rest match 0 today and are kept because
each is the plain word for an ask `has_ramp` answers.

`knee` / `hip` are diagnosis words in an otherwise ask-shaped list, and they
are admitted on evidence rather than shape: **every knee/hip mention in the
corpus, in every year, is a mobility limitation** — a joint replacement, a
post-surgical recovery, or a stated limit on how far someone can walk. Zero
false positives. (Aggregate only: the narratives are PHI and are not quoted,
here or in the code.)

**`hip`, `cane` and `ramp` match as WHOLE WORDS**, not as substrings. Each is
short enough to sit inside a common unrelated word — *relationship*,
*shipping*, *hurricane*, *cramp* — and firing on those is not the near-miss
recall-over-precision buys, it is a household that asked for nothing of the
kind. The other eleven keywords stay substrings, because `walk` has to catch
*walking* and *walkway*. Both false-positive words are pinned as negative
cases. On the 2026 snapshot the boundary rule changes no count: still 14
households, 6 / 5 / 3 across the narrative fields, 11 gated.

One candidate is **deliberately excluded**: bare `close to`, 5 households. It
is **proximity**, not step-free access — a different ask, which the registry
answers with `map_x`/`map_y` and `near_bathhouse` rather than with `has_ramp`,
so a household flagged on it would be hatched against a column that cannot
speak to what they asked for. The mobility-bearing one of the five is caught by
`knee`; the three that remain uncaught are a child who needs to be near
activities, a family recalling a past illness, and a bathroom-proximity request
`needs_private_bathroom` already carries. None is a step-free ask.

**API** — `LodgingUnitSummary` gains `has_ramp` (the row's own fact) and
`ramp_coverage` (the resolution); `AccessibilityFlagSummary` gains
`needs_step_free`. `_resolve_ramp_coverage` runs over the SAME leaf walk as
power and fridge; the walk is now parameterised on the vocabulary a room
answers in as well as on the field that receives the verdict, so there is still
exactly one traversal of the tree.

**`ramp_coverage` carries a fifth grade the boolean amenities have no word
for**, and settling it was this issue's one piece of real design work:

| grade | meaning |
|---|---|
| `all` | every answering room is fully step-free |
| `some` | at least one is, but not all |
| `partial` | **no** room is, but at least one has a qualified ramp |
| `none` | every answering room answered `no` |
| `unknown` | nothing answers — blank, unconfirmed, or no active room left |

`partial` folds into neither neighbour. Folding it into `none` would re-erase 5
of the 14 assessments in the exact direction the select exists to prevent — and
3 of those 5 carry the ramp qualifier in `notes`, which is the record of
somebody having gone and looked. Folding it into `some` would make the order
lie: "no room is step-free but one has a ramp with a lip" is a weaker claim
than "some rooms are step-free".

**⚠️ Blank resolves `unknown`, never `none`.** 104 of 118 units are blank, so
reading blank as "no ramp" would mark almost the whole registry
step-free-hostile on evidence nobody recorded. An unrecognised string rails to
blank too — **not** as a defence against the registry loader, which cannot
produce one (PocketBase validates the select on save), but against a later
migration widening the value list and against a record built before the column
existed. A grade this code has never heard of must read as "not assessed",
never fall through to a claim.

**Frontend** — one further `NEEDS_DIMENSIONS` entry, reading `ramp_coverage`
and **never** the raw `has_ramp`. That is not only the container trap the other
two entries spell out: `has_ramp` is a STRING, so `'no'` is **truthy**, and any
consumer filtering on truthiness renders "step-free" on the four cabins staff
assessed as explicitly having no ramp. A test pins both directions.

`someIs: 'unmet'`, **not** the fridge reading. This takes the `is_accessible`
shape the module doc already reasons out: a building advertising two step-free
rooms out of ten invites precisely the placement that lands in one of the other
eight. What separates it from fridge is the shared-fridge ruling's own logic in
reverse — a fridge one room over is still usable, a ramp one room over is not.

This dimension cost one object literal **and one branch** in `resolveNeedsFit`,
which is the honest counter-example to that module's "a second dimension is a
further entry in this array and nothing else". The extra branch is a property
of the REGISTRY — a three-value supply column — not a sign the first dimension
was built wrong, and the module doc now says so. Still no new glyph, no new
colour, no new chip.

## ⚠️ Post-merge step, required

**`family_camp_derived` must be RE-RUN for 2026 or `needs_step_free` stays
empty on every row.** The migration is additive and computes nothing — the
input is a person-partition custom value that needs the household collapse, so
no migration could. Verified: the migration applies cleanly to a 771 MB
production-shaped database, preserves all 3,932 registration rows, and leaves
the column at 0 until the sync runs.

## ⚠️ 2026-forward only. There is no history to recover

`family_camp_medical` carries narrative text for **2026 alone** — 86 households
— and every prior year is empty. So this is a 2026-forward signal, exactly as
`needs_fridge` is. **A backfill would recover nothing**, and a prior-year
`false` is the absence of an input, never a measured "did not ask". Do not read
one as a false negative.

## PHI containment

The mobility narrative names individuals and their conditions. **Only the
boolean leaves the sync layer.** The sentences stay in `family_camp_medical`,
which is admin-gated on all five rules and absent from every export config. A
test pins that no registration text column ever receives the narrative, and
mutation-checking it turns it red.

## Deliberately NOT done

- **No rendering work.** No new glyph, no new colour, no new chip, no admin
  control for `has_ramp`. #2072 owns the card's visual treatment, and its ruled
  glyph set is bathroom · power · fridge — step-free is not in it. Adding
  `has_ramp` to `AMENITY_FLAGS` would have hit the truthiness inversion above;
  it needs its own three-state control, which is not this PR.
- **`is_accessible` is untouched.** #2327's ruling — accessible draws only when
  true — stands, and the two columns stay independent because neither
  determines the other.
- **`rosterAttention` and `placementCandidates` are not wired**, for the same
  reason #2224 left fridge out of them: both read a RAW unit field and neither
  of these needs can, and wiring them puts new words in front of staff. Their
  divergence comments are updated to name the third need rather than left
  claiming the tables differ by one entry.
- **No backfill**, per the section above.

## Verification

- `bash scripts/pre-push-verify.sh` → **ALL CHECKS PASSED** (exit 0), including
  `golangci-lint` and `eslint`.
- Migration `1500000167` applied to a scratch database built from
  `pb_migrations/` alone, reverted with `migrate down`, and re-applied — the
  column is gone after the down and back after the up, and
  `scripts/dev/verify-family-camp-request-fields.sh` reports OK at each step.
  That script now covers `needs_step_free`, so a dropped column is caught.
- Also applied to a 771 MB production-shaped dev database: OK, 3,932
  registration rows intact.
- `pocketbase-types.ts` regenerated from the scratch migrated database (not
  from local dev `data.db`); `types.gen.ts` regenerated from the OpenAPI
  schema. Both diffs are additive only.
- Every new test was written failing first and run red before implementation,
  then mutation-checked. Dropping the bathroom route, dropping
  `needs_step_free` from the survival guard, dropping `walk` from the keyword
  surface, dropping `Bathroom-Yes` from the shared routing list, letting blank
  answer as a value instead of `None`, dropping the resolver call, unrailing an
  unrecognised `has_ramp`, folding `partial` into `none`, dropping the flag
  read, flipping `someIs` to `partial`, mapping the `partial` grade to `fits`,
  and pointing the dimension at the raw `has_ramp` each turned the intended
  test red, and only that test.

## Review record

Scanned on-branch and reviewed by CodeRabbit. Four findings in this PR's own
code, all fixed here; three findings against sibling commits already merged to
`main` were out of scope and left alone.

1. **PHI in source comments (major).** The keyword-surface comment quoted three
   health narratives verbatim from the production corpus to justify admitting
   `knee` / `hip`, and two test fixtures were verbatim corpus text. The repo
   rule is absolute — no real personal information in code, tests, comments,
   docs, commits or PRs — so the justification is now a categorical description
   with no quoted text, the fixtures are clearly synthetic, and the branch was
   **force-pushed as a single squashed commit** so the quoted text is in no
   reachable commit. The PR body and the audit comment on the issue were
   scrubbed too.
2. **`hip` matched as a substring (minor, correctness).** It fired on
   *relationship* and *shipping*; `cane` on *hurricane* and `ramp` on *cramp*.
   Those three now match as whole words while the other eleven keywords stay
   substrings. All four false-positive words are pinned as negative cases, and
   the boundary rule changes no count on the snapshot.
3. **A false premise in a docstring.** The railing on an unrecognised `has_ramp`
   was justified by "PocketBase stores it as a plain string and the loader
   writes whatever the JSON holds" — wrong: migration `1500000131` declares a
   `select` with a fixed value list and PB validates it on save. The railing is
   still right, for two reasons that are true (a later migration widening the
   list, and a record built before the column existed), and the docstrings, the
   test's own justification and this body now say so.
4. **A stale docstring this PR created.** `fetch_family_camp_registrations`
   still said **five** housing flags after this PR added a sixth — in the very
   docstring that tells callers to read the columns rather than re-derive them.

## Issue body accuracy

The issue body's own measurement said 10 mobility households and 84 with any
narrative. Re-run on the same snapshot at the same grain, this branch measures
**11** on the body's narrower surface and **86** narrative households; the body
flagged 10-vs-11 as unresolved and said to re-run rather than trust it, which
is what happened.

The shipped surface is wider than the body's, at **14**, and that is a
deliberate choice the body invited ("pick the surface deliberately … take
recall over precision"). It also corrects two figures the body gives for the
narrower surface: the bathroom-only split is **5**, not 4, and its claim that
the `close to` households which are step-free asks are already caught by `walk`
is **false** — one of them is caught only by `knee`. The gate split the body
predicted (3 ungated) reproduces exactly at the wider count: 11 of 14 gated.
Nothing about the ruling turns on any of it; the code comments carry the
re-measured figures.

The migration number in the body (`1500000165`) was contested between three
issues and re-derived at branch time; `1500000167` is the first free number on
this branch.

Closes #2438.

### DIFF
--- a/api/schemas/lodging.py
+++ b/api/schemas/lodging.py
@@ -115,6 +115,30 @@ Shareability = Literal["unknown", "shareable", "single_party"]
 # means "nobody has said", not "there is no power". See `amenity_coverage`.
 AmenityCoverage = Literal["all", "some", "none", "unknown"]
 
+# The registry's own step-free assessment, mirrored from the `has_ramp` select
+# added by migration 1500000131 (kindred#2438). THREE values plus blank, and
+# the blank is the load-bearing one: it means NOT ASSESSED, which is neither
+# "yes" nor "no". That migration's comment says why it is not a bool -- "a bool
+# maps every unassessed cabin to false, which asserts 'no ramp' about cabins
+# nobody has looked at" -- and reading it as one reports 0 of 118 units while
+# erasing all 14 staff assessments.
+RampAssessment = Literal["", "yes", "no", "partial"]
+
+# How much of a slot is step-free, resolved over its LEAF descendants
+# (kindred#2438). FIVE grades rather than `AmenityCoverage`'s four, because the
+# supply column is three-valued and a qualified ramp is a real answer with
+# nowhere to go in a boolean grain:
+#
+#   all      every answering room is fully step-free
+#   some     at least one is, but not all
+#   partial  NO room is, but at least one has a qualified ramp
+#   none     every answering room answered `no`
+#   unknown  nothing answers -- blank, unconfirmed, or no active room left
+#
+# See `ramp_coverage` in `api/services/lodging_rules.py` for why `partial`
+# folds into neither of its neighbours.
+RampCoverage = Literal["all", "some", "partial", "none", "unknown"]
+
 # The STAFF-OWNED weekend status, from lodging_session_status (1500000142).
 # Unlike every other vocabulary in this module it mirrors no Go ingest, because
 # there is no ingest: CampMinder's Sessions API has no status concept at all,
@@ -229,6 +253,29 @@ class LodgingUnitSummary(BaseModel):
     # units carry `has_fridge`, four of those also carry `has_shared_fridge`,
     # and none carries shared without the parent.
     fridge_coverage: AmenityCoverage = "unknown"
+    # The registry's own step-free fact about THIS ROW, published beside its
+    # resolution exactly as `has_power` and `has_fridge` are (kindred#2438).
+    #
+    # ⚠️ NOT the field a drop is judged against -- `ramp_coverage` below is,
+    # and reading this instead reintroduces the container trap AND the
+    # truthiness trap at once: `has_ramp` is a STRING, so `'no'` is truthy, and
+    # a consumer filtering on it renders "step-free" on the four cabins staff
+    # assessed as explicitly having no ramp. Blank means NOT ASSESSED (104 of
+    # 118 production rows) and is never coerced to "no".
+    has_ramp: RampAssessment = ""
+    # The step-free twin of `power_coverage` and `fridge_coverage`, resolved
+    # over the same leaf walk and defaulting to "unknown" for the same reason:
+    # a payload built without the resolution pass must not claim an unmet need.
+    # Fourteen of 118 production units carry an assessment (5 yes / 5 partial /
+    # 4 no), so almost every unit resolves "unknown" today -- which is the
+    # honest answer, not a gap.
+    ramp_coverage: RampCoverage = "unknown"
+    # ⚠️ INDEPENDENT of `has_ramp` above, measured: `has_ramp = 'yes'` splits 2
+    # `is_accessible` true / 3 false, and the 4 explicit `no`s plus 5
+    # `partial`s are invisible to this flag, which reads them identically to
+    # the 104 rows nobody has looked at. No fold in either direction is
+    # information-preserving (kindred#2438), and #2327's ruling -- accessible
+    # draws only when TRUE -- is unchanged by the ramp dimension.
     is_accessible: bool = False
     is_confirmed: bool = False
     is_active: bool = False
@@ -532,6 +579,24 @@ class AccessibilityFlagSummary(BaseModel):
     # board hatches a unit for this and never dims one -- see `needsFit.ts`,
     # where the mark it feeds changes `background-image` and nothing else.
     needs_fridge: bool = False
+    # kindred#2438. The second need resolved out of the housing narrative, and
+    # the one the registry has always been able to answer: `lodging_units`
+    # has carried `has_ramp` since migration 1500000131 and nothing read it.
+    #
+    # ⚠️ ROUTED DIFFERENTLY from `needs_fridge` above, and the difference is
+    # measured. Fridge reads the accommodation narrative alone; this reads the
+    # BATHROOM narrative too, because 5 of the 14 mobility households on the
+    # 2026 snapshot narrate only through that field and 3 through both -- a
+    # family explaining why it needs a private bathroom is often explaining
+    # that someone cannot walk to the shared one.
+    #
+    # ⚠️ NOT GATED on `needs_accommodation`. All 6 fridge households are gated;
+    # only 11 of the 14 mobility households are.
+    #
+    # DERIVED IN THE SYNC LAYER and ADVISORY, on the same terms as the flag
+    # above: only the boolean crosses into this payload, and the mark it feeds
+    # hatches a unit card rather than refusing a drop.
+    needs_step_free: bool = False
     # True when the family said they can only attend WITH the accommodation
     # in place: `FAM CAMP-Opt Out VIP` = "Yes, please register regardless of
     # cabin type" means they will come either way, so the need is a warning;
--- a/api/services/lodging_repository.py
+++ b/api/services/lodging_repository.py
@@ -845,9 +845,10 @@ class LodgingRepository:
 
         Carries the ingest-derived request layer -- share_cabin_gate,
         wants_near / wants_with / wants_similar_ages, request_text -- and the
-        five narrative-free housing flags -- needs_private_bathroom,
-        needs_power, needs_accommodation, has_infant, and needs_fridge since
-        kindred#2224. Read those columns; do not re-derive them from
+        six narrative-free housing flags -- needs_private_bathroom,
+        needs_power, needs_accommodation, has_infant, needs_fridge since
+        kindred#2224, and needs_step_free since kindred#2438. Read those
+        columns; do not re-derive them from
         share_cabin_preference / shared_cabin_modes_raw, which are the raw
         profile values kept for provenance.
 
--- a/api/services/lodging_roster_service.py
+++ b/api/services/lodging_roster_service.py
@@ -24,7 +24,6 @@ from api.constants.lodging import INFANT_BED_EXEMPT_MONTHS, is_attending_adult_n
 from api.schemas.lodging import (
     MEDICAL_NARRATIVE_FIELD_NAMES,
     AccessibilityFlagSummary,
-    AmenityCoverage,
     EffectiveBathroom,
     HouseholdJourneyResponse,
     HouseholdJourneySession,
@@ -35,6 +34,7 @@ from api.schemas.lodging import (
     PartyAdult,
     PartyChild,
     ProximityKind,
+    RampAssessment,
     RequestTextBlock,
     RequestTextEntry,
     RosterCounts,
@@ -61,6 +61,7 @@ from api.services.lodging_rules import (
     container_bathroom,
     effective_bathroom,
     is_family_available,
+    ramp_coverage,
     request_text_authorship,
     request_text_source_order,
     unit_capacity,
@@ -135,6 +136,29 @@ def _i(record: Any, field: str, default: int = 0) -> int:
         return int(value)
     except TypeError, ValueError:
         return default
+
+
+def _ramp_assessment(value: str) -> RampAssessment:
+    """Rail a raw `has_ramp` string to the select's own vocabulary.
+
+    NOT a defence against the registry loader: migration 1500000131 declares
+    `has_ramp` as a PocketBase `select` with `values: ['yes','no','partial']`,
+    and PB validates that on save, so `registry.go`'s write of a typo'd value
+    fails the save rather than persisting it. This rails two OTHER directions,
+    both real. A later migration may WIDEN the value list -- that is how a
+    select grows -- and a grade this code has never heard of must read as NOT
+    ASSESSED rather than fall through `ramp_coverage`'s chain to `none`, which
+    is a claim. And `_s` is total over a record that lacks the attribute
+    entirely, which is what a summary built before the column existed looks
+    like.
+
+    Blank already means NOT ASSESSED -- 104 of the 118 production rows are
+    blank -- and coercing either case to "no" is the inversion the select
+    exists to prevent.
+    """
+    if value in ("yes", "no", "partial"):
+        return cast(RampAssessment, value)
+    return ""
 
 
 def _b(record: Any, field: str) -> bool:
@@ -1055,7 +1079,9 @@ def _resolve_power_coverage(units: list[LodgingUnitSummary], index: _BathroomInd
     pointing at the pre-resolution copies, which is exactly the kind of drift
     the one-index-per-call rule above exists to prevent.
     """
-    _resolve_amenity_coverage(units, index, answer=lambda room: room.has_power, target="power_coverage")
+    _resolve_amenity_coverage(
+        units, index, answer=lambda room: room.has_power, grade=amenity_coverage, target="power_coverage"
+    )
 
 
 def _resolve_fridge_coverage(units: list[LodgingUnitSummary], index: _BathroomIndex) -> None:
@@ -1090,18 +1116,60 @@ def _resolve_fridge_coverage(units: list[LodgingUnitSummary], index: _BathroomIn
         units,
         index,
         answer=lambda room: room.has_fridge or room.has_shared_fridge,
+        grade=amenity_coverage,
         target="fridge_coverage",
     )
 
 
-def _resolve_amenity_coverage(
+def _resolve_ramp_coverage(units: list[LodgingUnitSummary], index: _BathroomIndex) -> None:
+    """Fill in every unit's `ramp_coverage` in place — kindred#2438.
+
+    The third resolver over the one leaf walk, and every rule
+    `_resolve_power_coverage` established applies here unchanged: a container's
+    registry row describes the CONTAINER, an unconfirmed row means "nobody has
+    said", a deactivated room does not answer for its building, and a container
+    with no active room left reports `unknown` rather than falling back to its
+    own flag. One of the 14 production assessments IS on a container, and it is
+    ignored for exactly that reason.
+
+    TWO things are this function's own, and both come from `has_ramp` being a
+    three-value select rather than a bool:
+
+    1. A room answers in a THREE-VALUE vocabulary, so the verdict is graded by
+       `ramp_coverage` rather than `amenity_coverage` and carries a fifth
+       grade, `partial`. See that function for why `partial` folds into neither
+       `none` nor `some`.
+    2. BLANK IS NOT `no`. `_resolve_power_coverage`'s `None` case covers only
+       the unconfirmed ROW; here the field itself can be unanswered on a
+       confirmed row, and 104 of 118 production units are. So blank maps to
+       `None` — not assessed — and the building reports `unknown`. Reading it
+       as `no` would mark almost the whole registry step-free-hostile on
+       evidence nobody recorded, which is the inversion migration 1500000131
+       made the column a select to prevent.
+
+    An unrecognised string maps to `None` too — see `_ramp_assessment` for the
+    two directions that can produce one, neither of which is the registry
+    loader. An unreadable answer is no answer, never a claim in either
+    direction.
+    """
+    _resolve_amenity_coverage(
+        units,
+        index,
+        answer=lambda room: room.has_ramp or None,
+        grade=ramp_coverage,
+        target="ramp_coverage",
+    )
+
+
+def _resolve_amenity_coverage[Answer](
     units: list[LodgingUnitSummary],
     index: _BathroomIndex,
     *,
-    answer: Callable[[LodgingUnitSummary], bool],
+    answer: Callable[[LodgingUnitSummary], Answer | None],
+    grade: Callable[[Sequence[Answer | None]], str],
     target: str,
 ) -> None:
-    """The one leaf walk both resolvers above run.
+    """The one leaf walk all three resolvers above run.
 
     Parameterised on WHICH flag a room answers with and WHICH field receives
     the verdict, so a second amenity is a call site rather than a second
@@ -1109,6 +1177,14 @@ def _resolve_amenity_coverage(
     inactive rooms do not answer, unconfirmed rooms answer `None` — live here
     once; the reasoning for each lives in `_resolve_power_coverage`, which is
     the function that established them.
+
+    `Answer` is the vocabulary a ROOM answers in — `bool` for the boolean
+    flags, `str` for `has_ramp`'s three-value select — and `grade` is
+    parameterised alongside it for ONE reason, not for generality
+    (kindred#2438): `has_ramp` is a three-value select, so its rooms answer
+    `"yes"` / `"partial"` / `"no"` rather than a bool and need
+    `ramp_coverage`'s five-grade verdict. The WALK is what must not be
+    duplicated; which vocabulary a room answers in is the caller's business.
     """
     for unit in units:
         rooms = [
@@ -1120,14 +1196,11 @@ def _resolve_amenity_coverage(
         setattr(
             unit,
             target,
-            cast(
-                AmenityCoverage,
-                # `None` where nobody has confirmed the row: an unconfirmed
-                # `has_power = False` means "nobody has said", never "there is
-                # no power" -- the same gate `rosterAttention` already applies
-                # to the roster's own fit check.
-                amenity_coverage([answer(room) if room.is_confirmed else None for room in answering]),
-            ),
+            # `None` where nobody has confirmed the row: an unconfirmed
+            # `has_power = False` means "nobody has said", never "there is
+            # no power" -- the same gate `rosterAttention` already applies
+            # to the roster's own fit check.
+            grade([answer(room) if room.is_confirmed else None for room in answering]),
         )
 
 
@@ -1374,6 +1447,9 @@ class LodgingRosterService:
         # reasoning as above: `build_summary` carries no `units`, so resolving
         # coverage there would be work no response can read.
         _resolve_fridge_coverage(unit_summaries, unit_index)
+        # The same walk, one dimension over (kindred#2438), and the third and
+        # last read of it. Same path-only reasoning as above.
+        _resolve_ramp_coverage(unit_summaries, unit_index)
         # A SECOND PASS, and it has to be: a unit's cover can come from a row
         # on a unit built after it, so there is no order in which one pass over
         # `_build_units` would see every own-row it needs.
@@ -1993,6 +2069,11 @@ class LodgingRosterService:
                     has_ac=_b(unit, "has_ac"),
                     has_fridge=_b(unit, "has_fridge"),
                     has_shared_fridge=_b(unit, "has_shared_fridge"),
+                    # RAILED to the select's own vocabulary, not cast, so a
+                    # grade this code has never heard of reads as NOT ASSESSED
+                    # rather than leaking into the payload's Literal
+                    # (kindred#2438). See `_ramp_assessment`.
+                    has_ramp=_ramp_assessment(_s(unit, "has_ramp")),
                     is_accessible=_b(unit, "is_accessible"),
                     is_confirmed=_b(unit, "is_confirmed"),
                     is_active=_b(unit, "is_active"),
@@ -2457,6 +2538,11 @@ class LodgingRosterService:
             # above are: the derivation runs over RAW per-person narrative
             # values in the sync layer, and this service cannot see them.
             needs_fridge=_b(registration, "needs_fridge"),
+            # kindred#2438, and read from the column for the same reason: the
+            # derivation runs over RAW per-person narrative values in the sync
+            # layer, across BOTH housing narratives, and this service cannot
+            # see them.
+            needs_step_free=_b(registration, "needs_step_free"),
         )
 
     # --------------------------------------------------------------- counts
--- a/api/services/lodging_rules.py
+++ b/api/services/lodging_rules.py
@@ -223,6 +223,65 @@ def amenity_coverage(values: Sequence[bool | None]) -> str:
     return "none"
 
 
+def ramp_coverage(values: Sequence[str | None]) -> str:
+    """How much of a slot is step-free — kindred#2438.
+
+    The twin of `amenity_coverage` above, and a SEPARATE function for one
+    reason: `has_ramp` is a three-value select (`yes` / `no` / `partial`, blank
+    = NOT ASSESSED) rather than a bool, so a room can answer "qualified" and
+    the boolean grain has nowhere to put that. Migration 1500000131 made it a
+    select deliberately -- "a bool maps every unassessed cabin to false, which
+    asserts 'no ramp' about cabins nobody has looked at" -- and a boolean read
+    of it reports 0 of 118 units, erasing all 14 staff assessments (5 `yes`,
+    5 `partial`, 4 `no`, 104 blank on the production snapshot).
+
+    It does not walk anything, exactly as `amenity_coverage` does not: the
+    caller resolves which rooms answer, and `_BathroomIndex.leaf_codes_under`
+    is the ONE walk over that tree.
+
+    Args:
+        values: one entry per unit that answers for the slot -- `"yes"`,
+            `"partial"` or `"no"`. `None` means NO ANSWER, and it covers both
+            an unconfirmed row and an assessed-nobody blank; the caller maps
+            an unrecognised string to `None` too, because an unreadable answer
+            is not a claim in either direction.
+
+    Returns:
+        "all" | "some" | "partial" | "none" | "unknown".
+
+        FIVE grades, one more than the boolean amenities carry, and each is a
+        different claim:
+
+            all      every answering room is fully step-free
+            some     at least one is, but not all -- the reading that invites
+                     the placement landing in one of the others
+            partial  NO room is, but at least one has a qualified ramp
+            none     every answering room answered `no`
+            unknown  nothing answers
+
+        `partial` does not fold into `none`, because that would re-erase 5 of
+        the 14 assessments in the exact direction the select exists to prevent,
+        and 3 of those 5 carry the qualifier text in `notes`. It does not fold
+        into `some` either: "no room is step-free but one has a ramp with a
+        lip" is a weaker claim than "some rooms are step-free", and collapsing
+        them would make the grade order lie.
+
+        `unknown` whenever ANY contributing unit is unanswered -- the same
+        all-or-nothing evidence bar `amenity_coverage` states, and it matters
+        far more here: 104 of 118 units are blank, so a looser bar would grade
+        buildings step-free on the strength of the rooms somebody got to.
+    """
+    if not values or any(value is None for value in values):
+        return "unknown"
+    if all(value == "yes" for value in values):
+        return "all"
+    if any(value == "yes" for value in values):
+        return "some"
+    if any(value == "partial" for value in values):
+        return "partial"
+    return "none"
+
+
 def container_bathroom(leaf_bathroom_groups: frozenset[str]) -> tuple[str, str]:
     """A container's bathroom, inherited from its leaf descendants.
 
--- a/docs/reference/lodging-registry.md
+++ b/docs/reference/lodging-registry.md
@@ -210,6 +210,7 @@ loud on purpose, rather than a silently empty registry.
       "has_lights": true,
 
       "has_ramp": "partial",    // yes | no | partial; "" or absent = NOT ASSESSED
+                                //   read by the roster as ramp_coverage (#2438)
       "max_beds": 14,           // total sleeping spots. NOT sleeps — see below
 
       "shareability": "shareable"  // shareable | single_party; absent = not curated
@@ -333,6 +334,17 @@ default and splits fields by how much damage writing them could do:
 Two further rules: an empty `has_ramp` never overwrites a real assessment, and
 `notes` are filled only when the database has none — replacing free text a staff
 member wrote would destroy it.
+
+Since kindred#2438 `has_ramp` is no longer write-only: the weekend roster
+publishes it as `ramp_coverage`, resolved over a unit's LEAF descendants exactly
+as `power_coverage` and `fridge_coverage` are, and the board grades a family's
+`needs_step_free` flag against it. Two things follow for anyone editing this
+file. **Blank still means NOT ASSESSED and resolves `unknown`, never `none`** —
+104 of the 118 units are blank, so writing `no` into them would mark almost the
+whole registry step-free-hostile on evidence nobody recorded. And **`partial` is
+carried through** as its own grade rather than folded into either neighbour, so
+the ramp qualifier a `partial` unit records in `notes` stays the thing staff are
+being pointed at.
 
 `parent_unit` is compared as a **code**. The database stores it as a relation —
 a record id — so comparing raw values reports every parented unit as needing a
--- a/frontend/src/components/weekend/needsFit.test.ts
+++ b/frontend/src/components/weekend/needsFit.test.ts
@@ -27,6 +27,8 @@ function unit(overrides: Partial<LodgingUnitRow> = {}): LodgingUnitRow {
     has_fridge: false,
     has_shared_fridge: false,
     fridge_coverage: 'none',
+    has_ramp: '',
+    ramp_coverage: 'none',
     ...overrides,
   }
 }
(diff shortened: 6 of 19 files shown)


adamflagg:
@coderabbitai review

adamflagg:
@coderabbitai review

adamflagg:
Both CodeRabbit findings acted on, and the branch was force-pushed as a single squashed commit — the inline comments above point at a SHA that no longer exists, so the record is in the PR body's **Review record** section.

- **Corpus health narratives in source comments** — removed. The justification for admitting `knee`/`hip` is now a categorical description with nothing quoted, two verbatim test fixtures were replaced with synthetic ones, and the squash means the quoted text is in no reachable commit. The PR body and the audit comment on #2438 were scrubbed too.
- **`hip` as a substring** — fixed, and widened to the same class: `hip`, `cane` and `ramp` now match as whole words (*relationship*, *shipping*, *hurricane*, *cramp* were all live false positives); the other eleven keywords stay substrings because `walk` has to catch *walking* and *walkway*. All four are pinned as negative cases. No measured count changes.

CI Summary is green on the new head.

adamflagg:
## Independent scan (triage-attack Part B)

Re-verified against the current head (`39d5879e`), not just the builder's report.

**CI**: CI Summary green, 27/27 checks pass. Confirmed via `git fetch` + remote ref, not rtk's verdict.

**CodeRabbit**: real review landed — `**Actionable comments posted: 2**` (verified with `verify-pr-state.sh coderabbit 2494`), not a no-op notice. Both findings (PII in source comments/test fixtures; `hip` substring false-positive) are already remediated on this squashed head — confirmed by reading the diff directly, no corpus text or verbatim narrative is reachable in any commit, and `stepFreeWordPattern` now whole-word-matches `hip`/`cane`/`ramp`.

**Independent re-verification of the builder's own claims:**
- Ran `go test ./sync/... -run 'StepFree|RampCoverage|BathroomExplain' -v` — 5/5 pass.
- Ran `uv run pytest tests/unit/api/services/test_lodging_roster_service.py tests/unit/api/services/test_lodging_rules.py` — 334/334 pass.
- Read the has-some-data guard in `family_camp_derived.go` directly — `reg.needsStepFree` is OR'd in, matching the claimed 3-ungated-household fix.
- Read `ramp_coverage()` in `lodging_rules.py` and its 5-grade precedence (all/some/partial/none/unknown) — matches the doc comment and the pinned tests.
- Confirmed the issue #2438 body was corrected in place (audit comment 5340003871) and the PR body carries no quoted narrative text, only aggregate tables.
- Confirmed the branch is a single squashed commit (`git log --oneline` shows one PR commit on top of merge-base) — no PII-bearing commit is reachable.

**No new findings.** Scan floor 60, high effort — clean beyond what CodeRabbit already caught and the builder already fixed.

**Holding, not merging** — per campaign gate: `needsFit.ts` carries real GUI-affecting logic (a new branch + dimension entry) and two judgement calls need the owner's eyes:
1. The `some`-grades-worse-than-`partial` inversion in the ramp dimension (a building with SOME step-free rooms marks `unmet`; a building with a qualified-but-imperfect ramp marks the softer `partial`).
2. The width of the keyword surface — `knee`/`hip` admitted on corpus-wide evidence, `close to` deliberately excluded as a proximity ask.

Dev preview: `localhost:3176` (Vite). Nothing populates until `family_camp_derived` is re-run for 2026 — the migration is additive only.

Issue #2433, opened 2026-08-18
feat(frontend): family-facing weekend roster export (PDF and spreadsheet)


adamflagg:
## Status — approach settled 2026-08-19, supersedes the original framing

The template this issue was "blocked on" **has been found** — it is a hand-made Google Sheet, and it is
**ten columns, not the seven this body used to describe**. It was reviewed against production and the
output format is now settled. The earlier plan in this body — add `email` to `PartyAdult`, add a city to
the roster payload, and decide an RBAC gate on `/api/lodging/roster` — **is no longer the approach and
should not be implemented.** See "Architecture" below for why.

## What this is

A **family-facing** roster for one Family Camp weekend, exported to Google Sheets on demand by staff.
One row per person — campers and adults alike — grouped into a boxed, banded block per household.

It is **not** an export of the weekend roster tab. That tab (`frontend/src/components/weekend/HouseholdRosterTable.tsx:90`)
is a staff view: one row per placeable party, columns `Party` / `Cabin` / `Requests` / `Housing needs`.
This artifact is person-grained, carries contact details, and carries none of the operational columns.

## The format

Header: two merged rows — `{session title} {YYYY} Roster`, then the date range. Nothing else; no totals.

Columns, in order: **`Name` · `Adult / Camper` · `Age` · `Email` · `City`**

| Rule | Value |
|---|---|
| Row grain | one person |
| Order within a household | campers **youngest → oldest**, then adults **1 → 5** |
| Household order | A→Z on the **first camper's** surname |
| `Adult / Camper` | `Camper`, or `Adult N` from `family_camp_adults.adult_number` |
| Age, 12 months and over | whole years, no decimal — `8` |
| Age, under 12 months | whole months — `11 mos`, `1 mo` (singular is a real case) |
| Age basis | **computed at export time**, not at session start — staff want the roster to state ages as of the day it was pulled |
| City | first row of each household block only |
| Banding | alternates **per household**, not per row; plus a block border |
| Linked households | households sharing an adult name get a shared highlight colour — meaning *"these are linked, look here"*, never *"these should be merged"* |

There is **no family-name label and no relationship/role column.** Surnames within a household are too
often inconsistent to build a label from, and `relationship_to_camper` is blank on ~45% of 2026 rows.

Note the ordering above is what the hand-made sheet already does — this makes it explicit rather than changing it.

## Where it goes

A **dedicated Drive folder, separate from the data-workbook exports** — these carry family contact
[...]

PR #2495, opened 2026-08-19
feat(google): plumbing for the Family Camp roster export

adamflagg:
Groundwork for #2433. **No user-visible change and nothing writes a roster yet** — this lands the four capabilities the export needs and the schema that can hold its workbooks. PR 1 of 3 (2 = builder + endpoint, 3 = the toolbar button).

## A second Drive folder, addressed by id

`GOOGLE_DRIVE_ROSTER_FOLDER_ID` joins `GOOGLE_DRIVE_FOLDER_ID` in `.env.example` and **both** docker-compose service blocks, with `GetRosterFolderID()` reading it.

Rosters carry family contact details and Drive folder sharing is the privacy control for them, so they must not inherit the Exports folder's audience — hence a second variable rather than a tree derived from one root. Both roster folders are configured **by id** rather than resolved by name: a rename in Drive then errors instead of silently creating a duplicate. Only the `{year}` subfolder is created by code, because code owns that name and it is deterministic.

## Folder-scoped Drive helpers

`CreateSpreadsheetInFolder`, `FindSpreadsheetInFolder` and `FindOrCreateFolder` take an explicit parent. `CreateSpreadsheet` and `FindSpreadsheetByName` become thin wrappers passing `GetFolderID()`, so **the working data export resolves the same Drive operation as before** — same guards in the same order, same `SupportsAllDrives`/`IncludeItemsFromAllDrives`, same `trashed=false` and `Fields(…)`. The Drive query builder is extracted and tested, including the `mimeType` clause that separates a folder search from a spreadsheet search and the escaping.

One observable difference, deliberately accepted: on the `NewDriveClient` failure path the error gains a prefix — `failed to create drive client: X` becomes `failed to create spreadsheet in folder: failed to create drive client: X`. The messages on the realistic API-failure paths are byte-identical. Nothing matches on these strings, every caller either re-wraps with `%w` (so `errors.Is`/`errors.As` traverse intact) or logs, and the prefix names which operation failed.

`FindOrCreateFolder` refuses an empty parent id: an empty parent would create the folder in the service account's own Drive, where no member of staff can see it. It is not concurrency-safe and says so — two simultaneous callers can create two folders of one name. Exports are staff-triggered and synchronous, so the window is not worth a lock.

## A formatting capability on `SheetsWriter`

`ApplyFormatting` takes one tab's whole appearance — merges, frozen rows, column widths, cell styles, borders — and issues a **single** `spreadsheets.batchUpdate`. One call per tab, not per block: a 63-household roster is one round trip, not 63. It routes through `RateLimitedSheetsWriter` like every other method, so it inherits the 429 retry.

**Additive only.** The existing seven methods are untouched, so the 15-tab export that shares this writer cannot be perturbed.

The directive types (`SheetFormat`, `CellStyle`, `BorderRule`, …) keep `google.golang.org/api/sheets` out of the interface, as `TabColor` and `TabPropertyUpdate` already do. That puts the translation — with its field masks and its zero-value traps — in one pure, tested function.

Two traps are pinned by tests because both fail *silently*:

- **The `repeatCell` field mask names only what a style sets.** A blanket `userEnteredFormat` mask clears every property the rule is silent about, which for the roster means one style pass erasing the previous one's fills.
- **Colours and `bold: false` carry `ForceSendFields`.** Go's JSON encoder drops zero values, so `#000000` would otherwise serialize to `{}` and read as "unset" rather than black.

A malformed directive fails **before** any API call rather than being quietly dropped, and a zero-valued `GridRange` is refused outright: the Sheets API reads an absent end index as "to the end of the sheet", so an accidentally-empty range would format the entire tab instead of nothing.

## The workbook registry gains a session dimension

Migration `1500000165_sheets_workbooks_session_dimension.js`: `session_cm_id`, `fc_roster` added to the `workbook_type` vocabulary, and — **the half the design missed** — the unique index re-keyed from `(workbook_type, year)` to `(workbook_type, year, session_cm_id)`.

The original index admits exactly one workbook per year. 2026 has ten Family Camp weekends, eight with enrolled campers, each getting its own workbook. Without the re-key the *first* roster workbook of a season would save and every one after it would be rejected by the database, with a uniqueness error that names nothing useful.

No backfill is needed, and that is a property of PocketBase rather than luck: it declares number columns `NUMERIC DEFAULT 0 NOT NULL`, so the twelve existing rows take `0` rather than `NULL`. A nullable column would have quietly *disabled* uniqueness for exactly the rows that still rely on it, since SQLite treats each `NULL` as distinct.

`workbook_type` stays a **closed** vocabulary. Minting one type per session was the obvious alternative and is wrong — it groups rows in the admin Sheets tab and drives `GetWorkbookByType`'s filter, so an unbounded set turns that filter into id string-matching.

`frontend/src/types/pocketbase-types.ts` is regenerated against a scratch database built from `pb_migrations/` alone (the recipe CI prints), adding exactly `fc_roster` and `session_cm_id?`.

## Testing

TDD throughout — every test written and confirmed failing first.

- **Go unit**, no Google API: hex parsing at the black/white boundaries, field-mask construction per style permutation, request ordering, border edge omission, and every refusal path. Emitted requests are asserted as **JSON**, not as Go structs, so the assertions are about what Google actually receives.
- **`httptest`-backed**: `ApplyFormatting` issues exactly one HTTP call carrying all seven directives; an empty format issues none; an invalid directive issues none and names the offending value.
- **`scripts/dev/verify-sheets-workbooks-schema.sh`** asserts the field type, `onlyInt`, optionality, `max: null`, the vocabulary and the index columns against a throwaway migrated database — same family as `verify-lodging-schema.sh`, and manual for the same reason (it boots PocketBase). It reported four failures before the migration existed.
- **The down path was exercised separately**: two same-year `fc_roster` rows insert cleanly under the new index — the pair the old index would have rejected — and `migrate down 1` removes them along with the field, the vocabulary entry and the index.
- `bash scripts/pre-push-verify.sh` — all checks pass, including `golangci-lint` and `eslint`, which the pre-push hook does not run.

A floor-60 scan plus CodeRabbit ran over the branch afterwards. Two code defects came out of it and are fixed here: the Drive query escaping above, and a negative `CellStyle.FontSize` slipping through both `isEmpty`'s `== 0` test and the mask's `> 0` test to emit a `repeatCell` with an empty field mask — a 400 that would have taken the whole tab's single batchUpdate with it.

## Follow-up for PR 2 — read this before writing the roster workbook

**The Go write path is NOT ready, and it will not fail loudly.** This PR ships the schema half only; `pocketbase/sync/workbook_manager.go` is untouched and is still session-blind:

- `GetWorkbookByType` filters `workbook_type = 'X' && year = N` with no `session_cm_id` term and `limit 1`
- `WorkbookRecord` has no `SessionCMID` field, and `SaveWorkbookRecord` never sets the column
- `FormatWorkbookTitle(type, year)` has no session dimension either, so every roster workbook in a year would be handed the same title

A `GetOrCreateRosterWorkbook` copying the established `GetOrCreateYearWorkbook` shape would therefore save weekend #2's roster, match weekend #1's row (both `session_cm_id = 0`, since nothing sets it), take the **update** branch, and silently overwrite weekend #1's `spreadsheet_id`, title and URL — collapsing all eight enrolled 2026 weekends into one row. **The new UNIQUE index never fires**, because the ORM-level existence check short-circuits ahead of it. Anticipated in the design (§4.4: "it needs the session dimension or a new accessor"); inert here only because nothing writes `fc_roster` yet.

Also frontend, and lower stakes: `frontend/src/hooks/useSheetsWorkbooks.ts:16` hand-declares `workbook_type: 'globals' | 'year'`, and `SheetsTab.tsx` groups on globals-vs-not. Neither is wrong while no `fc_roster` rows exist. Left alone here to keep this PR free of rendered-surface changes.

Refs #2433

## Summary by CodeRabbit

* **New Features**
  * Added support for session-specific Family Camp roster workbooks.
  * Roster exports can use a dedicated Google Drive folder or be disabled when unconfigured.
  * Added automatic spreadsheet formatting, including merged cells, frozen rows, column widths, styles, and borders.
  * Improved folder-scoped spreadsheet creation and lookup in Google Drive.

* **Bug Fixes**
  * Improved formatting update reliability, including retries for temporary Google API rate limits.

* **Documentation**
  * Added configuration guidance and schema verification tooling for roster workbooks.


adamflagg:
@coderabbitai review

adamflagg:
@coderabbitai review

Focus, in order of blast radius:

1. `pocketbase/pb_migrations/1500000165_` re-keys a **UNIQUE index on a populated table** — `(workbook_type, year)` becomes `(workbook_type, year, session_cm_id)`. Is the up path safe against the 12 existing rows, and does the down path fully revert (field, select vocabulary, index, and the `fc_roster` rows that would collide under the restored index)?
2. `ApplyFormatting` is new surface on `SheetsWriter`, which the existing 15-tab data export shares. Confirm nothing in the existing seven methods changed behaviour.
3. `pocketbase/sync/sheet_format.go` builds Sheets API `fields` masks by hand. A mask that is too broad silently CLEARS formatting the rule does not mention. Also check `ForceSendFields` coverage — Go drops zero values, so an unforced zero colour or `bold:false` serialises to nothing.
4. `CreateSpreadsheet`/`FindSpreadsheetByName` were refactored into folder-scoped variants. They must resolve byte-identically to before for the existing export.

PR #2497, opened 2026-08-19
feat(sync): Family Camp roster builder and export endpoint

adamflagg:
feat(sync): Family Camp roster builder and export endpoint

Second of three PRs for the family-facing Family Camp roster (Refs #2433).
PR 1 (#2495) shipped the plumbing; this builds the artifact and exposes it at
`POST /api/custom/lodging/roster-export?year=&session=`, gated on
`bunking.manage`. PR 3 adds the toolbar button.

One row per person, campers and adults, grouped into banded blocks per
household, appended as a new date-stamped tab. It never overwrites or prunes a
tab: staff hand-edit every one, because paper registrations exist only as
lodging-board write-ins and never reach CampMinder, so the previous tab is where
they copy their work from. That is also why there is no scheduled run.

## The trap this PR had to avoid

`GetWorkbookByType` filtered `(workbook_type, year)` with `limit 1`, and
`SaveWorkbookRecord` used it as its existence check. Reusing that for
`fc_roster` would make the SECOND weekend of a season match the FIRST weekend's
row, take the UPDATE branch, and silently overwrite its `spreadsheet_id` --
collapsing all eight of 2026's enrolled weekends into one workbook. The new
unique index does not catch it, because the ORM check short-circuits ahead of
it. `WorkbookRecord` now carries `SessionCMID`, `GetWorkbookForSession` is the
session-aware lookup, and `GetWorkbookByType` delegates to it with
`session_cm_id = 0` so the data export's own lookups are unchanged. Verified
against dev: three weekends exported, three rows, three distinct spreadsheets.

## Two things found while building it

**The admin Sheets tab would have regressed with no frontend file in the diff.**
`useSheetsWorkbooks` reads `sheets_workbooks` directly and `SheetsTab` labels
every non-globals row `String(workbook.year)`, so eight roster workbooks would
have rendered as eight rows all reading "2026", sorted above the real 2026 data
workbook and each opening somewhere else. The hook now filters `fc_roster` out.
Roster workbooks are also excluded from `BuildIndexSheetData`: that index lives
in the globals workbook in the Exports folder, whose audience is deliberately
wider than the roster folder's, and listing one there would publish a
family-contact workbook's name and link to that audience.

**The design's tab-naming paragraph contradicted itself** -- it said tab names
reject `: \ / ? * [ ]` while specifying a `3:04 PM` format. Probed against the
live Sheets API: the colon is accepted, and a `Values.Update` reaches the tab
through both quoted and bare A1 notation, so `WriteToSheet` needs no change to
carry a name with spaces and commas. That rejected set is Excel's rule, not
Sheets'. The 31-character cap is real and is tested.

## Data traps respected

- Duplicate adults render RAW -- deduping here would hide #2483.
- Households sharing an adult are colour-linked, NEVER merged; merging would
  force one wrong city onto co-parents keeping two homes.
- `adult_number >= 3` rows carry a name and no email on 27 of 27 such 2026 rows;
  that blank is correct and permanent, not missing data.
- `family_camp_adults.last_name` is blank for all of 2026, so `name` is the
  column of record and the split pair is only the fallback.
- Households with campers and no adults render as-is.
- Ages are computed AT EXPORT TIME, not at session start -- the reference
  implementation did the latter and the design supersedes it.

## Verification

Unit tests cover the age boundaries (0/1/11/12/13 months and a leap-day
birthday), ordering, the placeholder filter, the name coalesce, city cleaning
and the household fallback, linked-household grouping, tab-name collisions, and
the emitted `batchUpdate` shape against a mock writer. Two of them were
mutation-checked: reinstating the session-blind lookup fails the collapse test,
and adding adult dedup fails the #2483 test. A third was rewritten after a
mutation survived it.

End to end against dev, hitting the real Google API: Family Camp 2 exported 63
households, 98 campers, 120 adults, 218 person rows -- the design's stated
figures for that weekend. Reading the sheet back confirmed 221 rows, both
merges, `frozenRows=3`, the five column widths, the header fill, 63 medium top
and 63 medium bottom borders (one per household), a centred age column, and two
linked-household groups. Every refusal path returns its intended status: 401
unauthenticated, 400 bad params / not a family weekend / no enrolled campers,
404 unknown session.

`scripts/pre-push-verify.sh` passes, `golangci-lint` clean.

Refs #2433

## Summary by CodeRabbit

* **New Features**
  * Added Family Camp roster exports to Google Sheets, including camper, household, adult, age, city, and linked-household details.
  * Exports create or reuse year- and session-specific workbooks, add dated tabs, and apply consistent formatting.
  * Added clear validation and error handling for invalid or unavailable sessions.
* **Bug Fixes**
  * Family Camp roster workbooks are excluded from general workbook listings and indexes.
  * Improved workbook tracking across separate camp sessions.
* **Tests**
  * Added comprehensive coverage for roster generation, formatting, exports, and validation.

### DIFF
--- a/frontend/src/hooks/useSheetsWorkbooks.test.ts
+++ b/frontend/src/hooks/useSheetsWorkbooks.test.ts
@@ -1,0 +1,78 @@
+/**
+ * TDD tests for useSheetsWorkbooks.
+ *
+ * The admin Sheets tab is about the DATA export workbooks — Globals and one per
+ * year. Family Camp roster workbooks (kindred#2433) share the sheets_workbooks
+ * collection but are a different surface: one per weekend, in a Drive folder
+ * with a deliberately narrower audience, reached through the roster toolbar
+ * rather than here.
+ *
+ * They must not reach this list. SheetsTab renders every non-globals row as
+ * `String(workbook.year)`, so 2026's eight enrolled weekends would arrive as
+ * eight rows all labelled "2026", indistinguishable from the real 2026 data
+ * workbook and each linking somewhere else.
+ */
+import { describe, it, expect, vi, beforeEach } from 'vitest'
+
+const mockGetFullList = vi.fn()
+
+vi.mock('../lib/pocketbase', () => ({
+  pb: {
+    collection: vi.fn(() => ({
+      getFullList: mockGetFullList,
+    })),
+  },
+}))
+
+describe('useSheetsWorkbooks', () => {
+  beforeEach(() => {
+    mockGetFullList.mockReset()
+    mockGetFullList.mockResolvedValue([])
+  })
+
+  describe('SHEETS_WORKBOOK_LIST_FILTER', () => {
+    it('excludes Family Camp roster workbooks', async () => {
+      const { SHEETS_WORKBOOK_LIST_FILTER } = await import('./useSheetsWorkbooks')
+      expect(SHEETS_WORKBOOK_LIST_FILTER).toContain('fc_roster')
+      expect(SHEETS_WORKBOOK_LIST_FILTER).toMatch(/workbook_type\s*!=/)
+    })
+
+    it('is applied to the sheets_workbooks query', async () => {
+      const { fetchSheetsWorkbooks, SHEETS_WORKBOOK_LIST_FILTER } =
+        await import('./useSheetsWorkbooks')
+
+      await fetchSheetsWorkbooks()
+
+      expect(mockGetFullList).toHaveBeenCalledTimes(1)
+      expect(mockGetFullList).toHaveBeenCalledWith(
+        expect.objectContaining({ filter: SHEETS_WORKBOOK_LIST_FILTER })
+      )
+    })
+
+    it('keeps the existing year-descending sort', async () => {
+      const { fetchSheetsWorkbooks } = await import('./useSheetsWorkbooks')
+
+      await fetchSheetsWorkbooks()
+
+      expect(mockGetFullList).toHaveBeenCalledWith(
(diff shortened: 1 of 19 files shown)


adamflagg:
@coderabbitai review

adamflagg:
All four addressed in e374963c. One correction for the record, on the city-fallback finding.

**The stated premise does not hold.** `households` is itself year-scoped — `idx_households_cm_id_year` is UNIQUE on `(cm_id, year)`, so there is one household *record per household per season* and a 2026 household id is unreachable from a 2025 person. Checked against the database rather than reasoned about:

- `households`: 22,326 rows across 10 years; every `cm_id` has exactly one row per year.
- households whose `persons` span more than one year: **0**
- households whose `family_camp_adults` rows span more than one year: **0**

So no prior-season row could ever have supplied the city, and the selected city never depended on cross-year id ordering.

**I made the change anyway**, for two reasons that are not the stated one: it states the year scoping at the query instead of relying on the reader knowing the `households` invariant, and it makes the two loaders consistent — `attachRosterAdults` filters the same way, and for the same belt-and-braces reason, not because either could leak. It is now pinned by `TestBuildFamilyCampRosterCityFallbackIsYearScoped`, which seeds the shape the invariant forbids so the filter is tested rather than merely present.

On the session-id finding: session CampMinder ids identify a *session*, not a person, and `sync_testsupport_test.go` already carries real ones as named constants on `main` (`cmIDFamilyCamp1`, `cmIDFamilyCamp6`, `cmIDWinterFamily`). I still made the swap, because a registry key-collision test names no weekend — a bare literal there implied a specificity it did not have. The builder tests keep the shared constants, since those fixtures do stand in for real weekends.

Issue #2252, opened 2026-08-12
feat(frontend): make the clear-write-in control's effect clearer


adamflagg:
## ⚠️ CORRECTED 2026-08-14 — phase C triage

Verified against `main` @ `3eb44ba4` and the production snapshot. Verdict: **VALID** / SCOPED.

> D9's stated precondition — 'verify the `WriteInCard` covers every zoom/collapsed state before deleting the chip' — is NOT met, so this is not mechanical. `WriteInCard` is imported by exactly one file, `LodgingUnitCard.tsx`. The 'Write-in' chip produced by `unitBadges.reservationBadge` is also consumed by `MapUnitPopover.tsx` at two places, neither of which renders a `WriteInCard`: the popover header badge (:179) and the per-room collapsed grid cell, whose own comment says 'The grid has no room for badges, so status rides in the tooltip' (:593-598). Deleting the chip from `reservationBadge` would make a write-in invisible on the whole map surface.

### Corrections to the body below

1. ADD at the top of the body: "> ## ★ D9 RULING — drop the write-in chip, relabel the single button 'Clear Write-in'. Supersedes readings 1-3; the reporter no longer needs to be asked. **But D9's precondition FAILS and this is therefore NOT mechanical.** D9 says 'verify the `WriteInCard` covers every zoom/collapsed state before deleting the chip'. It does not. `WriteInCard` has exactly one importer (`LodgingUnitCard.tsx:42`). `unitBadges.reservationBadge`'s 'Write-in' chip has two consumers, and the second — `MapUnitPopover.tsx` — renders no `WriteInCard` at all: the popover header badge (`:179`) and the collapsed per-room grid cell (`:593-598`, whose own comment reads 'The grid has no room for badges, so status rides in the tooltip'). Deleting the chip from `reservationBadge` makes a write-in invisible on the entire map surface. **Scope the deletion to `LodgingUnitCard`'s badge row only, or give `MapUnitPopover` a replacement signal first.**"
2. WRONG: "Worth asking the reporter which of these it is before building — the three have materially different fixes, and one of them is a confirm dialog this board may not want." → REPLACEMENT: "Superseded by D9, which supplied a fourth reading the report did not have — two affordances for one concept — and ruled on it. No confirm dialog. Do not ask the reporter."
[...]

PR #2505, opened 2026-08-19
feat(frontend): per-need glyphs on the family card, and one place a need is graded

adamflagg:
## What this changes

The family card's third line stops being a row of words and becomes what
kindred#2072 asked for: the four ruled needs — bathroom, power, fridge,
step-free — drawn as icon-only glyphs in a closed hue set, red-filled when the
cabin does not meet them, and omitted entirely when the household did not ask.
The single-parent mark moves up to line 2 where it describes the adult beside
it, and Returning / First-time collapses from two of the widest chips on the
card into one 16px icon pinned bottom right.

Underneath that, the more consequential half: **there was no per-need fit
resolver in this codebase, and the three tables that graded these needs
disagreed.** `rosterAttention`'s `VERIFIABLE_NEEDS` read the raw
`unit.has_power`; `needsFit`'s `NEEDS_DIMENSIONS` read the server-resolved
`power_coverage`; `FamilyCardChips` drew words off the flags directly. Twelve
of the fourteen 2026 family-pool containers record `has_power = 0` while every
leaf beneath them has power, so the roster reported "No power" about a building
the board's own drag-time hatch called fine. `needGlyphs.ts` is now the one
grading, and all three call it.

Relates to #2072. This is stage 1 of three — the glyph vocabulary. The Assign
modal (AS2 + W3) and the unit card (T2, the footer, the padding) follow, in
that order, because the modal needs the glyphs and not the reverse.

## Ruled, and where the ruling lives

Every decision is recorded in `docs/reference/weekend-card-vocabulary.md`,
which is tracked and is updated by this PR. That file exists because the same
decisions were written down once before in a **gitignored** spec that at least
nine shipped comments still cite by section number.

- **A3 — needs are icon-only chips.** Four dimensions, four hues, and the set
  is closed: `text-sky-500 dark:text-sky-400` and the same 500/400 step for
  purple, teal and orange. The review mock renders `#0ea5e9` / `#a855f7` /
  `#14b8a6` / `#f97316`; those values *are* those Tailwind steps. No hex is
  written anywhere, and two tests sweep for it.
- **N2 — a need the cabin does not meet turns the glyph red.** Warn fill, warn
  border, warn icon. The shape still says which need it is, which is what
  makes losing the hue affordable.
- **The absence rule.** A need nobody asked for is *omitted*, never dimmed.
  Pinned on the resolver and again on the card, because "draw all four and grey
  the unasked ones" is the helpful thing a later session adds.
- **S2 + Sa — single parent leaves the chip row.** It sat there wearing the
  muted `Near another family` grammar, which made a fact about who is in the
  room read as a preference the household expressed. It is now an amber mark
  on line 2 — First-time's amber, so amber means "notice this household"
  across both marks.
- **R3 — Returning / First-time is a 16px icon, no text.** The pair fires on
  every household card (279 returning against 123 first-time), so it was never
  absent and never discriminating in the way a chip's position implies.
- **`Accommodation required` → `Needs Accommodation`.** The label is
  **explicitly not locked** — it is one of five marks parked for staff input —
  so it lives in `ATTENTION_LABEL` alone. `HouseholdRosterRow` carried a
  hardcoded copy that had already diverged; it reads the constant now, or the
  rename would have left the roster and the card wording one fact two ways.

## The one contradiction this ships on purpose

The unit card (stage 3) will draw its bathroom mark as **presence** —
`bathroom != 'none'` — because that is the axis the CampMinder question
actually asks: *"a bathroom that doesn't require you to leave your cabin"*.
The family glyph still grades **exclusivity**, because that is what the product
has always graded and moving it is kindred#2501, which is itself gated on
reading the Adult form's wording — 19 of 66 flagged households, never audited.

So for one release a room can say it has a bathroom while the family on it
shows a red bathroom glyph. Owner ruling 2026-08-19: accept it and name it.
There is a comment at both sites, a row in the vocabulary doc, and a test in
`needGlyphs.test.ts` whose assertion is the one that flips when kindred#2501
lands.

## Kept deliberately

- **The four sharing-intent chips, the consent warning and the
  `Needs Accommodation` label stay as words.** All five are parked for staff
  input; nothing here pre-empts that.
- **`AccessibilityFlagList` was not renamed.** It carries a *pair* —
  `Accommodation required` / `Accommodation requested` — keyed on the same
  flag, and renaming half a pair breaks it. The card and that list therefore
  word one fact two ways today; the vocabulary doc says so, so both reach
  staff together.
- **The roster's own wording did not move.** The glyph is called *Bathroom in
  unit*, but `rosterAttention` still grades exclusivity, and "No bathroom in
  unit" would be a false sentence about a cabin with a shared bathroom in it.
  Both halves move when kindred#2501 lands.
- **`partial` keeps its hue.** Two glyph states are ruled, not three: a ramp
  with a lip, or a building where some rooms have power, is a qualification
  rather than a warning — the same reading `PlaceFamilyPicker` already takes.
  Degree lives on the drag hatch, which grades it on the hatch period.
- **The roster grades two of the four needs, and the drag hatch three.** The
  *grading* is now one rule; *which* needs a surface reports is per-surface and
  each scope is written down where it lives. Bathroom is out of the hatch
  because only 6 of 118 units carry a private bathroom, so it would hatch 112
  cards on any pick-up — and 90 even after kindred#2501. Fridge and step-free
  are out of the roster because those sections are a staff-facing
  classification with counts on them, and folding two more needs in moves
  parties from `settled` to `unmet`. That is a ruling to take, not a side
  effect to inherit.

## Also struck

Each of these was removed deliberately, and each now has a **negative pin**
that fails if it comes back. That mechanism is not decoration here: this
codebase has twice restored an element whose absence was undefended, which
`LodgingUnitCard.test.tsx`'s own sr-only pin records.

| Struck | Why | Pinned in |
| --- | --- | --- |
| `Private bathroom` / `Power` need chips | they *are* the glyphs now | `FamilyCard.test.tsx`, "the marks kindred#2072 STRUCK" |
| `No private bathroom` / `No power` | a bathroom glyph beside a chip saying the bathroom is missing states one fact twice | same |
| `Fit not verified` — the whole `unverified` arm | the name was wrong on both arms; arm (b) fires *because* the cabin is confirmed, so "reconfirm" asked for a check already done | same |
| `Whole building` | an earlier cut that had **never been landed** — still on the card in code. It survives on the map, the same split `Staff` takes | `FamilyCard.test.tsx` + `LodgingUnitCard.test.tsx`, three cases |
| `Single parent` as a chip | moved to line 2, not deleted | `FamilyCard.test.tsx` |
| `Returning` / `First-time` as words | now one 16px icon | `FamilyCard.test.tsx` |
| `ChipTone.need`, `.quiet`, `.building` | a tone with no chip is an invitation to invent one | the type itself |
| `holdsWholeBuilding` on `FamilyCard` | the prop went with the chip — defended by `tsc`, not by a test | type error at any call site |

Three comments describing superseded behaviour were **rewritten, not deleted**:
the single-parent test's scoping note (it recorded a real earlier scoping that
the 2026-08-19 rulings went past), `MapUnitPopover`'s "reproduced from the
board" note (the map is that chip's only home now), and `worseOf`'s reference
to a dimension table that no longer exists.

## Not done

- kindred#2501 — the bathroom fit rule. Named above; the assertion that flips
  is already written.
- kindred#2502, kindred#2499, kindred#2500 — untouched.
- kindred#2503 — the Assign modal's `People` field is scoped to it; the modal
  itself is stage 2.
- The unit card's own marks (T2, the footer, the meta row, the padding, and
  the board half of the cuts) are stage 3 and deliberately not in this diff.

## Verification

```
cd frontend && ./node_modules/.bin/vitest run     # 453 files, 7138 tests, all pass
cd frontend && ./node_modules/.bin/tsc --noEmit   # clean
bash scripts/pre-push-verify.sh                   # prettier, eslint, tsc, vitest — all pass
```

32 existing assertions moved. None was weakened to fit: fixtures that meant
"this cabin has no power" now say so through `power_coverage`, the field the
rule actually reads, and the assertions that only checked that the card
*printed* a verdict were deleted in favour of the negative pin plus
`rosterAttention.test.ts`, which owns the rule itself. 24 new pins were added,
five of them for the container-power fix alone.

Both cut-defending mechanisms were mutation-checked rather than assumed:
restoring the three struck chips fails four tests, and breaking the absence
rule fails nine.

⚠️ Hand-written `.tsx` throughout, so the visual gate applies — do not merge
until the board has been looked at.

## Summary by CodeRabbit

* **New Features**
  * Weekend cards now display accessibility and accommodation needs as consistent, color-coded icons with tooltips.
  * Added clearer indicators for unmet, partial, unknown, returning, and first-time household statuses.
  * Updated roster attention wording to “Needs Accommodation.”
  * Whole-building markers are removed from lodging cards and retained on the map.
* **Documentation**
  * Expanded weekend-card vocabulary, grading rules, labels, colors, and implementation guidance.


adamflagg:
@coderabbitai review

Focus, if it helps aim the review:

- **`needGlyphs.ts` is a consolidation of three tables that disagreed.** `rosterAttention.VERIFIABLE_NEEDS` (raw `has_power`), `needsFit.NEEDS_DIMENSIONS` (server-resolved `power_coverage`) and `FamilyCardChips` (text chips) all graded the same needs differently. A wrong verdict from the merged resolver is silent — it renders as a coloured glyph, not an error — so the truth table in `needGlyphs.test.ts` and the `someIs` semantics per need are the load-bearing parts.
- **It moves a staff-facing classification.** The roster's `settled`/`unmet` sections now grade power off `power_coverage`, so twelve entirely-powered containers stop reporting "No power". Intended, but worth a second pair of eyes on whether anything ELSE moved with it.
- **32 existing assertions were retargeted and 24 negative pins added.** The failure mode to look for is an assertion that was weakened rather than retargeted — a test that would now pass with either the old or the new behaviour.
- **Two surfaces still read the marks this PR cut from the card** (`MapUnitPopover`, `HouseholdRosterRow`). Worth checking the cuts really are board-only.

adamflagg:
### Code review

Found 5 issues:

1. The roster's bathroom verdict now flips from `unmet` to `settled` when `effective_bathroom` is `unknown` or absent — a verdict change beyond the intended container-power fix (bug: old `satisfiedBy: (_unit, party) => party.effective_bathroom === 'private'` failed for every non-`private` value; `bathroomCoverage` now maps `unknown`/`undefined` → `'unknown'`, and `needVerdict` returns `'fits'` for `'unknown'`, so `asked.filter(... !== 'fits')` drops the need). Verified by running it: `unknown` → `settled`, absent → `settled`, `shared` → `unmet`. The inline comment on those lines claims the opposite — "every case that used to flag still flags, except the fully-covered container the raw flag got wrong". The analogous `power_coverage: 'unknown'` case *is* pinned by a new test; the bathroom one is not, and it propagates to `MapUnitPopover`, which filters on `attention.level === 'unmet'`. Not reachable on today's data (all 118 units carry a `bathroom` value, no unresolvable placement codes), but it is the same class of staff-facing reclassification the PR deliberately declined to make for fridge and step-free.

https://github.com/adamflagg/kindred/blob/9871efa81f020baf36e8e04e62749696a7b12084/frontend/src/components/weekend/rosterAttention.ts#L202-L209

2. The vocabulary doc moves `Building` from §1 ("marks that are drawn") into §3 ("STRUCK") and states a code change this PR never made — "suppressed at the card the way #2252 suppresses `Write-in` — so on a board card `reservationBadge` now draws nothing at all". `LodgingUnitCard.tsx:312` still suppresses only the write-in case, `unitBadges.ts` still returns `{label:'Building'}` for a container, and a combined container is exactly the card that reaches it. Same pattern in §5a row 1 ("The row **stays, rendering only when `Inactive` or `Reconfirm space` is present**") and §3's meta-row entry ("renders on no live card today") — the meta row at `LodgingUnitCard.tsx:846` is unconditional. §9 correctly lists the unit card as stage 3 pending; the §3 row for `Whole building` is correctly hedged as "had never been landed". These are not. The row this replaced was accurate, and §8 says "If they disagree, this file is wrong — fix it here."

https://github.com/adamflagg/kindred/blob/9871efa81f020baf36e8e04e62749696a7b12084/docs/reference/weekend-card-vocabulary.md#L76-L78

3. A new comment says, present tense, "The unit card draws its bathroom mark as PRESENCE — `bathroom != 'none'`". It does not: `LodgingUnitCard.tsx` still renders two distinct badges, `Private` and `Shared`, which is the exclusivity distinction the comment says has already left the card. The PR body gets the tense right ("The unit card (stage 3) **will draw**"); the comment does not. Duplicated at `needGlyphs.test.ts:128`.

https://github.com/adamflagg/kindred/blob/9871efa81f020baf36e8e04e62749696a7b12084/frontend/src/components/weekend/needGlyphs.ts#L119-L122

4. Deleting `VERIFIABLE_NEEDS` leaves three present-tense references to it dangling in `placementCandidates.ts` (`:56`, `:114`, `:172`), and `:63` — "The power rule is stated in both places. CHANGE ONE, CHANGE BOTH" — now points at `needsFit.ts`, which this PR gutted; the rule lives in `needGlyphs.ts`, a file that comment never names. Every reference the PR itself wrote correctly uses the past tense. [#2462](https://github.com/adamflagg/kindred/pull/2462)'s review fixed exactly this class of drift in the PR that caused it.

https://github.com/adamflagg/kindred/blob/9871efa81f020baf36e8e04e62749696a7b12084/frontend/src/components/weekend/placementCandidates.ts#L55-L57

5. A new comment cites `LodgingUnitCard.test.tsx:2277` as recording the sr-only region "that came back after being ruled out, twice". Line 2277 is a combobox visibility assertion; the block described starts at `:2295` (`:2280` before this PR). CLAUDE.md §4 on stale references: "overwhelmingly not wrong *premises* but wrong *identifiers*".

https://github.com/adamflagg/kindred/blob/9871efa81f020baf36e8e04e62749696a7b12084/frontend/src/components/weekend/FamilyCard.test.tsx#L1455-L1458

Checked and found clean: the `power`/`fridge`/`step_free` truth tables reproduce the old `NEEDS_DIMENSIONS` exactly (same fields, same `someIs`, same `unknown → fits`); `ramp_coverage` is read instead of the truthy `has_ramp` string; the retargeted assertions in `FamilyCard.test.tsx`, `LodgingUnitCard.test.tsx` and `MapUnitPopover.test.tsx` invert or tighten rather than loosen; `wholeBuildingHolders` and the map's `WholeBuildingBadge` survive the card-side cut with no dangling references; and `FamilyCard.tsx`'s line-2 `min-w-0 flex-1` chain preserves truncation and the `ml-auto` cabin anchor.

PR #2506, opened 2026-08-20
feat(frontend): the Assign modal replaces the inline picker on every card

adamflagg:
## What this changes

`Assign` becomes a pill that opens a modal. Every candidate row in it carries
the party's bed count, the need glyphs **already coloured against this room**,
last year's cabin and a plain fit verdict — none of which fits in a 244px card,
which is the whole argument for the change. Typing a name that matches nobody
swaps the region below the box to a write-in form, and backspacing back into a
match swaps it straight back.

It also collapses **~82 mounted comboboxes to one**. The inline picker was
mounted on every placeable card, each holding the entire unplaced queue and
memoising an annotate-and-sort over up to 63 parties, re-run on any board
re-render. The modal exists only while it is open.

Relates to #2072. Stage 2 of three; stage 1 is the glyph vocabulary it depends
on, stage 3 moves this pill into the card's new footer.

**It also carries stage 1's seven post-merge findings and a second round of
owner rulings on both surfaces** — the two sections at the end of this body
cover them, including why the card fixes ride here rather than in a PR of their
own.

## This supersedes a standing ruling, deliberately

The 2026-08-09 ruling on kindred#2080 was explicit: **"not a popover and not a
second surface"**, option A, literally Hold's shape. AS2 (owner, 2026-08-19)
supersedes it **for this control only**, and the width is what buys the
supersession. `UnitAvailabilityControl` and the merge/split pills stay inline,
because none of them has information that wants width.

That it is *also* a convergence with summer is worth stating: `BunkSwapModal`
already solves "pick somebody for this container, starting from the container"
with a dialog. `lodging-board-vs-summer.md` §3's Actions row — which read
"open — decision, not a build" — is **closed** by this PR.

## W3, and why each rule is in a test rather than a comment

Each of these is the kind of thing a reviewer softens into a nicety, so each is
an assertion in `AssignFamilyModal.test.tsx`:

- **One live input, and it IS the occupant name.** No separate occupant field.
- **It never locks**, and the test pins the *same DOM node* through the flip —
  a remount would lose focus and the caret mid-word.
- **Only the region below it swaps.** Header, input and footer stay mounted, so
  the panel does not jump under the cursor.
- **Backspacing into a match swaps it back**, committing nothing in either
  direction — and destroying nothing either, so a typed note survives the round
  trip.
- **`Enter` saves from a FIELD, never from the search box.** This is the
  ruling, not a keybinding detail: a family name one character off matches
  nothing, and a write-in is silent about having been the wrong thing to do.
  The list rows are real `<button>`s, so a keyboard still has a path — that is
  affordable here in a way it was not across ~82 inline lists.

## The staff-facing number, in your words

The modal header states beds **free**, where the card states placed-of-capacity.
Ruling, verbatim:

> The modal states beds FREE because that is the question being asked at the
> point of placement — will this party fit in what is left. The card's N/M is
> unchanged and over-capacity still means placed exceeds capacity everywhere on
> the board.

Two framings of one arithmetic, neither redefined. The card's own figure is
untouched and still pinned.

## The fourth grading table

The kindred#2072 survey counted three tables grading the ruled needs. There
were **four** — `placementCandidates.ts` was the one it missed, and its own
module doc had already flagged the divergence: a family asking for a fridge or
step-free access was hatched mid-drag on a cabin that could not supply it and
annotated `fits` in the picker.

It calls `needGlyphs.ts` now. What it had **right**, and what the resolver had
to learn rather than flatten: a candidate has no placement, so its bathroom must
be graded against the cabin under consideration and never against
`party.effective_bathroom` — which would annotate every unplaced party
identically on every cabin. That is the resolver's new `prospective` reading,
and this module is why it exists.

## One extraction worth naming

`NeedGlyphMark` and the warn ink move out of `FamilyCard.tsx` into their own
`NeedGlyph.tsx`, because two surfaces draw the glyph now — the family card
against the cabin a household is IN, the modal against the cabin being
CONSIDERED.

That split is the lesson from the `Whole building` chip, which lived inside
`FamilyCard.tsx` unexported and was therefore **reproduced by hand** in
`MapUnitPopover.tsx` — "same label, same tokens, same icon", with a comment
admitting the copy. Two copies drift. This one cannot.

The move carries the glyph's accessible name and its 20×20 box (both landed in
stage 1) with it unchanged.

## Kept deliberately

- **Nothing is filtered out of the list.** 6 of 118 units carry a private
  bathroom against 45 parties asking for one, so a list narrowed to "what fits"
  would be empty most of the time and staff would go back to dragging. This is
  the one place the modal deliberately diverges from `BunkSwapModal`, which
  does filter — summer's gender rule is a hard constraint; amenity fit here is
  advisory.
- ~~**The row identity is the attending adults, not the children.**~~
  **Superseded 2026-08-20 — see the section below.** The reasoning was that
  matching the mock would mean a second implementation of the family card's
  children-run in the change that exists to collapse duplicated rules. The
  owner ruled for the children, and the run was *moved* rather than copied.
- **The `People` field is not built.** `lodging_write_ins` has `occupant_name`
  and `note` and nowhere to put a count. Per your ruling it ships without one,
  and kindred#2503's body now explicitly owns the field, the `Enter`-saves rule
  it must respect, and the beds-free arithmetic it feeds.

## Also struck

| Struck | Why | Defended by |
| --- | --- | --- |
| `PlaceFamilyPicker` — the inline combobox, 453 lines | superseded by AS2; the shape it had is exactly what a later reader would restore from the ruling it honoured | `LodgingUnitCard.test.tsx`, "mounts no inline combobox anywhere on the card" |
| the per-need NOTES on candidate rows (`No private bathroom`, `No power`, `Some rooms have power`) | the row draws the glyph beside them, and a sentence repeating it is the "one fact twice" N2 removed from the family card | `placementCandidates.test.ts`, "writes NO note for a need" |

`notes` is now what no glyph can say, which today is capacity alone.

Four comments pointing at the deleted picker were **repointed, not dropped** —
including `UnitAvailabilityControl.test.tsx`'s list of where each deleted
guarantee went, which is precisely the kind of pointer that stops a guarantee
being findable and then stops it being true.

## Known divergence this ships with, deliberately

**A combined container's card and this modal disagree about its capacity, for
one release.** The modal reads `effectiveSleeps` (the container's delta plus its
rooms); the unit card still reads the raw `unit.sleeps`. Measured on the
production snapshot: **all 15 containers record `sleeps = 0`**, which the API
maps to `null` and the card renders as an em dash — so on each of the four
combined containers the card says *"capacity not recorded"* while the modal
says, correctly, `8 of 8 beds free`.

The modal is the surface that is right: `countUnmeasuredSpaces` and the map peek
already use `effectiveSleeps`, and the card is the last reader of the raw value.
Raised by CodeRabbit; **owner ruled it fixed in stage 3**, since the card is
that stage's file. Marked at the site so it cannot be lost.

It also falsified a claim in `lodging-board-vs-summer.md` — that a container row
"carries its measured whole-house `sleeps`… one building records 7 against rooms
summing to 6". No container records anything. That paragraph is corrected in
this PR rather than left standing.

## Not done

- **The card half of the capacity divergence above** — stage 3.
- kindred#2503 — the `People` field and the write-in party size.
- Stage 3 — the footer this pill moves into, T2, the padding, and the board
  half of the cuts.
- `#2072-rail` — deleting the 3px coloured stripe from every household roster
  row. It is half of what kindred#2072's title promises and is in **no** stage,
  so that issue stays open after all three land.

## Verification

```
cd frontend && ./node_modules/.bin/vitest run   # 453 files, 7150 tests, all pass
bash scripts/pre-push-verify.sh                 # prettier, eslint, tsc, vitest — pass
```

30 new tests for the modal; 8 existing board and card tests retargeted from the
combobox to the pill, each still asking exactly what it asked before. One
existing test's name finally became true: "sends the unit staff clicked, with
the occupant **and note** they typed" asserted `reason: ''` with a comment
explaining that the inline box collected no note. It does now.

⚠️ Hand-written `.tsx`, so the visual gate applies — do not merge until the
modal has been opened and looked at.

---

# The 2026-08-20 rulings, applied

The owner compared this modal against the review artifact and issued six
rulings. All six are implemented, in `77f28598` and `6b481976`.

| # | Ruled | Was | Now |
| - | ----- | --- | --- |
| 3.1 | card width | 672px (`size="lg"`, a default nobody chose) | **520px** exactly, the artifact's own `.modalcard`, through a new opt-in `maxWidthClassName` on `ui/Modal` |
| 3.2 | the fit verdict goes green / red | 12px, normal weight, `text-muted-foreground` — the same ink as last year's cabin beside it | 11px bold, `green-700`/`green-300` and `red-800`/`red-300` |
| 3.3 | the whole vertical rhythm | `px-6 py-4`, `gap-3`, separator 12px below the box and 8px above the rows | the artifact's 14px inset and 9px gaps; separator 9px each side; rows 6px apart; both rules dashed |
| 3.4 | the row font | 14px | **13px**, name `font-semibold` |
| 3.5 | rows name the children | the attending adults | the children with ages, from `householdIdentity.childrenRun` — the SAME derivation `FamilyCard`'s `ChildList` calls |
| 3.6 | rows take the input's ground | transparent | `bg-background`, measured equal to the search box in both themes |

**3.1, stated as the ruling asked.** `max-w-[520px]` exactly, not `size="md"`'s
512. The width is the argument for AS2 superseding the "not a second surface"
ruling, and it is the one dimension in this dialog that was measured rather
than chosen — a row that truncates a name eight pixels earlier than the design
it is compared against is not the design. `ui/Modal`'s new prop is opt-in and
no other caller is touched.

**3.2's green is `green-700`, not the board's `forest`, and that is a departure
from the ruling's own steer.** Measured in Chromium against this app's tokens:
`forest-700` resolves to `#003917` against a `--foreground` of `#0c3125` — a
**1.08:1** contrast between the verdict and the row text beside it, which is
the exact defect the ruling names. `green-700` is `#008236`, 2.87:1. It is also
the ramp the other half of the verdict already comes from (`red-800`, the warn
ink `NeedGlyph` owns). Raised for the owner rather than settled here.

**3.4 is settled, and it was both things.** The FACE already matched — the
title is Fraunces and the body is DM Sans in both, because `index.css` applies
`--font-display` to `h1–h3`. The difference was SIZE (14px vs 13px), plus the
app's own −0.176px of body tracking, which the artifact does not have and which
is correct.

## Two focus defects, found while measuring, fixed

- **The dialog opened with focus on the Close button.** The search box carried
  React's `autoFocus` and never kept it: `ui/Modal`'s focus effect runs after
  commit and lands on `focusable[0]`, which is Close, because a custom header
  renders it above the body. So the dialog whose own doc says it "exists to be
  typed into" opened on a control that swallows printable keys and closes on
  Space or Enter.
- **Closing it dropped focus on the floor.** `previouslyFocusedRef` captured
  `document.activeElement` *after* `autoFocus` had moved it inside the dialog,
  so closing restored focus to the modal's own detached input and it fell to
  `<body>` rather than back to the Assign pill.

Both are closed by an opt-in `initialFocusRef` on `ui/Modal`; nothing changes
for a dialog that passes none. **Five other dialogs still declare `autoFocus`
and are still broken the same way** — `ManualResolutionModal`, `MergeDialog`,
`ResolveDialog`, `ScenarioEditModal`, `NewScenarioModal`. Untouched here; they
are separate surfaces with their own review.

## Also struck

| Struck | Why | Defended by |
| --- | --- | --- |
| `autoFocus` on the search box | it never won, and it broke focus restoration as a side effect | `AssignFamilyModal.test.tsx`, "gives focus back to whatever opened it" — mutation-checked: re-adding `autoFocus` fails it |
| `youngestFirst`'s export | ordering and dedupe have to see the same array, since the surname lift reads index-for-index against the sorted order | it is module-private; `childrenRun` is the only caller |

## Verification

```
cd frontend && ./node_modules/.bin/vitest run   # 453 files, 7183 tests, all pass
bash scripts/pre-push-verify.sh                 # prettier, eslint, tsc, vitest — pass
```

Six new class assertions pin the three rulings that shipped with no test at all
(the width, the rhythm, the type). Each was mutation-checked by reverting the
ruling and confirming the test goes red. Their comments say plainly that jsdom
cannot check pixels and name the browser number each class stands for.

⚠️ **Every one of those open calls is now answered — see the section below.**

---

# The second pass, 2026-08-20 — every open call answered

The owner drove the built dialog against the artifact a second time. Everything
below was measured in Chromium on the real module, mounted through Vite, and
each option was mocked by changing one property in the live DOM rather than
drawn.

| Ruled | Was | Now |
| ----- | --- | --- |
| **The row grades the beds LEFT** | `partyBeds <= effectiveSleeps` — the room's whole capacity | `partyBeds <= capacity - occupants`, the question the header has answered since 2026-08-19. The note counts free beds with it: `Over capacity · needs 3, 2 free` |
| **One child's age follows their first name** | `Isla Nguyen (3)` | `Isla (3) Nguyen` — the floor in `dedupeChildNames` moved from two children to one, so the family card and the details panel moved with it |
| **The second line is always drawn** | dropped when it held only a verdict, and the verdict rejoined the identity | always drawn. The verdict is at its longest exactly when the row has nothing else, so the collapse crushed the name it was protecting |
| **No glyph strip when there are no glyphs** | an empty flex child still took the line's 6px gap | omitted, so last year's cabin starts where the name above it does |
| **No rule under the header** | a `border-b` this dialog drew itself | 9px of undivided ground, as the artifact has it |
| **The close button is centred in its band** | `top-4` + a 36px box in a 51px header | `closeAlign="center"`, an opt-in on `ui/Modal`; no other dialog moves |
| **The card's border stays 1px** | 1px, unresolved | 1px, ruled. `ui/Modal`'s weight stands; the artifact's 2px is not this app's grammar |
| **The footer hint is centred in its band** | 4px under the rule, 14px above the card's edge | `py-[9px]`, same 34px band, line in the middle of it |
| **The Write in button moves under the field it commits** | in the footer, which grew 35px → 51px on the flip | in the write-in region. Card 456px in both modes, footer 35px in both |
| **The Returning / First-time mark is 20px** | 16px, bottom-aligned against 20px glyph chips | 20px, ink level with the chips. R3's row in the vocabulary doc carries the new size and the measurement |

**The write-in flip was the last of the jump W3 forbids, and it hid in the one
place nobody measured.** Header, search box, swap region and the footer's *top*
edge all travelled 0px — the number this PR originally reported — while the
footer itself grew to hold the button, dropping the card's bottom edge 16px.
Four fixes were mocked; the ruled one moves the button into the fixed-height
region, which is the only part of this dialog meant to change. It diverges from
the artifact, which draws the button in the footer and simply lets its card
grow.

**`spanWidth` gates the remainder exactly as it gates the header's
over-capacity claim.** A party holding several rooms is drawn on each of them
(#2010), so `occupants` counts the same people twice and legitimately
over-states; subtracting it would print `does not fit` on rows that fit, with
the header beside them saying nothing was wrong.

## Stage 1's post-merge findings ride here, and why

#2505 was reviewed again after it merged; seven findings came out of it and all
seven are fixed in this PR rather than in one of their own. They live here
because this branch already rewrites the files they touch — `NeedGlyphMark`
moved out of `FamilyCard.tsx` into `NeedGlyph.tsx`, and `ChildList`'s
derivation moved into `householdIdentity.ts`. Re-authoring them against main's
pre-extraction tree and then rebasing this branch onto that would have meant
hand-porting each one through the same conflict, which is how a fix in this
stack has silently vanished before.

| Finding | Fixed as |
| ------- | -------- |
| `unknown` coverage read as `fits` on every glyph | `unknown` grades **unmet** for all four glyphs (owner: *"unknown values should not equal fits, across all surfaces on the glyphs"*). The drag-time hatch keeps the old reading, deliberately — 102 of 118 cabins record `ramp_coverage: unknown`, so folding it in takes a step-free household from 32 hatched cabin-pairs to 848 |
| the Returning mark wore `forest-700` | `green-700`/`green-300` — 1.08 : 1 against the card's own text became 2.87 : 1 |
| the single-parent icon hung below the text it stood before | 9px on the baseline, ink 0.50px above the cap and 1.00px clear of the descender |
| a struck-chip pin that could not fail | the fixture arms the demand side too; mutation-checked by restoring the chip |
| two `unit` props whose deletion left 1415 tests green | one pin per surface, both mutation-checked |
| an amber coupling that read only one of the two marks it named | asserted as an equality |
| a false doc claim and two stale comments | corrected at four sites |

## Verification, second pass

```
cd frontend && ./node_modules/.bin/vitest run   # 453 files, 7213 tests, all pass
./node_modules/.bin/tsc --noEmit                # clean
./node_modules/.bin/eslint src/components/weekend  # clean
```

Every geometry ruling is pinned as a class carrying the browser number it
stands for, because jsdom has no layout engine and a test named for a layout
property that renders in jsdom asserts nothing — that is how a 133px jump
survived this file for a release. Each pin was mutation-checked.

**Filed rather than built:** #2507 — standardise every dialog's close button on
the artifact's 18px in-flow mark. The owner preferred it to the centring above
but wants it landed across ~20 dialogs at once; `closeAlign` should be deleted
when it lands.

**Still not verified:** the scrolling list. Four rows never scroll, so forty
two-line rows, the scrollbar's effect on row width, and whether the last row
clears the footer rule are untested. CodeRabbit's review predates every ruling
on this page.


adamflagg:
@coderabbitai review

Focus, if it helps aim the review:

- **This deletes `PlaceFamilyPicker` (453 lines + its 524-line test) and replaces it with a modal.** The risk is a guarantee that lived only in those tests and did not get re-homed — the write-in offer's three conditions, the per-card saving gate, and the refusal to pre-filter the candidate list by fit.
- **`placementCandidates.ts` was a fourth grading table** and now calls the shared resolver in a new `prospective` reading — a candidate has no placement, so its bathroom must be graded off the cabin under consideration, never off `party.effective_bathroom`. Worth checking that reading is applied consistently and that no verdict moved other than the intended fridge/step-free addition.
- **The modal's header states beds FREE while the unit card states placed-of-capacity.** Deliberate, owner-ruled, and the card's figure is untouched — but it is two framings of one arithmetic on one screen, so a second opinion on whether anything can now disagree is useful.
- **`Enter` in the search box must never save.** It is what stops a mistyped family name silently becoming a write-in instead of a placement. If there is any path where a keystroke commits from that box, that is the highest-severity thing in this diff.

adamflagg:
### Code review

Found 2 issues:

1. Clicking a need glyph on a candidate row silently PLACES the family. The glyphs are rendered inside the row's own `<button onClick={() => choose(party)}>`, and `NeedGlyphMark` renders a `ui/Tooltip` whose trigger is itself a real `<button>` with no `stopPropagation()` — so a click on the glyph bubbles up and commits the placement. Verified by probe: `fireEvent.click(need-glyph-bathroom)` calls `onSelect` once. `NeedGlyph.tsx` scopes its own validity to the other consumer — "Valid HTML on the family card because the chip row is a SIBLING of the card's own `<button>`, never its child (kindred#2222)" — and `FamilyCard.tsx` keeps that sibling shape (commit `bc5d2ba4`, #2227). No test clicks a glyph, and neither eslint nor jsx-a11y sees the nesting because it is built from components rather than parsed HTML.

https://github.com/adamflagg/kindred/blob/855f1d2c7f644d23163445731e68d919b776379d/frontend/src/components/weekend/AssignFamilyModal.tsx#L386-L393

https://github.com/adamflagg/kindred/blob/855f1d2c7f644d23163445731e68d919b776379d/frontend/src/components/ui/Tooltip.tsx#L254-L259

https://github.com/adamflagg/kindred/blob/855f1d2c7f644d23163445731e68d919b776379d/frontend/src/components/weekend/NeedGlyph.tsx#L73-L79

2. `candidateFit`'s JSDoc now contradicts the function under it: it says "there is no step-free dimension to add — `is_accessible` exists on the UNIT with nothing on the party side asking for it", while the body five lines below calls `resolveNeedGlyphs(party, unit, 'prospective')`, whose `NEED_GLYPHS` includes `{ key: 'step_free', flag: 'needs_step_free' }`. It also contradicts two comments this PR added in the same file — the module doc's "this grades all four plus capacity" and the inline "All four ruled needs, where this table used to carry two".

https://github.com/adamflagg/kindred/blob/855f1d2c7f644d23163445731e68d919b776379d/frontend/src/components/weekend/placementCandidates.ts#L135-L141

Issue #2510, opened 2026-08-20
feat(pb,api): the registry does not model whether two bathroom groups in one building are reachable without going outside


adamflagg:
## The gap

The registry models **which rooms share a bathroom** (`bathroom_group`) and **which rooms are inside which building** (`parent_unit`). It does not model **whether you can get from one room to another without going outside**.

That third fact is the one that decides whether a whole-building let genuinely gives a family the bathroom they asked for, and it is currently unrepresented — so any rule we write about multi-group containers is guessing.

## Why it came up

`container_bathroom` (`api/services/lodging_rules.py`) inherits a bathroom for a container from its leaves. Today it answers `("none", "")` whenever the leaves span **two** bathroom groups, even when every room has a bathroom. That is a false negative on **four buildings** in the 2026 registry, all of which report "no bathroom" while every room in them has one. ⚠️ **Figure corrected 2026-08-24: the room count is 18, not 22.** The distinct leaf-room count under the four multi-group containers is **18**; **24** is the double-counted sum across the nested pair; **22** appears to have been 18 leaves plus the 4 container rows and does not reproduce at any grain. The "four buildings" half — the load-bearing half — reproduces exactly.

The obvious fix — "a whole-let covers every group inside it, so report a bathroom" — was **rejected on owner review**, because it is only true if the groups are mutually reachable.

> **Owner, 2026-08-20:** *"this depends on if 'rooms' are physically inaccessibly separated or not, which is something we aren't currently accounting for."*

The worked example, paraphrased to keep unit names out of source (see CLAUDE.md §4):

- One building's **upper pair** share an internal living space, and their bathroom is inside it.
- Its **lower pair** share an internal corridor, and their bathroom is in that corridor.
- **To get from the upper pair to the lower pair you go outside.**

So each group genuinely has a bathroom *within it*, and the building has two bathrooms — but a family let the whole building does not have one bathroom they all share; they have two, in two places, one of which requires going outdoors to reach. For a request that is explicitly *"a bathroom that doesn't require you to leave your cabin, for a medical or accessibility-related reason"* (`FAM CAMP-bathroom`, cm_id 274056), that distinction is the entire question.

## The open question, which must be answered before the code

**Is a bathroom group ALWAYS internally reachable, and are two groups in one building ALWAYS separated?**

If both hold universally, the rule is derivable from what we already store and no new field is needed. If either is sometimes false, we need to record reachability per group (or per group-pair) and the rule has to read it.

[...]

Issue #2503, opened 2026-08-19
feat(pb,frontend): optional party size on a write-in, and use it as the occupancy numerator


adamflagg:
## What

Give a write-in an **optional integer party size**, and use it as the unit card's occupancy numerator.

**Gated behind the kindred#2072 board GUI PR landing first.** This is a follow-on, not part of that PR.

## Today

`lodging_write_ins` has exactly two content columns: `occupant_name` and `note`. There is no size, and that absence is deliberate — `LodgingUnitCard.tsx:560-561`:

```ts
const wholesaleWriteIn = writtenInto && occupants === 0
const occupancyFigure = wholesaleWriteIn ? '—' : String(occupants)
```

A write-in occupies the room *wholesale*: the card draws an **em dash** as the numerator rather than assert a number it does not have. The card's own comment states the reasoning — `0/5` beside a full room is a lie, and `5/5` is a different one.

## The change

- Add a nullable integer to `lodging_write_ins` (and `_draft`).
- Offer it as an optional field wherever a write-in is created or edited — including the new Assign modal from #2072.
- When set, it becomes the occupancy numerator instead of the em dash.

## Consequences worth deciding in the issue, not in review

1. **A write-in can become over capacity.** Today it structurally cannot — the em dash short-circuits `capacityKnown && occupants > sleeps`. With an integer, a hand-typed number can turn a card's figure red. That is probably right, but it is a new red state sourced from free typing rather than from roster data.
2. **Both states stay drawable.** Optional means a write-in with no size still shows the em dash, so the card needs to handle both, and so does anything aggregating occupancy.
3. **Aggregates.** Anything summing placed people per weekend has to decide whether write-in sizes count. They are not roster people, so a naive sum changes reported occupancy.
4. **The three-way merge.** `effective_bathroom`-style container inheritance does not apply here, but a combined container holding a written-into room needs a stated rule for whose size shows.

## Acceptance

- A write-in may carry a size or not; both render correctly.
- The over-capacity path is exercised by a test with a hand-set size above `sleeps`.
- Whatever is decided about aggregates is written down in the issue before it ships.

adamflagg:
Body edited to record the 2026-08-19 owner ruling on kindred#2072's stage 2, so the next reader implements from it rather than from the mock:

[...]

PR #2511, opened 2026-08-20
fix(api): a container reports the amenities its rooms actually have

### LINKED ISSUE #2501: fix(frontend,api): bathroom need is graded on exclusivity but asked on in-cabin access
## ★ OWNER RULING 2026-08-20 — THE RULE IS PRESENCE, NOT EXCLUSIVITY. This is no longer a proposal.

> **Owner:** *"the glyph should not grade exclusivity, just 'do they have a bathroom (shared or private)'."*

The fix below is approved. What remains is execution and the Adult-form gate at the bottom, which
is a question about what the FLAG means on adult weekends, not about the rule.

⚠️ **This body's code anchors were written against a tree that no longer exists.** kindred#2072 /
#2505 deleted `rosterAttention.ts`'s `VERIFIABLE_NEEDS` table and its `satisfiedBy` functions, and
moved every verdict into one grader. The premise and the evidence below are unchanged and still
correct — only the identifiers moved. Corrected in place 2026-08-20; see "The fix" for the live one.

## The bug

`needs_private_bathroom` is graded on an **exclusivity** axis. The question that sets it asks about an **in-cabin** axis. They are not the same axis, and the mismatch marks families as unmet while they sit in exactly the accommodation they asked for.

Verbatim, from `docs/reference/family-camp-field-provenance.md:180` — the staff-supplied record of the Camper Registration form, field `FAM CAMP-bathroom` (cm_id 274056):

> "Does your family require access to a bathroom that doesn't require you to leave your cabin for a medical or accessibility-related reason?"

That is *in-cabin vs walk-to-the-bathhouse* — in registry terms `bathroom != 'none'`. A `shared` unit satisfies it as fully as a `private` one.

The live rule is `frontend/src/components/weekend/needGlyphs.ts:205`, inside `bathroomCoverage`:

```ts
const value = reading === 'prospective' ? (unit.bathroom ?? 'unknown') : party.effective_bathroom
if (value === 'private') return 'all'
if (value === undefined || value === 'unknown') return 'unknown'
return 'none'          // 'shared' lands here -> unmet -> red
```

and `api/services/lodging_rules.py:138-167` defines `shared` as a bathroom *inside the unit* that two families split, upgraded to `private` only when the placement covers the whole `bathroom_group`.

The word "private" appears nowhere in the source. It entered at the column name and propagated to the UI label and the unmet text. Camp's own housing copy uses the other axis: "minimal housing with **attached** bathrooms".

## Evidence that families mean "attached", not "exclusive"

90 free-text answers sit behind the bathroom question itself (65 households, 2026). Re-read by hand:

- **At most 5 rows — 3 to 5 households, ~5%** — describe something a shared bathroom cannot meet.
- The rest describe **speed and distance**: 18 mention night waking, plus incontinence, gastrointestinal conditions, infants, elderly parents, crutches. 9 ask only to be sited near the bathhouse.
- 0 contain "en suite"/"ensuite"; 0 contain "own bathroom".

## The fix

⚠️ **`bathroomCoverage` has TWO readings and the fix must cover BOTH.** An earlier version of this
body proposed a placed-lane change only, which would have left the Assign modal and the picker still
grading exclusivity while the roster graded presence — replacing a card/glyph contradiction with a
board/modal one.

- **placed** reads `party.effective_bathroom`
- **prospective** reads `unit.bathroom`

Both must treat `'shared'` as satisfying, i.e. the test becomes presence (`!== 'none'`) rather than
`=== 'private'`, on each reading.

**Already shipped, do not redo:** the glyph label is `'Bathroom in unit'` as of #2505. What is left
on the wording side is `ROSTER_NEED_WORDING` in `rosterAttention.ts:102`, still `"No private
bathroom"`.

⚠️ **Fix `container_bathroom` first or in the same PR.** Flipping the roster wording to "No bathroom
in unit" while `container_bathroom` still returns `('none','')` for a multi-group building would make
the roster positively assert *"no bathroom in unit"* about buildings where every room has one. See
the confirmed defect at the bottom of this body.

Consider renaming the column itself; `needs_in_cabin_bathroom` is what 274056 actually asks. That is a bigger change and can be a follow-up.

## What changes

Measured on the production snapshot, 8 non-cancelled family sessions, 45 requesters / 42 households:

| lane | red mismatches today | cleared by the fix |
|---|---|---|
| live | 10 | **7** |
| scenario | 25 | **16** |
| union | 27 | **18** |

Re-verified against the 2026-08-20 snapshot: **7 live households flip red -> green.** Two further
households that look like candidates on the raw column already resolve `private`, because they hold
both rooms of their bathroom group and the exclusivity branch upgrades them — they are green today
and are not part of the 7.

**Every case that stays red is `effective_bathroom = 'none'`** — a cabin with no bathroom in it at all. Clean partition, and a real problem no rule change fixes.

It also returns capacity: today **2 of the 3 live satisfactions are manufactured by merging a whole bathroom group** — spending 2 to 4 rooms to house a party of 1 to 3 — purely to meet a strictness the form never set. And supply stops being the binding constraint: 28 attached units serve the need instead of 6, of which only 1 is even reachable by a family (the other 5 are `staff_default`, and `lodging_availability` is empty in every year).

## The cost, stated plainly

The ~3 to 5 households a year who genuinely need exclusivity stop getting an automatic red mark. That is acceptable because **every one of them wrote a narrative explaining why** — they are a read-the-note case, not a boolean case.

Owner accepted this tail explicitly in the 2026-08-20 ruling. Note the shape of what is accepted:
`shared` means two families split a bathroom *inside the cabin*, so a medical requester can now be
graded satisfied while sharing with strangers.

## Tests that pin the current behaviour — BOTH must flip

| file | asserts today |
|---|---|
| `frontend/src/components/weekend/needGlyphs.test.ts:142` | `'shared'` grades `'none'`, commented "until kindred#2501" |
| `frontend/src/components/weekend/rosterAttention.test.ts:218` | the same, at the roster level |

The second was not previously listed here and is easy to miss — it fails only after the first is fixed.

## ⚠️ Gate before merging

`Adult-Bathroom` (cm_id 274053) supplies **19 of the 66 flagged households** and **its question wording has never been read** — the Adult registration form has not been audited, and only bare Yes/No values are stored. Get the verbatim prompt off that form first:

- If it asks the same in-cabin question, this fix covers 100% of the flag.
- If it asks about exclusivity, adult weekends need their own answer and this fix must be scoped to the camper partition.

The 2026-08-20 ruling settles the RULE; it does not answer what the adult form asked.

## Also worth logging while in this code

- 6 households mention a bathroom in the *accommodation* narrative and carry no flag at all — the sync routes that field to fridge and step-free, never to bathroom.
- ✅ **CONFIRMED 2026-08-20, and it is worse than "one merge away".** The 2-room building whose rooms
  both record `bathroom = 'none'` while sharing a non-empty `bathroom_group` is the only such group in
  the registry, and both rooms are flagged as near a bathhouse. `container_bathroom` returns
  `('shared', group)` on group identity alone — it never reads any leaf's bathroom value — and
  `effective_bathroom` then upgrades a whole-let to **`private`**, i.e. a GREEN glyph on a medical
  bathroom request for two rooms with no bathroom. Reproduced directly:
  `container_bathroom(frozenset({g})) -> ('shared', g)`; `effective_bathroom(...) -> 'private'`.
  The same blindness produces 5 false negatives in the other direction. Tracked separately; it is a
  defect in `api/services/lodging_rules.py:318`, not in this rule.

### LINKED ISSUE #2502: fix(api): _build_units scores a container's bathroom on its own blank row, not its covered rooms
## ⚠️ THE PREMISE IS RIGHT; THE PROPOSED FIX IS NOT. Read this before implementing.

Corrected 2026-08-20 against the tree and the production snapshot. The bug is real and reproduces
exactly as described. Two things changed:

1. **The fix below ("feed the drawn slot's covered leaf codes into `_build_units`") cannot answer for
   an UNOCCUPIED container**, and that is the majority case — most containers are not placed into.
   Slot codes only exist once something is drawn. Build it as a **fourth in-place resolver** instead,
   beside the three that already exist. Details under "The fix".
2. Two identifiers named below do not exist. Corrected in place.

## The bug

`_build_units` in `api/services/lodging_roster_service.py:2057` scores each unit's bathroom by calling
`effective_bathroom(unit's own 'bathroom', group, group_members, frozenset({code}))` — a **one-element**
slot containing the unit itself.

A container's own registry row carries `bathroom = 'none'` — 15 of the 118 2026 rows are containers, and
**all 15 publish `'none'`, while 13 of the 15 have at least one room that records a bathroom.**
`lodging_rules.py:162` short-circuits (`if bathroom != "shared": return bathroom`), so the stored
`'none'` reaches the wire untouched.

`container_bathroom` (`lodging_rules.py:285`) is the function written to inherit from leaves. It has
**exactly one caller in the repo** — `_resolve_party_bathroom` at `:1019`, on the *party* path. Nothing
on the unit path calls it.

**Why the other three amenities do not have this bug:** `power_coverage`, `fridge_coverage` and
`ramp_coverage` are resolved by a **second pass** that runs after the tree index exists —
`build_roster:1445/1449/1452`, all funnelling into `_resolve_amenity_coverage:1164`, whose loop walks
`index.leaf_codes_under()`. Bathroom predates that pattern and was never moved across. The mechanical
blocker is ordering: the index does not exist yet when `_build_units` runs.

## What staff see

⚠️ The original wording here ("1 card reads `Private`, 12–14 read `Shared`") predates the vocabulary
change — **the card prints no bathroom word at all now**; it draws or omits an icon
(`LodgingUnitCard.tsx:926`, gated on `bathroom !== 'none'`). The count is unchanged; the rendering is.

- combined containers show **no bathroom icon at all**, including buildings where every room has one
- the picker annotates those same containers as not meeting the need, on the very units that satisfy it,
  because a whole-bathroom-group merge is exactly what `effective_bathroom` upgrades to `private`

⚠️ `placementCandidates.bathroomVerdict` **does not exist**. The real path is
`placementCandidates.ts:186` -> `resolveNeedGlyphs(party, unit, 'prospective')` -> `needGlyphs.ts:204`
(`bathroomCoverage`), whose prospective reading takes `unit.bathroom` — the same unresolved field.

So the card, the picker and the roster disagree about the same unit.

**There is a second, sharper form of the disagreement**: prospective vs placed. A whole-let container
resolves `private` on the party path (via `container_bathroom`) but `'none'` on the unit path. The same
container therefore grades UNMET in the Assign modal and MET once the family is placed in it.

## The fix

Add `_resolve_bathroom(units, index)` as a **fourth in-place resolver**, called beside the existing
three at `build_roster:1445-1452`, using the same `index.leaf_codes_under()` walk and inheriting the
same `is_active` / `is_confirmed` gates. Then delete the inline `bathroom=effective_bathroom(...)` from
`_build_units`.

Two things to get right:

- ⚠️ **`build_summary:1634` calls `_build_units` and NONE of the resolvers.** The inline computation is
  currently the only thing setting bathroom on that path, so deleting it without wiring the resolvers
  into `build_summary` regresses the summary counts. Wire all four into both paths.
- ⚠️ **Depends on `container_bathroom` being fixed first.** It grades on group IDENTITY and never reads
  any leaf's bathroom value, so today it returns 5 false negatives and 1 false positive across the 15
  containers — including a GREEN verdict on a medical bathroom request for a two-room building whose
  rooms both record `bathroom = 'none'`. Rolling that up to the unit card without fixing it would
  publish the false positive to a second surface. See kindred#2501's closing section.

## Why it matters now

kindred#2501 is **ruled** (2026-08-20): the rule becomes presence, not exclusivity. That makes the
attached-vs-bathhouse distinction the one the board draws, so a unit scored on a blank container row
will read "no bathroom" on a building whose rooms all have one. This bug gets *more* visible after that
change, not less.

⚠️ Correction: an earlier reading of this issue suggested a forced ordering with #2501. There is none in
the grading direction — measured, **#2501 changes container grading by exactly zero**, because every
bathroom group sits wholly inside one parent, so a whole-let container can only ever resolve `private`
or `none`, never the `shared` that #2501 moves. The original line below is correct as written:

> It is independent of #2501's outcome — the container scoring is wrong under either rule, so this can
> ship on its own.

The one real coupling is the **wording** flip, and it runs the other way: #2501 changes
`ROSTER_NEED_WORDING` to "No bathroom in unit", which would then positively assert that about a
fully-plumbed building. Fix `container_bathroom` before or with that flip.

## Acceptance

- A combined container reports the bathroom its covered rooms actually provide, **while unoccupied**.
- A three-deep container inherits through an intermediate container (18 of 118 units are grandchildren).
- The unit card, the placement picker and the roster agree for every drawn slot.
- The prospective and placed readings agree for the same container.
- `build_summary` counts agree with `build_roster` counts for the same weekend.
- A test covering a combined container whose leaves are `shared` and whose group is fully covered — the
  case that should resolve `private` — and one whose leaves record no bathroom at all, which must not.

adamflagg:
Three amenity-resolution defects found by auditing the container/leaf/merge/split logic against the needs-vs-meets grading, after two were reported from production — **plus the bathroom grading axis itself**, which moves from exclusivity to presence on the owner's 2026-08-20 ruling.

## ⚠️ A shared bathroom now SATISFIES a request that previously showed red

This is a change to what a staff-facing mark means, so it goes first rather than in a footnote.

Until now, a household flagged `needs_private_bathroom` was graded met only by a `private` bathroom. A cabin with a bathroom **inside it that two parties split** drew a red glyph and printed "No private bathroom" on the roster. It no longer does: presence is the axis, and the wording moves with it to "Bathroom in unit" / "No bathroom in unit".

Owner ruling, verbatim: *"the glyph should not grade exclusivity, just 'do they have a bathroom (shared or private)'"* and *"sharing a bathroom for whatever reason still provides people a bathroom."*

The axis was never exclusivity to begin with. The CampMinder question behind the flag asks for *"a bathroom that doesn't require you to leave your cabin"* — so `shared` means a bathroom **inside** the cabin, and walking to a bathhouse is not `shared`, it records as `none` and still reads unmet. The column's name is the thing that is wrong here; renaming it is a deliberate follow-up.

**What it moves, measured:** across 2026's three weekends that have bathroom-asking families placed, **7 placed families** move from *unmet* to *settled* on the roster's attention sections — 5, 1 and 1. On the largest of the three, "No private bathroom" went 8 → 0 and "No bathroom in unit" 0 → 3, verified live against both stacks. **What it costs, accepted on the record:** the ~3–5 households a year who need a bathroom nobody else uses lose an automatic red mark and are found by reading the request instead.

This also ends a one-release contradiction rather than creating one: the unit card already drew its bathroom mark on presence, so a room could show "has a bathroom" while the family placed on it showed a red bathroom glyph, off the same field. One axis now, on every surface — card, glyph, roster wording, Assign modal and map peek.

## What was wrong

**A whole-house card showed no bathroom while both its rooms showed one the moment staff split it.** All 15 registry containers store `bathroom = "none"` on their own row — a building is not a room — and 13 of them have at least one room that records one. `_build_units` scored each unit against a one-element slot containing itself, so `effective_bathroom` short-circuited on the blank value and it reached the wire untouched.

Power, fridge and step-free are immune because they are resolved by a **second pass** that runs once the tree index exists. Bathroom predates that pattern and was never moved across — the blocker was ordering, not design. `_resolve_bathroom` is that missing twin.

It answers for an **empty** container, which is the point. The party path was already correct, but only once something was placed, so the same building graded unmet in the picker and met once the family landed in it. Most containers are never placed into. This is also why the issue's original proposal — thread the drawn slot's codes into `_build_units` — could not work; #2502's body has been corrected.

**`container_bathroom` graded on group identity and never read a single leaf's bathroom value.** A `bathroom_group` says which rooms share one bathroom; it never says the bathroom is inside any of them. One registry group names a bathhouse its two rooms walk to, and both record `bathroom = "none"` — identity alone returned a shared group, which the exclusivity branch upgraded to `private` on a whole-let. That is a **satisfied verdict on a medical in-cabin bathroom request for two rooms with no bathroom in them**. It now takes `(bathroom, group)` pairs.

**`has_ac` had no resolver at all.** It sat in the schema between two fields that both have twins, and three surfaces read it raw. Seven of the 15 containers record `has_ac = 0` with AC-bearing rooms, so merging a house hid a mark both its rooms carry.

## Measured against the 2026-08-20 production snapshot

118 units, 15 containers, 103 rooms, all active and all confirmed. Every figure below was re-derived by running the shipped functions over the real registry, not taken from a summary.

| | before | after |
|---|---|---|
| containers reporting the bathroom their rooms have | 0 of 15 | **8 of 15** |
| the bathhouse case | `private` (satisfied) | `none` |
| containers that should report none and changed | — | **0** |
| units answering the bathroom need | 6 of 118 | **36 of 118** |

That last row moved twice, and the two halves are separable: the presence ruling takes 6 → 28, and `_resolve_bathroom` takes 28 → 36. Six comments and two reference docs quoted the old **6 of 118** and the **112 of 118** cards a bathroom hatch would fire on; the live figures are **36** and **82**, and all eight sites are corrected here. The registration side is now 479 rows, of which 66 ask for a bathroom and 48 for power.

## Also shipping — four staff-visible changes beyond the three defects

**1. The details panel and roster toolbar gain the two needs they were silently dropping.** `AccessibilityFlagList` held a second, hard-coded four-branch needs table beside `NEED_GLYPHS`; it now derives from it. The panel gains **Fridge** and **Step-free** rows and the roster toolbar gains the two matching filter chips (4 → 6). Six 2026 households ask for a fridge and no roster surface said so — one of them sits on a card drawing a red fridge glyph while the panel's "Housing needs" section did not mention a fridge. `needs_step_free` is 0 on all 479 rows until a sync runs, so that chip matches nobody today; the wiring is correct and waiting on data (#2072).

**2. The map peek trades a raw boolean for the resolved step-free grade — and shows MORE, not less.** The `Accessible` item read the raw `is_accessible` column, which has no resolver and so carries the same container trap as every other raw read on that card. It is replaced by `Step-free`, gated on `ramp_coverage === 'all'`. Measured over all 118 units: `is_accessible` is true on **2**, `ramp_coverage: 'all'` on **6**, and both of the 2 are inside the 6 — a strict superset, nothing lost. Both `is_accessible` rows are rooms inside a house the board draws whole by default, so on the default draw level the old item rendered on **0** map slots and the new one renders on **4**. Which column staff mean by "accessible" is a genuine product question this does not settle; it belongs to #2327, and this change only stops the peek reading a column with no container resolution.

**3. `rosterAttention.resolvePartyUnit` no longer coin-flips on a multi-code placement.** It returned `members[0]` — the storage order of the `units` relation — which is CodeRabbit's finding on PR #2505. It now builds the set of drawn cards a placement is actually represented by (`representingCards`) and grades against the **worst** of them (`worstCard`).

**Zero verdicts move, re-measured over all 122 of 2026's placements** rather than argued: not one placement's candidate cards disagree on `power_coverage`, and no household that asks for a fridge or step-free sits on a set that disagrees on those either. Nine placements do have cards that disagree on `fridge_coverage` and four on `ramp_coverage` — none of them is asked for by the family in it. The roster's two graded needs are safe either way, because bathroom is read off the server's `effective_bathroom` and never off the card. So this is a latent fix: one merge across a fridge-disagreeing pair makes it visible, and swapping two ids in the relation would have flipped the answer.

**4. The map peek gains an over-capacity gate it never had.** `mapModel` now carries `spanWidth`, and the amber "beds over capacity" verdict is withheld for a party spread across several rooms — the same gate both board surfaces already take. Zero spanning parties exist in 2026, so nothing on screen moves; this is a guard on a reachable state.

**Roster section counts:** `ROSTER_NEEDS` is unchanged at `['bathroom', 'power']`, so items 1–4 above move **no** roster section count. The bathroom axis change at the top of this body does, by exactly the 7 placed families named there — that is the only number on this PR that moves for a reason other than arithmetic, and it is the point of #2501 rather than a side effect.

## What is deliberately NOT fixed here

Four buildings whose every room has a bathroom still report none, because their rooms span two `bathroom_group`s. Widening that was **proposed and rejected on owner review**: it is only sound if the groups are mutually reachable, and the registry does not model that. One building's upper pair share an internal living space while its lower pair share a corridor, and you go outside to get between them — so each group has a bathroom, but a family let the whole building does not have one bathroom they all share.

Filed as **#2510**. The false negative is pinned by a test and called out in both docstrings so it cannot be "fixed" by accident.

A fully-occupied whole-house card still counts its beds as free — that one moves printed numbers on the stats bar and embeds a policy, so it waits for a ruling under **#2503**.

## Review follow-up

A review of this branch found one real defect and five falsified comments; all are fixed in `af0bb825`.

**The defect: two lanes, one field, different answers.** `_resolve_bathroom` filters its leaf walk to `is_active` — a room nobody can be placed in does not answer for its building — and `_resolve_party_bathroom` walked the same tree unfiltered. A container with one live room recording no bathroom and one **retired** room recording a shared one, both in the same group, graded `none` on the card and `private` on the family placed into it: the retired room counted twice over, once to supply a bathroom nobody can use and again to complete the group the exclusivity branch checks. Dormant today (the registry has zero inactive rows) and live the moment staff retire a room. Both lanes take the same filter now, and both hand the **unfiltered** group membership to `effective_bathroom`, so a group with a retired member holds at `shared` on both paths rather than upgrading on either.

**The comments.** Three "only on THIS path" notes and a "third and last read of it" sat directly above the commit that wired all five resolvers into both orchestrators. `_resolve_amenity_coverage` said "all three resolvers above" while the new one called itself the fourth caller. And the summary-path note said four of five resolvers move no number there, leaving the fifth unexplained — the gap CodeRabbit read as grounds to delete `_resolve_bathroom`'s call. The true answer is **all five**, now stated with both halves verified: `_build_counts` reads no coverage field and no bathroom, and `_resolve_party_bathroom` recomputes a container's answer from its leaves rather than reading the container's own column. The call stays, because the asymmetry between the two orchestrators is the defect rather than a consequence of it.

## Also

Removes a real lodging unit name from a comment in `api/`. **The Lodging Name Guard passes with it present** — real unit and area names are also sitting in `pocketbase/sync/*_test.go` — so the guard has a blind spot. Reported rather than swept; not in scope here.

## Testing

TDD throughout — every test written and watched fail first, including the bathhouse case failing as `('shared', ...)` and the retired-room case failing as `'private' == 'none'`, which is each defect reproducing. The two mixed-active guard tests were mutation-checked by removing the filter and reading the exit code.

- 6618 pass, 41 skipped on the pre-push runner (6645 / 23 with a live PocketBase). Frontend: 452 files, 7247 tests, all green.
- ruff, mypy, tsc, eslint, prettier, markdownlint, api-types freshness all green. `scripts/pre-push-verify.sh` exits 0.
- Verified against every production container by running the fixed functions over the real registry, not only against fixtures.

Closes #2501.
Closes #2502.

Refs #2510, #2503, #2327.

### DIFF
--- a/api/schemas/lodging.py
+++ b/api/schemas/lodging.py
@@ -238,6 +238,16 @@ class LodgingUnitSummary(BaseModel):
     # built without the resolution pass must not claim an unmet need.
     power_coverage: AmenityCoverage = "unknown"
     has_ac: bool = False
+    # The AC twin of `power_coverage`, resolved over the same leaf walk and
+    # defaulting to "unknown" for the same reason: a payload built without
+    # the resolution pass must not claim an absent amenity. Seven of the 15
+    # production containers record `has_ac = 0` with AC-bearing rooms, so
+    # merging a house hid a mark both its rooms carry.
+    #
+    # DISPLAY ONLY. Air conditioning has no demand glyph -- ruled on 0 of 184
+    # housing narratives mentioning it, against 54 for a bathroom -- so this
+    # exists to keep the amenity strip honest, not to grade a need.
+    ac_coverage: AmenityCoverage = "unknown"
     has_fridge: bool = False
     # NARROWS `has_fridge` -- it can never contradict its parent, so a consumer
     # reading only `has_fridge` stays correct (the registry's own contract,
--- a/api/schemas/lodging.py
+++ b/api/schemas/lodging.py
@@ -106,9 +106,11 @@ Shareability = Literal["unknown", "shareable", "single_party"]
 # How much of a slot carries one amenity, resolved over its LEAF descendants
 # (kindred#1912). Three grains rather than a bool because both boolean
 # policies fall out of it for free -- `OR == != "none"`, `AND == == "all"` --
-# and because what SOME means differs per criterion: for `is_accessible`, some
-# is worse than none, since a building advertising two step-free rooms out of
-# ten invites the placement that lands in one of the other eight.
+# and because what SOME means differs per criterion: for step-free, some is
+# worse than none, since a building advertising two step-free rooms out of ten
+# invites the placement that lands in one of the other eight. (That grain is
+# graded from `has_ramp`, NOT `is_accessible` -- this named the wrong column
+# until kindred#2502; the two are independent and disagree on five rows.)
 #
 # "unknown" is the absence of evidence, exactly as EffectiveBathroom and
 # Shareability spell their own. `has_power = False` on an unconfirmed row
--- a/api/services/lodging_roster_service.py
+++ b/api/services/lodging_roster_service.py
@@ -1209,10 +1209,12 @@ def _resolve_bathroom(units: list[LodgingUnitSummary], index: _BathroomIndex) ->
     four-value enum every surface already reads and giving it a parallel
     `bathroom_coverage` would have left two fields to keep in step. So it is
     the twin of those four in its walk and its `is_active` filter, and the
-    last amenity that was still answered from the unit's own row. All 15 production containers
-    store `bathroom = "none"` (a building is not a room) while 13 of them
-    have at least one room that records one, so every whole-house card drew
-    no bathroom while both its rooms drew one the moment staff split it.
+    last amenity that was still answered from the unit's own row.
+
+    All 15 production containers store `bathroom = "none"` (a building is not
+    a room) while 13 of them have at least one room that records one, so every
+    whole-house card drew no bathroom while both its rooms drew one the moment
+    staff split it.
 
     ⚠️ CONTAINERS ONLY, unlike the three above. A leaf's own row is already
     the right answer and `_build_units` already computes it correctly --
--- a/api/services/lodging_roster_service.py
+++ b/api/services/lodging_roster_service.py
@@ -1713,9 +1713,21 @@ class LodgingRosterService:
             # Same one-index-per-call rule `build_roster` follows -- see the
             # comment there and `_BathroomIndex`'s own docstring.
             unit_index = _BathroomIndex.build(unit_summaries)
-            # `build_roster` runs the amenity resolvers right here; this path
-            # ran none of them, so a container's bathroom stayed at its own
-            # blank row on the summary while the board resolved it.
+            # ⚠️ THIS PATH RAN NONE OF THE AMENITY RESOLVERS UNTIL kindred#2502.
+            # `build_roster` runs all five right after building its index; this
+            # one built the index and went straight to `_build_counts`, so every
+            # coverage field stayed at its `"unknown"` default and a container's
+            # bathroom stayed at its own blank row while the board resolved it.
+            #
+            # Nothing `_build_counts` reads today is a coverage field, so four of
+            # these five change no number on this path right now. They are wired
+            # anyway, because the ASYMMETRY is the defect: one path resolving and
+            # the other not is exactly how the bathroom gap survived unnoticed,
+            # and the next reader of a coverage field here would inherit it.
+            _resolve_power_coverage(unit_summaries, unit_index)
+            _resolve_fridge_coverage(unit_summaries, unit_index)
+            _resolve_ramp_coverage(unit_summaries, unit_index)
+            _resolve_ac_coverage(unit_summaries, unit_index)
             _resolve_bathroom(unit_summaries, unit_index)
             parties = self._build_parties(
                 session_type=_s(session, "session_type"),
--- a/api/services/lodging_roster_service.py
+++ b/api/services/lodging_roster_service.py
@@ -925,7 +925,7 @@ class _BathroomIndex(NamedTuple):
 
     units_by_code: dict[str, LodgingUnitSummary]
     # Immediate children only, keyed by the PARENT's code. Nesting (a
-    # container inside a container, e.g. Doctor's House under a larger
+    # container inside a container, e.g. an apartment under a larger
     # block) is walked at read time in `leaf_codes_under`, mirroring
     # `drawn_units`' own upward walk of the same `parent_code` relation.
     children_by_parent: dict[str, tuple[LodgingUnitSummary, ...]]
@@ -1015,8 +1015,10 @@ def _resolve_party_bathroom(unit_codes: list[str], index: _BathroomIndex) -> str
         if unit.is_container:
             leaves = index.leaf_codes_under(code)
             occupied |= leaves
-            leaf_groups = frozenset(index.units_by_code[leaf].bathroom_group for leaf in leaves)
-            inherited_bathroom, inherited_group = container_bathroom(leaf_groups)
+            leaf_bathrooms = frozenset(
+                (index.units_by_code[leaf].bathroom, index.units_by_code[leaf].bathroom_group) for leaf in leaves
+            )
+            inherited_bathroom, inherited_group = container_bathroom(leaf_bathrooms)
             if not bathroom:
                 bathroom, group = inherited_bathroom, inherited_group
         else:
@@ -1159,6 +1161,59 @@ def _resolve_ramp_coverage(units: list[LodgingUnitSummary], index: _BathroomInde
         grade=ramp_coverage,
         target="ramp_coverage",
     )
+
+
+def _resolve_bathroom(units: list[LodgingUnitSummary], index: _BathroomIndex) -> None:
+    """Fill in every CONTAINER's `bathroom` from its leaves -- kindred#2502.
+
+    The fourth twin of the three resolvers above, and the last amenity that
+    was still answered from the unit's own row. All 15 production containers
+    store `bathroom = "none"` (a building is not a room) while 13 of them
+    have at least one room that records one, so every whole-house card drew
+    no bathroom while both its rooms drew one the moment staff split it.
+
+    ⚠️ CONTAINERS ONLY, unlike the three above. A leaf's own row is already
+    the right answer and `_build_units` already computes it correctly --
+    `effective_bathroom` against a one-element slot leaves a `shared` leaf
+    `shared`. Overwriting leaves here would be a no-op at best and would
+    duplicate the one place that logic lives.
+
+    ⚠️ WHY A RESOLVER AND NOT SLOT-CODE THREADING. The party path answers
+    this correctly already (`_resolve_party_bathroom`), but only once a
+    placement exists -- so the same building graded unmet in the picker and
+    met once the family landed in it. Most containers are never placed into.
+    A resolver has no placement to read, which is the whole point: it answers
+    for an EMPTY building, which is when staff are choosing one.
+
+    The exclusivity upgrade is honest here rather than generous: booking a
+    whole container covers every room under it by construction, so if the
+    rooms' shared group has no members outside the container,
+    `effective_bathroom` is being handed a genuinely complete slot.
+    """
+    for unit in units:
+        if not unit.is_container:
+            continue
+        leaves = [
+            leaf
+            for code in sorted(index.leaf_codes_under(unit.code))
+            if (leaf := index.units_by_code.get(code)) is not None and leaf.is_active
+        ]
+        inherited, group = container_bathroom(frozenset((leaf.bathroom, leaf.bathroom_group) for leaf in leaves))
+        if inherited == "none":
+            # Nothing to inherit -- the container keeps exactly what its own
+            # registry row says, which is what it already holds. This is also
+            # the bathhouse case: rooms sharing a group that names somewhere
+            # they WALK to, with no bathroom in any of them.
+            continue
+        unit.bathroom = cast(
+            Any,
+            effective_bathroom(
+                inherited,
+                group,
+                index.group_members.get(group, frozenset()),
+                frozenset(leaf.code for leaf in leaves),
+            ),
+        )
 
 
 def _resolve_amenity_coverage[Answer](
@@ -1450,6 +1505,7 @@ class LodgingRosterService:
         # The same walk, one dimension over (kindred#2438), and the third and
         # last read of it. Same path-only reasoning as above.
         _resolve_ramp_coverage(unit_summaries, unit_index)
+        _resolve_bathroom(unit_summaries, unit_index)
         # A SECOND PASS, and it has to be: a unit's cover can come from a row
         # on a unit built after it, so there is no order in which one pass over
         # `_build_units` would see every own-row it needs.
@@ -1640,6 +1696,10 @@ class LodgingRosterService:
             # Same one-index-per-call rule `build_roster` follows -- see the
             # comment there and `_BathroomIndex`'s own docstring.
             unit_index = _BathroomIndex.build(unit_summaries)
+            # `build_roster` runs the amenity resolvers right here; this path
+            # ran none of them, so a container's bathroom stayed at its own
+            # blank row on the summary while the board resolved it.
+            _resolve_bathroom(unit_summaries, unit_index)
             parties = self._build_parties(
                 session_type=_s(session, "session_type"),
                 # THIS weekend's start. The six year-scoped fetches above are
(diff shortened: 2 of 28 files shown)


adamflagg:
@coderabbitai review

adamflagg:
@coderabbitai review

Focus: `_resolve_bathroom` / `_resolve_has_ac` in api/services/lodging_roster_service.py and the `(bathroom, group)` pair logic in `container_bathroom` (api/services/lodging_rules.py). The risk is a *satisfied* verdict on a medical in-cabin-bathroom request when no descendant actually has one.

PR #2512, opened 2026-08-20
fix(scripts): the lodging name guard stopped exempting comments

adamflagg:
The guard reported `OK` while **eighteen real building and area names sat in tracked, public source**. Every one was in a comment, and comments were dropped before the report.

Found while auditing something else (#2511), by running the guard's own needle list by hand and diffing against what it reported.

## Why the exemption existed, and why it was wrong

It held that prose naming a unit to explain a rule is documentation rather than the registry living in code — added in #1891 after a docstring reddened an unrelated phase.

Right about intent, wrong about consequence. **A comment is source.** It ships in the repo, and a public repo leaks it exactly as a literal does. Nor is the tradeoff real: prose can say what it needs to say without naming a building. All eighteen were **rewritten, not deleted**, and none lost its explanatory force.

#2367 had already reached this conclusion and scoped its fix to `frontend/src/**`, stating it was *"leaving the pocketbase/ side for a future pass, rather than guessing at files another issue may be mid-edit on."* This is that pass.

## The rule now

| | code hits | comment hits |
|---|---|---|
| not a test file | fail | **fail** (was: dropped) |
| a test file | exempt | **fail** (was: dropped) |

…for files under `SCAN_ROOTS`, which is five directories and **does not include the pytest tree** — see the scope note below.

**Fixture code in a test file stays exempt, deliberately.** Real names as fixture *data* are legitimate — an alias resolver test has to resolve the strings staff actually wrote — and narrowing it would fail the many lodging test files built on exactly that. #1909 found two needle-matching literals in fixture code and fixed them at the source rather than narrowing the filter; that trade is unchanged. Only comments moved.

## What was scrubbed

- one `api/` docstring
- three migration headers
- **one `frontend/src/**` component comment** — a non-test frontend file, so outside #2367's scope and never in scope for anything
- eleven `pocketbase/` Go test comments

…plus everything the review round below turned up.

## Three tests are REWRITTEN, not adapted

TEST 4, 7 and 12 each pinned the old exemption. The **specification** changed; the implementation did not drift from it, so adapting them would have been backwards. Each carries why it flipped and what it now proves.

TEST 4 and 7 additionally assert the report **names the offending file**. A guard that caught one probe and silently dropped the other would have passed a bare exit-code check.

---

# Review round

Ten findings. Every fix is verified by planting a needle and reading the exit code — a green guard is not evidence.

### The widening introduced two regressions of its own

**The test/not-test split was decided by the whole grep line, not the path.** `grep -In` emits `path:lineno:text`, and the new pattern was unanchored, so a line's own **content** could route its file into the test bucket and win the code-hit exemption:

| planted line | old pattern | new |
|---|---|---|
| a fixture-path constant containing `tests/`, in ordinary application code | **exempted** | fails |
| a version-path string containing `latest/` (which carries `test/`) | **exempted** | fails |
| a needle inside a path-like literal containing `_test.` | **exempted** | fails |

The pattern this replaced anchored both alternatives to the path with `^`; that is the property that got lost, and the five-case mutation table in the first cut could not catch it because every case varied file *kind* and none varied line *content*. The split now runs through `awk` on field 1, and `tests?/` must be a whole path segment again. TESTs 14-16.

**The test-bucket pipeline had no `|| true`.** Under `set -euo pipefail`, a `grep -v` that removed every line exited 1, pipefail promoted it, and `set -e` killed the guard mid-run: **exit 1 with nothing printed at all**. Reachable because that `grep -v` matches its path anywhere in the line, including inside a comment's text. The escape is scoped to grep with a brace group so a `python3` crash stays fatal. TEST 17 asserts the guard never exits 1 silently.

### The comment dropper promised fail-open and delivered fail-closed

`drop_comment_hits.py`'s docstring says a file that will not parse or read is *"treated as ALL CODE, so its hits survive"*. Under `--only-comments` — the only mode the guard still uses — "all code" means **all dropped**. Two live consequences:

- the guard greps `*.sh`, but the module had no scanner for it, so a needle in a `#` comment in a test shell script came back classified as code and was exempted. The rule table's *"a test file → comment hits fail"* was simply untrue for shell;
- a `.py` file that would not tokenize lost its comment hits the same way.

`noncode_lines` now returns `None` for "we learned nothing" and such hits are written through in **both** modes; `.sh` gets a `#`-comment scanner on the same whole-line rule the C-style one uses, so fixture code in a test `.sh` stays exempt. TESTs 18-20, plus six unit tests.

### One test's rationale was false, and the thing it claimed to cover had no cover

TEST 7's new comment said the dropper *"still decides whether a line is reported as a comment hit or a code hit"* and that its assertion *"keeps the #2181 regression covered"*. Both wrong, and measured rather than argued: TEST 7's probe is a **non-test** `.tsx`, which since this change never reaches the dropper at all, and TEST 7 still passes with the JSX fused-brace logic deleted. Nothing end-to-end covered that logic. The comment now says what the test actually proves, and **TEST 21** plants the same shape in a *test* `.tsx` so the hit has to go through the dropper — with the fusion disabled it reports exit 0 while TEST 7 still reports exit 1.

### The scrub was grep-driven, not principle-driven

The guard's grep was **case-sensitive**, and its own header says the needle list is only a representative sample. Prose in this codebase names a unit by its lowercase **code**, so a title-case scan read straight past it: seven real comment hits were sitting in tracked source — a migration header, two frontend components and three test files.

The raw scan is now case-insensitive. That adds **64 raw hits** repo-wide: **63** are lowercase codes in test *fixture* data, which stays exempt, and **1** is an ordinary English false positive (a needle's folded form appears inside a normal English word). That one needle, and only that one, gets a `\b` — it contains a space, so it never matched a camelCase identifier anyway. `\b` is deliberately **not** applied to the whole list: it would stop catching identifiers built from a unit name, which is a leak shape the guard does catch today.

Every real name left in a **comment** in a file this PR touches is rewritten, checked against the **full private registry** — names, codes, area names and alias strings — rather than the needle sample. Fixture code is untouched. Three comments the first pass left as broken English (an orphaned "Both", a dangling number, a "four strings" list with one item removed) read as sentences again.

Turning on case-insensitivity is what pulled two frontend components and two frontend test files into this PR that were not in it before: their comments went red, so they had to be scrubbed here. All four changes are comment-only.

### Scope stated honestly, and one issue filed

`SCAN_ROOTS` is `pocketbase/ api/ bunking/ frontend/src/ scripts/`. **The pytest tree at `tests/` is not scanned at all** and never has been, so the rule table above was reading as universal when it is not. It carries **71 needle hits** (80 case-insensitively) across three files, one of which restates in prose the exact fact this PR scrubbed out of `api/services/lodging_rules.py` — the scrub moved the sentence into an unscanned file rather than out of the repository.

Not fixed here: adding a sixth scan root changes what every future PR is measured against and deserves its own diff. The guard's comment now states its roots honestly and points at **#2515**, which carries the measurement and the scrub-or-exempt decision.

### A tracked doc outside the scan roots, which the guard can never see

`docs/` is not a scan root, so this was invisible to every check: **`docs/investigation-not-bunk-with-bug.md` is tracked and public and carried real camper names and real CampMinder IDs throughout**, including a verbatim dump of all 31 `not_bunk_with` records — each naming a requester and the campers they asked not to be bunked with. Far more than the single needle-matching name that surfaced it.

The dump is **removed rather than translated**: a verbatim list of who does not want to bunk with whom has no engineering value once its counts are recorded, and the counts are kept. Every other name and id in the file is now the repo's fictional set.

### Stale comments corrected

The guard's header claimed *"comments are dropped below"*, which this PR makes false; the retained #2367 block described frontend-only scoping as current and asserted the pattern anchors to a path segment — the exact property removing `^` had destroyed, now true again; and the test suite's TEST 13 comment cited two variables this PR deleted.

## Verified by mutation, not by green

| planted | expect | before | after |
|---|---|---|---|
| comment in a non-test `.py` | 1 | 1 | 1 |
| comment in a Go test | 1 | 1 | 1 |
| literal in non-test code | 1 | 1 | 1 |
| fixture code in a Go test | 0 | 0 | 0 |
| code in a non-test file whose text holds `tests/` | 1 | **0** | 1 |
| code in a non-test file whose text holds `latest/` | 1 | **0** | 1 |
| code in a non-test file whose text holds `_test.` | 1 | **0** | 1 |
| a filter that empties the test bucket | 0 + a message | **1, zero bytes** | 0 |
| comment in a test `.sh` | 1 | **0** | 1 |
| fixture code in a test `.sh` | 0 | 0 | 0 |
| comment in an unparseable `.py` test | 1 | **0** | 1 |
| JSX `*/}` comment line in a test `.tsx` | 1 | 1 | 1 (0 with the fusion deleted) |
| lowercase needle in a comment | 1 | **0** | 1 |
| baseline | 0 | 0 | 0 |

**22/22 guard self-tests** (13 existing, 9 new), 8 `drop_comment_hits` unit tests, plus the repo's scoped pre-push superset.

## Note on overlap with #2511

The earlier claim of one shared file was wrong. There are **five**, and the number grew because turning on case-insensitivity pulled two more frontend files in:

| file | overlap |
|---|---|
| `api/services/lodging_roster_service.py` | byte-identical hunk — #2511 found and scrubbed the same comment |
| `api/services/lodging_rules.py` | different regions |
| `frontend/src/components/weekend/MapUnitPopover.tsx` | different regions; #2511 shifts the file substantially above this PR's hunk |
| `frontend/src/components/weekend/AssignFamilyModal.tsx` | different regions |
| `frontend/src/components/weekend/LodgingUnitCard.tsx` | different regions |

Trial-merged against #2511's current head with `git merge-tree`: **all five auto-merge, no conflicts, either order.** #2511 adds **zero** needle-bearing lines, case-insensitively checked, so it will not turn the guard red once this lands.

Refs #2511
Refs #2515

### DIFF
--- a/api/services/lodging_roster_service.py
+++ b/api/services/lodging_roster_service.py
@@ -925,7 +925,7 @@ class _BathroomIndex(NamedTuple):
 
     units_by_code: dict[str, LodgingUnitSummary]
     # Immediate children only, keyed by the PARENT's code. Nesting (a
-    # container inside a container, e.g. Doctor's House under a larger
+    # container inside a container, e.g. an apartment under a larger
     # block) is walked at read time in `leaf_codes_under`, mirroring
     # `drawn_units`' own upward walk of the same `parent_code` relation.
     children_by_parent: dict[str, tuple[LodgingUnitSummary, ...]]
--- a/api/services/lodging_rules.py
+++ b/api/services/lodging_rules.py
@@ -143,9 +143,10 @@ def effective_bathroom(
 ) -> str:
     """Spec §3.2.1 — private vs shared depends on the merge state.
 
-    Tioga 1 and Tioga 2 are each `shared`, because two families normally
-    split them. Merge both and the same bathroom becomes `private`, so
-    merging can itself be the accommodation for a medical bathroom request.
+    Two rooms sharing one bathroom are each `shared`, because two families
+    normally split them. Merge both and the same bathroom becomes `private`,
+    so merging can itself be the accommodation for a medical bathroom
+    request.
 
     Args:
         bathroom: the unit's own value ("", "none", "private", "shared").
--- a/docs/investigation-not-bunk-with-bug.md
+++ b/docs/investigation-not-bunk-with-bug.md
@@ -2,29 +2,34 @@
 
 ## Date: 2026-01-20
 
+> **Redacted 2026-08-20 (kindred#2512).** This file is tracked and public and used to
+> carry real camper names and real CampMinder IDs throughout, including a verbatim dump
+> of all 31 not_bunk_with records. Names and IDs below are the repo's fictional set; the
+> record dump is gone. See the "All 31 not_bunk_with Records" section for what it held.
+
 ## Original Problem
 
-Camper 6325675 (Gabriella Bauer-Kahan) has a raw `not_bunk_with` field containing "Amalia Smith", but processed records showed siblings as targets (Aiden and Rebecca Bauer-Kahan) instead.
+Camper 1000001 (Emma Johnson) has a raw `not_bunk_with` field containing "Riley Sam", but processed records showed siblings as targets (Liam and Olivia Johnson) instead.
 
 ## Root Cause Found and Fixed
 
 ### Issue: AI Hallucinating "SIBLING" Placeholder
 
-The AI (GPT-4.1-nano) was incorrectly outputting `target_name = "SIBLING"` for plain name inputs like "Amalia Smith".
+The AI (GPT-4.1-nano) was incorrectly outputting `target_name = "SIBLING"` for plain name inputs like "Riley Sam".
 
 **Evidence from debug logs:**
 ```text
-[AI-PARSE] Input: field='Do Not Share Bunk With' requester='Gabriella Bauer-Kahan' text='Amalia Smith'
+[AI-PARSE] Input: field='Do Not Share Bunk With' requester='Emma Johnson' text='Riley Sam'
 [AI-PARSE] Output: targets=['SIBLING'] request_types=['not_bunk_with']  # WRONG!
 ```
 
-The SIBLING placeholder then triggered `placeholder_expander.py` to expand it to actual siblings (Aiden and Rebecca Bauer-Kahan).
+The SIBLING placeholder then triggered `placeholder_expander.py` to expand it to actual siblings (Liam and Olivia Johnson).
 
 ### Fix Applied
 
 Removed the SIBLING placeholder entirely from `config/prompts/parse_not_bunk_with.txt` because:
 
-1. **Analysis of all 31 not_bunk_with records showed ZERO actual sibling references** - All records contain explicit names like "Amalia Smith", "Kyla Udell", etc.
+1. **Analysis of all 31 not_bunk_with records showed ZERO actual sibling references** - All records contain explicit names like "Riley Sam", "Samuel Johnson", etc.
 2. **The SIBLING placeholder was causing false positives** - The AI was hallucinating it for normal names
 
 **Changes made to `config/prompts/parse_not_bunk_with.txt`:**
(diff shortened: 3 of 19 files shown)


adamflagg:
@coderabbitai review

Focus: scripts/dev/verify-no-hardcoded-lodging.sh — the file-split vs comment-dropper interaction. The prior bug was that the non-test path still ran its hits back through the comment dropper, so the guard passed while comments held real names. Check for any remaining path where a hit is discovered and then silently dropped.

PR #2518, opened 2026-08-21
fix(frontend): the housing board sorts units alpha within an area, not by tree depth

### LINKED ISSUE #2514: bug(frontend): the housing board orders units by tree depth, so containers draw before rooms instead of alphabetically within an area
## What was reported

> containers should be alpha within categories, not containers first

Filed from the weekend housing board.

## What I found — my analysis, not the reporter's. Verify before implementing.

`frontend/src/components/weekend/boardLayout.ts:1027-1028` states that units are drawn in the
payload's order and never re-sorted, because "the repository's own query already sorts
`area.sort_order,name`".

That assumption is broken upstream. `frontend/src/components/weekend/unitLevel.ts:231`
`drawnUnits` walks the registry tree **breadth-first**:

```ts
const queue = [...roots]
while (queue.length > 0) {
  const next = queue.shift()
  ...
  const isLeaf = next.is_container !== true
  if (isLeaf || next.is_combined === true) drawn.push(next)
  else queue.push(...kids)          // ← descendants go to the TAIL
}
```

`roots` is every unit with no resolvable parent. Because descendants are appended to the tail, a
unit's **depth in the registry tree dominates its name**: every root-level unit — parentless rooms
and combined containers alike — is emitted before any room that lives beneath a non-combined
container, whatever it is called. The payload's alphabetical order survives only *within* a depth
band, and `boardLayout` then buckets that already-scrambled sequence without re-sorting.

From the staff side that reads exactly as the report: buildings cluster ahead of the rooms of other
buildings.

## Probable fix

Sort at the point the board buckets, rather than relying on a payload order the walk has already
destroyed — `boardLayout.ts` around :1027, by `(area_sort_order, name)`. Correct the stale comment
at :1027-1028 in the same change, since that comment is what made the loss invisible.

`unitLevel.test.ts` has `drawnUnits` assertions that pin the current emission order. If the
ordering becomes a specified behaviour, those are a **rewrite** with a comment saying why, not an
adaptation — the repo forbids editing a test to match an implementation.

## Needs confirming first

**What "categories" means here.** The board buckets by **area** (`area_sort_order`), but units also
carry an `inventory_class`. The fix differs depending on which one the reporter has in mind, and
the report does not say.

*Reported via kindred-feedback#129 — the private original, which holds the reporter, the page it was filed from, and the app version.*

adamflagg:
`buildBoard` bucketed units in whatever order the payload's `drawn` list arrived in, on the strength of a stale comment claiming the repository's own query already sorted `area.sort_order,name`. That assumption was broken upstream: `unitLevel.ts`'s `drawnUnits` walks the registry tree breadth-first, appending descendants to the tail of its queue, so a unit's DEPTH in the tree decided its position, not its name. Every root-level unit — parentless rooms and combined containers alike — drew ahead of any room nested under a non-combined container, whatever either was called. Staff saw containers cluster ahead of other buildings' rooms within an area, exactly as reported ("containers should be alpha within categories, not containers first").

## Fix

Sort each area's bucket by `(area_sort_order, name)` after the payload is bucketed (`boardLayout.ts`, `buildBoard`), rather than trusting payload order. Corrected the stale comment at that spot describing the old (wrong) assumption.

## "Categories" — resolved as area, not `inventory_class`

The issue's open question was which grouping "categories" meant. Read out of the tree: `inventory_class` groups nothing on the board today — its only uses are visibility and badge predicates (`boardLayout.ts:914`, `unitBadges.ts:77/:105/:143`). There is no code path that buckets, sorts, or headers by it. Reading "categories" as `inventory_class` would mean inventing a grouping the board has never had, not fixing a sort — so this reads it as area, the board's actual (and only) grouping.

## Tests

- Rewrote `never reorders the units WITHIN an area` (`boardLayout.test.ts`) — it pinned the old, wrong invariant that payload order should survive untouched. It now feeds units in reverse-alphabetical order (the shape a BFS walk produces when depth runs opposite name) and asserts the board still sorts them alphabetically. Comment says why it's a rewrite, not an adaptation.
- Updated one incidental-order assertion in `keeps staff housing that still holds a party` — both fixture units land in the same default area, so the correct alphabetical order flipped; the assertion under test (membership — a party never disappears) is unaffected.
- Mutation-checked: removing the new sort turns both tests red.

## Deliberately did not do

- Did not touch `unitLevel.ts`'s breadth-first walk or `drawnUnits`, and did not touch `unitLevel.test.ts`'s pinned emission-order assertions. The walk order is unchanged; only where `buildBoard` re-sorts before drawing changed. Those tests still pass untouched.

Closes #2514.

### DIFF
--- a/frontend/src/components/weekend/boardLayout.test.ts
+++ b/frontend/src/components/weekend/boardLayout.test.ts
@@ -1170,25 +1170,25 @@ describe('buildBoard — area grouping and colour', () => {
   })
 
   it('sorts a unit numbered past 9 numerically, not as a string (kindred#2518 review)', () => {
-    // Plain `localeCompare` treats the number as a string, so "Manzanita 10"
+    // Plain `localeCompare` treats the number as a string, so "Cedar 10"
     // sorts as if its first character were "1" and lands ahead of
-    // "Manzanita 2" — every production area's numbering is single-digit
-    // today, which is exactly why this was latent, but summer's equivalent
-    // sort (`BunkingBoardByArea.tsx`) already numeric-compares trailing
-    // digits so "B-2" precedes "B-10", and the board must not regress that
-    // the moment an area's numbering crosses into double digits.
+    // "Cedar 2" — every production area's numbering is single-digit today,
+    // which is exactly why this was latent, but summer's equivalent sort
+    // (`BunkingBoardByArea.tsx`) already numeric-compares trailing digits so
+    // "B-2" precedes "B-10", and the board must not regress that the moment
+    // an area's numbering crosses into double digits.
     const board = buildBoard(
       [],
       [
-        unit({ unit_id: 'u1', code: 'manzanita-10', name: 'Manzanita 10' }),
-        unit({ unit_id: 'u2', code: 'manzanita-9', name: 'Manzanita 9' }),
-        unit({ unit_id: 'u3', code: 'manzanita-2', name: 'Manzanita 2' }),
+        unit({ unit_id: 'u1', code: 'cedar-10', name: 'Cedar 10' }),
+        unit({ unit_id: 'u2', code: 'cedar-9', name: 'Cedar 9' }),
+        unit({ unit_id: 'u3', code: 'cedar-2', name: 'Cedar 2' }),
       ]
     )
     expect(board.areas[0]?.slots.map((slot) => slot.unit.code)).toEqual([
-      'manzanita-2',
-      'manzanita-9',
-      'manzanita-10',
+      'cedar-2',
+      'cedar-9',
+      'cedar-10',
     ])
   })
 
--- a/frontend/src/components/weekend/boardLayout.test.ts
+++ b/frontend/src/components/weekend/boardLayout.test.ts
@@ -397,7 +397,11 @@ describe('buildBoard — which units get a card', () => {
       [unit(), staffUnit({ unit_id: 'u2', code: 'aspen-lodge', name: 'Aspen Lodge' })]
     )
     const slots = board.areas.flatMap((area) => area.slots)
-    expect(slots.map((slot) => slot.unit.code)).toEqual(['cedar-1', 'aspen-lodge'])
+    // Both fall in the same default area, so they now sort alphabetically by
+    // name (kindred#2514) — "Aspen Lodge" before "Cedar 1". The membership
+    // this test is actually about is unaffected: staff housing that still
+    // holds a party never disappears.
+    expect(slots.map((slot) => slot.unit.code)).toEqual(['aspen-lodge', 'cedar-1'])
     expect(board.offBoard).toHaveLength(0)
     expect(board.unplaced).toHaveLength(0)
   })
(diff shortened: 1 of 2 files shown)


adamflagg:
## Scan report — PR #2518 (issue #2514)

**Level:** requested `high`; the reviewer's own environment (no Agent tool available in this
subagent) fell back to a single-pass inline review labeled `xhigh` in its own transcript — no
multi-agent fan-out, but broader-recall angles run inline instead of narrower ones. Reporting
this plainly per the "name the level it actually ran at" instruction: it is not the multi-agent
`high` pipeline, but it is not a silent downgrade either — recall angles ran, just single-pass.

### Verified independently before trusting the review
- Ran the actual mutation check myself: deleting the new `(area_sort_order, name)` sort turns
  both `boardLayout.test.ts` assertions red; restored cleanly.
- Confirmed the root-cause claim against `unitLevel.ts:231` `drawnUnits` — genuinely
  breadth-first, descendants appended to queue tail.
- Confirmed `unitLevel.ts` / `unitLevel.test.ts` are untouched by this PR (`git diff --stat`
  empty), matching the "did not touch the BFS walk" claim in the PR body.
- Confirmed `inventory_class` groups nothing on the board (`boardLayout.ts:914`,
  `unitBadges.ts:77/105/143` — visibility/badge predicates only), so reading "categories" as
  area rather than `inventory_class` is correct.
- Checked `#2076`'s owner ruling ("intra unit remains alpha") — this PR doesn't invent a new
  rule, it makes the code actually deliver the ruling the stale comment only claimed to.
- Full frontend suite: 7247 tests pass. `tsc --noEmit`: clean.

### Finding — fixed in-branch (code this PR wrote)
The review caught a real defect in the new sort itself: `a.unit.name.localeCompare(b.unit.name)`
compares the trailing unit number as a **string**, not numerically. I reproduced it directly
(`node -e`): `"Cedar 10".localeCompare("Cedar 2")` sorts `10` ahead of `2`. Summer's equivalent
per-area sort (`BunkingBoardByArea.tsx`) already numeric-compares trailing digits for exactly
this reason (`kindred#2514`'s own "Family Camp Models Summer" rule). I checked both `data.db`
and `data-prod.db` — every numbered unit today is single-digit, so this was latent, not visible
yet, but it regresses the instant any area's numbering reaches double digits.

**Fix:** `localeCompare(b.unit.name, undefined, { numeric: true })` — one option added to the
existing call, no new dependency, no divergence from the established pattern once you count the
option as "the same idea, built into `Intl.Collator`" rather than reproducing summer's manual
digit-splitting. Wrote a failing test first (`sorts a unit numbered past 9 numerically...`),
confirmed it failed against the unpatched sort, applied the fix, confirmed green, mutation-
checked by reverting the fix and confirming red again.

**Self-caught follow-up:** my first attempt at this fix used a real production unit name
("Manzanita") in the test and comment as the illustrative example, which the CI Lodging Name
Guard correctly failed. Swapped to the fictional "Cedar N" already used throughout this test
file for every literal and every comment; guard now passes locally and in CI.

### Not filed, not extended
Nothing else surfaced. The finding was entirely within code this PR wrote (rule 1: fix here,
always), so there was nothing to file. No sibling extension attempted — summer's more elaborate
dash-splitting numeric parser (handling `B-2`/`B-10` plus alpha suffixes) was not ported
wholesale; `numeric: true` gets the same correctness for weekend's space-separated naming
without inventing scope the finding didn't ask for.

### Merge gate — HELD per campaign classification
This PR changes the visible order of every card on the weekend housing board — visual impact,
campaign-flagged. **Not merged.** Holding for the owner's eyes per the standing rule, independent
of how clean CI or the scan came back.

- CI Summary: **green** on `e46c8134` (latest commit).
- `closingIssuesReferences`: `#2514` (verified via GraphQL).
- Push verified by remote ref (`git rev-list --count origin/...HEAD` = 0).
- Numbers a staff member reads: **none change meaning.** This reorders cards on a board; no
  metric, count, rate, or classification is touched.

PR #2519, opened 2026-08-21
fix(sync): close the family-camp bounded custom-values jobs' lock, phase, and log gaps

### LINKED ISSUE #2491: fix(sync): the family-camp bounded custom-values jobs share collections with the unbounded ones but not their lock, phase slot, or log label
## ⚠️ CORRECTED 2026-08-21 — the last serialisation constraint is DISCHARGED. Fully unblocked.

Verified against `main` @ `a6ffb313`.

**PR #2493 (#2482's capture half) has MERGED as `caa0fe29`.** The "serialises against" row below
called it *open*; nothing is in flight against `pocketbase/sync/person_custom_field_values.go` or
`household_custom_field_values.go` any more. With #2489 already merged as `1b9b2814`, this issue has
**no predecessors and nothing to serialise against** — rebase, do not wait.

**The defect itself is unchanged and still live.** Re-verified at `a6ffb313`: the guard is at
`pocketbase/sync/api.go:1474` — still the literal
`orchestrator.IsRunning("person_custom_values") || orchestrator.IsRunning("household_custom_values")`
— and the two bounded instances are still registered under
`person_custom_values_family_camp` / `household_custom_values_family_camp`, which those two strings
do not match.

---

## ⚠️ WIDENED 2026-08-19 — this is THREE defects from one root cause, not one

Filed originally against `api.go`'s guards only. A second, independent review of PR #2494 (which
was stacked on #2489) found two more consequences of the same registration change, plus a third
face of the lock gap that `api.go` does not cover. **All three verified against `main` @ `1b9b2814`.**
Fix them together — they are the same two files and the same root cause, and filing them
separately would cost three branches for one afternoon's work.

**Root cause, stated once:** #2489 registered `person_custom_values_family_camp` /
`household_custom_values_family_camp` as service instances *distinct from* `person_custom_values` /
`household_custom_values` (`orchestrator.go:110,112` in `syncJobMeta`; `:2417,2421` in
`RegisterService`). They write the **same** PocketBase collections but carry a **different lock
identity, a different phase-enumeration slot, and no distinct log label.**

### Face B — the cron-vs-cron overlap, which `api.go` does not cover

The orchestrator's in-progress guard is keyed by service name
(`runSingleSyncInternal` → `o.runningJobs[syncType]`), so the bounded and unbounded jobs do not
exclude each other there either. The daily cron fires `0 3 * * *` and the unrestricted
custom-values sweep `0 4 * * 0`; robfig/cron runs each entry on its own goroutine. On a Sunday
where the 03:00 run is still in its expensive phase at 04:00, both fetch the same family-camp
person, and `idx_person_cf_vals_unique (year, person, field_definition)` (migration `1500000028`)
makes one of the two writes fail.

⚠️ **Severity, stated honestly:** the failing write increments `s.Stats.Errors` and `continue`s
past `sweptOwners[personPBId] = true`, so that person is skipped by the orphan sweep for that run.
That is **fail-safe, not data loss** — `sweptOwners` exists precisely so a partial fetch is not read
as deletions (`person_custom_field_values.go:181`, `:505`), and the next run self-heals. It also
requires the daily run to exceed 60 minutes against a ~16.5 min projection. **Real, low, and worth
closing with Face A's fix rather than on its own.**

⇒ The `IsAnyJobRunning()`-style collection-scoped check this issue already proposes for Face A
should be applied at the **orchestrator** level so it covers the cron path too, not only the three
`api.go` handlers.

### Face C — the admin "Custom Values" phase run now does the family-camp cohort TWICE

`GetJobsForPhase(PhaseExpensive)` is built from `syncJobMeta`, which now lists **four** jobs
(`orchestrator.go:105,106,110,112`). `EnqueuePhaseSync(year, PhaseExpensive, …)` — reachable from
the phase UI labelled *"Sync custom field values (slow, 1 API call per entity)"* — therefore runs
the unrestricted sweep and then immediately re-fetches the ~447 family-camp households and their
781 persons. **≈11.5 min of rate-limited CampMinder quota, burned every time an admin runs that
phase**, for values fetched minutes earlier. Nothing is corrupted; it is pure duplicated quota.
If the bounded jobs are only meant for the daily cron, exclude them from the phase enumeration.

### Face D — the bounded pass is indistinguishable from the year-wide sweep in the logs

`Name()` returns `serviceNamePersonCustomValues` (`person_custom_field_values.go:55`) and
`LogSyncStart(serviceNamePersonCustomValues)` is unconditional (`:77`), so the 03:00 family-camp
pass emits `Starting sync service=person_custom_values` — byte-identical to the Sunday 04:00 full
sweep. The following cohort log prints `session=all`, because the bounded instances leave `Session`
at `DefaultSession` while ignoring it. **An operator diagnosing stale dietary/transportation data
sees `person_custom_values` completing nightly and concludes the year-wide sweep runs daily.**
It does not. Pass the registered job name (or a `FamilyCampBounded`-aware label) into
`LogSyncStart` and the cohort log.

### Graph wiring

| | |
|---|---|
| **files_touched** | `pocketbase/sync/api.go`, `pocketbase/sync/orchestrator.go`, `pocketbase/sync/person_custom_field_values.go`, `pocketbase/sync/household_custom_field_values.go` + tests |
| **hard_predecessors** | none (#2489 is merged as `1b9b2814`) |
| **serialises against** | ~~PR #2493 (#2482 capture half) — open, touches both `*_custom_field_values.go`~~ ⛔ **none — #2493 MERGED as `caa0fe29`** |
| **classification** | SCOPED |
| **closes** | #2491 |

---

## Problem

`handleCustomValuesSync`, `handlePersonCustomFieldValuesSync`, and
`handleHouseholdCustomFieldValuesSync` in `pocketbase/sync/api.go` guard against a duplicate
run by checking `orchestrator.IsRunning(...)` (or `RunSingleSyncWithService`'s internal check)
against the **literal** service names `"person_custom_values"` / `"household_custom_values"`:

```go
// api.go:1474
if orchestrator.IsRunning("person_custom_values") || orchestrator.IsRunning("household_custom_values") {
```

kindred#2482's cadence half (#2489) registered two **additional** service instances of the same
Go types under different names — `person_custom_values_family_camp` and
`household_custom_values_family_camp` — as part of the daily cron
(`getDailySyncJobs()`), scoped to family-camp attendees via the new `FamilyCampBounded` field.
Both the original and the new bounded instances read and write the **same** underlying
PocketBase collections (`person_custom_values`, `household_custom_values`).

The guards above don't know about the two new names, so they no longer prevent overlap: while
the daily cron's bounded pass is in flight (~11.5 min projected, see #2489), an operator
triggering "Sync Custom Values" from the admin panel (or the equivalent
`?session=`-scoped person/household custom-values endpoints) is not blocked.

## Why this can lose data, not just waste API calls

Each `Sync()` run computes its own `sweptOwners` (the set of persons/households its own fetch
actually covered) and deletes, via `deleteOrphans`, any record for a swept owner that is not in
its own `ProcessedKeys` this run. The bounded cohort is always a subset of the unrestricted
cohort (family-camp attendees ⊆ all persons/households for the year), so every person the
bounded pass touches is also in the unrestricted run's `sweptOwners`. If the two runs interleave
such that one run's write for a person lands after the other run's own read/orphan-computation
for that same person, the later run's orphan sweep can delete a row the concurrent run just
wrote and has not yet accounted for.

This is speculative in its exact interleaving (not reproduced against production) but the
underlying premise — the mutual-exclusion guard no longer covers all writers to the collection —
is confirmed by reading the source; only the exact conditions of a lost write need further
verification before fixing.

## Suggested fix

Give the orchestrator a name-agnostic way to ask "is anything that writes to
`person_custom_values`/`household_custom_values` currently running", analogous to the existing
`IsAnyJobRunning()` used correctly elsewhere in `api.go` (e.g. the individual-job and phase
handlers) — a small allowlist/prefix check covering `person_custom_values`,
`household_custom_values`, `person_custom_values_family_camp`, and
`household_custom_values_family_camp` would close the gap without a broader redesign. Update the
three handlers named above to use it.

## Scope note

Deliberately not fixed in #2489: that PR's file scope was restricted to
`scheduler.go`/`orchestrator.go`/`session_resolver.go`/`person_custom_field_values.go`/
`household_custom_field_values.go` + tests, and `api.go` is outside that allowlist. Filed as a
follow-up per triage-attack's size-vs-scope rule.

Found during code review of #2489 (kindred#2482's cadence half).

adamflagg:
## What

`#2489` registered the bounded daily family-camp custom-values instances
(`person_custom_values_family_camp`, `household_custom_values_family_camp`) as service names
*distinct from* the unrestricted `person_custom_values` / `household_custom_values` jobs they
sit alongside. Both bounded instances write the **same** PocketBase collections as their
unrestricted counterparts, but under a different registered name — so anything that keyed
mutual exclusion, phase execution, or its own log line on the literal job name treated them as
unrelated writers. Four faces of that one root cause, fixed together (same two sync files, same
afternoon's work):

- **Face A** — `api.go`'s three on-demand handlers (`handleCustomValuesSync`,
  `handlePersonCustomFieldValuesSync`, `handleHouseholdCustomFieldValuesSync`) guarded against a
  duplicate run with `IsRunning("person_custom_values")` / `IsRunning("household_custom_values")`
  — the literal unrestricted names only. An operator triggering an on-demand sync while the
  daily cron's bounded pass was in flight was not blocked, so the two runs could interleave
  against the same collection and race a concurrent write against the other's `deleteOrphans`
  read.
- **Face B** — the orchestrator's own in-progress guard (`runningJobs[syncType]`, checked in
  `runSingleSyncInternal`) has the identical gap for the **cron-vs-cron** case `api.go` doesn't
  cover: the daily 03:00 cron's bounded pass and the Sunday 04:00 unrestricted sweep run under
  different registered names and did not exclude each other. Low-severity in practice — a
  failing write increments `Stats.Errors` and skips that owner from the orphan sweep for the
  run (fail-safe, not data loss; the next run self-heals) — but the same root cause as Face A.
- **Face C** — the admin "Custom Values" phase button runs `GetJobsForPhase(PhaseExpensive)`,
  which now lists all **four** jobs. Triggering it re-fetches the ~447 family-camp households
  and 781 persons the daily cron already covered minutes earlier — **≈11.5 min of rate-limited
  CampMinder quota burned for values already fresh**, every time an admin runs that phase.
  Nothing corrupted; pure duplicated quota.
- **Face D** — `Sync()` unconditionally logged
  `LogSyncStart(serviceNamePersonCustomValues)` regardless of which instance was running, so the
  03:00 family-camp pass emitted `Starting sync service=person_custom_values` — byte-identical
  to the Sunday full sweep, with `session=all` in the cohort log because the bounded instance
  ignores `Session`. An operator diagnosing stale data from the logs alone would conclude the
  year-wide sweep runs nightly. It does not.

## Fix

- **`orchestrator.go`**: a `customValuesCollectionGroup` map ties each of the four job names to
  its underlying collection ("person" or "household"), plus
  `customValuesGroupRunningLocked` (used inside existing lock-held check-and-mark sections) and
  its exported self-locking form `IsCustomValuesCollectionRunning`. `runSingleSyncInternal`
  (the path both `getDailySyncJobs`'s cron and `RunCustomValuesSync`'s weekly sweep take) and
  `RunSingleSyncWithService` (the path the two session-scoped handlers take) now check the
  group, not the literal name — closing Face A and Face B in the same two call sites that
  already held the lock, so the atomicity `#1881`/`#2105` established is unchanged.
  `person_custom_values` and `household_custom_values` stay in separate groups, so
  `RunCustomValuesSync`'s documented "safe to run in parallel" behavior (independent
  collections, independent CampMinder endpoints) is preserved.
- **`api.go`**: `handleCustomValuesSync`'s front guard (the one handler that doesn't route
  through `RunSingleSyncWithService`) now calls `IsCustomValuesCollectionRunning` for both
  names, closing Face A there too. A new `phaseExecutionJobs(phase)` filters the two bounded
  jobs out of `PhaseExpensive`'s *execution* list at the two call sites that actually run a
  phase (`handleRunPhase`, `processQueuedSyncs`) — closing Face C.
- **`person_custom_field_values.go` / `household_custom_field_values.go`**: a new `logJobName()`
  method returns the bounded instance's own registered name when `FamilyCampBounded` is set.
  `Sync()` uses it for `LogSyncStart` and adds a `job` field to both cohort-size log lines —
  closing Face D.

### Deliberate divergence from the issue's literal Face C wording

The issue's suggested fix was to "exclude [the bounded jobs] from the phase enumeration" —
i.e., change what `GetJobsForPhase(PhaseExpensive)` (built from `syncJobMeta`) returns. That
function is pinned by `family_camp_daily_cadence_test.go`'s
`TestSyncJobMeta_FamilyCampBoundedJobsAreExpensivePhase`, added deliberately in `#2489` to
assert the two bounded jobs *are* classified as `PhaseExpensive`, consistent with their
unrestricted siblings. Changing `GetJobsForPhase` itself would have silently retired that
pinned invariant.

Instead, `GetJobsForPhase` — and the phase metadata `handleGetPhases` reports for the admin
UI — is left untouched; only the two call sites that actually **execute** a phase filter
through the new `phaseExecutionJobs`. Same measured defect closed (no duplicated quota burn on
an admin-triggered run), no existing pinned test touched.

## What this deliberately does not do

- Does not change `GetJobsForPhase`/`syncJobMeta`'s classification of the two bounded jobs as
  `PhaseExpensive` — see the divergence note above.
- Does not fully re-architect `runSingleSyncInternal`'s check-then-mark sequence into one
  atomic critical section for the general case. The `existingStatus == nil` branch's check and
  its later `runningJobs[syncType] = status` write are still two separate lock acquisitions
  (pre-existing shape, unrelated to this issue) — the group-aware check closes the specific
  name-mismatch gap this issue reports without widening the change into that pre-existing,
  narrower TOCTOU window, which the issue itself calls speculative and out of scope
  ("only the exact conditions of a lost write need further verification before fixing").
- Does not touch `MarkSyncRunning`'s own `IsRunning` check — its only production callers are
  the `process_requests` handlers, unrelated to the four custom-values job names.
- Frontend: none touched. This is Go-only, as scoped.

Closes #2491.

### DIFF
--- a/pocketbase/main_test_parallelism_test.go
+++ b/pocketbase/main_test_parallelism_test.go
@@ -111,6 +111,8 @@ var serialGroups = []struct {
 			"TestProcessAssignment_SaveFailure_IsInfraError",
 			"TestProcessEnrollment_MissingSessionID_IsRejected",
 			"TestProcessEnrollment_SaveFailure_IsInfraError",
+			"TestPersonCustomFieldValuesSync_CompletionLogUsesBoundedJobName",
+			"TestHouseholdCustomFieldValuesSync_CompletionLogUsesBoundedJobName",
 		},
 	},
 	{
--- a/pocketbase/sync/api.go
+++ b/pocketbase/sync/api.go
@@ -1242,7 +1242,7 @@ func processQueuedSyncs(orchestrator *Orchestrator) {
 	case "phase":
 		// Run all jobs in the phase sequentially
 		phase := Phase(qs.Service)
-		jobs := GetJobsForPhase(phase)
+		jobs := phaseExecutionJobs(phase)
 		slog.Info("Queued phase sync: Running jobs",
 			"phase", phase, "year", qs.Year, "jobs", jobs, "debug", qs.Debug)
 
@@ -1470,8 +1470,13 @@ func handleWeeklySync(e *core.RequestEvent, scheduler *Scheduler) error {
 func handleCustomValuesSync(e *core.RequestEvent, scheduler *Scheduler) error {
 	orchestrator := scheduler.GetOrchestrator()
 
-	// Check if custom values sync is already running
-	if orchestrator.IsRunning("person_custom_values") || orchestrator.IsRunning("household_custom_values") {
+	// Check if custom values sync is already running. Collection-group-aware (kindred#2491
+	// Face A): the daily cron's bounded family-camp jobs ("person_custom_values_family_camp",
+	// "household_custom_values_family_camp" -- kindred#2489) write these exact same
+	// collections under different registered names, so a check against only the two literal
+	// unrestricted names missed them.
+	if orchestrator.IsCustomValuesCollectionRunning("person_custom_values") ||
+		orchestrator.IsCustomValuesCollectionRunning("household_custom_values") {
 		return e.JSON(http.StatusConflict, map[string]any{
 			"error": "Custom values sync already in progress",
 		})
@@ -2144,6 +2149,39 @@ func handleHouseholdDemographicsSync(e *core.RequestEvent, scheduler *Scheduler)
 	})
 }
 
+// familyCampBoundedCustomValuesJobs names the two daily-cron-only custom-values instances
+// (kindred#2489) that phaseExecutionJobs excludes from an admin-triggered PhaseExpensive run.
+var familyCampBoundedCustomValuesJobs = map[string]bool{
+	"person_custom_values_family_camp":    true,
+	"household_custom_values_family_camp": true,
+}
+
+// phaseExecutionJobs returns the jobs actually run for phase -- the list handleRunPhase and
+// processQueuedSyncs iterate to start syncs, as opposed to GetJobsForPhase's classification
+// list (used for phase metadata/UI in handleGetPhases and pinned as including all four
+// custom-values jobs by family_camp_daily_cadence_test.go's TestSyncJobMeta_
+// FamilyCampBoundedJobsAreExpensivePhase).
+//
+// For PhaseExpensive specifically, it drops the two bounded family-camp jobs
+// (kindred#2489): they are always covered minutes earlier by the daily cron
+// (getDailySyncJobs), so an admin running the "Custom Values" phase on demand would otherwise
+// re-fetch the identical family-camp cohort a second time -- kindred#2491 Face C, measured at
+// ~11.5 min of rate-limited CampMinder quota burned for values already fresh. GetJobsForPhase
+// itself is left untouched; only the execution list is filtered.
+func phaseExecutionJobs(phase Phase) []string {
+	jobs := GetJobsForPhase(phase)
+	if phase != PhaseExpensive {
+		return jobs
+	}
+	filtered := make([]string, 0, len(jobs))
+	for _, j := range jobs {
+		if !familyCampBoundedCustomValuesJobs[j] {
+			filtered = append(filtered, j)
+		}
+	}
+	return filtered
+}
+
 // handleGetPhases returns list of available sync phases with metadata
 func handleGetPhases(e *core.RequestEvent) error {
 	phases := GetAllPhases()
@@ -2234,7 +2272,7 @@ func handleRunPhase(e *core.RequestEvent, scheduler *Scheduler) error {
 	}
 
 	// Get jobs for this phase
-	jobs := GetJobsForPhase(phase)
+	jobs := phaseExecutionJobs(phase)
 	if len(jobs) == 0 {
 		return e.JSON(http.StatusBadRequest, map[string]any{
 			"error": "No jobs found for phase: " + string(phase),
--- a/pocketbase/sync/api_test.go
+++ b/pocketbase/sync/api_test.go
@@ -1995,3 +1995,200 @@ func TestNormalizeSession(t *testing.T) {
 		})
 	}
 }
+
+// =============================================================================
+// kindred#2491: the family-camp bounded custom-values jobs share collections with the
+// unbounded ones but not their lock, phase slot, or log label. Face A + Face C below.
+// =============================================================================
+
+// TestHandleCustomValuesSyncRejectsWhenFamilyCampBoundedGroupmateRunning pins Face A's
+// original report: handleCustomValuesSync's front guard used to check only the literal names
+// "person_custom_values" / "household_custom_values" via orchestrator.IsRunning, so it did not
+// see the bounded daily family-camp jobs (kindred#2489) as writers of the same collections and
+// would return 200 "Custom values sync triggered" while one was still in flight.
+func TestHandleCustomValuesSyncRejectsWhenFamilyCampBoundedGroupmateRunning(t *testing.T) {
+	t.Parallel()
+
+	cases := []struct {
+		name          string
+		boundedJob    string
+		registerBound func(o *Orchestrator)
+	}{
+		{
+			name:       "person bounded pass blocks the endpoint",
+			boundedJob: "person_custom_values_family_camp",
+			registerBound: func(o *Orchestrator) {
+				o.RegisterService("person_custom_values_family_camp",
+					&MockService{name: "person_custom_values_family_camp"})
+			},
+		},
+		{
+			name:       "household bounded pass blocks the endpoint",
+			boundedJob: "household_custom_values_family_camp",
+			registerBound: func(o *Orchestrator) {
+				o.RegisterService("household_custom_values_family_camp",
+					&MockService{name: "household_custom_values_family_camp"})
+			},
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			scheduler := NewScheduler(nil)
+			orchestrator := scheduler.GetOrchestrator()
+			tc.registerBound(orchestrator)
+
+			if err := orchestrator.MarkSyncRunning(tc.boundedJob); err != nil {
+				t.Fatalf("MarkSyncRunning(%q): %v", tc.boundedJob, err)
+			}
+
+			re := &core.RequestEvent{}
+			re.Request = httptest.NewRequest(http.MethodPost, "/", http.NoBody)
+			rec := httptest.NewRecorder()
+			re.Response = rec
+
+			if err := handleCustomValuesSync(re, scheduler); err != nil {
+				t.Fatalf("handler returned error: %v", err)
+			}
+
+			if rec.Code != http.StatusConflict {
+				t.Errorf("expected %d (already running), got %d: %s",
+					http.StatusConflict, rec.Code, rec.Body.String())
+			}
+		})
+	}
+}
+
+// TestHandleCustomValuesSyncAllowsRunWhenNothingRunning is the control case: with nothing
+// running (bounded or unrestricted), the endpoint must still succeed.
+func TestHandleCustomValuesSyncAllowsRunWhenNothingRunning(t *testing.T) {
+	t.Parallel()
+	scheduler := NewScheduler(nil)
+
+	re := &core.RequestEvent{}
+	re.Request = httptest.NewRequest(http.MethodPost, "/", http.NoBody)
+	rec := httptest.NewRecorder()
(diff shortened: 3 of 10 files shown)


adamflagg:
## Scan result: MERGE-ELIGIBLE, fixed in-branch, merging

Ran `code-review` at **xhigh** effort (the skill's own reported level; the campaign asked for high, xhigh is what actually ran). Full transcript verified independently — mutation-checked, `-race` clean, `golangci-lint` clean, full `go test ./...` green.

### Findings — both fixed in-branch (commit `47edc428`)

Both are the same root cause this PR's own fix addresses (a job name registered distinct from its unrestricted sibling is invisible to anything keyed on the literal string) — a fourth and fifth consumer of that same gap, not this PR's premise extended somewhere new.

1. **`LogSyncComplete` still hardcoded the pre-Face-D literal.** `Sync()`'s two `LogSyncComplete` call sites (the "nothing to sync" early return and the end-of-run summary, in both `person_custom_field_values.go` and `household_custom_field_values.go`) still passed `"PersonCustomFieldValues"` / `"HouseholdCustomFieldValues"` even after `logJobName()` started feeding `LogSyncStart` and the cohort log. That's the exact line Face D's own problem statement names — *"an operator diagnosing stale data...sees person_custom_values completing nightly"* — the word "completing" is this line, the one carrying the created/updated/errors counts. Fixed: both call sites now use the already-computed `jobName` variable.
   - Pinned with two new tests (`TestPersonCustomFieldValuesSync_CompletionLogUsesBoundedJobName`, `TestHouseholdCustomFieldValuesSync_CompletionLogUsesBoundedJobName`) driving the bounded instance's real "no family-camp sessions this year" early return — no CampMinder mock needed, since `GetFamilyCampSessionCMIDs` only queries `camp_sessions` and returns empty before any network call. Mutation-checked: reverting either fix turns its test red with the exact old literal in the log output.

2. **`SyncJobToCollections` (table_exporter.go) never gained entries for the two bounded job names.** `GetChangedCollections()` iterates every completed job by its registered name; without an entry, the bounded pass's real writes were silently dropped from the Sheets export skip-optimization — the daily cron's fresh family-camp data would never trigger an export of `person_custom_values`/`household_custom_values`. Fixed: added `"person_custom_values_family_camp": {"person_custom_values"}` and the household twin.
   - Pinned with a new subtest on the existing `TestOrchestrator_GetChangedCollections` table (`"bounded family-camp jobs map to the same collections as their unrestricted siblings"`). Mutation-checked: reverting the two map entries turns it red.

Also added the two new tests to `main_test_parallelism_test.go`'s `serialTests` registry (they use `t.Setenv` for a real `campminder.Client`, same reason as the existing `TestAttendeesLogStatusChangeDryRunWritesNothing` group) — required companion registration, not a new finding.

### Verified, not just trusted

- `main` claims in the PR body cross-checked against source: cron schedule (`0 3 * * *` / `0 4 * * 0`), `MarkSyncRunning`'s only production callers (`process_requests`, grepped), `handleGetPhases` still calling unfiltered `GetJobsForPhase`, `TestSyncJobMeta_FamilyCampBoundedJobsAreExpensivePhase` still green — all matched exactly.
- Predecessors #2489 and #2493 both confirmed MERGED via `gh pr view`.
- `go build ./...`, `go test ./...` (full suite, all packages), `go test -race` on the touched concurrency tests, `golangci-lint run` (0 issues), `bash scripts/pre-push-verify.sh` (ALL CHECKS PASSED) — all run locally, all clean.

### Merge gate

1. No decision remains — both findings were mechanical, no design judgment.
2. CI Summary actually ran and is green (confirmed via `gh api .../check-runs`, not just the rollup).
3. Scan complete, findings fixed in-branch.
4. Push verified by the remote ref moving (`git rev-list --count origin/...HEAD` = 0).
5. `closingIssuesReferences` = `[2491]`, confirmed via GraphQL.
6. Diff is still exactly the item the campaign described (Faces A–D of kindred#2491), plus the two adjacent gaps above.
7. No visual impact — zero files under `frontend/src/` in the diff.
8. No staff-facing number changes meaning — this is sync-orchestration locking, phase execution, and log labeling only; nothing here touches what any metric counts or how it's computed.

Merging.

Issue #2476, opened 2026-08-19
feat(frontend,api): Share Bunk With moves to the bottom of the six request blocks, starts collapsed, and its chevron dismisses the modal


adamflagg:
## What was reported

> share bunk with - put at bottom, start closed. Change arrow on side to click

Reported from the weekend housing board's family details panel.

### 1. "put at bottom" — a BACKEND ordering change

Block order is fixed server-side in `REQUEST_TEXT_SOURCES` (`api/services/lodging_rules.py:323-330`) and emitted in that order at `api/services/lodging_roster_service.py:366`. The frontend does not re-sort.

**⚠️ The block names are ALIASES and do not describe what they are.** This matters for anyone reading the list cold:

| # | source constant | what it actually is | authorship |
|---|---|---|---|
| 1 | `COVID-19 Bunking Requests` | **the reg bunk form** — not a COVID-era artefact | family |
| 2 | `Shared-request` | a shared-request form | family |
| 3 | `FAM CAMP-Share Comments` | a shared-request form | family |
| 4 | `BunkingNotes Notes` | staff note | staff |
| 5 | `Internal Bunk Notes` | staff note | staff |
| 6 | `Share Bunk With` | the bunking-CSV lane's `bunk_request_form` | family |

**One of the two shared-request forms (2 or 3) is dead post-2025**, which is why staff see five blocks rather than six.

**★ OWNER RULING:** `Share Bunk With` moves to the **bottom of the six** — below both staff notes. Blocks 1–3 are the family-specific ones and stay above the staff notes.

**Reason, recorded because it is the whole justification:** `Share Bunk With` is more often **summer**-specific, whereas the two staff-type notes are summer and/or family, and the top blocks come from custom field values that are explicitly Family Camp.

### 2. "start closed" — needs new machinery, and revises a prior ruling

Today **every** block starts expanded: `ShareRequestPanel.tsx:239` seeds `folded` to an empty `Set`, and `:292` computes `expanded={!folded.has(...)}`.

**★ OWNER RULING:** `Share Bunk With` starts **collapsed**; the other four stay open.

The doc block at `ShareRequestPanel.tsx:30-34` records the 2026-08-17 ruling that *"blocks start EXPANDED, because a scanning eye must not miss request text… The expectation is explicitly that this ships and gets tuned against real staff use."* **This is that tuning** — so the comment must be rewritten, not left contradicting the screen.

[...]

PR #2521, opened 2026-08-21
feat(api,frontend): Share Bunk With moves to the bottom and starts collapsed

adamflagg:
## Summary

Two of #2476's three asks. `Share Bunk With` moves to the last of the six
request-text blocks (a staff-specified order, not derived from authorship or
volume), and it now starts collapsed while its five siblings stay expanded.

**The third ask — the chevron hit-target / dismiss-region bug — is NOT
included. See "Deliberately not done" below. This issue stays OPEN.**

## 1. Block order (`api/services/lodging_rules.py`)

`REQUEST_TEXT_SOURCES` placed `Share Bunk With` second of six. Owner ruling
2026-08-21: staff asked for it last — `COVID-19 Bunking Requests`,
`Shared-request`, `FAM CAMP-Share Comments`, `BunkingNotes Notes`,
`Internal Bunk Notes`, `Share Bunk With`. One-line reorder; nothing else in
the tuple moves.

This retires two previously pinned test invariants on purpose:
- The exact-order assertion on `REQUEST_TEXT_SOURCES`.
- The authorship invariant that every family-authored block sorted ahead of
  every staff-authored one — `Share Bunk With` is family-authored and now
  sorts after both staff notes.

Both are rewritten with the new rationale, written into the test bodies: the
order is what staff asked for, and it must not be re-derived from
authorship or volume by a later reader. On 2026 family-camp households
(denominator 479), `Share Bunk With` is the SECOND MOST POPULATED of the six
blocks (234) — ahead of `Shared-request` (114) and `FAM CAMP-Share Comments`
(0) — yet staff put it last.

## 2. Starts collapsed (`frontend/src/components/weekend/ShareRequestPanel.tsx`)

`Share Bunk With` now starts folded; the other five blocks keep the
2026-08-17 "start expanded" default. The default has to be applied in TWO
places — the initial `useState` seed and the reset-on-household-change
branch — because the panel is never remounted between families; applying it
only to the initializer would re-open the block the moment staff click the
next household. A dedicated test (`THE TRAP: a household switch
re-collapses Share Bunk With rather than opening it`) pins this, and was
verified to fail against that half-fix before the full fix landed.

The stale 2026-08-17 doc comment ("blocks start EXPANDED … explicitly that
this ships and gets tuned against real staff use") is rewritten to record
this as that tuning.

Three pre-existing tests used `Share Bunk With` as incidental fixture data
for assertions unrelated to fold state (per-field split, per-child split,
"leaves its neighbour open"). Swapped to a different family field so they
keep testing what they always tested without becoming entangled with the
new default-fold exception.

## Deliberately NOT done

**Ask 3 — "change arrow on side to click" (the chevron hit-target /
dismiss-region bug) — is not implemented here.** #2476 frames this as an
interaction bug: clicking the chevron near the panel's left edge dismisses
`FamilyDetailsPanel` instead of collapsing the block, while clicking the
field name (further right, same `<button>`) works correctly.

I could not identify a code path that reproduces this from static analysis
with the confidence needed to fix it correctly:

- `FamilyDetailsPanel`'s own backdrop (`family-panel-backdrop`) is a
  self-closing `pointer-events-none` div — it can never be the real target
  of a browser click, so its `onClick={handleClose}` cannot be what fires.
- The dismiss mechanism actually wired at all three `FamilyDetailsPanel`
  call sites is `useDismissOnDeadSpace` + `shouldKeepPanelsOpen`
  (`utils/clickoutsidePredicate.ts`), which decides on DOM ancestry
  (`target.closest('[data-panel="family-details"]')`,
  `target.closest('button, …')`). The chevron toggle is a `<button>`
  nested inside `[data-panel="family-details"]`, so any click whose real
  DOM target is that button is protected by two independent checks in that
  predicate, regardless of how close to the visual edge it sits.
- The only account I can square with the reported symptom is a genuine
  pixel-level miss — the 12px chevron icon, sitting near the panel's own
  left padding, occasionally receiving a click whose real coordinates land
  outside the panel's DOM subtree entirely. Confirming that (and picking
  the right fix — hit-target margin vs. something else) needs a live
  browser, not jsdom (no real layout engine) and not more code reading.

I attempted to verify this empirically against a live environment via the
Playwright MCP browser, but the shared browser instance was held by a
concurrent session for the duration of this work
(`Error: Browser is already in use for .../mcp-chrome-9cfc516`) and never
freed up. Rather than ship a speculative fix for a mechanism I can't pin
down or test-drive, I'm leaving Ask 3 for a follow-up with real-browser
verification.

## Also flagged, not fixed here (file out of this item's scope)

`api/services/lodging_roster_service.py:350` documents the roster-block
order as "family-authored first, staff notes last" — now stale after this
reorder. That file isn't in this item's file list; noting it here so the
staleness isn't lost.

Part of #2476.

### DIFF
--- a/api/schemas/lodging.py
+++ b/api/schemas/lodging.py
@@ -551,9 +551,11 @@ class ShareRequestSummary(BaseModel):
     # family-authored blocks and `request_text` are not gated. See
     # `_may_read_staff_notes` in api/routers/lodging.py.
     #
-    # Ordered by `REQUEST_TEXT_SOURCES` -- family-authored blocks first, staff
-    # notes last. A source field with no text produces NO block: kindred#2255's
-    # ruling for this same modal leaves no "nothing applicable" clutter behind.
+    # Ordered by `REQUEST_TEXT_SOURCES` -- a staff-specified order (kindred#2476,
+    # owner ruling 2026-08-21), NOT derived from authorship or volume. `Share
+    # Bunk With` is family-authored and sorts last, after both staff notes. A
+    # source field with no text produces NO block: kindred#2255's ruling for
+    # this same modal leaves no "nothing applicable" clutter behind.
     request_blocks: list[RequestTextBlock] = Field(default_factory=list)
 
     # What the board actually places on. `preference` above stays the raw
--- a/api/services/lodging_rules.py
+++ b/api/services/lodging_rules.py
@@ -373,11 +373,12 @@ class RequestTextSource(NamedTuple):
 
     `label` is the CampMinder field name VERBATIM, including the misnamed
     `COVID-19 Bunking Requests` -- the field that has plainly been repurposed
-    as the general bunking-request question and carries 205 of the 382
-    households rostered into a 2026 family session. Owner ruling 2026-08-17:
-    "call them the original fieldnames for now until staff can weigh in after
-    it's live". A display-names issue gets filed once they have. Do not
-    "improve" these strings.
+    as the general bunking-request question and carries 238 of the 479
+    households rostered into a 2026 family-camp registration (kindred#2476,
+    re-measured 2026-08-21). Owner ruling 2026-08-17: "call them the original
+    fieldnames for now until staff can weigh in after it's live". A
+    display-names issue gets filed once they have. Do not "improve" these
+    strings.
     """
 
     label: str
--- a/api/services/lodging_rules.py
+++ b/api/services/lodging_rules.py
@@ -384,14 +384,23 @@ class RequestTextSource(NamedTuple):
     authorship: RequestTextAuthorship
 
 
-# ORDERED, and the order is the panel's block order. Family-authored fields
-# first so a household's own ask leads, staff-authored notes last; within a
-# lane, by 2026 coverage, so the field staff read most often is at the top.
+# ORDERED, and the order is the panel's block order.
+#
+# kindred#2476, owner ruling 2026-08-21: this order is what STAFF ASKED FOR.
+# It is NOT derived from authorship or volume, and it must NOT be re-derived
+# from the data by a later reader. On 2026 family-camp households
+# `Share Bunk With` is the SECOND MOST POPULATED of the six blocks (234 of
+# 479) -- ahead of `Shared-request` (114) and `FAM CAMP-Share Comments` (0)
+# -- yet staff placed it last. The order used to be "family-authored fields
+# first, staff notes last, each lane sorted by 2026 coverage"; that rule no
+# longer holds (`Share Bunk With` is family-authored and sorts after both
+# staff notes) and must not be reinstated by "correcting" the order back to
+# match volume or authorship.
 #
 # Counts are rostered households on `pocketbase/pb_data/data-prod.db`,
-# denominator 382 (`status_id = 2`, 2026's eight family sessions):
-# COVID-19 Bunking Requests 205, Share Bunk With 104, Shared-request 100,
-# BunkingNotes Notes 28, Internal Bunk Notes 8.
+# denominator 479 (2026 family-camp registrations):
+# COVID-19 Bunking Requests 238, Share Bunk With 234, Shared-request 114,
+# BunkingNotes Notes 111, Internal Bunk Notes 10.
 #
 # `FAM CAMP-Share Comments` is 0 for 2026 and is carried anyway. It is one of
 # the three fields the Go ingest already joins into `request_text` (live
@@ -411,11 +420,11 @@ class RequestTextSource(NamedTuple):
 #   an immaterial source field.
 REQUEST_TEXT_SOURCES: tuple[RequestTextSource, ...] = (
     RequestTextSource("COVID-19 Bunking Requests", "family"),
-    RequestTextSource("Share Bunk With", "family"),
     RequestTextSource("Shared-request", "family"),
     RequestTextSource("FAM CAMP-Share Comments", "family"),
     RequestTextSource("BunkingNotes Notes", "staff"),
     RequestTextSource("Internal Bunk Notes", "staff"),
+    RequestTextSource("Share Bunk With", "family"),
 )
 
 # The family-camp lane: CampMinder custom-field ids, matching the three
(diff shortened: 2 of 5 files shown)


adamflagg:
## Scan result: HELD for owner review (visual impact) — not merged

This item is flagged by the campaign as visual impact (reorders a staff-facing
panel, changes which block starts collapsed) and holds regardless of scan
outcome. Recording what the scan found below; **PR stays open.**

Ran `code-review` at **high** effort (single-pass, 10 review angles, no
subagent fan-out — reported by the reviewer itself, noted here per the
"report the level it actually ran at" instruction).

### Fixed in-branch (3 commits, all doc-staleness, no behavior change)

1. **`api/services/lodging_rules.py:376`** (pre-existing, same file this PR
   already edits) — the `RequestTextSource` class docstring still said
   `COVID-19 Bunking Requests` carries "205 of the 382" households, left
   behind when the `REQUEST_TEXT_SOURCES` comment block a few lines below it
   was re-measured to "238 of the 479" for this PR. Updated to match.
   Commit `4466827`. (Found independently before the scan; the reviewer
   found the same defect mid-review and dropped it from its own report once
   it saw the fix land on the branch.)
2. **`api/schemas/lodging.py:554`** (pre-existing, not in this PR's original
   file list — small, doc-only, no schema/contract change, so fixed here per
   the fix-by-default rule) — `ShareRequestSummary.request_blocks`'s wire-
   contract docstring still said "family-authored blocks first, staff notes
   last," the exact invariant this PR retires. Updated to describe the
   actual current order (staff-specified, kindred#2476).
3. **`ShareRequestPanel.test.tsx`** module docstring (this PR's own file) —
   still asserted blocks start EXPANDED as a blanket rule with no caveat for
   the new `Share Bunk With` exception the file's own new `describe` block
   exists to pin.

Both (2) and (3) found by the `code-review` high-effort pass; both are the
same class of staleness this PR already fixed in two other spots
(`lodging_rules.py`, `ShareRequestPanel.tsx` component doc comment) but
missed here. Commit `b2833e4`.

### Independently verified (read-only, `pocketbase/pb_data/data-prod.db`)

Re-derived every count this PR's comments/tests cite, joining
`original_bunk_requests` (bunking-CSV lane) and `person_custom_values` →
`custom_field_defs` (family-camp lane) to `persons.household` and
`family_camp_registrations` for `year = 2026`, non-blank content, distinct
households:

| field | claimed | measured |
|---|---:|---:|
| denominator (2026 family-camp registrations) | 479 | **479** |
| `COVID-19 Bunking Requests` | 238 | **238** |
| `Share Bunk With` | 234 | **234** |
| `Shared-request` | 114 | **114** |
| `FAM CAMP-Share Comments` | 0 | **0** |
| `BunkingNotes Notes` | 111 | **111** |
| `Internal Bunk Notes` | 10 | **10** |

All exact matches. `code-review` independently re-derived the same figures
and confirmed the same result.

### Mutation-checked the pinned trap test

`THE TRAP: a household switch re-collapses Share Bunk With rather than
opening it` — reverted the reset branch to the pre-PR
`if (folded.size > 0) setFolded(new Set())` (i.e. simulated the half-fix the
test exists to catch) and reran: the test failed as expected (`Share Bunk
With` reopened on the second household — "The second family's fridge ask"
rendered when it should still be folded). Restored the real fix, reran:
44/44 pass. The test pins what it claims to pin. `code-review` separately
hand-traced the `DEFAULT_FOLDED`/`isDefaultFolded` logic through every
household-switch/toggle permutation and reached the same conclusion.

### Ran full verification (after all 3 fix commits)

- `uv run pytest tests/unit/api/` (lodging-scoped) — 742 passed
- `frontend: vitest run ShareRequestPanel.test.tsx` — 44 passed
- `tsc --noEmit` — clean
- `./node_modules/.bin/eslint` / `./node_modules/.bin/prettier --check` on
  touched frontend files — clean
- `ruff check` / `ruff format --check` / `mypy` on touched Python files — clean
- Full `pre-push-verify.sh` pipeline (pytest 6624 passed / 41 skipped, mypy,
  ruff, api-types-freshness, tsc) — clean on both fix commits
- Push verified by remote ref both times, latest:
  `origin/feature/share-bunk-with-order` == local `HEAD` (`b2833e4b`),
  0 commits ahead/behind
- CI Summary: green on the final commit

### Premise review

Both shipped changes (block order, start-collapsed) trace to an explicit,
dated owner ruling recorded directly in #2476's body (2026-08-21), including
an unusual and valuable step: the *original* rationale (volume/provenance)
was measured, found false, and the body says so and replaces it with "staff
asked for it, don't re-derive from data." That is not a hidden product
decision riding on a bug fix — it is a decision that was already made,
recorded, and the PR faithfully implements it without re-litigating or
smuggling in extra scope.

### Deliberately not extended

- **Ask 3 (chevron hit-target bug)** is correctly left out of this PR. Traced
  the reasoning in the PR body against the actual code
  (`FamilyDetailsPanel`'s `pointer-events-none` backdrop, `useDismissOnDeadSpace`
  + `shouldKeepPanelsOpen` in `clickoutsidePredicate.ts`) — the described
  mechanism is real and the conclusion (needs live-browser verification, not
  more static reading) is sound. Not filing a sibling issue; #2476 stays open
  and already carries this as unfinished.
- **`lodging_roster_service.py:350`** stale doc comment — already flagged in
  the PR body as out of this item's scope. Agreed, and left alone — it's a
  bigger, multi-file staleness note (not the same 1-line mechanical fix as
  the two picked up above), and whoever picks up #2476's remaining ask can
  fold it in.

### Merge gate

Not evaluated for merge — **campaign classification holds this PR for the
owner's eyes** regardless of gate status. `closingIssuesReferences` is
correctly empty: the PR body deliberately does not close #2476 (only 2 of
3 asks are done), so the issue stays open until the chevron fix lands.

**What the owner would be looking at:** on the weekend family-details panel,
the `Share Bunk With` request-text block moves from 2nd to last (6th) of the
six blocks, and it now renders collapsed by default (staff must click to
expand it) while the other five blocks keep rendering expanded, as before.
No other block's order or default state changes.

PR #2522, opened 2026-08-21
feat(frontend): centre the modal close button in its header band

### LINKED ISSUE #2507: feat(frontend): standardise modal close buttons on the artifact's 18px in-flow mark
## What

Standardise every `ui/Modal` custom-header dialog on the design artifact's close mark: an **18×18 circled `×`, in flow at the end of the header row** (`margin-left: auto`), rather than the 36×36 rounded square floated at `top-4`.

From the artifact's own CSS (kindred#2072's review artifact, `.wbtn`):

```css
.wbtn {
  display: inline-flex; align-items: center; justify-content: center; flex: 0 0 auto;
  width: 18px; height: 18px; border: 1px solid var(--border); border-radius: 9999px;
  background: transparent; color: var(--muted-foreground);
  font-size: 11px; line-height: 1; padding: 0;
}
.mhead .wbtn { margin-left: auto; }
```

## Why

Owner ruling 2026-08-20, after comparing the built Assign modal against the artifact: *"I like option B, but I think we file that as a future issue enhancement because I'd want to standardise the modals on it and that would be widely impactful."*

Three things it buys, beyond looking like the design:

1. **It cannot overhang.** The floated button's `top-4` is a constant that assumes a header at least 52px tall (16 + a 36px box). Header height belongs to the caller, so a dialog that tightens its own header makes the button hang past it — measured at 5px on `AssignFamilyModal` while it still drew a header rule, where the hover fill painted across the rule and the hit area covered the top edge of the search box. An in-flow mark is laid out by the header, so it cannot come apart.
2. **It gives the header its width back.** The floated button forces `pr-14` (56px) of reserved right padding on every custom header. In flow, a long title or unit name gets that space and wraps later.
3. **One grammar.** ~20 dialogs currently draw the same floated square; whichever way this goes it should be all of them, not one.

## Scope

- `frontend/src/components/ui/Modal.tsx` — the custom-header branch, and the two other close-button branches (simple-title and no-header) should be reviewed at the same time so the app does not end up with two marks.
- Every caller using the `header` slot: drop the now-unnecessary `pr-14`-style reservations.
- `closeAlign` (added on the #2506 branch as the interim fix for the overhang) becomes dead once this lands — delete it with this change rather than leaving a second mechanism behind.
- `Modal.test.tsx` pins the current default (`top-4`) and the centred opt-in; both need retargeting, and the accessible name (`Close modal`) must survive, because tests across the app query the close control by role and name.

## Interim state

PR #2506 centres the Assign modal's floated button in its header band (`closeAlign="center"`, opt-in, default unchanged). That closes the overhang on the one dialog that hit it; it is not this issue.

Relates to #2072.

adamflagg:
**Reworked.** This branch previously built #2507 as filed — the artifact's 18px in-flow circled
mark. That was shown to the owner and **rejected on sight**. The 36×36 control is kept; only
*where it sits* changes.

## The defect

`top-4` is a **constant**; header height is not. A top-anchored button assumes a header at least
52px tall (16px + a 36px box), so it reads high in a tall header and hangs past a short one. The
owner spotted it on Manage Scenarios: the X *fits*, but sits high rather than centred between the
top of the dialog and the rule under the title.

The wrapper the button positions against is the **header slot itself**, so centring in the band
cannot come apart whatever height a caller's header turns out to be.

## Measured, not predicted

Chromium, against this component, with each caller's own header markup — gap **above / below**
inside the band:

| Dialog | band | `top-4` today | centred |
|---|---|---|---|
| Manage Scenarios | 81px | 16 / 29 — **13px high** | **22.5 / 22.5** |
| Lodging Units | 88px | 16 / 36 — **20px high** | **26 / 26** |
| Heads Up | 74.5px | 16 / 22.5 | **19.25 / 19.25** |
| Assign Family | 51px | 16 / **−1** — overhangs | **7.5 / 7.5** |

Centring returns **equal gaps at every dialog measured**. The other 20 never read the setting: the
simple-title branch is already an `items-center` flex row, the no-header branch has no band, and
both measured **byte-identical** at either setting — outside the change, not merely unharmed.

## Why this wasn't already the default

`closeAlign` has offered centring since #2506, and its own docstring said centring *"would be right
for any header"* — it stayed opt-in only because *"moving them is its own review."* **That review
ran.** The docstring is rewritten rather than left asserting "DEFAULT UNCHANGED, deliberately",
which becomes false the moment this lands.

## A latent bug the flip would have exposed

`hasCustomHeader` was `header !== undefined`, and `null !== undefined` is **true** — so
`PostValidationResultsModal`, which passes `header={null}`, rode the custom-header branch with a
**zero-height** band. `top-4` survived by luck. Centring on 0px computes `0 − 18 = −18px`, putting
half the button above the panel's `overflow-hidden` edge: invisible and unclickable. Guarded here,
because this change is what makes it load-bearing.

## Deliberately not done

- **`closeAlign` is kept, not deleted.** The earlier cut removed it (in-flow made "top vs centre"
  meaningless). A header that wants the old anchor can still say `closeAlign="top"` — pinned by a
  test — rather than having to re-litigate it.
- **No caller pins `"top"` yet.** Three two-row headers (Manual Resolution, Cohort Drill-Down,
  Household Year Members) are computed but not photographed. A math-only pass called Manual
  Resolution "worse"; that grades against *title* alignment, which is not the stated criterion —
  the ask was centring within the band. Left on the default; if one looks wrong, it's a one-line
  opt-out.
- **No `pr-14` padding was touched.** The button is still floated, so those reservations are still
  correct — unlike under the in-flow approach, which required removing them everywhere.

## Tests

The default assertion is **rewritten, not adapted** — the spec changed, not the code. Added: the
`"top"` opt-out, and the `header={null}` guard. That last one is asserted on the **branch** (the
button carries `absolute` only in the custom-header branch) rather than on `top-1/2`, which is
absent under the top anchor too and would have passed against the very bug it pins.
Mutation-checked: reverting the default turns 1 red; removing the null guard turns 1 red.

Closes #2507.

### DIFF
--- a/api/schemas/lodging.py
+++ b/api/schemas/lodging.py
@@ -551,9 +551,11 @@ class ShareRequestSummary(BaseModel):
     # family-authored blocks and `request_text` are not gated. See
     # `_may_read_staff_notes` in api/routers/lodging.py.
     #
-    # Ordered by `REQUEST_TEXT_SOURCES` -- family-authored blocks first, staff
-    # notes last. A source field with no text produces NO block: kindred#2255's
-    # ruling for this same modal leaves no "nothing applicable" clutter behind.
+    # Ordered by `REQUEST_TEXT_SOURCES` -- a staff-specified order (kindred#2476,
+    # owner ruling 2026-08-21), NOT derived from authorship or volume. `Share
+    # Bunk With` is family-authored and sorts last, after both staff notes. A
+    # source field with no text produces NO block: kindred#2255's ruling for
+    # this same modal leaves no "nothing applicable" clutter behind.
     request_blocks: list[RequestTextBlock] = Field(default_factory=list)
 
     # What the board actually places on. `preference` above stays the raw
--- a/api/services/lodging_rules.py
+++ b/api/services/lodging_rules.py
@@ -373,25 +373,35 @@ class RequestTextSource(NamedTuple):
 
     `label` is the CampMinder field name VERBATIM, including the misnamed
     `COVID-19 Bunking Requests` -- the field that has plainly been repurposed
-    as the general bunking-request question and carries 205 of the 382
-    households rostered into a 2026 family session. Owner ruling 2026-08-17:
-    "call them the original fieldnames for now until staff can weigh in after
-    it's live". A display-names issue gets filed once they have. Do not
-    "improve" these strings.
+    as the general bunking-request question and carries 238 of the 479
+    households rostered into a 2026 family-camp registration (kindred#2476,
+    re-measured 2026-08-21). Owner ruling 2026-08-17: "call them the original
+    fieldnames for now until staff can weigh in after it's live". A
+    display-names issue gets filed once they have. Do not "improve" these
+    strings.
     """
 
     label: str
     authorship: RequestTextAuthorship
 
 
-# ORDERED, and the order is the panel's block order. Family-authored fields
-# first so a household's own ask leads, staff-authored notes last; within a
-# lane, by 2026 coverage, so the field staff read most often is at the top.
+# ORDERED, and the order is the panel's block order.
+#
+# kindred#2476, owner ruling 2026-08-21: this order is what STAFF ASKED FOR.
+# It is NOT derived from authorship or volume, and it must NOT be re-derived
+# from the data by a later reader. On 2026 family-camp households
+# `Share Bunk With` is the SECOND MOST POPULATED of the six blocks (234 of
+# 479) -- ahead of `Shared-request` (114) and `FAM CAMP-Share Comments` (0)
+# -- yet staff placed it last. The order used to be "family-authored fields
+# first, staff notes last, each lane sorted by 2026 coverage"; that rule no
+# longer holds (`Share Bunk With` is family-authored and sorts after both
+# staff notes) and must not be reinstated by "correcting" the order back to
+# match volume or authorship.
 #
 # Counts are rostered households on `pocketbase/pb_data/data-prod.db`,
-# denominator 382 (`status_id = 2`, 2026's eight family sessions):
-# COVID-19 Bunking Requests 205, Share Bunk With 104, Shared-request 100,
-# BunkingNotes Notes 28, Internal Bunk Notes 8.
+# denominator 479 (2026 family-camp registrations):
+# COVID-19 Bunking Requests 238, Share Bunk With 234, Shared-request 114,
+# BunkingNotes Notes 111, Internal Bunk Notes 10.
 #
 # `FAM CAMP-Share Comments` is 0 for 2026 and is carried anyway. It is one of
 # the three fields the Go ingest already joins into `request_text` (live
@@ -411,11 +421,11 @@ class RequestTextSource(NamedTuple):
 #   an immaterial source field.
 REQUEST_TEXT_SOURCES: tuple[RequestTextSource, ...] = (
     RequestTextSource("COVID-19 Bunking Requests", "family"),
-    RequestTextSource("Share Bunk With", "family"),
     RequestTextSource("Shared-request", "family"),
     RequestTextSource("FAM CAMP-Share Comments", "family"),
     RequestTextSource("BunkingNotes Notes", "staff"),
     RequestTextSource("Internal Bunk Notes", "staff"),
+    RequestTextSource("Share Bunk With", "family"),
 )
 
 # The family-camp lane: CampMinder custom-field ids, matching the three
--- a/frontend/src/components/AllCamperRequestsModal.tsx
+++ b/frontend/src/components/AllCamperRequestsModal.tsx
@@ -265,7 +265,7 @@ export function AllCamperRequestsModal({
 
   const header = (
     <div
-      className="border-border/60 border-b pt-[22px] pb-[18px] pl-7"
+      className="border-border/60 border-b pt-[22px] pr-16 pb-[18px] pl-7"
       style={{
         backgroundImage:
           'radial-gradient(120% 160% at 0% 0%, color-mix(in oklch, var(--color-forest-50, oklch(97% 0.01 145)) 85%, transparent) 0%, transparent 60%)',
--- a/frontend/src/components/AllCamperRequestsModal.tsx
+++ b/frontend/src/components/AllCamperRequestsModal.tsx
@@ -265,7 +265,7 @@ export function AllCamperRequestsModal({
 
   const header = (
     <div
-      className="border-border/60 border-b pt-[22px] pr-16 pb-[18px] pl-7"
+      className="border-border/60 border-b pt-[22px] pb-[18px] pl-7"
       style={{
         backgroundImage:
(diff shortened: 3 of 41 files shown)


adamflagg:
@coderabbitai review

adamflagg:
## Scan result — HELD for visual review, not merged

Per the campaign's classification this PR changes the close control's appearance on ~20
dialogs across the app (a shared UI primitive), so it holds for the owner's eyes regardless
of scan/CI outcome. It is **not** being merged by this pass.

### Review (code-review skill, high effort, floor 60)

Ran once as a single-pass review (no `Agent`-tool subagent fan-out available in this run —
narrower coverage than a full multi-agent pass, noting it so the level is honest). Verified
independently: read all 9 touched files in full, ran `tsc --noEmit` (clean), `eslint` on every
touched file (0 errors), full Vitest suite (452 files / 7248 tests green pre-fix), traced the
grid-stacking mechanics against every `headerOnDark` caller for occlusion (none found), and
hand-checked `SolverDebugImpossibilityModal`'s Copy-JSON button clears the new mark by 14px.

Four findings came back. Two fixed in-branch, one logged as a disagreement (not extending file
scope), one refuted alongside its twin:

**Fixed in-branch** (commit `303b012e`, code this PR itself wrote — Modal.tsx):
1. **`header={null}` fallthrough had no test.** `hasCustomHeader`'s `header !== null` clause
   (added by this PR so `PostValidationResultsModal`'s `header={null}` still gets the floating
   mark rather than a collapsed ~18px grid row) was unguarded — a later "simplification" that
   drops the clause would regress that caller silently. Added
   `treats header={null} as "no custom header" …` to `Modal.test.tsx`.
   **Mutation-checked**: reverted the clause to `header !== undefined` only, reran — the new
   test failed exactly as expected (`expected 'grid grid-cols-1' to contain 'absolute'`), then
   reverted the mutation and confirmed green again.
2. **The 18px close mark's Tailwind classes were duplicated near-verbatim across all three
   close-button branches** (custom header / simple title / no header) — a future resize would
   need three hand-edits with no compiler signal if one were missed, reopening exactly the
   "two grammars" problem this PR exists to close. Extracted
   `CLOSE_MARK_BASE_CLASSES` / `CLOSE_MARK_LIGHT_CLASSES` / `CLOSE_MARK_DARK_CLASSES`. Every
   individual class token is unchanged — only where it's declared — confirmed by the existing
   `toContain('h-[18px]')`-style assertions in `Modal.test.tsx` still passing untouched.

**Logged as disagreement, not fixed here:**
3 & 4. `HouseholdYearMembersModal.tsx`'s `pr-12` and `AllCamperRequestsModal.tsx`'s `pr-16`
   are stale reservations sized for the old 36px button. Real, but **out of this PR's file
   scope** — and the PR body's own "Deliberately not done" section already names both of these
   exact files and defers them on purpose. Extending file scope to cover "the same design
   decision applies over here too" is new scope, not an in-branch fix, per this run's own
   ground rules. Left untouched; no separate issue filed either — #2507 already tracks the
   full ~20-dialog rollout and the remaining callers are enumerated in this PR's body, so a new
   issue would just duplicate that list.

**Post-fix verification:** full suite **452 files / 7249 tests green** (452/7249, the +1 being
the new test), `tsc --noEmit` clean, `scripts/pre-push-verify.sh` all green (prettier/eslint/
tsc/vitest), pushed and confirmed by the remote ref (`git rev-list --count origin/…​..HEAD` = 0).

### CodeRabbit

Requested explicitly (`@coderabbitai review`) and verified a real review landed:
**Actionable comments posted: 1**, not a rate-limit/skip notice.

Its one finding: the shared 18px mark is below the ~24px minimum interactive-target
recommendation, and it suggests keeping the 18px *visual* mark inside a ≥24px button.
This is a legitimate general point, but the 18×18 sizing is not an oversight — it's copied
verbatim from the design artifact's own `.wbtn` CSS, quoted in #2507's body, from an explicit
owner ruling ("I like option B..."). Changing hit-target size is itself a visual/design change
across the same ~20 dialogs this PR already touches conceptually, so I did not act on it
unilaterally — flagging it for the owner's visual review pass alongside the PR's own
grid-stack-vs-flex judgment call (see the PR body's "one deliberate deviation" section), rather
than overriding an explicit design ruling on my own authority.

### Merge gate

- CI Summary: **SUCCESS** (confirmed after the fix-commit push, not the earlier one).
- `closingIssuesReferences`: **non-empty** — `#2507` (confirmed via GraphQL, not by reading the
  body).
- Diff still recognisably the item the campaign described: yes — same 9 files, same mechanism.
- **GUI/visual impact: yes** — `Modal.tsx` + 6 hand-written callers, all under
  `frontend/src/`, none matched by the generated-type exclusion. This is exactly why the PR
  stays open.
- Staff-facing number meaning: **no** — pure UI/styling change, no metric, count, rate, or
  status classification touched.

**Status: held open for owner visual review.** What the owner will be looking at: every modal
in the app that used a custom header (~20 dialogs — request/scenario/validation modals, the
lodging units panel, the weekend Assign-Family modal, etc.) now draws its close `×` as a small
18px circled outline sitting in the header's own row, flush to the right edge, instead of the
larger 36px floating square that sat in the top-right corner independent of the header. Two
things specifically worth an eyeball: (1) whether 18px reads as too small a click target on a
real monitor — CodeRabbit's point above — and (2) the grid-stack-vs-flex judgment call the PR
body documents for dark-gradient headers (`LodgingUnitsPanel`, `HouseholdYearMembersModal`),
which was measured against one dialog and never against a second reviewer's eyes.

adamflagg:
## ⚠️ CodeRabbit's finding arrived AFTER the scan finished — it is unaddressed, and it needs your call

**Timing.** CodeRabbit posted at `07:52:42Z`. This branch's last commit is `303b012e` at
`07:49:35Z` — **three minutes earlier**. The scan agent never saw it. Flagging that explicitly so
the green check is not read as "reviewed and resolved".

**What it says.** `Modal.tsx:127` — *Major*: the close `<button>` is 18×18 CSS px in every branch,
which does not provide a 24px target. It proposes keeping the visible 18px mark but nesting it
inside a ≥24px button, and updating `Modal.test.tsx` and `AssignFamilyModal.test.tsx` to
distinguish target from mark.

**Why it was not applied automatically — two reasons that pull the same way:**

1. **It is a WCAG target-size finding, and this repo opts out of that class deliberately.**
   `CLAUDE.md` §4: three users, desktop, mouse and keyboard, *"no assistive technology and none
   planned"*. Applying a generic a11y rule here is the same shape of miss as the lodging-table
   `cm_id` "Major" that was also wrong for this codebase.
2. **18px is the owner-pinned design, not an implementation choice.** #2507 quotes the artifact's
   literal `.wbtn` CSS (`width: 18px; height: 18px`) and the 2026-08-20 ruling. Nesting it in a
   24px target changes the control's hit geometry away from what was approved.

**So this is a decision point with visual impact — which per the campaign's own rules means it
HOLDS rather than being resolved by an agent in either direction.** It is not being silently
dismissed and it is not being silently applied.

**The narrow question for you:** an 18px mouse target is small even setting accessibility aside.
Is that acceptable at the pinned size, or do you want the mark to stay 18px visually with a larger
invisible hit area? The second is a small change and does not alter the appearance.

PR #2523, opened 2026-08-21
refactor(sync): read socialize_with from its CampMinder custom field, not the CSV

### LINKED ISSUE #2484: refactor(sync): read socialize_with from its CampMinder custom field instead of the manually uploaded CSV
## ⚠️ CORRECTED 2026-08-21 — the predecessor is CLOSED. Schedulable now.

Verified against `main` @ `a6ffb313`.

**#2482 is CLOSED/COMPLETED.** The Classification section below says this *"should follow #2482's
daily custom-values cadence rather than precede it"* — that constraint is **discharged**; #2482's
cadence half shipped as #2489 (`1b9b2814`) and its capture half as PR #2493 (`caa0fe29`). The
"is becoming daily" line under "Why it is worth doing" reads as future tense and is now **past**.

**One decision inside this issue has a stated default, so it does not gate the work:** whether
cm_id `85804` (the *Explain* free text) should also be ingested is **out of scope for #2484**. File
it separately if wanted; do not block the dropdown swap on it. The build note below is kept because
the question is real, not because it is a gate.

---

## The ask

Retire the CSV as the source for `socialize_with` and read it from the CampMinder custom field
instead, so one of the five bunking-CSV columns stops depending on a manual upload.

## Why this one, and only this one

`original_bunk_requests` is fed by a **manually uploaded CampMinder report** (`csvFieldMap` in
`pocketbase/sync/bunk_requests.go`). That upload is the freshness bottleneck behind #2481 — a note
entered in CampMinder after the last upload is invisible indefinitely while the daily sync stays green.

Measured 2026-08-19 against the production snapshot, all five CSV columns against
`person_custom_values`:

| CSV column → field | custom-field equivalent | 2026 coverage | replaceable? |
|---|---|---|---|
| `RetParent-Socializewithbest` → **`socialize_with`** | **cm_id 85803** `Ret Parent-Socialize with best` | custom values **1,145** persons vs CSV **1,134**; overlap **1,134** | ✅ **yes — a strict superset** |
| `Share Bunk With` → `bunk_request_form` | `279009 TBM SHBTN - Bunk Request` | 76 vs 1,089, overlap 36 | ❌ different field |
| `Do Not Share Bunk With` → `staff_not_bunk_with` | none | — | ❌ no definition exists |
| `Internal Bunk Notes` → `internal_notes` | none | — | ❌ no definition exists |
| `BunkingNotes Notes` → `bunking_notes` | none | — | ❌ no definition exists |

**`socialize_with` is the only column with a clean custom-field equivalent, and the custom field has
BETTER coverage than the CSV** — every CSV person is present in the custom values, plus 11 more.

There is also **cm_id 85804** `Ret Parent-Socialize with best Explain` (830 persons, 2026), the
free-text companion, which the CSV lane does not carry at all.

## Why it is worth doing

- **It is the highest-volume column** — 911 households / 1,134 persons in 2026, more than any other.
- **It needs no AI.** `socialize_with` is a dropdown value parsed directly, not LLM-parsed
  (`process_requests.go`: *"socialize_with -> direct parsing (dropdown values)"*), so moving it off the
  CSV removes volume without touching the AI pipeline.
- ~~**The custom-values sync ~~is becoming~~ IS daily** (#2482, **CLOSED/COMPLETED** — cadence in
  #2489 `1b9b2814`, capture in PR #2493 `caa0fe29`), so this field would go from
  "as fresh as the last manual upload" to "at most a day old" for free.~~
  ⚠️ **CORRECTED 2026-08-21 (build pass) — the daily part is wrong.** #2482/#2489's daily
  cadence is the **FamilyCampBounded** instance only (`person_custom_values_family_camp`,
  scoped to family-camp attendees via `SessionResolver.GetFamilyCampPersonIDsAnyStatus`).
  `socialize_with` (cm_id 85803) is a **summer** bunking field — it is captured only by the
  unrestricted `person_custom_values` instance, which `orchestrator.go`'s
  `GetCustomValuesSyncJobs` still runs on the **weekly** cron (`0 4 * * 0`,
  `Scheduler.runCustomValuesSync`), unchanged by #2482/#2489. So this field goes from "as
  fresh as the last manual upload" (arbitrary, human-dependent) to **at most a week old**
  (automatic), not "at most a day old." Still a genuine freshness win, and does not change
  whether the swap is worth building — just corrects the number.
- It shrinks the manual-upload surface from five columns to four, which is a real step toward
  retiring it.

## What this does NOT solve

⚠️ **It does not fix #2481 or #2478.** Both note fields are CampMinder **built-ins**, not custom fields,
and have no definition to read. The CSV lane must stay for them. This issue narrows the dependency; it
does not remove it.

## Build notes

- **Verify the value shape matches before switching.** The CSV column and the custom field must produce
  the same parsed dropdown values — compare the two for the overlapping 1,134 persons and confirm they
  agree, row for row, before cutting over. A silent format difference would corrupt the social graph.
- Decide whether `85804` (the *Explain* free text) should also be ingested; it is a genuinely new
  signal the CSV never carried, and it *would* be LLM-parsed, so it is a scope decision rather than a
  freebie.
- Keep the CSV column readable during a transition so the two can be diffed in production.
- `socialize_with` feeds the social graph — see `bunking/graph/social_graph_builder.py`. Any change to
  its source needs the graph's tests green, not just the sync's.

## Classification

**SCOPED**, backend-only, no UI. Not urgent — it is a simplification, not a fix. ~~Should follow #2482's
daily custom-values cadence rather than precede it.~~ ⛔ **#2482 is CLOSED/COMPLETED — this sequencing
constraint is discharged and the issue is schedulable now.**

adamflagg:
`socialize_with` — the highest-volume of the five bunking-CSV columns (911 households / 1,134
persons in 2026) — was sourced entirely from a manually uploaded CampMinder report
(`csvFieldMap` in `pocketbase/sync/bunk_requests.go`). A note or dropdown answer entered in
CampMinder after the last manual upload stayed invisible indefinitely, with no automatic
refresh path, even while the daily sync stayed green.

CampMinder exposes the same answer as a custom field (cm_id `85803`, "Ret Parent-Socialize with
best"), already synced into `person_custom_values` by the existing weekly
`person_custom_values` sync. A 2026-08-19 production measurement (in the issue) found the
custom field is a strict superset of the CSV for the field's population: 1,145 persons in
custom values vs. 1,134 in the CSV, with all 1,134 CSV persons among them.

**Fix:** `bunk_requests.go`'s `processRow` now sources `socialize_with`'s content from
`person_custom_values` (via a new `loadPersonCustomFieldValuesByCMID` helper added to
`person_custom_field_values.go`). The custom field is authoritative only where there is no CSV
value to compare it against — the 11-person coverage-increase population the issue's own
numbers document (1,145 vs. 1,134, full overlap) — so those 11 people get a `socialize_with`
signal they never had before; that is a real coverage increase, not merely a swap. Where both
sources carry a non-empty value, the CSV value stays authoritative and any disagreement is
logged so the two can be diffed against production during the rollout. A person with no
custom-field answer yet (or a genuinely blank one) falls back to the CSV column unchanged. The
other four CSV columns (`bunk_request_form`, `staff_not_bunk_with`, `internal_notes`,
`bunking_notes`) are untouched; none has a usable custom-field equivalent.

**Why CSV wins on disagreement, review fix:** `socialize_with`'s sole consumer,
`orchestrator.py`'s `_parse_socialize_preference`, exact-matches the value against exactly two
literal strings ("Kids their own grade and one grade above/below") with no AI fallback. The
issue's own build note asked to verify the two sources produce identical text before cutting
over ("a silent format difference would corrupt the social graph"). Verified directly against
the production snapshot for 2026 (the current sync year): every custom-field value in that
population is one of the two literals the parser expects, and there is zero disagreement
between the CSV and the custom field for any of the overlapping persons — so today's rollout is
safe. But the code itself did not enforce that guarantee: the original version of this PR made
the custom field win even on disagreement, which would silently drop a request out of the
social graph the day CampMinder ever returns a differently-worded answer through the
custom-field API than through the CSV export. Fixed so the already-live, already-verified CSV
value wins on disagreement instead, with the mismatch still logged as the production safety net.

**Also fixed:** `loadPersonCustomFieldValuesByCMID` now trims the custom field's raw value.
CampMinder's custom-field API is a different export path than the CSV, whose column is already
trimmed (`getColumn`'s `strings.TrimSpace`); left untrimmed, a whitespace-only difference would
have read as a real disagreement and logged a spurious warning (plus unnecessary
`content_hash`/`processed` churn) on every sync run.

**Freshness, corrected:** the issue originally claimed this would make the field "at most a day
old," reasoning from #2482/#2489's new daily custom-values cadence. That daily cadence is the
`person_custom_values_family_camp` instance only, scoped to family-camp attendees.
`socialize_with` is a summer bunking field, captured by the *unrestricted* `person_custom_values`
instance, which is still on the weekly cron (`0 4 * * 0`,
`pocketbase/sync/scheduler.go:runCustomValuesSync`) — untouched by #2482/#2489. So the real
improvement is "arbitrary, human-dependent staleness" → "at most a week old, refreshed
automatically." Corrected the issue body and left a comment explaining why; still worth doing,
just not the number originally claimed.

**Deliberately not done:**
- The `85804` free-text "Explain" companion field is out of scope per the issue's own stated
  default — not ingested here.
- The CSV column (`csvFieldMap["RetParent-Socializewithbest"]`) is left in place and still
  parsed, specifically so the mismatch warning above has something to diff against in
  production during the transition; removing the CSV column entirely is a follow-up once that
  diff comes back clean, not part of this PR.
- No change to the other four CSV-sourced fields, the Python dropdown parser
  (`orchestrator.py`'s `_parse_socialize_preference`), or the social graph builder — the shape
  of the value going into `original_bunk_requests.content` is unchanged, only its source is.
- No automated pre-cutover verification job. The production check described above was done by
  hand during review; automating it (a scheduled diff that flags if the custom field ever starts
  returning a value outside the parser's two known literals) is a reasonable follow-up but is new
  scope, not part of this fix.

## Testing

- `pocketbase/sync/person_custom_field_values_by_cmid_test.go`: pins
  `loadPersonCustomFieldValuesByCMID`'s cm_id → PB id join through `custom_field_defs` (not
  `person_tag_defs`), year-scoping, the "field not synced yet → empty map, not an error"
  fallback contract, and (added in review) that surrounding whitespace is trimmed.
- `pocketbase/sync/bunk_requests_socialize_with_test.go`: pins the swap behavior in
  `processRow` — custom field is used when the CSV is absent, CSV wins on disagreement (added
  in review, replacing the original "custom field always wins" pin — see the file's top comment
  for why), falls back to CSV when the custom value is absent or blank, and the other four CSV
  fields are unaffected.
- All top-level tests written with `t.Parallel()` first per the package's
  `main_test_parallelism_test.go` guard; one planned log-line assertion was dropped rather than
  reaching for `captureSweepLogs` (a process-global `slog` swap incompatible with
  `t.Parallel()`), since exempting it required editing a guard file outside this issue's scope.
- `go build ./...`, `go vet ./...`, `gofmt -l`, `golangci-lint run`, `go test ./...` (pocketbase
  module) all clean.
- Mutation-checked: reverting the disagreement fix (making the custom field win again) turns
  `TestProcessRow_SocializeWith_DisagreementKeepsCSV` red; reverting the CSV-absent branch turns
  `TestProcessRow_SocializeWith_UsesCustomFieldWhenCSVAbsent` red — confirming both pin real
  behavior, not tautologies.

Closes #2484.

### DIFF
--- a/pocketbase/sync/bunk_requests.go
+++ b/pocketbase/sync/bunk_requests.go
@@ -40,10 +40,9 @@ type BunkRequestsSync struct {
 
 	// socializeWithByPerson maps person PB id -> the CampMinder custom field's raw
 	// value (cm_id 85803, "Ret Parent-Socialize with best") for the current sync year.
-	// kindred#2484: this is now the source of truth for the "socialize_with" field --
-	// the CSV column is still read in processRow (so a disagreement can be logged
-	// during the transition), but the value written to original_bunk_requests comes
-	// from here whenever a non-empty entry exists. A person absent from this map
+	// kindred#2484: processRow prefers this over the CSV column when the CSV has no
+	// value at all; when both are present and disagree, the CSV wins and the
+	// disagreement is logged (see processRow for why). A person absent from this map
 	// (custom field not yet synced for them) falls back to the CSV column unchanged.
 	socializeWithByPerson map[string]string
 }
@@ -259,20 +258,33 @@ func (s *BunkRequestsSync) processRow(row []string, columnIndex map[string]int, 
 		content := s.getColumn(row, columnIndex, csvColumn)
 
 		// kindred#2484: socialize_with's source of truth is the CampMinder custom
-		// field (cm_id 85803), not this CSV column, from here on. The column is still
-		// read above so a disagreement between the two can be logged during the
-		// transition -- the custom field wins when it carries a non-empty value; a
-		// person absent from the map, or present with an empty value (field synced
-		// but genuinely blank), falls back to the CSV content unchanged.
+		// field (cm_id 85803) whenever there is no CSV column to compare it against --
+		// the coverage-increase population the issue's own numbers document (1,145
+		// custom-field persons vs 1,134 CSV, full overlap). A person absent from the
+		// map, or present with an empty value (field synced but genuinely blank),
+		// falls back to the CSV content unchanged, same as before.
+		//
+		// PR #2523 review: when BOTH sources carry a non-empty value and they
+		// disagree, the CSV value wins, not the custom field. socialize_with's sole
+		// consumer, orchestrator.py's _parse_socialize_preference, exact-matches the
+		// value against exactly two literal strings with no AI fallback -- trusting an
+		// unverified custom-field value on disagreement risks silently dropping the
+		// request out of the social graph. The disagreement is still logged so the two
+		// can be diffed in production, per the issue's own build note ("verify the
+		// value shape matches before switching... before cutting over").
 		if fieldName == "socialize_with" {
 			if customValue, ok := s.socializeWithByPerson[personPBID]; ok && customValue != "" {
-				if content != "" && content != customValue {
-					slog.Warn("socialize_with: CSV and custom field values disagree, using custom field",
+				switch content {
+				case "":
+					content = customValue
+				case customValue:
+					// Agreement -- nothing to do.
+				default:
+					slog.Warn("socialize_with: CSV and custom field values disagree, keeping CSV value",
 						"person_cm_id", personID,
 						"csv_value", content,
 						"custom_field_value", customValue)
 				}
-				content = customValue
 			}
 		}
 
--- a/pocketbase/sync/bunk_requests.go
+++ b/pocketbase/sync/bunk_requests.go
@@ -37,6 +37,15 @@ type BunkRequestsSync struct {
 	BaseSyncService
 	validPersonIDs map[int]string // Maps CampMinder person ID to PocketBase person ID
 	csvPersonIDs   map[int]bool   // Tracks all enrolled person IDs seen in the CSV
+
+	// socializeWithByPerson maps person PB id -> the CampMinder custom field's raw
+	// value (cm_id 85803, "Ret Parent-Socialize with best") for the current sync year.
+	// kindred#2484: this is now the source of truth for the "socialize_with" field --
+	// the CSV column is still read in processRow (so a disagreement can be logged
+	// during the transition), but the value written to original_bunk_requests comes
+	// from here whenever a non-empty entry exists. A person absent from this map
+	// (custom field not yet synced for them) falls back to the CSV column unchanged.
+	socializeWithByPerson map[string]string
 }
 
 // NewBunkRequestsSync creates a new sync service
@@ -124,6 +133,20 @@ func (s *BunkRequestsSync) RunSync(csvPath string, _ int) error {
 
 	// Get current year from config
 	currentYear := s.getCurrentYear()
+
+	// kindred#2484: preload the socialize_with custom field values once for this run,
+	// rather than per-row -- the field definition and every person's answer for the
+	// year are the same for every row processRow will see. A load failure (or the
+	// field definition not being synced yet) is not fatal to the CSV sync: it just
+	// means every row falls back to sourcing socialize_with from the CSV column, as
+	// it did before this issue.
+	socializeWithValues, err := loadPersonCustomFieldValuesByCMID(s.App, cmIDSocializeWithBest, currentYear)
+	if err != nil {
+		slog.Warn("Loading socialize_with custom field values failed, sourcing socialize_with from CSV only",
+			"error", err)
+		socializeWithValues = map[string]string{}
+	}
+	s.socializeWithByPerson = socializeWithValues
 
 	// Process rows
 	rowNumber := 1 // Start at 1 since we already read headers
@@ -234,6 +257,24 @@ func (s *BunkRequestsSync) processRow(row []string, columnIndex map[string]int, 
 	// Process each CSV field that maps to our field options
 	for csvColumn, fieldName := range csvFieldMap {
 		content := s.getColumn(row, columnIndex, csvColumn)
+
+		// kindred#2484: socialize_with's source of truth is the CampMinder custom
+		// field (cm_id 85803), not this CSV column, from here on. The column is still
+		// read above so a disagreement between the two can be logged during the
+		// transition -- the custom field wins when it carries a non-empty value; a
+		// person absent from the map, or present with an empty value (field synced
+		// but genuinely blank), falls back to the CSV content unchanged.
+		if fieldName == "socialize_with" {
+			if customValue, ok := s.socializeWithByPerson[personPBID]; ok && customValue != "" {
+				if content != "" && content != customValue {
+					slog.Warn("socialize_with: CSV and custom field values disagree, using custom field",
+						"person_cm_id", personID,
+						"csv_value", content,
+						"custom_field_value", customValue)
+				}
+				content = customValue
+			}
+		}
 
 		// Check if record exists for this person/year/field combination
 		existingRecords, err := s.App.FindRecordsByFilter(
--- a/pocketbase/sync/bunk_requests_socialize_with_test.go
+++ b/pocketbase/sync/bunk_requests_socialize_with_test.go
@@ -1,0 +1,217 @@
+// kindred#2484: socialize_with's source of truth moves from the manually uploaded
+// bunk-requests CSV column ("RetParent-Socializewithbest") to the CampMinder custom
+// field it mirrors (cm_id 85803), read via person_custom_values. The CSV column is
+// still parsed during the transition so a disagreement can be logged in production
+// (the issue's own build note: "compare the two ... confirm they agree, row for row,
+// before cutting over"), but the value actually written to original_bunk_requests now
+// comes from the custom field whenever one is present.
+package sync
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/pocketbase/pocketbase/core"
+	pbtests "github.com/pocketbase/pocketbase/tests"
+)
+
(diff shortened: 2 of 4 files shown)


adamflagg:
## Triage review — PR #2523 (kindred#2484)

**Scan:** `code-review` skill at HIGH effort. It ran as a single-pass inline review
(no Agent-tool fan-out available in this nested context, so it worked all 10 angles itself
in one pass, capped at ≤15 findings) — reporting this explicitly per the campaign's
"report the level it actually ran at" instruction. Not a downgrade in coverage, just no
multi-agent verify pass.

### Findings and disposition

1. **[Fixed in-branch, code this PR wrote]** The swap made the custom field authoritative over
   the CSV *even on disagreement*. `socialize_with`'s sole consumer,
   `orchestrator.py`'s `_parse_socialize_preference`, exact-matches the value against exactly two
   literal strings with no AI fallback — an unverified custom-field value that disagreed with the
   known-good CSV could silently drop the request out of the social graph. I independently
   verified against the production snapshot (`pb_data/data-prod.db`) for 2026, the current sync
   year: every custom-field value in the population is one of the two literals the parser
   expects, and there is **zero disagreement** between the CSV and the custom field for the
   overlapping persons today — so the immediate rollout was already safe. But the code did not
   *enforce* that, so I fixed it: the CSV now wins on disagreement (still logged, per the issue's
   own "diff in production" ask), and the custom field is authoritative only when the CSV has no
   value at all. Rewrote the test that pinned the old ("custom field always wins") invariant —
   deliberate retirement, rationale is in the test file's top comment — and added a new test for
   the CSV-absent branch. Both mutation-checked (see below).

2. **[Fixed in-branch, code this PR wrote]** `loadPersonCustomFieldValuesByCMID` didn't trim the
   custom field's raw value, while the CSV side is already trimmed
   (`getColumn`'s `strings.TrimSpace`). A whitespace-only difference from CampMinder's
   custom-field API (a different export path than the CSV) would have read as a spurious
   disagreement — logging a false-positive warning and unnecessarily rewriting
   `content_hash`/`processed` on every sync run. Fixed with a one-line trim; added a dedicated
   test.

3. **[Fixed via PR body edit]** The body claimed "this is a value-swap, not a coverage change,"
   but the issue's own numbers (1,145 custom-field persons vs. 1,134 CSV, full overlap) mean 11
   persons who had no CSV answer now get one — a real coverage increase for that population.
   Corrected the language in the PR body (also fixed a pre-existing wrong reference to
   `orchestrator.go`'s `_parse_socialize_with_direct` — the real file/function is
   `orchestrator.py`'s `_parse_socialize_preference`).

### Premise check

Not a product decision wearing a bug's clothes — the issue names a real contradiction (CSV
freshness bottleneck vs. an already-synced, superset custom field) and the fix follows without
persuasion. The one embedded scope question (cm_id 85804, the free-text companion) has a stated
default in the issue and is correctly left out.

### Mutation-check transcript

- `TestProcessRow_SocializeWith_DisagreementKeepsCSV`: reverting the fix (making the custom
  field win on disagreement again) turns it red with
  `socialize_with content = "YOUNGER", want "OLDER (stale csv answer)"`. Confirmed, then reverted
  the mutation.
- `TestProcessRow_SocializeWith_UsesCustomFieldWhenCSVAbsent`: reverting the CSV-absent branch
  turns it red with `found 0 socialize_with OBR rows for pb_person_1005/2026, want 1`. Confirmed,
  then reverted the mutation.

### Verification

- `go build ./...`, `go vet ./...`, `gofmt -l`, `golangci-lint run --config ../.golangci.yml`
  (0 issues after fixing 3 lint findings my own new test/switch code introduced: a `goconst`
  hit, an `lll` line-length hit, and a `staticcheck` tagged-switch suggestion), `go test ./...`
  (pocketbase module, including `-race`) all clean via `scripts/pre-push-verify.sh` —
  **ALL CHECKS PASSED**.
- CI Summary: SUCCESS, confirmed by an actual run (all Go Tests shards, Go Lint, Migration Smoke
  Test, CodeQL, etc. green).
- Push verified by the remote ref moving (`git rev-list --count origin/...HEAD` = 0), not by
  tool exit code.
- `closingIssuesReferences` (GraphQL): `#2484`, non-empty.
- No hand-written frontend files in the diff — visual-impact gate does not fire.
- Staff-facing number check: this PR changes the *source* of one bunk-request field's content
  (an age-preference dropdown answer feeding the solver/social graph), not any metric, count,
  rate, or population classification staff reads. No number's meaning moves.

### Deliberately not extended

- Did not build the automated pre-cutover verification the issue's build note asked for
  (a scheduled diff job). The one-time check I ran by hand during review satisfies the note for
  today's rollout; automating it is new scope, noted as a follow-up in the PR body rather than
  filed as a separate issue (it's small enough to fold into the CSV-column-removal follow-up
  already mentioned in "Deliberately not done").
- Did not touch the other four CSV-sourced fields, the Python parser, or the social graph
  builder — out of scope per the issue itself.

**Merging** — all 8 gate conditions hold: no remaining decision, CI Summary green (confirmed
run), scan findings fixed in-branch, push verified by ref, `closingIssuesReferences` non-empty,
diff still recognizably this item, no visual impact, no staff-facing number redefinition.

